PR #136, opened 2014-01-29
Using PhantomJS on Travis to run Qunit tests

mwjames:


mwjames:
I don't have time for this and I don't see any one else taking the lead on this, so bye bye!

PR #225, opened 2014-03-13
Introduce @sortkey to support sorting annotation for a subobject

mwjames:
- `@sortkey` is a restricted keyword which allows to set an arbitrary sort value for a particular subobject
- `@sortby` is a restricted keyword which allows to set a reference to a property which in return will use the value of the reference property as sort value

`@sortby` and `@sortkey` are case sensitive, values are stored as a "string", and further will not appear visible to the end-user but subobject with a dedicated `_SKEY` will work the same way as a normal wiki page with a `_SKEY`.
## Subobject example

```
{{#subobject:
 |Has group=test
 |Has value=1111
 |@sortby=Has value --> _SKEY=1111
}}

{{#subobject:
 |Has group=test
 |Has value=9999
 |@sortkey=A --> _SKEY=A
}}
```
## Recurring events example

```
{{#set_recurring_event:property=Has date
 |start=January 4, 2010 05:00
 |end=June 8, 2011
 |unit=week
 |period=1
 |@sortby=Has date
}}
```

EDIT:PR only adds `@sortkey`

### DIFF
--- a/includes/parserhooks/SubobjectParserFunction.php
+++ b/includes/parserhooks/SubobjectParserFunction.php
@@ -18,6 +18,8 @@ use Parser;
  */
 class SubobjectParserFunction {
 
+	const PARAM_SORTKEY = '@sortkey';
+
 	/** @var ParserData */
 	protected $parserData;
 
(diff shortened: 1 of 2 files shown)


kghbln:
@mwjames Just to get this right: There is no way to manipulate the order of sorting, i.e. results are always shown in descending order. So if one would want to change the sort order than the sortkeys will have to be changed.

mwjames:
> Just to get this right: There is no way to manipulate the order of sorting, i.e. results are always shown in ascending order. So if one would want to change the sort order than the sortkeys will have to be changed.

This change only deals with the storage of a sorkey to a subobject it does not alter any existing behaviour on how results (using a default sorting) are represented.

By representing, I refer to any process (query, Special:Browse) that uses the data from the store (including the sorkey) to build a displayable  format (table, list etc.).

I would have to dig but I think the default order is ascending if no explicit order is given but you should be able to manipulate the display of a query using the `order` parameter for a subject.

kghbln:
Oops, indeed ascending is the default and manipulating the sort order works too. I messed up things at my end with a typo so nothing worked because this. One of these days ... :(

PR #494, opened 2014-08-14
Add QueryBuilder, QueryContainer, and QueryCompiler

mwjames:
This PR does nothing more than moving isolatable code into its own class

- Move `SMWSQLStore3Query` to its own `SMW\SQLStore\QueryEngine\QueryContainer` class
- Move code that builds query artifacts from `SMWSQLStore3QueryEngine` to `SMW\SQLStore\QueryEngine\QueryContainerBuilder`


JeroenDeDauw:
Ah damn. I totally forgot about this. The small cleanup things I did today in the code which is refactored here where to get a better understanding of what it did in order to determine a better design. What I was thinking of is pretty close to what you did here.

JeroenDeDauw:
I'd like to move forwards with this change. Mind if I rebase it against master and make some follow up commits?

mwjames:
> Ah damn. I totally forgot about this. The small cleanup things I did today in the code which is refactored here where to get a better understanding of what it did in order to determine a better design. What I was thinking of is pretty close to what you did here.

The `SPARQLStore` re-factoring [0, 1] in 2.1 was easier then it would be for the `SQLStore` therefore not having both in one release was a conscious decision.

> I'd like to move forwards with this change. Mind if I rebase it against master and make some follow up commits?

:+1: 

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/master/includes/src/SPARQLStore/QueryEngine

[1] https://scrutinizer-ci.com/g/SemanticMediaWiki/SemanticMediaWiki/code-structure/master/package/SMW%5CSPARQLStore%5CQueryEngine

JeroenDeDauw:
@mwjames can you merge #718 and #721? I don't think they will conflict, but in case they do, I'll only have to rebase once.

mwjames:
> I'd like to move forwards with this change. Mind if I rebase it against master and make some follow up commits?

To make it easier for you (as there were some non-obvious changes, it took me 2h to find out why one particular integration test didn't pass), the rebase is made against the 2.1.x branch (left all existing tests untouched, indicating a minimized risk for regressions).

mwjames:
I'm not full satisfied with the naming, instead of `QueryCompiler` it probably should be called `QueryContainerBuilder` as each individual `...ContainerBuilder` class actual returns a `QueryContainer` object.

Furthermore the `QueryContainerBuilder` registration should be moved to something like [ConditionBuilderStrategyFinder.php](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/src/SPARQLStore/QueryEngine/ConditionBuilderStrategyFinder.php).

JeroenDeDauw:
Heh. You beat me to the rebase. Why did you rebase against 2.1.x instead of master though? Now this still cannot be merged.

I agree the naming can be a bit better and some other improvements can be made. I do however suggest we merge this first, so we are able to work on various aspects in parallel, without stuffing it all into one big PR. Having some unfortunate naming on master for some time is not a big deal - we already have a lot of that, and this PR fixes a pile of issues already. As long as we do not create a release with a new badly named public interface or something like that, we should be good.

mwjames:
> Why did you rebase against 2.1.x instead of master though? Now this still cannot be merged.

OK, doing `git --rebase` was a nightmare but doing `git merge` against 2.2 showed a more promising result.

JeroenDeDauw:
@mwjames can you create issues for the things you think should still be done?

mwjames:
> can you create issues for the things you think should still be done?

#743

PR #495, opened 2014-08-14
Table format: Add printout parameter 'sep'

s7eph4n:
This parameter allows to specify a separator between values different from '<br />'

Also fixed use of '&lt;br />' to '&lt;br>'.

I'd like to add some testing, but I have no clue at all how this whole setup works. Just specifying arrays instead of singular values does not cut it.

### DIFF
--- a/includes/query/SMW_PrintRequest.php
+++ b/includes/query/SMW_PrintRequest.php
@@ -244,13 +244,13 @@ class SMWPrintRequest {
 
 	/**
 	 * Returns the value of a named parameter.
-	 * 
+	 *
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
@mwjames hey! Don't steal all the credit! I also contributed to that test. And it's Markus his achievement that the production code is near impossible to test.

@s7eph4n rebase needed. If you manually verified that this works, then we can merge it. Though getting a verification from a second person would be fantastic.

mwjames:
Solved by #988

Issue #191, opened 2014-02-10
Multiple entries for an entity with same the subject is returned by the Store


mwjames:
https://bugzilla.wikimedia.org/show_bug.cgi?id=48707

jcostello1:
Not sure it will help but here it is anyways. Perhaps some new clues....

I recently was hit with a mass of duplicate entries on my site. I recently upreved to SMW 2.3.1 but can't say that's when it happened. For quite a while I had been carrying a single duplicate page result, a real page name as well as a previous version of the page name that contained an accented character. I chalked it up to the accent and the name change until last night when I discovered a large number of duplicates in a particular query, about 1/3-1/2 of the results. 

All duplicates were strange in that they had odd capitalization of the real page names. Either just the first, the first couple or sometimes all of the words in the titles were fully capitalized but otherwise correct. My original as well as these latest duplicates were all pages with the same infobox template inclusion which sets the semantic properties that were being queried and I haven't detected duplicates in any other area of my site. 

Rebuilding did not correct the issue and nor did editing and saving the effected pages so in the end I removed all the smw_ tables (after backing up of course), reinitialized and rebuilt. With that, all duplicates were removed. Thank goodness my site is small enough!

I'm currently running: MW 1.23, SMW 2.3.1, PHP 5.5.33

mwjames:
> All duplicates were strange in that they had odd capitalization of the real page names. Either just the first, the first couple or sometimes all of the words in the titles were fully capitalized but otherwise correct.

That seems odd and an example or screenshot would have been helpful.

> My original as well as these latest duplicates were all pages with the same infobox template inclusion which sets the semantic properties

One hypothesis of mine was that it has something to do with a race condition during the update (and templates play a vital role in how/when) or DB transaction rollbacks somehow interfering with the update process but this issue is very hard to pin point and with no exact way of reproducing it chances to find the condition as to when and where a duplication occurs are greatly diminished. 

 2.4 might help with the transaction lag since updates are pushed into a deferred queue (#1435) together will all the other MW DB commits.

> Rebuilding did not correct the issue and nor did editing and saving the effected pages so in the end I removed all the smw_ tables (after backing up of course), reinitialized and rebuilt. With that, all duplicates were removed.

Hopefully this is fixed by detecting "null" entities (those that have no references, #1216) and being removed during the `rebuildData.php` run.

jcostello1:
[...]

PR #510, opened 2014-08-20
Display `distinct` query results

mwjames:
Relates to #191

This change does not solve the problem of #191 (where duplicates entries are created in `smw_objects_ids` etc.) but it will avoid having duplicate entries being displayed.

### QueryEngine

~~Use `GROUP BY` to allow for INNER JOIN values the be distinctive as well. See also [distinct-vs-group][distinct-vs-group].~~

Using the `!isset( $dataItemCache[ $dataItem->getHash() ] )` to determine whether the subject has already been selected or not.

### fetchSemanticData

Using a `hash` to determine whether a specific column value has been already processed and avoid its output duplication during `getPropertyValues()` or `getSemanticData`.

[distinct-vs-group]: http://sqlmag.com/database-performance-tuning/distinct-vs-group

### DIFF
--- a/includes/storage/SQLStore/SMW_SQLStore3_Queries.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3_Queries.php
@@ -420,13 +420,29 @@ class SMWSQLStore3QueryEngine {
 			return 0;
 		}
 
+		$db = $this->m_store->getDatabase();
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Needs rebase

PR #613, opened 2014-11-07
Remove ambiguous responsibility from TitleCreator, add DeepRedirectTarge...

mwjames:
...tResolver
- Add `DeepRedirectTargetResolver`
- Protect against circular redirects
- Update `BrowseBySubject` to use `DeepRedirectTargetResolver`
- Add subobject parameter to `BrowseBySubject`
- Add `BrowseBySubjectTest::testRawJsonPrintOutput`
- Follow-up on #401

### DIFF
--- a/includes/src/MediaWiki/DeepRedirectTargetResolver.php
+++ b/includes/src/MediaWiki/DeepRedirectTargetResolver.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace SMW\MediaWiki;
+
+use Title;
+use RuntimeException;
+
+/**
+ * @license GNU GPL v2+
+ * @since 2.1
+ *
+ * @author mwjames
+ */
+class DeepRedirectTargetResolver {
+
+	/**
+	 * @var PageCreator
+	 */
+	private $pageCreator = null;
+
+	/**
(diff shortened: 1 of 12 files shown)


kghbln:
@mwjames I finally [documented the subobject parameter](https://semantic-mediawiki.org/w/index.php?title=Ask_API&diff=36959&oldid=35800) Hope that I got it right.

includes/src/MediaWiki/DeepRedirectTargetResolver.php
```diff
@@ -0,0 +1,97 @@
...
+			return $title;
+		}
+
+		throw new RuntimeException( "Redirect target is unresolvable" );
+	}
+
+	private function addToKnownRedirects( $title ) {
+
+		if ( !isset( $this->knownRedirects[ $title->getPrefixedDBkey() ] ) ) {
+			return $this->knownRedirects[ $title->getPrefixedDBkey() ] = false;
```
JeroenDeDauw: `return a = b` detected. Even if this does what you wanted, please write it in a form that does not like it's a bug

mwjames: Addressed with #792

PR #659, opened 2014-11-30
Add EditPage help / $smwgEnabledEditPageHelp

mwjames:
Finding help entry points is not always easy for novice user therefore displaying help on the edit page will help those yet unfamiliar with SMW.

Depending on the type of page, messages will change. Setting `$smwgEnabledEditPageHelp = false` will remove all smw-related message from the edit page.

Messages will change based on the context such as:

editpage-1

editpage-2

editpage-3

editpage-4

editpage-5


kghbln:
[Documented at a second spot](https://www.semantic-mediawiki.org/wiki/Help:Edit_help)

PR #673, opened 2014-12-11
Ignore $wgCapitalLinks setting for the property namespace

mwjames:
If `$wgCapitalLinks` = false [0] then annotations like `[[has type:: ... ]]` will fail because of the inactive conversion rule `has -> Has` [1].

[0] https://www.mediawiki.org/wiki/Manual:$wgCapitalLinks
[1] https://www.semantic-mediawiki.org/wiki/Thread:Help_talk:Properties_and_types/Is_any_part_of_the_data_type_declaration_case-sensitive%3F


JeroenDeDauw:
The fuck @coveralls?

kghbln:
@JeroenDeDauw That's marketing gone wrong.

Issue #702, opened 2015-01-09
Wrong HTTP headers sent by SPARQLStore - cannot work properly with Sesame triplestore


tfrancart:
In extensions/SemanticMediaWiki/includes/sparql/SMW_SparqlDatabase.php, method doQuery :

    curl_setopt( $this->m_curlhandle, CURLOPT_URL, $this->m_queryEndpoint );
    curl_setopt( $this->m_curlhandle, CURLOPT_HTTPHEADER, array('Accept: application/sparql-results+xml,application/xml;q=0.8' ));
    curl_setopt( $this->m_curlhandle, CURLOPT_POST, true );
    $parameterString = "query=" . urlencode( $sparql ) .
    ( ( $this->m_defaultGraph !== '' )? '&default-graph-uri=' . urlencode( $this->m_defaultGraph ) : '' );
    curl_setopt( $this->m_curlhandle, CURLOPT_POSTFIELDS, $parameterString );
    curl_setopt( $this->m_curlhandle, CURLOPT_HTTPHEADER, array('Content-Type: application/x-www-form-urlencoded;charset=UTF-8'));

    $xmlResult = curl_exec( $this->m_curlhandle );

CURLOPT_HTTPHEADER cannot be set 2 times. Thus the first header ("Accept: application/sparql-results+xml,application/xml;q=0.8'") is not send correctly. Code should be :

    curl_setopt( $this->m_curlhandle, CURLOPT_URL, $this->m_queryEndpoint );
    curl_setopt( $this->m_curlhandle, CURLOPT_HTTPHEADER, array(
         'Accept: application/sparql-results+xml,application/xml;q=0.8',
         'Content-Type: application/x-www-form-urlencoded;charset=UTF-8'
    ));

    curl_setopt( $this->m_curlhandle, CURLOPT_POST, true );
    $parameterString = "query=" . urlencode( $sparql ) .
    ( ( $this->m_defaultGraph !== '' )? '&default-graph-uri=' . urlencode( $this->m_defaultGraph ) : '' );
    curl_setopt( $this->m_curlhandle, CURLOPT_POSTFIELDS, $parameterString );

    $xmlResult = curl_exec( $this->m_curlhandle );

mwjames:
> In extensions/SemanticMediaWiki/includes/sparql/SMW_SparqlDatabase.php, method doQuery :

[...]

PR #703, opened 2015-01-09
Sesame complaint HTTP header, refs #702

mwjames:
refs #702


mwjames:
@tfrancart Can you verify that this would work for your `openrdf-sesame`.

tfrancart:
correct.

PR #710, opened 2015-01-14
Better approach to match `=` in query condition

mwjames:
The #640 apporach was too simplistic, it broke the smw.org@`Template:Invert-property` which produces strings like `sep=| ]][[Location of::` as part of the template transclustion during the `#ask` execution.


mwjames:
@kghbln If you found that `Template:Invert-property` (England etc.) was broken then this should fix the issue.

PR #715, opened 2015-01-18
Remove apparently dead code

JeroenDeDauw:


mwjames:
This is now a bit dangerous because this code is currently not tested and removing it before 2.1 release could make the release unstable without knowing.

includes/storage/SQLStore/SMW_SQLStore3_Queries.php
```diff
@@ -858,53 +858,45 @@ protected function compileValueDescription(
 
 		// Try comparison based on value field and comparator,
 		// but only if no join with SMW IDs table is needed.
-		if ( $diType != SMWDataItem::TYPE_WIKIPAGE ) {
```
mwjames: Where did this go?

https://scrutinizer-ci.com/g/SemanticMediaWiki/SemanticMediaWiki/code-structure/master/class/%2Bglobal%5CSMWSQLStore3QueryEngine

JeroenDeDauw: If the else branch is dead, then the if part always got execute, so removing this check should not change anything.

JeroenDeDauw:
> this code is currently not tested

Really? Don't we have integration tests that execute queries?

> before 2.1 release could make the release unstable without knowing

Well yeah. How about we branch from the commit before then?

mwjames:
> Really? Don't we have integration tests that execute queries?

Yes, and we do have quite a few of them but `SQLStore3_Queries` is such mess that any changes can still have hidden effects that currently are not detected by our tests.

> Well yeah. How about we branch from the commit before then?

Probably the best idea.

JeroenDeDauw:
You can haz `2.1.x` branch: https://github.com/SemanticMediaWiki/SemanticMediaWiki/branches

mwjames:
Master is currently broken.

JeroenDeDauw:
How do you figure that?

mwjames:
> How do you figure that?

Was just looking at the red Travis button ;-)

PR #717, opened 2015-01-18
Cleanup in SMWSQLStore3QueryEngine

JeroenDeDauw:


JeroenDeDauw:
Pushing some other cleanup commits onto here since they'd otherwise merge conflict

PR #718, opened 2015-01-18
Imrpove function signature docs

JeroenDeDauw:


JeroenDeDauw:
Looks like it does not conflict with #718

PR #720, opened 2015-01-18
Fix type hints

JeroenDeDauw:
Most importantly, this class depends on SMWSQLStore3, not SQLStore


tests/phpunit/includes/storage/sqlstore/StatisticsCollectorTest.php
```diff
@@ -41,7 +41,7 @@ private function newInstance( $count = 55, $cacheEnabled = false, $hash = 'foo'
 			'isFixedPropertyTable' => true
 		) );
 
-		$store = $this->newMockBuilder()->newObject( 'Store', array(
+		$store = $this->newMockBuilder()->newObject( 'SMWSQLStore3', array(
```
JeroenDeDauw: @mwjames do I first need to register additional  stuff in this mock builder or what?

mwjames: One of the old tests that has been converted to use the standard `Mock` invocation. It is an independent PR therefore I will put this on my list and use the opportunity to fix the test.

mwjames:
>Why did you change this back? Now it's wrong again...

No, not really because SMWSQLStore3 is -> `\SMW\SQLStore\SQLStore` not the other way around.

JeroenDeDauw:
`SQLStore` does not have the method `getStatisticsTable`, which this class expects to be present on the field set via this constructor parameter

JeroenDeDauw:
Also, the class is named `SMWStore`, not `SQLStore`.

JeroenDeDauw:
And `getConnection` on `SMWStore` has a required parameter which is not provided. This code looks very broken to me now.

mwjames:
According to my reading it does because of [0] where `class_alias( 'SMWSQLStore3', 'SMW\SQLStore\SQLStore' );` (because of the legacy) [1]

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/storage/SQLStore/SMW_SQLStore3.php#L866

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/SemanticMediaWiki.php#L91

JeroenDeDauw:
As far as I can tell, only calls to the store field are `getConnection` and `getstatisticsTable`. In which case we should just provide those and not the store.

mwjames:
> And getConnection on SMWStore has a required parameter which is not provided. This code looks very broken to me now.

The default parameter changes with the implementation [0, 1] but depending on whether a specific store implementation needs to call a specific connection (as for the `SPARQLStore` as it needs the `db` on some calls such as the whole concept table/statistics stuff) can call `getConnection( 'mw.db' )` but the `SQLStor`e will never be able to call `getConnection( 'sparql' )` as this would raise an exception.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/SPARQLStore/SPARQLStore.php#L399
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/storage/SQLStore/SMW_SQLStore3.php#L890

mwjames:
> As far as I can tell, only calls to the store field are getConnection and getstatisticsTable. In which case we should just provide those and not the store.

I tried that but then I would had to change the interface because the collector needs details on the property tables by calling `findPropertyTableByType` which requires the `SQLStore`.

I just wanted to fix the tests not the murky interface ;-)

mwjames:
The setup registers 3 known methods [0] to access the `DB` and if necessary can add more depending on whether or not a connection handler is available (as in case of the `SPARQLStore`). This allows in future to do something like `getConnection( 'pdo' )` without having to change the whole interfaces (the `ConnectionManager` "manages" all available connections).

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/Setup.php#L90

JeroenDeDauw:
> The default parameter changes with the implementation

I was under the impression you are not allowed to do this (adding a default value to a method that does not have one in the parent class) with PHP strict standards.

> I tried that but then I would had to change the interface because the collector needs details on the property tables by calling findPropertyTableByType which requires the SQLStore.

Ah... I did not see this since it's in the parent class. Not sure the template approach taken is the best one here.

Anyway. #730

PR #741, opened 2015-01-21
Use Serialization component - fix #740

### LINKED ISSUE #740: Usage of the Serialization component
We have copies of old versions of the Serializer and Deserializer interfaces from the [Serialization component](https://github.com/wmde/Serialization). If we use the component instead, we gain interoperability with other things that use it, such as the new [Ask serialization component](https://github.com/wmde/AskSerialization/).

Any objections against, or concerns with, switching to these interfaces?

JeroenDeDauw:


JeroenDeDauw:
Build passed, except the one that's incorrectly in the require section...

mwjames:
> Build passed, except the one that's incorrectly in the require section...

Well then ...

JeroenDeDauw:
You can also admire my rage in #751 :)

PR #744, opened 2015-01-21
Fix type hint in PropertiesCollector

JeroenDeDauw:


mwjames:
Travis is having a field day (https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/builds).

JeroenDeDauw:
Don't worry, it'll catch up in a few years, at which point it will probably be broken due to 5.3 no longer being supported.

slow

Issue #756, opened 2015-01-24
#set to be usable with a template parameter


mwjames:
If a user wants to make a `QuickQuery` for an annotated value normally s(he) has to go for some length and employ another extension (SF) to be used in combination for creating something like`
{{#ifeq:{{{Keyword|keyword}}}||{{!}} {{#arraymap:{{{keyword|{{{Keyword|}}} }}}|;|x|[[Has keyword::{{ucfirst:x}}| ]]{{#ask: [[Has keyword::{{ucfirst:x}}]]|?has description|limit=0|searchlabel={{ucfirst:x}} }}|&#32;·&#32; }} }}
` which is somewhat impractical and performance extensive (using arraymap).

Extending `#set` with `|template=` allows to:
- `{{#set: Has keyword=one;two;three|+sep=;|template=SetQuickQueryTemplate}}` with a template parameter `|template=SetQuickQueryTemplate` that can include
- `<includeonly>{{#ask: [[{{{property}}}::{{{value}}}]]|limit=0|searchlabel={{{value}}} }}&#32;</includeonly>` where `{{{property}}}` and `{{{value}}}` are fixed named arguments.

This eliminates the need for SF `arraymap` as dependency, `|+sep=;` is already part of `#set` to divide an annotation string, and `|template=` allows for customized links or queries.

mwjames:
Changes to `SetParserFunction` are rather simple: 

```
		$parametersToArray = $parameters->toArray();

		if ( $this->parser !== null && isset( $parametersToArray[ 'template'] ) ) {
			$template = $parametersToArray[ 'template'][0];
			unset( $parametersToArray[ 'template'] );
		}

		foreach ( $parametersToArray as $property => $values ){

			foreach ( $values as $value ) {

				if ( $template !== '' ) {
					$html .= $this->createTemplateOutput( $template, $property, $value );
				}
```
```
		if ( $html !== '' ) {
			$html = $this->parser->recursiveTagParse( $html );
		}

		$html .= $this->messageFormatter
			->addFromArray( $this->parserData->getErrors() )
			->addFromArray( $parameters->getErrors() )
			->getHtml();

		return array( $html, 'isHTML' => true );
	}

	private function createTemplateOutput( $template, $property, $value ) {
		return '{{' . $template . '|property=' . $property . '|value=' . $value . '}}';
	}
```

mwjames:
[...]

PR #758, opened 2015-01-24
#set to support template parameter, refs #756

mwjames:
refs #756

### DIFF
--- a/includes/parserhooks/SetParserFunction.php
+++ b/includes/parserhooks/SetParserFunction.php
@@ -2,6 +2,8 @@
 
 namespace SMW;
 
+use SMW\MediaWiki\HtmlTemplateRenderer;
+
 use Parser;
 
 /**
@@ -20,7 +22,7 @@ use Parser;
 class SetParserFunction {
 
 	/**
-	 * @var ParserDate
+	 * @var ParserData
 	 */
 	private $parserData;
 
@@ -30,14 +32,21 @@ class SetParserFunction {
 	private $messageFormatter;
 
 	/**
+	 * @var HtmlTemplateRenderer
+	 */
+	private $templateRenderer;
+
+	/**
 	 * @since 1.9
 	 *
 	 * @param ParserData $parserData
 	 * @param MessageFormatter $messageFormatter
+	 * @param HtmlTemplateRenderer $templateRenderer
 	 */
-	public function __construct( ParserData $parserData, MessageFormatter $messageFormatter ) {
+	public function __construct( ParserData $parserData, MessageFormatter $messageFormatter, HtmlTemplateRenderer $templateRenderer ) {
 		$this->parserData = $parserData;
 		$this->messageFormatter = $messageFormatter;
+		$this->templateRenderer = $templateRenderer;
 	}
 
 	/**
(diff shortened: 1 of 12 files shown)


includes/parserhooks/SetParserFunction.php
```diff
@@ -30,6 +30,11 @@ class SetParserFunction {
 	private $messageFormatter;
 
 	/**
+	 * @var Parser
+	 */
+	private $parser = null;
```
JeroenDeDauw: Why init to `null` explicitly? The doc is also kinda wrong, since it can be`null`

includes/parserhooks/SetParserFunction.php
```diff
@@ -41,6 +46,15 @@ public function __construct( ParserData $parserData, MessageFormatter $messageFo
 	}
 
 	/**
+	 * @since  2.2
+	 *
+	 * @param Parser $parser
+	 */
+	public function setParser( $parser ) {
```
JeroenDeDauw: Somehow it should be made very clear to the users of this class that `setParser` needs to be called before `parse`, since else you can end up with a fatal

mwjames: Because I don't have to change the signature (means all test remain as is) and I only have to check for `$this->parser !== null` once.

JeroenDeDauw: Hm... you indeed don't run into a problem if the template param is not set. Still, I find this highly suspicious

mwjames:
Going for a more generic solution (isolating the Parser injection and to be re-used later by the `ListResultPrinter` to untangle the template generation):

```php
	$templateGenerator= new TemplateGenerator( $this->parser );

	$this->templateGenerator->addField( 'property', $property );
	$this->templateGenerator->addField( 'value', $value );
	$this->templateGenerator->addField( '#', $count++ );
	$this->templateGenerator->collecFieldsForTemplate( 'Foo' );

	//...

	$this->templateGenerator->doParse();
	$this->templateGenerator->getHtml();
```

JeroenDeDauw:
```php
    $this->templateGenerator->doParse();
    $this->templateGenerator->getHtml();
```

So you'd generate and store the HTML during `doParse` and then fetch it via `getHtml`? If so, why store the state in between at all? Why not just return it when it's created?

JeroenDeDauw:
I see how putting the Parser in there makes the earlier problem go away, though I'm a bit skeptical about how cohesive it is to put the parser and the template wikitext building together.

JeroenDeDauw:
Looking at the work being done, it follows this format:

[semi structured input (aka, the template params)] => [some representation of the input]

That happens when going from params to wikitext, and when going from params to html. So how about

```php
interface TemplateRenderer {
    public function render( Template $t );
}


class WikitextTemplateRenderer implements TemplateRenderer {
}

class HtmlTemplateRernderer implements TemplateRenderer {
    public function __construct( WikitextTemplateRenderer $r ) {}
}
```

Somehow this seems like a lot of things for little work though

mwjames:
> HTML during doParse and then fetch it via getHtml?

Because one need `doParse` (`$this->parser->recursiveTagParse( $this->template );`) for #set but not for the `ListResultPrinter `. Parsing is done outside of the `ListResultPrinter` and applied to the whole result string by the `ResultPrinter`.

> though I'm a bit skeptical about how cohesive it is to put the parser and the template wikitext building together.

Only use an object where it is actually required and #set itself don't need to know about the `Parser` because only the `TemplateGenerator` uses the object.

mwjames:
> Somehow this seems like a lot of things for little work though

shooting pigeons here ...

JeroenDeDauw:
What?

huh

mwjames:
OK, now went for the `WikitextTemplateRenderer` \ `HtmlTemplateRenderer` without the interface.

PR #762, opened 2015-01-28
adding https to fix SSL Compatibility warning

jehovahsays:
http can go with https but https can't go with http.


mwjames:
Thanks

jehovahsays:
No trouble.

PR #771, opened 2015-02-01
Move tem table creation to service and introduce DBAL

JeroenDeDauw:


JeroenDeDauw:
Guess it now breaks since the MW magical test table stuff is not happening for the temp tables now

mwjames:
```
1) SMW\Tests\Integration\MediaWiki\Import\RecordDataTypeTest::testImportOfRecordValues
RuntimeException: Import failed with 
A database error has occurred. Did you forget to run maintenance/update.php after upgrading?  See: https://www.mediawiki.org/wiki/Manual:Upgrading#Run_the_update_script
Query: 
Function: SMW::executeQueries
Error: 
```
```
4) SMW\Tests\Integration\Query\DisjunctionQueryDBIntegrationTest::testDisjunctionSubqueryForPageTypePropertyChainThatComparesEqualToValue
Doctrine\DBAL\Exception\TableExistsException: An exception occurred while executing 'CREATE TEMPORARY TABLE sunittest_t0 (id INTEGER UNSIGNED DEFAULT NULL)':
SQLSTATE[HY000]: General error: 1 table sunittest_t0 already exists
```

JeroenDeDauw:
In Wikibase Query we also create a DBAL connection from MW code. (In fact I copied the ConnectionBuilder code.)

For some reason I forgot we need to run the tests with `--use-normal-tables`

JeroenDeDauw:
Talked to @beberlei and now figure DBAL does not presently have what we need here. Will just split of the temp table code into its own service and keep the DBAL introduction for something else.

JeroenDeDauw:
Created item for follow up: #774

PR #777, opened 2015-02-01
Introduce ConceptCache

JeroenDeDauw:


JeroenDeDauw:
777, do I win something?

JeroenDeDauw:
Just split some code so far. Found the deletion stuff was duplicated and got rid of that. Also returned some array() and instead of empty returns, which where wrong. Should not have changed the actual behavior though. Obvious follow up work still needed to actually clean up the code, though this can be merged if it passes as far as I'm concerned.

src/SQLStore/QueryEngine/ConceptCache.php
```diff
@@ -0,0 +1,159 @@
...
+
+		$cid = $this->store->smwIds->getSMWPageID(
+			$concept->getDBkey(),
+			SMW_NS_CONCEPT,
+			'',
+			'',
+			false
+		);
+
+		$db->deleteConceptById( $cid );
```
mwjames: `$db->deleteConceptById( $cid );` -> `$this->deleteConceptById( $cid );`

includes/storage/SQLStore/SMW_SQLStore3_Queries.php
```diff
@@ -93,113 +86,39 @@ public function __construct( SMWSQLStore3 $parentstore, $dbslave ) {
...
 
-			if ( $qid < 0 ) {
-				return;
-			}
+	/**
+	 * @param string $desctxt
+	 *
+	 * @return SqlQueryPart
+	 */
+	public function doSomeStuffAndReturnSomething( $desctxt ) {
```
mwjames: doSomeStuffAndReturnSomething ??

src/SQLStore/QueryEngine/ConceptCache.php
JeroenDeDauw: Yeah I just noticed, kinda fail :)

includes/storage/SQLStore/SMW_SQLStore3_Queries.php
JeroenDeDauw: It is used. I factored this code out from ConceptCache::refresh.

You're right the name is unclear. For me to give it a more meaningful name I first need to understand what the code actually does.

mwjames: `$desctxt` contains the raw `[[Modification date::+]]` query and the code within `doSomeStuffAndReturnSomething` is to resolve the query tree and return a sql transformation which is to be used to build the final joins/from/where sql condition.

mwjames: Maybe `resolveQueryTreeForQueryCondition` (QueryCondition being the textual representation of the concept description)

JeroenDeDauw: Renamed

JeroenDeDauw:
Tests ought to pass now

mwjames:
:+1:

mwjames:
Follow-up #779

PR #784, opened 2015-02-04
Use WikitextTemplateRenderer in ListResultPrinter

mwjames:
- If `named args=yes` and the subject does not have a named parameter it can now be accessed through something like `{{{1}}}`; all parameters that have not a name will get a continued number and handled if it was `named args=no`
- `sep=;` can now be used (or any another separator) to divide multiple values of a property

### DIFF
--- a/includes/queryprinters/ListResultPrinter.php
+++ b/includes/queryprinters/ListResultPrinter.php
@@ -147,6 +147,8 @@ class ListResultPrinter extends ResultPrinter {
 			return '';
 		}
 
+		$this->templateRenderer = ApplicationFactory::getInstance()->newMwCollaboratorFactory()->newWikitextTemplateRenderer();
+
 		$this->initializePrintingParameters( $queryResult );
 
 		$result = '';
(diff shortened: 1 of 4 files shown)


mwjames:
Interesting postgres will output
- `[0]:テスト:123; 456:[$%&amp;*==42]:[1]:Bar:123; 456:[$%&amp;*==42]:[2]:Foo:123; 456:[$%&amp;*==42]:` while all others do
- `[0]:Bar:123; 456:[$%&amp;*==42]:[1]:Foo:123; 456:[$%&amp;*==42]:[2]:テスト:123; 456:[$%&amp;*==42]:`

`[0]:テスト` and `[2]:テスト` are reversed

mwjames:
This ins't a problem of this PR because we never actually tested any resultprinter output. By having made the sorting explicit `|sort=Has named argument|order=desc` (postgres now returns the same results):
- `[0]:Foo:123; 456:[$%&amp;*==42]:[1]:Bar:123; 456:[$%&amp;*==42]:[2]:テスト:123; 456:[$%&amp;*==42]:`

but now Fuseki and Virtuoso show different behaviour (Sesame returns above) in terms of the returned subjects:

- Fuseki: `[0]:テスト:123; 456:[$%&amp;*==42]:[1]:Bar:123; 456:[$%&amp;*==42]:[2]:Foo:123; 456:[$%&amp;*==42]:`
- Virtuoso: `[0]:Bar:123; 456:[$%&amp;*==42]:[1]:Foo:123; 456:[$%&amp;*==42]:[2]:テスト:123; 456:[$%&amp;*==42]:`

mwjames:
After it became clear that `postgres` needs special attention to recognize `UTF-8`:

- `...::testTemplateOutputUsingUnnamedArgumentsForNonUnicode` is run with postgres while
- `...::testTemplateOutputUsingNamedArgumentsForUnicodeIncludedSubject` excludes postgres

mwjames:
@kghbln FYI, see `named args=yes` + `sep=;` comments above

PR #788, opened 2015-02-05
Assert that Title::moveTo() returned true

legoktm:
If the move fails, it will return an array of errors, which will be useful in debugging.


mwjames:
In order to run this against the current master (1.25), travis.yml[0] needs to use `MW=master` instead of `MW=master@5cc1f1d;`. 

https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/.travis.yml

legoktm:
This will still fail against current master (working on a patch), I just wanted to submit this to make debugging in the future easier.

PR #793, opened 2015-02-10
Add suggest section with SRF and SF

JeroenDeDauw:


composer.json
```diff
@@ -47,6 +47,10 @@
 	"replace": {
 		"mediawiki/semantic-mediawiki": "*"
 	},
+	"suggest": {
```
mwjames: Why would we suggest those and not any other possible candidate it seems rather selective! I don't really mind it but SF seems a rather odd candidate given the fact that it no longer requires SMW.

JeroenDeDauw: These two extensions are the most commonly used ones alongside SMW. I've not added things such as Semantic Maps, because they are rather specific. There are ones for which I'm really not sure, such as SESP, which while general, is still not something used commonly. If you think something else should be added, go ahead.

JeroenDeDauw: Well?

mwjames: Alright ;-)

PR #806, opened 2015-02-13
Method simplification in SomePropertyCompiler

JeroenDeDauw:


mwjames:
So I guess

```
if ( $diType !== DataItem::TYPE_WIKIPAGE ) {
...
} else { // exact match (like comparator = above, but not using $valueField		
	throw new RuntimeException("Debug -- this code might be dead.");		
}
```
was never executed right?

JeroenDeDauw:
If `$diType !== DataItem::TYPE_WIKIPAGE` was false, you'd have gotten an exception. So I'm guessing the comment is right, and it never is

PR #814, opened 2015-02-13
Remove unused private method from SMWConceptValue

JeroenDeDauw:


mwjames:
Copy the code since the Exporter has a big `TODO` for the concept stuff [0] and therefore needs a follow-up to ensure it gets covered.

```php
	/**
	 * Function not called anywhere but code kept for reference before
	 * migrating it to SMWExporter.
	 */
	private function getExportData() {
		if ( $this->isValid() ) {
			$qp = new SMWQueryParser();
			$desc = $qp->getQueryDescription( str_replace( array( '&lt;', '&gt;', '&amp;' ), array( '<', '>', '&' ), $this->m_dataitem->getConceptQuery() ) );
			$exact = true;
			$owldesc = $this->descriptionToExpData( $desc, $exact );
			if ( !$exact ) {
				$result = new SMWExpData( new SMWExpResource( '' ) );
				$result->addPropertyObjectValue( SMWExporter::getSpecialNsResource( 'rdf', 'type' ),
				                                new SMWExpData( SMWExporter::getSpecialNsResource( 'owl', 'Class' ) ) );
				$result->addPropertyObjectValue( SMWExporter::getSpecialNsResource( 'rdfs', 'subClassOf' ), $owldesc );
				return $result;
			} else {
				return $owldesc;
			}
		} else {
			return null;
		}
	}
```
[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/export/SMW_Exporter.php#L698

PR #819, opened 2015-02-14
Improvements to DataRebuilder

JeroenDeDauw:


JeroenDeDauw:
@mwjames I do not see any behaviour change... yet there must be one since the test fails now. Any idea?

mwjames:
> yet there must be one since the test fails now. Any idea?

I haven't looked at it but according to the error a mock was expected to run but wasn't reached.

```
1) SMW\Tests\Maintenance\DataRebuilderTest::testRebuildSelectedPagesWithPageOption
Expectation failed for method name is equal to <string:makeTitleOf> when invoked at sequence index 0.
The expected invocation at index 0 was never reached.
```

JeroenDeDauw:
Yeah I looked at the test failure already

mwjames:
> Yeah I looked at the test failure already

`makeTitleOf` was made private hence the mock failed. `TitleCreator` is now injected which allows for proper mocking during `testRebuildSelectedPagesWithPageOption`.

Issue #347, opened 2014-06-24
MW 1.23 / User-defined properties disappearing from pages


jthingelstad:
On WikiApiary I’m seeing a large number of pages losing properties for no reason. This bug was originally highlighted at thingles/WikiApiary#114 and has persisted now through a few iterations of master.

I've been seeing an increase in the number of data errors that bots are getting because properties that should be there are not. Looking at [properties for WikiApiary](https://wikiapiary.com/wiki/Special:Browse/WikiApiary) for example shows:

![screenshot](https://cloud.githubusercontent.com/assets/5351/3372834/930bfb94-fbae-11e3-853f-d48f68355f3c.gif)

That should have an "Is active" and "Is audited" property. They are in the template and in the page. If I manually force a "refresh" of the page the properties will appear again.

![screenshot2](https://cloud.githubusercontent.com/assets/5351/3372881/d1682304-fbae-11e3-8d72-fd921d97a58f.gif)

I’m seeing this a lot. Various tasks executing from cron or routinely throwing exceptions because mandatory properties have gone missing.

PR #820, opened 2015-02-14
Introduce SemanticDataCache

mwjames:
Parsing pages is an expensive endeavour and therefore liable to the amount of time spent on each page during an update (re-parse).

Using the `SemanticDataCache` can reduce the burden and processing times significantly by making a decision on whether a page needs a "real" update (revId to identify the content status) or it is merely enough to re-fetch the data from cache and provide the input.

The premis is that if the revision hasn't changed then the `SemanticData` haven't changed either hence the data from cache can be used.

- If `rebuildData.php` is run with option `--no-cache` then no `SemanticDataCache` is used.
- In case the de-serialization returns with an error a re-parse of a page is automatically triggered during the `UpdateJob` run.
- Currently `smwgCacheType` is used to identify the cache type for building the `CompositeCache` for the `SemanticDataCache` object.

Sample using `rebuildData.php -s 1 -e 40 --runtime`:
- without-cache: Memory used: 24309648 (b: 8997048, a: 33306696) with a runtime of 82.89 sec (1.38 min)
- with-cache: Memory used: 14102776 (b: 8997048, a: 23099824) with a runtime of 36.71 sec

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -557,6 +557,9 @@ $GLOBALS['smwgCacheType'] = CACHE_ANYTHING;
 # - smwgStatisticsCache Enable to serve statistics from cache
 # - smwgStatisticsCacheExpiry Number of seconds before the cache expires
 #
+# - smwgSemanticDataCache to improve internal read/write access to SemanticData during
+# an update process.
+#
 # @since 1.9
 ##
 $GLOBALS['smwgCacheUsage'] = array(
@@ -568,6 +571,7 @@ $GLOBALS['smwgCacheUsage'] = array(
 	'smwgPropertiesCacheExpiry' => 3600,
 	'smwgStatisticsCache' => true,
 	'smwgStatisticsCacheExpiry' => 3600,
+	'smwgSemanticDataCache' => true
 );
 
 ###
--- a/composer.json
+++ b/composer.json
@@ -42,6 +42,7 @@
 		"mediawiki/validator": "~2.0,>=2.0.4",
 		"serialization/serialization": "~3.2",
 		"onoi/message-reporter": "~1.0",
+		"onoi/cache": "~1.0",
 		"doctrine/dbal": "~2.5"
 	},
 	"replace": {
--- a/includes/SemanticData.php
+++ b/includes/SemanticData.php
@@ -551,6 +551,24 @@ class SemanticData {
 	}
 
 	/**
+	 * @since 2.2
+	 *
+	 * @param DIProperty $property
+	 *
+	 * @return boolean
+	 */
+	public function hasProperty( DIProperty $property ) {
+
+		foreach ( $this->getProperties() as $semanticDataProperty ) {
+			if ( $semanticDataProperty->equals( $property ) ) {
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
 	 * Whether the SemanticData has a SubSemanticData container and if
 	 * specified has a particular subobject using its name as identifier
 	 *
--- a/includes/Setup.php
+++ b/includes/Setup.php
@@ -81,10 +81,19 @@ final class Setup {
 	}
 
 	private function registerSettings() {
+
 		$this->applicationFactory->registerObject(
 			'Settings',
 			Settings::newFromGlobals( $this->globalVars )
 		);
+
+		$cacheFactory = $this->applicationFactory->newCacheFactory();
+
+		$compositeCache = $cacheFactory->newMediaWikiCompositeCache(
+			$GLOBALS['smwgCacheType']
+		);
+
+		$cacheFactory->getSemanticDataCache()->setCache( $compositeCache );
 	}
 
 	private function registerConnectionProviders() {
--- a/maintenance/rebuildData.php
+++ b/maintenance/rebuildData.php
@@ -5,8 +5,8 @@ namespace SMW\Maintenance;
 use SMW\Maintenance\DataRebuilder;
 use SMW\Maintenance\MaintenanceHelper;
 use Onoi\MessageReporter\MessageReporterFactory;
+use SMW\ApplicationFactory;
 use SMW\StoreFactory;
-use SMW\Settings;
 
 $basePath = getenv( 'MW_INSTALL_PATH' ) !== false ? getenv( 'MW_INSTALL_PATH' ) : __DIR__ . '/../../..';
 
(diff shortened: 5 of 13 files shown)


mwjames:
Before the PR:

```
==========================================================================================
JobQueue benchmarks
------------------------------------------------------------------------------------------
- Dataset: BaseLoremIpsumDataset.v1.xml
- MediaWiki: 1.25alpha
- Store: SMWSQLStore3
- ShowMemoryUsage: false
- ReuseDatasets: true
- PageCopyThreshold: 1000
- RepetitionExecutionThreshold: 1
------------------------------------------------------------------------------------------
- SMW\RefreshJob: 1.2924681 (n) 1.2924681 (mean) 1.2924681 (total) (sec)
- SMW\UpdateJob: 0.0270573 (n) 27.0573311 (mean) 27.0573311 (total) (sec)
==========================================================================================

SMW\Tests\Benchmark\JobQueueBenchmarkTest doBenchmark ran for 28.413 seconds
.
==========================================================================================
RebuildData benchmarks
------------------------------------------------------------------------------------------
- Dataset: BaseLoremIpsumDataset.v1.xml
- MediaWiki: 1.25alpha
- Store: SMWSQLStore3
- ShowMemoryUsage: false
- ReuseDatasets: true
- FullDelete: true
- PageCopyThreshold: 1000
- RepetitionExecutionThreshold: 1
------------------------------------------------------------------------------------------
- SMW\Maintenance\RebuildData: 0.0427976 (n) 42.797627 (mean) 42.797627 (total) (sec)
==========================================================================================
```

After the PR:
```
==========================================================================================
JobQueue benchmarks
------------------------------------------------------------------------------------------
- Dataset: BaseLoremIpsumDataset.v1.xml
- MediaWiki: 1.25alpha
- Store: SMWSQLStore3
- ShowMemoryUsage: false
- ReuseDatasets: true
- PageCopyThreshold: 1000
- RepetitionExecutionThreshold: 1
------------------------------------------------------------------------------------------
- SMW\RefreshJob: 1.007448 (n) 1.007448 (mean) 1.007448 (total) (sec)
- SMW\UpdateJob: 0.0145479 (n) 14.5478842 (mean) 14.5478842 (total) (sec)
==========================================================================================

SMW\Tests\Benchmark\JobQueueBenchmarkTest doBenchmark ran for 15.604 seconds
.
==========================================================================================
RebuildData benchmarks
------------------------------------------------------------------------------------------
- Dataset: BaseLoremIpsumDataset.v1.xml
- MediaWiki: 1.25alpha
- Store: SMWSQLStore3
- ShowMemoryUsage: false
- ReuseDatasets: true
- FullDelete: true
- PageCopyThreshold: 1000
- RepetitionExecutionThreshold: 1
------------------------------------------------------------------------------------------
- SMW\Maintenance\RebuildData: 0.024953 (n) 24.952961 (mean) 24.952961 (total) (sec)
==========================================================================================
```

joelkp:
> The premis is that if the revision hasn't changed then the SemanticData haven't changed either hence the data from cache can be used.

I see one problem with this for a certain kind of SMW extension, though I think there's also a simple solution.

Semantic Dependency is used for specific pages where (unlike other pages) semantic data does depend on other pages. The way it works is that update jobs are created for the dependent pages when the pages they depend on are updated.

I think such functionality could work together with the cache feature, if it were possible to disable reading from the cache for specific update jobs. The jobs created by the extension could then use this, and the cache would be updated with the new data, and could still be used to improve overall performance.

JeroenDeDauw:
Ah. @joelkp beat me to it. I was about to point out the same thing. Some pages might indeed have values that depend on something not part of the page, be it semantic data in another page, or something else entirely. Not recomputing can cause unexpected behaviour.

I'm thinking it's somewhat unfortunate that such dependencies are supported on all pages by default. If users had to explicitly allow it per page, then we could do several things more efficiently.

mwjames:
> I see one problem with this for a certain kind of SMW extension, though I think there's also a simple solution.

I currently can't see how an extension is involved here, `SemanticDataCache` is for internal use and in case an extension runs a SMW specific hook it will do so for the `UpdateJob` no matter of the source of the data (either being parsed or retrieved from the `SemanticDataCache`).

`SemanticDataCache` is similar to the `ParserCache` to avoid having the  `ContentParser` do unnecessary parsing. Neither `ContentParser` nor the `InternalParseBeforeLinks` hook contains any auxiliary hook for extensions to modify data during the in-text annotation parsing which also means that no extension can modify content during the parse process.

Data added through a hook is done after the parsing and is being run in any event independent of the source of the data.

The main objective for `SemanticDataCache` is to mitigate #347 in `LinksUpdateConstructed` and improve memory/runtime for the `UpdateJob`. ( `SemanticDataCache !== QueryCache !== ParserCache` )

By default `$GLOBALS['smwgCacheUsage']['smwgSemanticDataCache']` is set true.

Possible ways to intervene with the  `SemanticDataCache` usage:
- Set `ApplicationFactory::getInstance()->getSettings()->set( 'smwgSemanticDataCache', false );` (as done for `rebuildData.php` with option `--no-cache`)
- Set `SemanticData::setUpdateIdentifier` to an extension dependent identifier (by default it contains the `getLatestRevID`) 

> Semantic Dependency is used for specific pages where (unlike other pages) semantic data does depend on other pages.

I don't know Semantic Dependency therefore I can't comment specifically here.

> if it were possible to disable reading from the cache for specific update jobs. The jobs created by the extension could then use this, 

The `UpdateJob` requires `'smwgSemanticDataCache'` to be `true` in order to evaluated whether the cache can be used or not. I don't know if we need an extra `UpdateJob` parameter `--no-cache` because currently the `UpdateJob` does not allow for extra parameters to be invoked (legacy interface) and adding it on-the-fly could make it difficult for existing jobs (those already in the `JobQueue`) to recognize the added interface parameter (well, if it comes to adding `parameters` then it needs testing).

I would certainly appreciate extensive testing to identify possible yet uncovered dependencies.

JeroenDeDauw:
> I don't know Semantic Dependency therefore I can't comment specifically here.

You can easily create this problem with SMW itself: assign the result of a query to a property value

mwjames:
> You can easily create this problem with SMW itself: assign the result of a query to a property value

Query usage tracking is a whole other thing which needs addressing in order to be able to disable the `ParserCache` of a subject to ensure query results are updated but also the allow for `SemanticDataCache` to be disabled.

JeroenDeDauw:
Yeah sure. The issue is already there to some extend, yet this new cache might well make it worse and thus break things currently working.

Don't misunderstand me: I like the stuff being done here. If data rebuilding indeed gets twice as fast, that's awesome

mwjames:
Instead of the `LinksUpdateConstructed` hook as trigger, the `SMWStore::updateDataAfter` is going to be used to ensure that all available data after processing are to represent the cache item for the selected revision.

JeroenDeDauw:
@mkroetzsch can you think of something we're forgetting about that will cause problems if this change is made?

mwjames:
Superseded by #1127 and #1035.

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help:Maintenance_script_%22rebuildData.php%22&curid=5500&diff=57603&oldid=54745)

PR #821, opened 2015-02-14
Add PropertyTableInfoFetcher

mwjames:
Move all property table/definition methods into a separate isolated class

### DIFF
--- a/src/SQLStore/PropertyTableInfoFetcher.php
+++ b/src/SQLStore/PropertyTableInfoFetcher.php
@@ -19,9 +19,9 @@ class PropertyTableInfoFetcher {
 	 * Array for keeping property table table data, indexed by table id.
 	 * Access this only by calling getPropertyTables().
 	 *
-	 * @var TableDefinition[]
+	 * @var TableDefinition[]|null
 	 */
-	private static $propertyTableDefinitions = null;
+	private $propertyTableDefinitions = null;
 
 	/**
 	 * Array to cache "propkey => table id" associations for fixed property
@@ -30,7 +30,7 @@ class PropertyTableInfoFetcher {
 	 *
 	 * @var array|null
 	 */
-	private static $fixedPropertyTableIds = null;
+	private $fixedPropertyTableIds = null;
 
 	/**
 	 * Keys of special properties that should have their own
@@ -38,14 +38,14 @@ class PropertyTableInfoFetcher {
 	 *
 	 * @var array
 	 */
-	private static $customizableSpecialProperties = array(
+	private $customizableSpecialProperties = array(
 		'_MDAT', '_CDAT', '_NEWP', '_LEDT', '_MIME', '_MEDIA',
 	);
 
 	/**
 	 * @var array
 	 */
-	private static $fixedSpecialProperties = array(
+	private $fixedSpecialProperties = array(
 		// property declarations
 		'_TYPE', '_UNIT', '_CONV', '_PVAL', '_LIST', '_SERV',
 		// query statistics (very frequently used)
@@ -67,7 +67,7 @@ class PropertyTableInfoFetcher {
 	 *
 	 * @var array
 	 */
-	private static $defaultDiTypeTableIdMap = array(
+	private $defaultDiTypeTableIdMap = array(
 		DataItem::TYPE_NUMBER     => 'smw_di_number',
 		DataItem::TYPE_BLOB       => 'smw_di_blob',
 		DataItem::TYPE_BOOLEAN    => 'smw_di_bool',
@@ -108,8 +108,8 @@ class PropertyTableInfoFetcher {
 	 */
 	public function findTableIdForDataItemTypeId( $dataItemId ) {
 
-		if ( array_key_exists( $dataItemId, self::$defaultDiTypeTableIdMap ) ) {
-			return self::$defaultDiTypeTableIdMap[$dataItemId];
+		if ( array_key_exists( $dataItemId, $this->defaultDiTypeTableIdMap ) ) {
+			return $this->defaultDiTypeTableIdMap[$dataItemId];
 		}
 
 		return '';
@@ -127,14 +127,14 @@ class PropertyTableInfoFetcher {
 	 */
 	public function findTableIdForProperty( DIProperty $property ) {
 
-		if ( self::$fixedPropertyTableIds === null ) {
+		if ( $this->fixedPropertyTableIds === null ) {
 			$this->getPropertyTableDefinitions();
 		}
 
 		$propertyKey = $property->getKey();
 
-		if ( array_key_exists( $propertyKey, self::$fixedPropertyTableIds ) ) {
-			return self::$fixedPropertyTableIds[$propertyKey];
+		if ( array_key_exists( $propertyKey, $this->fixedPropertyTableIds ) ) {
+			return $this->fixedPropertyTableIds[$propertyKey];
 		}
 
 		return $this->findTableIdForDataTypeTypeId( $property->findPropertyTypeID() );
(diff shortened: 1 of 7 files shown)


src/SQLStore/PropertyTableInfoFetcher.php
```diff
@@ -0,0 +1,197 @@
...
+ */
+class PropertyTableInfoFetcher {
+
+	/**
+	 * Array for keeping property table table data, indexed by table id.
+	 * Access this only by calling getPropertyTables().
+	 *
+	 * @var TableDefinition[]
+	 */
+	private static $propertyTableDefinitions = null;
```
JeroenDeDauw: I think it's harmful to make such a cache global.

src/SQLStore/PropertyTableInfoFetcher.php
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * @author mwjames
+ */
+class PropertyTableInfoFetcher {
+
+	/**
+	 * Array for keeping property table table data, indexed by table id.
+	 * Access this only by calling getPropertyTables().
+	 *
+	 * @var TableDefinition[]
```
JeroenDeDauw: Type hint is wrong, it can also be null

src/SQLStore/PropertyTableInfoFetcher.php
mwjames: We need the cache here because of performance, this information is accessed several hundreds times per request and the current way the` SQLStore` is using the information can't be circumvented.

> I think it's harmful to make such a cache global.

I don't want the `SQLStore` to make this sort of caching therefore `PropertyTableInfoFetcher` is the solely keeper.

First step is to isolate and keep status quo but allow for a possibility adjustments at a later stage.

src/SQLStore/PropertyTableInfoFetcher.php
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		$settings = ApplicationFactory::getInstance()->getSettings();
+
+		$enabledSpecialProperties = self::$fixedSpecialProperties;
+		$customizableSpecialProperties = array_flip( self::$customizableSpecialProperties );
+
+		$customFixedProperties = $settings->get( 'smwgFixedProperties' );
+		$customSpecialProperties = $settings->get( 'smwgPageSpecialProperties' );
+
```
JeroenDeDauw: I'd try to get rid of the global access here and demand those two settings in the constructor when you are splitting this stuff. Also makes it easier to test for these things.

src/SQLStore/PropertyTableInfoFetcher.php
JeroenDeDauw: Is it not feasible to have an instance cache and than just make sure the same instance of the service is used?

src/SQLStore/PropertyTableInfoFetcher.php
mwjames: I'm now fan for those types to be used in the constructor, more likely doing a `setter` instead (can be empty).

src/SQLStore/PropertyTableInfoFetcher.php
mwjames: Can be done but for this I have to look at Xdebug to ensure we don't start making holes.

src/SQLStore/PropertyTableInfoFetcher.php
JeroenDeDauw: If it's something the class needs to function, then the constructor is the appropriate place. If you do it via a setter, you end up with code that breaks if some method did not get called before, which is something to avoid, esp across class boundaries/

mwjames:
proptableinfofetcher

src/SQLStore/PropertyTableInfoFetcher.php
```diff
@@ -0,0 +1,219 @@
...
+		DataItem::TYPE_WIKIPAGE   => 'smw_di_wikipage',
+		//DataItem::TYPE_CONCEPT    => '', // _CONC is the only property of this type
+	);
+
+	/**
+	 * @since 2.2
+	 *
+	 * @param array $customFixedProperties
+	 */
+	public function setCustomFixedPropertyList( array $customFixedProperties ) {
```
JeroenDeDauw: I really think this ought to be in the constructor. What makes you intent on using setters?

The one annoying thing I can think of is that it's less readable, since you get positional unnamed parameters. (Would be so much nicer if PHP had named parameters...) That's not nice, though not enough reason to give up on constructors altogether

src/SQLStore/PropertyTableInfoFetcher.php
```diff
@@ -0,0 +1,219 @@
...
+			DataTypeRegistry::getInstance()->getDataItemId( $dataTypeTypeId )
+		);
+	}
+
+	/**
+	 * Find the id of a property table that is normally used to store
+	 * data items of the given type. The empty string is returned if
+	 * no such table exists.
+	 *
+	 * @since 1.8
```
JeroenDeDauw: This method, `PropertyTableInfoFetcher::findTableIdForDataItemTypeId`, only exists since this commit

src/SQLStore/PropertyTableInfoFetcher.php
```diff
@@ -0,0 +1,219 @@
...
+	 *
+	 * It is ensured that the keys of the returned array agree with the name of
+	 * the table that they refer to.
+	 *
+	 * @since 2.2
+	 *
+	 * @return TableDefinition[]
+	 */
+	public function getPropertyTableDefinitions() {
+
```
JeroenDeDauw: You can do 

```php
if ( $this->propertyTableDefinitions === null ) {
    $this->buildCache();
}

return $this->propertyTableDefinitions;
```

Then you don't mix the caching part in with regular access, and avoid repeating the return of the field.

src/SQLStore/PropertyTableInfoFetcher.php
```diff
@@ -0,0 +1,219 @@
...
+		$this->propertyTableDefinitions = $definitionBuilder->getTableDefinitions();
+		$this->fixedPropertyTableIds = $definitionBuilder->getFixedPropertyTableIds();
+
+		return $this->propertyTableDefinitions;
+	}
+
+	/**
+	 * @since 2.2
+	 */
+	public function clear() {
```
JeroenDeDauw: `clearCache` is probably better. I'm thinking that having such a method in the interface is not ideal to begin with though. It lets the user know you have a cache. What if you want an implementation that does not have a cache? Not saying this should be fixed, just noting I find it suspicious

JeroenDeDauw:
Yay \o/

src/SQLStore/PropertyTableInfoFetcher.php
mwjames: > I really think this ought to be in the constructor. What makes you intent on using setters?

Because those custom lists are optional and not crucial in defining the behaviour of this class of course if injected it will alter the amount of table definitions (provided by `getPropertyTableDefinitions`) but even without any of those lists `getPropertyTableDefinitions` would return valid results.

JeroenDeDauw: Ok, then I probably misunderstand their role

mwjames: In this case I renamed them to something like `setEnabledCustomSpecialPropertyList`

JeroenDeDauw: I understand `setCustomFixedPropertyList` but not `setEnabledCustomSpecialPropertyList`

The "set enabled" part is odd and ambiguous phrasing

I have no problem with the original name btw, and am not saying it's unclear.

PR #823, opened 2015-02-15
Remove dead code from SMWImportValue

JeroenDeDauw:


mwjames:
Is this code really not being used? (This class has no coverage and isn't checked by any integration test)

JeroenDeDauw:
Nothing uses `$elemtype`, and that is the only thing computed in the loop

PR #825, opened 2015-02-15
Small tweaks in Highlighter

JeroenDeDauw:


mwjames:
Just as a note the Highlighter class and the js/tooltip stuff are a candidate to be separated from SMW-core and loaded as its ow component.

JeroenDeDauw:
Sure, no reason to not clean it up though

PR #826, opened 2015-02-15
Tweaks in SimplePropertyStatisticsRebuilder

JeroenDeDauw:


includes/storage/SQLStore/SimplePropertyStatisticsRebuilder.php
```diff
@@ -41,11 +40,7 @@ class SimplePropertyStatisticsRebuilder implements \SMW\Store\PropertyStatistics
 	 */
 	public function __construct( Store $store, MessageReporter $reporter = null ) {
 		$this->store = $store;
-		$this->reporter = $reporter;
-
-		if ( $this->reporter === null ) {
-			$this->reporter = MessageReporterFactory::getInstance()->newNullMessageReporter();
-		}
+		$this->reporter = $reporter !== null ?: MessageReporterFactory::getInstance()->newNullMessageReporter();
```
JeroenDeDauw: Oops. Like this it assigns the field to itself when not $reporter is not null, instead of to $reporter. Fail

mwjames:
Some clean-up went into f6d8b46c2dd.

PR #829, opened 2015-02-15
Add `--runtime` to `rebuildConceptCache`

mwjames:


maintenance/rebuildConceptCache.php
```diff
@@ -100,6 +99,7 @@ protected function addDefaultParams() {
 		$this->addOption( 's', '<startid> Process only concepts with page id of at least <startid>', false, true );
 		$this->addOption( 'e', '<endid> Process only concepts with page id of at most <endid>', false, true );
 
+		$this->addOption( 'runtime', 'Will display the runtime environment of the script', false );
```
JeroenDeDauw: > runtime environment

This makes me think it's about the PHP version or something like that. Improved in #830

JeroenDeDauw: The option name itself is still ambiguous though. http://en.wikipedia.org/wiki/Run_time

`--display-runtime` is probably better, and this also makes it clear it's an option, and not something expecting a value

Then again, it is not just about the execution time. `--display-resource-consumption` is more accurate, though kinda long

kghbln:
This is now documented on wiki.

Issue #732, opened 2015-01-20
CacheableResultCollector::findPropertyTableByType uses undefined field


JeroenDeDauw:
It accessed a $store field, which is not defined in the class itself. It implicitly relies on the deriving classes to define one.

mwjames:
> It accessed a $store field, which is not defined in the class itself. It implicitly relies on the deriving classes to define one.

I know it is a rather unfortunate design and I'd like to get rid of it sooner rather than later.

mwjames:
> I know it is a rather unfortunate design and I'd like to get rid of it sooner rather than later.

#831 will start to remove dependency on `CacheableResultCollector` and in its final process will remove this class.

mwjames:
RIP

PR #831, opened 2015-02-17
Add SimpleListLookup interface + ListLookupCache

mwjames:
First part to remove `CacheableResultCollector` and the inheritance abuse to those connected classes while making unit tests easier and splitting responsibilities. Composition is favoured over inheritance, isolating the cache layer from the actual lookup operations and having all classes involved share the the same `SimpleListLookup` interface.

refs #732


src/Cache/CacheFactory.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+class CacheFactory {
+
+	/**
+	 * @since 2.2
+	 *
+	 * @return integer
+	 */
+	public function getMainCacheType() {
+		return ApplicationFactory::getInstance()->getSettings()->get( 'smwgCacheType' );
```
JeroenDeDauw: Circular dependency!

cat-chasing-tail

src/Cache/CacheFactory.php
```diff
@@ -0,0 +1,67 @@
...
+
+	/**
+	 * @since 2.2
+	 *
+	 * @param integer $cacheSize
+	 *
+	 * @return Cache
+	 */
+	public function newFixedInMemoryCache( $cacheSize = 500 ) {
+		return OnoiCacheFactory::getInstance()->newFixedInMemoryCache( $cacheSize );
```
JeroenDeDauw: How can this return an `SMW\Cache`?

src/SQLStore/SimpleListLookup.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A basic interface for fetching a simple list either from a DB or cache
+ * instance
+ *
+ * @license GNU GPL v2+
+ * @since 2.2
+ *
+ * @author mwjames
+ */
+interface SimpleListLookup {
```
JeroenDeDauw: Surprise: I find this interface a bit strange. (As I always do right? :))

It has a single method inherently relevant to list lookup stuff. The other 3 make sense in caching context. I do not think they belong to the base interface. For what reasons did you put them there?

One reason could be that `ListLookupCache` is a decorator, and that you want other caching implementations of the base interface to work correctly inside there. I'm not sure why that'd be needed though, esp since an alternate cache implementation would be odd due to the current usage of a caching strategy. 

How about dropping the caching related methods from `SimpleListLookup`, and moving that responsibility into `ListLookupCache`? This thing itself can get the timestamp, and it can create or require (in the constructor) an unique identifier.

mwjames: > It has a single method inherently relevant to list lookup stuff. The other 3 make sense in caching context. I do not think they belong to the base interface. For what reasons did you put them there?

I thought about it and the way I envisioned this was that `SimpleListLookup` contains all methods that are necessary the communicate with a `SimpleListLookup` instance whether it is a concrete class that does the lookup or a caching layer (such as `ListLookupCache`)  but one is indifferent to the concrete implementation detail.

Knowing that ` ... instance of SimpleListLookup` you have a contract to four methods without the need for a consumer to know that is coming from cache or a DB lookup (well you know it indirectly after the execution and not before by checking on `isCached`).

If `SimpleListLookup` has only the `fetchResultList` method then one has always to check that it is `SimpleListLookup` or a concrete `ListLookupCache` instance because `isCached` is`getTimestamp` is only provided by `ListLookupCache`.

src/Cache/CacheFactory.php
mwjames: `SMW\Cache\Cache` and `SMW\Cache\FixedInMemoryCache` will be removed after the migration is done (`SMW\Cache\Cache` == `Onoi\Cache\Cache`)

JeroenDeDauw: Well sure. Bad to use incorrect type hints now though.

JeroenDeDauw:
You didn't reply to my best comment!

sad

src/SQLStore/SimpleListLookup.php
JeroenDeDauw: > If SimpleListLookup has only the fetchResultList method then one has always to check that it is SimpleListLookup or a concrete ListLookupCache instance because isCached is getTimestamp is only provided by ListLookupCache.

What makes it so that users of this thing inherently need to be able to interact with the cache stuff? If you can avoid this, then what I suggest works. And I do not see why you'd not be able to avoid it.

mwjames: > What makes it so that users of this thing inherently need to be able to interact with the cache stuff?

No, the way `SimpleListLookup` is specified s(he) never has to know whether it is coming from a cache or not that's why `SimpleListLookup` contains all necessary methods so a consumer such as [0] only knows it is `SimpleListLookup`.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/311dba9f34b2f080282095d1bac0dd8185edad91#diff-af75f55a8f874a95bcdbcdd5d07da2c7L52

PR #834, opened 2015-02-20
Use travis_retry for composer commands

JeroenDeDauw:


JeroenDeDauw:
Huh - guess bash cannot access it because it's being called from a different file

Issue #764, opened 2015-01-29
SQLStore / sqlite, postgres RAND() is an unknown function


mwjames:
SMW 2.2-alpha/sqlite doing something like:
```
{{#ask:[[Page content language::+]]
|?Page content language
|format=broadtable
|limit=1
|link=all
|order=random
|headers=show
|mainlabel=-
|searchlabel=… その他の結果
|class=sortable wikitable smwtable
}}
```
will yield
```
Unexpected non-MediaWiki exception encountered, of type "RuntimeException"
[fd2cf2a3] /sandbox-import/index.php?title=%E7%89%B9%E5%88%A5%3A%E5%95%8F%E3%81%84%E5%90%88%E3%82%8F%E3%81%9B&q=%5B%5BPage+content+language%3A%3A%2B%5D%5D&po=%3FPage+content+language%0D%0A&eq=yes&p%5Bformat%5D=broadtable&sort_num=&order_num=ASC&p%5Blimit%5D=1&p%5Boffset%5D=&p%5Blink%5D=all&p%5Bsort%5D=&p%5Border%5D%5Brandom%5D=1&p%5Bheaders%5D=show&p%5Bmainlabel%5D=-&p%5Bintro%5D=&p%5Boutro%5D=&p%5Bsearchlabel%5D=%E2%80%A6+%E3%81%9D%E3%81%AE%E4%BB%96%E3%81%AE%E7%B5%90%E6%9E%9C&p%5Bdefault%5D=&p%5Bclass%5D=sortable+wikitable+smwtable&eq=yes Exception from line 176 of C:\xampp\htdocs\sandbox-import\extensions\SemanticMediaWiki\includes\src\MediaWiki\Database.php: A database error has occurred. Did you forget to run maintenance/update.php after upgrading? See: https://www.mediawiki.org/wiki/Manual:Upgrading#Run_the_update_script
Query: SELECT DISTINCT t2.smw_id AS id,t2.smw_title AS t,t2.smw_namespace AS ns,t2.smw_iw AS iw,t2.smw_subobject AS so,t2.smw_sortkey AS sortkey FROM smw_object_ids AS t2 INNER JOIN smw_di_blob AS t0 ON t2.smw_id=t0.s_id WHERE (t0.p_id='262') ORDER BY RAND() LIMIT 6 
Function: SMW::getQueryResult
Error: 1 no such function: RAND

[...]

Issue #779, opened 2015-02-01
SQLStore / Improve ConceptCache test coverage


mwjames:
refs #777

> Obvious follow up work still needed to actually clean up the code, though this can be merged if it passes as far as I'm concerned.

JeroenDeDauw:
Well, I already did the follow up work in that PR now...

JeroenDeDauw:
I guess you can improve how the ConceptCache gets constructed and deprecate or remove the methods from SQLStore and QueryEngine. Though that is not specific to ConceptCache, and esp the construction stuff is a wider issue affecting different parts of the SQLStore.

mwjames:
`ConceptCache` is not yet directly tested (indirectly thought the integration tests)

JeroenDeDauw:
There is no new behavior. If there where unit tests for the store that called the old methods, you could refactor these as well to use the new class.

Obviously it'd still be good to add nice unit tests, to the extend this is possible. (Thinking about it, I suspect integration tests are more cost effective for this code in its current form.)

JeroenDeDauw:
Do you want a ticket to track that?

mwjames:
Yea, because [ConceptQueryDBIntegrationTest] (https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/Query/ConceptQueryDBIntegrationTest.php) currently does not run for postgres (this is the only one as far as I remember).

JeroenDeDauw:
"Improve ConceptCache" is not a good title or description for that though

mwjames:
One issue is the missing temp table/column generation for `postgres` in `ConceptCache`:

```
php maintenance/rebuildConceptCache.php --create --debug

Notice: pg_send_query(): There are results on this connection. Call pg_get_result() until it returns FALSE in ...\includes\db\DatabasePostgres.php on line 490
```

```
Unexpected non-MediaWiki exception encountered, of type "RuntimeException"
[df84578f] [no req]   Exception from line 231 of ..\src\MediaWiki\Database.php: A database error has occurred. Did you forget to run maintenance/update.php after upgrading?  See
: https://www.mediawiki.org/wiki/Manual:Upgrading#Run_the_update_script
Query: INSERT INTO smw_concept_cache SELECT DISTINCT t0.s_id AS s_id, 54 AS o_id FROM "smw_fpt_mdat" AS t0 WHERE NO
[...]

PR #835, opened 2015-02-20
Fix RAND query for sqlite, refs #764

mwjames:
refs #764

### DIFF
--- a/includes/storage/SQLStore/SMW_SQLStore3_Queries.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3_Queries.php
@@ -770,6 +770,13 @@ class SMWSQLStore3QueryEngine {
 					throw new RuntimeException( "Expected a string value as sortkey" );
 				}
 
+				// #835
+				// SELECT DISTINCT and ORDER BY RANDOM causes an issue for postgres
+				// Disable RANDOM support for postgres
+				if ($this->store->getConnection()->getType() === 'postgres' ) {
+					$smwgQRandSortingSupport = false;
+				}
+
 				if ( ( $order != 'RANDOM' ) && array_key_exists( $propkey, $qobj->sortfields ) ) { // Field was successfully added.
 					$result['ORDER BY'] = ( array_key_exists( 'ORDER BY', $result ) ? $result['ORDER BY'] . ', ' : '' ) . $qobj->sortfields[$propkey] . " $order ";
 				} elseif ( ( $order == 'RANDOM' ) && $smwgQRandSortingSupport ) {
--- a/src/MediaWiki/Database.php
+++ b/src/MediaWiki/Database.php
@@ -161,6 +161,16 @@ class Database {
 		// consistently ensure that the select doesn't add an extra prefix
 		if ( $this->getType() === 'sqlite' ) {
 			$tablePrefix = $this->readConnection()->tablePrefix( '' );
+
+			if ( isset( $options['ORDER BY'] ) ) {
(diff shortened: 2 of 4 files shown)


mwjames:
It turns out not only sqlite has an issue with `RAND` but also postgres.

```
SMW\Tests\Integration\Query\RandomQueryResultOrderIntegrationTest::testRandomOrder
pg_query(): Query failed: ERROR:  function rand() does not exist
LINE 1: ....smw_id=t0.s_id  WHERE (t0.p_id='134')  ORDER BY  RAND()  LI...
                                                             ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

mwjames:
Next issue for `postgres` is `  for SELECT DISTINCT, ORDER BY expressions must appear in select list`:

```
1) SMW\Tests\Integration\Query\RandomQueryResultOrderIntegrationTest::testRandomOrder
pg_query(): Query failed: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list
LINE 1: ....smw_id=t0.s_id  WHERE (t0.p_id='134')  ORDER BY  RANDOM()  ...
```

http://stackoverflow.com/questions/11401229/how-to-use-select-distinct-with-random-function-in-postgresql

mwjames:
@JeroenDeDauw Using `RANDOM` with postgres is a bit more complicated than switching the function, apparently `SELECT DISTINCT` and `RANDOM` do not work together [0] in ways specified by `sqlite` or `mysql`. I propose we do not support `RANDOM` (function just being set `` to avoid any failure) on postgres otherwise you would have to re-write [1].

[0] http://www.postgresql.org/message-id/27009.1171559417@sss.pgh.pa.us

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/storage/SQLStore/SMW_SQLStore3_Queries.php#L385

JeroenDeDauw:
Two questions:

1. what is this used for anyway? sort=random in ask queries?

2. have you looked at DBAL to see if it provides a usable abstraction?

mwjames:
> what is this used for anyway? sort=random in ask queries?

Yes (order=random) and it would create a failure for `sqlite` and `postgres` as seen in #764.

> have you looked at DBAL to see if it provides a usable abstraction?

No

JeroenDeDauw:
So far for our complete SQLite and Postgres support :)

JeroenDeDauw:
watz

mwjames:
> So far for our complete SQLite and Postgres support :)

SQLite is fine but postgres is a pain (similar to the  `Concept` https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/779#issuecomment-73592464 issue).

JeroenDeDauw:
I can't find anything in the DBAL docs. Asking on IRC now

mwjames:
The postgres issues hasn't been solved satisfactory but to avoid any failures the current approach is to disable random support for postgres.

JeroenDeDauw:
> JeroenDeDauw | Does DBAL have a suitable abstraction that I can use?

> zeroedin-bill | afaik no, doctrine doesnt have a convenient way to do that, as it varies between vendors a lot

yoonghm:
MySQL allowing **non-select column to be used for sorting provided** only **safe** or **correct** under certain condition. Other RDBMS's do not allow this is actually doing the right thing. [0].

Does it really make sense to have this feature?

[0] http://www.programmerinterview.com/index.php/database-sql/sql-select-distinct-and-order-by/

PR #836, opened 2015-02-21
Add PHPCS and PHPMD support

JeroenDeDauw:
There are a ton of violations against the included rules,
so we cannot make this voting yet. Including the support
allows us to easily see the existing violations and incrementally
fix them.


JeroenDeDauw:
The composer based commands are just running against src/ for now,
since includes/ and tests/ will have many more failures

PR #860, opened 2015-03-01
Various fixes for escape character usage on SPARQLStore, SQLStore

mwjames:


mwjames:
Newly added tests `testRegexSearchForCharactersThatRequireSpecialEscapePattern`, `testRegexSearchForEscapeCharacter` are skipped for `postgres` because of "Query failed: ERROR: duplicate key value violates unique constraint"

PR #867, opened 2015-03-03
ByJsonDataQueryRunnerTest

mwjames:
Make query tests more discoverable and accessible to non-developers by:

- Defining query test cases in `json` format
- Automatic creation and clean-up of defined test entities
- Test case description follows the wikitext annotation style
- Separation of query tests, test case setup, and internal object handling

Query tests are crucial in order to verify that data are provided as excepted. Writing meaningful tests is a time consuming tasks and having tests be easily modellable is equally challenging.

PS: I grew tired on writing tests for each new query case by creating a set of rules of things like data sets, boundaries, settings etc..

`ByJsonDataQueryRunnerTest` will do all to work using test cases defined in `json` format by reading data from a file and execute queries and verify the results with the is expected item list.

Adding a new test case is as easy as to copy an existing `json` file, define the data parameter together with the queries it should run without the need for any PHP skills.

- `ByJsonDataQueryRunnerTest` runs all tests that are auto-discoverable by the `BulkFileProvider`
- `DataToQueryDefinitionFileHandler` instance to handle each separate file
- `QueryDefinitionInterpreter` interprets the `json` in respect to the `queries` section
- `QueryDefinitionTestCaseProcessor` to process, execute, and assert queries against the data provided by `QueryDefinitionInterpreter`

## Example

```json
{
	"description": "Blob value assignment/query test case using simple in-text annotation",
	"properties": [
		{
			"name": "Has blob property",
			"contents": "[[Has type::Text]]"
		},
		{
			"name": "Has another blob property",
			"contents": "[[Has type::Text]]"
		}
	],
	"subjects": [
		{
			"name": "Simple blob subject",
			"contents": "[[Has blob property::Value One]], [[Has another blob property::Value One]]"
		},
		{
			"name": "Another blob subject",
			"namespace": "NS_HELP",
			"contents": "[[Has blob property::Value Two]], [[Has another blob property::Value Two]]"
		}
	],
	"queries": [
		{
			"about": "Simple disjunctive blob query",
			"condition": "[[Has blob property::Value One]] OR [[Has another blob property::Value Two]]",
			"printouts" : [ "Has blob property" ],
			"parameters" : {
			  "limit" : "10"
			},
			"queryresult": {
				"results": [
					"Simple blob subject#0##",
					"Another blob subject#12##"
				],
				"count": "2",
				"dataitems": [
					{
						"type": "_txt",
						"value": "Value Two"
					},
					{
						"type": "_txt",
						"value": "Value One"
					}
				]
			}
		}
	],
	"settings": {
		"smwgNamespacesWithSemanticLinks": {
			"NS_MAIN": true,
			"NS_HELP": true,
			"SMW_NS_PROPERTY": true
		}
	},
	"meta": {
		"version": "0.1"
	}
}
```


mwjames:
@kghbln This should allow you to add new query tests without any PHP skills or to verify what kinf of queries are currently tested. I hope the `json` is somewhat self-explanatory.

- `properties` define what properties should be used during the test
- `subjects` define the pages that contain the data/annotations 
- `queries` define what sort of queries should be run against the data set

PR #877, opened 2015-03-08
DataRebuilder to run property rebuild first

mwjames:
For a full rebuild always ensure that property definitions are rebuild before any other page is going to be selected

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -57,6 +57,7 @@
         </whitelist>
     </filter>
     <php>
+       <var name="smwgSparqlDefaultGraph" value="http://example.org/phpunit-testrun"/>
        <var name="benchmarkQueryRepetitionExecutionThreshold" value="5"/>
        <var name="benchmarkQueryLimit" value="500"/>
        <var name="benchmarkQueryOffset" value="0"/>
--- a/src/Maintenance/DataRebuilder.php
+++ b/src/Maintenance/DataRebuilder.php
@@ -120,14 +120,10 @@ class DataRebuilder {
 		}
 
 		$this->useIds = array_key_exists( 's', $options ) || array_key_exists( 'e', $options );
-
 		$this->verbose = array_key_exists( 'v', $options );
 
-		$this->filters = $this->getFiltersFromOptions( $options );
-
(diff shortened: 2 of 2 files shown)


mwjames:
@kghbln FYI

kghbln:
Great, saw this on Monday and did not really dig into there this great change came from. Now I know! :) This change triggers the property refresh every time "rebuildData.php" is run. It might be an idea to have a way to disable this under certain circumstances, e.g. if one is e.g. using `-s`. On the other hand the property refresh is rather fast doing it's job ...

mwjames:
>  It might be an idea to have a way to disable this under certain circumstances, e.g. if one is e.g. using -s. On the other hand the property refresh is rather fast doing it's job ...

I thought about it but then you would need another option and the issue was that you normally would not make a fuss about running the property rebuild first (as you could by just doing `-p`).

Not forcing it can make the rebuild of property pages appear somewhere in between (as its page id is fetched in ascending order) which gives you some "normal" rendered pages without and some with the "new" typing (I came across some cases wondering about its inconsistencies just to realize the depending order relative to the property page id).

Let's see if we really need an option to avoid having the property rebuild first, for now we choose to be more restrictive.

kghbln:
Indeed, fair enough!

mwjames:
@kghbln I just realized [0], this didn't make on the release-notes list.

http://wikimedia.7.x6.nabble.com/Semantic-MediaWiki-2-2-has-been-released-tp5047747p5047753.html

kghbln:
PeterP on mailing list:

My vote would be for the opt-out switch but that's because I have gotten used to rebuilding 3-5K ID's at a time. I've never run the full 100K or so in one go because of the 'memory-leak' warning. Since that is still there my guess is that most would heed it and then get fed up having to rebuild the properties over and over again. It's no big deal though.

kghbln:
> I just realized [0], this didn't make it on the release-notes list.

Well it now made it [to the wiki](https://semantic-mediawiki.org/w/index.php?title=Help%3ARebuildData.php&diff=39256&oldid=39250) :) Thanks for the note.

mwjames:
> My vote would be for the opt-out switch

Added with #1106, `--skip-properties`

Issue #858, opened 2015-02-24
Merge system messages to make them reusable


kghbln:
On [tw.n](https://translatewiki.net/wiki/Thread:Support/About_MediaWiki:Smw-pa-property-predefined_sil_container/en_and_MediaWiki:Smw-pa-property-predefined_sbl_parentpage/en) there was the suggestion to move messages like `Smw-pa-property-predefined sil container` and `Smw-pa-property-predefined sbl parentpage` from the respective extensions into SMW core and do something like:

"$1" is a predefined property (also known as special property) which is provided by the [$2 $3] extension.

Is there an easy way to make SMW aware of special properties provided by extra extensions and to figure out which one it is a the same time?

mwjames:
Not again this discussion (see my comments [0]). Anyway this is where the magic happens [1].

> to move messages like Smw-pa-property-predefined sil container and Smw-pa-property-predefined sbl parentpage from the respective extensions into SMW core

While the text maybe similar nevertheless there are only provided by an extension that invokes those properties having them in SMW core seems as an odd practice.

> Is there an easy way to make SMW aware of special properties provided by extra extensions and to figure out which one it is a the same time?

No. The hook [2] registers properties without keeping any record by whom (extension) there are provided. Theoretically you could establish a convention that `__sbl_*`, `__sil_*` are provided by an extension (but as it happens SESP or SG [3] doesn't follow such convention ).

[0] https://gerrit.wikimedia.org/r/#/c/69330/, https://gerrit.wikimedia.org/r/#/c/67609/7/languages/SMW_Messages.php

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/articlepages/SMW_PropertyPage.php#L58

[2] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/PropertyRegistry.php#L330

[3] https://github.com/SemanticMediaWiki/SemanticGlossary/blob/master/src/PropertyRegistry.php#L18

kghbln:
Thank you for your detailed insight! Good to know. Basically that's how I felt about it.

PR #879, opened 2015-03-08
Show improved context for predefined property info

mwjames:
prop-info


mwjames:
@kghbln relates to #858; split the text into a `common` part and a part that solely describes a property task while removing repetition from some of the text elements mentioned in #858. Main objective was to improve the context displayed when hovering over a predefined property.

kghbln:
Yeah, this is a nice improvement solving more than one "issue". Thank you! Just added some qqq's which should roll in at the next repo sync.

PR #884, opened 2015-03-12
Isolate vocabulary fetching / rdf-003-import-foaf.json

mwjames:
Isolate vocabulary fetching via `ControlledVocabularyImportFetcher`

#refs #823


mwjames:
I can't believe this due to the previous `	$msglines = preg_split( "([\n][\s]?)", wfMessage( "smw_import_$onto_ns" )->inContentLanguage()->text() );` the import message would have not been fetched for any other language than `en` (or depending on what `$wgLanguageCode = "ja";` is set)

Issue #464, opened 2014-07-26
Owl / Import vocabulary / Use of `Imported from` property


mwjames:
I tried to create a `Property:Homepage` with `[[imported from::foaf:homepage]]` just to encounter that fundamental settings are missing. After looking at `SMW_DV_Import.php`:

- The `smw_unknown_importns` is missing from i18n
- Individual namespace definition such as `smw_import_foaf` is missing as well

Adding `"smw_import_foaf": "http://xmlns.com/foaf/0.1/|[http://www.foaf-project.org/ Friend Of A Friend] \n name|Type:String\n homepage|Type:URL\n mbox|Type:Email\n mbox_sha1sum|Type:String\n depiction|Type:URL\n phone|Type:String\n Person|Category\n Organization|Category\n knows|Type:Page\n member|Type:Page"
` to the en.json, copied from http://semanticweb.org/index.php?title=MediaWiki:Smw_import_foaf,  I was finally able to use `[[imported from::foaf:homepage]]`

### i18n
- [X] Add `smw-pa-property-predefined_impo` explain the property and point to http://semantic-mediawiki.org/wiki/Help:Import_vocabulary
- [x] Add `smw-datavalue-import-unknownns` for better help when a namespace definition such as `foaf` is missing and needs a message key such as `smw_import_foaf`

mwjames:
@kghbln Maybe in 2.2 we could sort things a bit. I found another site (smw.coopaxis.fr) that relies on import vocab. @tfrancart would be great if you could share your experience and suggestions how to improve the documentation or feature in general. 

> I just never understood (but I never asked) why the types declared in MediaWiki:smw_import_... don't work.
Because of that, I have to declare them on every page property anyway.
So, to me, I could never obey this part of the documentation: "If you import a property, remove its "has type" statement."

@jaideraf A more detailed analysis on this issue would help as well.

tfrancart:
Actually, smw.coopaxis.fr and the blog post on blog.sparna.fr all comes from me.
I second jaideraf by saying this vocabulary import feature is important for me too.
I also wasn't able to make this part of the documentation work : "If you import a property, remove its "has type" statement."

I don't really have enough experience for now to make suggestions on how to improve, our SMW+RDF project is just starting, but be sure that I will let you know if I have !

jaideraf:
@mwjames, just to explain some things:
[...]

PR #886, opened 2015-03-13
ImportValue to check for a valid type definition

mwjames:
refs #464


mwjames:
@kghbln @jaideraf FYI

This now checks if an imported line is correctly annotated (formatted). Imports without a head or terms without a type are no longer permitted.

Issue #885, opened 2015-03-13
result format "template" / link to further results wrapped in "pre" tag


kghbln:
When using the result format "template" the link to "further results" is wrapped into a "pre" tag. Changing  class="smw-template-furtherresults" does not help in this case. I think this is a SMW 1.9.x regression.

A public example for this is the [page on extensions at WikiAPIary](https://wikiapiary.com/wiki/Extension:Extensions)

mwjames:
Using a query like:

```
{{#ask: [[Modification date::+]]
 |?Modification date 
 |limit=1
 |template=TryAskTemplateNesting
 |outrotemplate=SomeOutro
 |format=template
 |searchlabel=..see more
}}
```
I'm unable to confirm the `<pre>` tag for the normal vector skin.

![template-output](https://cloud.githubusercontent.com/assets/1245473/6643792/80bf955a-c9f2-11e4-9323-b969157b0f50.png)

kghbln:
Actually I got this on a private wiki with Monobook. Was just happy to also see it on WikiApiary which of course uses Foreground. Dunno if it matters but I am on Firefox.

mwjames:
> Actually I got this on a private wiki with Monobook. Was just happy to also see it on WikiApiary which of course uses Foreground. Dunno if it matters but I am using Firefox.

I need an example that actually creates the issue (because currently I'm unable to replicate the issue)!

kghbln:
Ah, found one on [smw.o](https://semantic-mediawiki.org/wiki/Help:Template_format#Usage_for_userparam) using Vector

mwjames:
> Ah, found one on smw.o using Vector. Edit: e.g. limit=3 is causing this. See here

Yes, that was an easy enough example for me to replicate the issue.

## Solution

How about you find `return $this->rowstart . ' ' .`in [0] and remove the `. ' ' ` part. (`' '` is an extra space that let's MW interpreting it as a `<pre>` tag parser requirement)

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/queryprinters/ListResultPrinter.php#L455

kghbln:
Indeed, it was not the limit option. Applied you code change and it now works. However I accidently detected another on wiki ["fix" for this](https://semantic-mediawiki.org/w/index.php?title=Template:Query_output_demo&diff=prev&oldid=37872) which proves that it was a SMW 1.9.x regression.

[...]

PR #887, opened 2015-03-13
Fix pre tag rendering in template output, refs #885

mwjames:
refs #885, An extra space caused the MW Parser to interpret the `FurtherResults` element to be tagged with a `<pre>`.


mwjames:
@kghbln FYI

Issue #880, opened 2015-03-10
Validator::ParserHook is causing StripState/Invalid marker during UpdateJob::run


DestructionOfPlanetJupiter:
```
Backtrace:
#0 /home/ccccc/htdocs/w/includes/parser/StripState.php(66): StripState->addItem(string, string, string)
#1 /home/ccccc/htdocs/w/includes/parser/Parser.php(999): StripState->addGeneral(string, string)
#2 /home/ccccc/htdocs/w/extensions/Validator/src/legacy/ParserHook.php(302): Parser->insertStripItem(string, StripState)
#3 [internal function]: ParserHook->renderFunction(Parser, string, string, string, string, string)
#4 /home/ccccc/htdocs/w/extensions/Validator/src/legacy/ParserHook.php(336): call_user_func_array(array, array)
#5 /home/ccccc/htdocs/w/extensions/Validator/src/legacy/ParserHook.php(610): ParserHook->renderFunctionObj(Parser, PPFrame_DOM, array)
#6 [internal function]: ParserHookCaller->runFunctionHookObj(Parser, PPFrame_DOM, array)
#7 /home/ccccc/htdocs/w/includes/parser/Parser.php(3753): call_user_func_array(array, array)
#8 /home/ccccc/htdocs/w/includes/parser/Parser.php(3471): Parser->callParserFunction(PPFrame_DOM, string, array)
#9 /home/ccccc/htdocs/w/includes/parser/Preprocessor_DOM.php(1179): Parser->braceSubstitution(array, PPFrame_DOM)
#10 /home/ccccc/htdocs/w/includes/parser/Parser.php(3281): PPFrame_DOM->expand(PPNode_DOM, integer)
#11 /home/ccccc/htdocs/w/includes/parser/Parser.php(1239): Parser->replaceVariables(string)
#12 /home/ccccc/htdocs/w/includes/parser/Parser.php(405): Parser->internalParse(string)
#13 /home/ccccc/htdocs/w/extensions/SemanticMediaWiki/includes/ContentParser.php(189): Parser->parse(string, Title, ParserOptions, boolean, boolean, integer)
#14 /home/ccccc/htdocs/w/extensions/SemanticMediaWiki/includes/ContentParser.php(129): SMW\ContentParser->fetchFromParser()
[...]

PR #891, opened 2015-03-14
Strict type declaration for properties with import reference

mwjames:
A user-defined property with a reference to an imported property `Imported from` could define a different datatype than that of the imported property. This can introduce inconsistencies for type declarations in properties.

`TypeByImportPropertyAnnotator` will use the type definition if it finds an import property reference `Imported from` and replace it with any other datatype specified.

`[[Has type::Page]] [[Imported from::foaf:name]]` with `foaf:name` declared as `Text` will take precedence over the `Page` type declaration.

refs #886, #464

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -314,8 +314,9 @@
 	"log-name-smw": "Semantic MediaWiki log",
 	"log-description-smw": "Activities for [https://www.semantic-mediawiki.org/wiki/Help:Logging enabled event types] that have been reported by Semantic MediaWiki and its components.",
 	"smw-datavalue-import-unknownns":"The import namespace \"$1\" is unknown. Please ensure that owl import details are available via [[MediaWiki:Smw import $1]]",
-	"smw-datavalue-import-missing-nsuri":"Uable to find a \"$1\" namespace URI in the [[MediaWiki:Smw import $1]] import.",
-	"smw-datavalue-import-missing-type":"No type definition was found for \"$1\" in the [[MediaWiki:Smw import $2]] import.",
+	"smw-datavalue-import-missing-nsuri":"Uable to find a \"$1\" namespace URI in the [[MediaWiki:Smw import $1|$1 import]].",
+	"smw-datavalue-import-missing-type":"No type definition was found for \"$1\" in the [[MediaWiki:Smw import $2|$2 import]].",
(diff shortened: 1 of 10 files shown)


mwjames:
@SemanticMediaWiki/testers I'd appreciated some testing for this feature! For example, if the `foaf` imported vocabulary describes `homepage|Type:URL` then a property with reference to `foaf:homepage` should be typed with `URL` regardless of whether a `Has type` annotation exists or not.

For details on imported vocabularies, see [smw.org](https://www.semantic-mediawiki.org/wiki/Help:Import_vocabulary). Tips for manual testing, can be found [here](https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/master/tests#manual-testing).

jaideraf:
@mwjames I tested, it seems to be working, but...

When `{{#set:Page has default form=$someForm}}` (property from Semantic Forms) is on the property page, it shows that "UNIQ" thing of the parser (UNIQ58cd27dd581a0c9f-item-0--QINU, for example).

I dont know what is the relation of the two things (the declaration of the property type in the import vocabulary feature and the (almost deprecated) "Page has default form" property of the SF).

I think this is not a bloquer, because the new {{#default_form:$someForm}} parser function (in SF 3.0 and above) that substitute the "Page has default form" property seems to be working fine.

I just want to put this information here for documentation.

mwjames:
> I tested, it seems to be working,

Thanks

> When {{#set:Page has default form=$someForm}} (property from Semantic Forms) is on the property page, it shows that "UNIQ" thing of the parser (UNIQ58cd27dd581a0c9f-item-0--QINU, for example).

Lately I have seen those "UNIQ" in connection with SF https://github.com/WikiApiary/WikiApiary/issues/210 (which has no connection to this PR) but maybe https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/880#issuecomment-78348008 (strip marker issue in the mw-parser) can help.

jaideraf:
Oh, :+1:.
And thanks for working in the vocabulary import feature and fixing the issue. 
I think this is an important feature of SMW.

mwjames:
Moving on there are other things that needs clean-up.

mwjames:
@kghbln @jaideraf It would be great if we could revitalize and flesh out the content of the smw.org@Import vocabulary page.

PR #893, opened 2015-03-17
Add "improper value for" statistics

mwjames:
Improve visibility of the "improper value for" property by allowing better access to the error statistic and
help to improve value annotation gardening by linking to the appropriate property page.

error-stats


mwjames:
@kghbln @thingles If you grab the information from the api `&errorcount` you could start to build a "semantic content health" or curation coefficient by comparing the error value relative to the overall property values of a wiki. A percentile base line can be used to mark a semantic wiki as "excellent", "good", "fair", "poor" in terms of curated values.

thingles:
@mwjames Great idea! Created WikiApiary/WikiApiary#228 to track this.

kghbln:
This is one of the things I wonder why nobody has come up with the idea earlier. This is actually great for "annotation gardeners". Thanks for providing this!

Issue #896, opened 2015-03-19
Boolean format#x for "false" produces empty cell


mwjames:
An empty string (for false) will produce no table cells due to being interpret as empty printout.

https://semantic-mediawiki.org/w/index.php?title=Help_talk:Type_Boolean&offset=20150319095532&lqt_mustshow=1671#format_.23x_doesnt_work_1671

mwjames:
https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/408c59967b2d354e92155fafe84d04b9bc512e2b

PR #897, opened 2015-03-19
Fix empty string in boolean x-format, refs #896

mwjames:
Add `format-table-boolean-query-001.json` to check table content output for different formatting options.

refs #896


mwjames:
Travis is killing me ... once the queue is full, the performance drops like a stone (especially the coverage/Xdebug run).

Issue #899, opened 2015-03-19
[Tracking ticket] Fix coding style issues


JeroenDeDauw:
We want our code to be compliant with our own coding style rules. Recently we created an automated way of detecting violations, making it trivial to get a list of violations. Most violations are very easy to fix.

To get a list of violations, first get a git clone of the repository. If you want to be able to contribute back, you probably first want to create a fork, and clone that.

```
git clone git@github.com:YourGitHubAccount/SemanticMediaWiki.git
```

Then change into the directory and run `composer install`

```
cd SemanticMediaWiki
composer install
```

If you do not have Composer installed yet, [first install it](http://getcomposer.org/).

Next install [PHPCS](https://github.com/squizlabs/PHP_CodeSniffer) in whatever manner suits you best. For instance like this:

```
curl -OL https://squizlabs.github.io/PHP_CodeSniffer/phpcs.phar
```

Once that is done, you can get a list of violations by running the following in the root SMW directory:

```
php phpcs.phar src/* includes/* tests/* --standard=phpcs.xml --extensions=php -sp
```

The output will look as follows (though will have the above command instead of `compsoser cs-standalone`)

![Console output](https://cloud.githubusercontent.com/assets/146040/6734757/5459c4d8-ce5a-11e4-8562-28247706f5b3.png)

![Console output](https://cloud.githubusercontent.com/assets/146040/6734761/56fafa72-ce5a-11e4-93d3-762e9ebafa7e.png)

You do not need to actually install SMW itself. So you do not need a working MediaWiki installation, a database and a webserver. Even if you have an actual SMW install, it might still be easier to work with a dedicated development version for this task.

Once you fixed a violation or a set of related violations, create a commit, push it to your fork, and create a pull request.
### More advanced usage

If you create an actual working install of SMW, you can run `composer cs`, which will execute additional code style checks.

`composer cs` and `composer cs-standalone` only run style checks in the `src` directory. You can [look at how those commands are defined](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/composer.json#L80-L97) and easily see how you can run them for other directories, such as `includes` and `tests`. Example:

```
vendor/bin/phpcs includes/* --standard=phpcs.xml --extensions=php -sp
```

[...]

PR #903, opened 2015-03-20
Fixing spacing and line length in a few files #899

ckoerner:
Trying my hand at a few more small fixes for coding style. Feedback appreciated.

### DIFF
--- a/src/SPARQLStore/SparqlDBConnectionProvider.php
+++ b/src/SPARQLStore/SparqlDBConnectionProvider.php
@@ -68,18 +68,23 @@ class SparqlDBConnectionProvider implements DBConnectionProvider {
 	 * @param string|null $updateEndpoint
 	 * @param string|null $dataEndpoint
 	 */
-	public function __construct( $connectorId = null, $defaultGraph = null, $queryEndpoint = null, $updateEndpoint = null, $dataEndpoint = null ) {
-		$this->connectorId = $connectorId;
+	public function __construct( 
+		$connectorId = null,
+		$defaultGraph = null,
+		$queryEndpoint = null,
+		$updateEndpoint = null,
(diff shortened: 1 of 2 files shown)


src/SPARQLStore/SparqlDBConnectionProvider.php
```diff
@@ -68,19 +68,24 @@ class SparqlDBConnectionProvider implements DBConnectionProvider {
 	 * @param string|null $updateEndpoint
 	 * @param string|null $dataEndpoint
 	 */
-	public function __construct( $connectorId = null, $defaultGraph = null, $queryEndpoint = null, $updateEndpoint = null, $dataEndpoint = null ) {
-		$this->connectorId = $connectorId;
-
-		if ( $this->connectorId === null ) {
-			$this->connectorId = $GLOBALS['smwgSparqlDatabaseConnector'];
+	public function __construct(
```
mwjames: I probably won't agree with the constructor change since this is not really readable.

mwjames: @JeroenDeDauw Not sure about this ruleset but I can't agree on the strict setting in terms of constructor length.

ckoerner: Yeah, I was kind of stumped on the right thing to do here. Should I ignore those line length warnings for now?

JeroenDeDauw: The line length warnings are indeed tricky. Lots of false positives.

JeroenDeDauw:
@ckoerner so we do not want to merge the constructor change. The rest is good though. Can you modify the PR? Thanks a lot for the effort btw!

mwjames:
I'll fix the constructor my self.

PR #904, opened 2015-03-20
Max line lenght 140 -> 180

JeroenDeDauw:


mwjames:
Given the current performance of Travis it would take 2 hours before all tests pass therefore we GO FULL RISK ;-)

PR #909, opened 2015-03-23
In reference to #899: add missing and remove unnecessary spacing

hermannschwaerzlerUIBK:
I tried to remove all errors from cs-standalone regarding spacing.
But I intentionally left the 2 or 3 cases where a function is called inside square brackets as I was not sure about those.


JeroenDeDauw:
Fantastic! Thanks a lot!

PR #914, opened 2015-03-23
Add DIProperty::isUnrestrictedForUse

mwjames:
Follow-up on #913

User-defined properties are by default unrestricted for usage but certain pre-defined properties are not (e.g `Has subobject`, `Has query`).

When registering a pre-defined property, the annotable state (default is `true`) has to be declared as well.

Properties like `Has subobject`, `Has query` are by definition to be set internally and therefore not free for usage in formats like:
- `[[Has subobject:: ...]]`, `[[Has query:: ... ]]`
- `{{#set: |Has subobject= ... |Has query= ... }}`
- `{{#subobject: |Has subobject= ... |Has query= ... }}`

prop-rest


mwjames:
@kghbln Besides `Has subobject`, `Has query` are there any other pre-defined properties that should be restricted for free annotation usage?

kghbln:
Some come to my mind: Modification date, Creation date, Is a new page, Last editor is, Media type, MIME type, Query string, Query format, Query size, Query depth, Query duration. Probably all which are generated automatically. The first 6 are probably debatable since I will perhaps not harm to set a value manually though this would not make sense to do so. Thus ... There are also some on SESP around.

mwjames:
> Modification date, Creation date, Is a new page, Last editor is, Media type, MIME type, Query string, Query format, Query size, Query depth, Query duration.
> The first 6 are probably debatable

Yes, but most of them will be overridden by the system even though a user sets them manually. Subobject like properties are handled differently therefore need extra attention but of course with this PR any potential property can be restricted.

> here are also some on SESP around.

SESP (`EXIF`) has to be adjusted later on (to set `false`) and SIL (`Has interlanguage link`).

kghbln:
Fair enough!

mwjames:
> Some come to my mind: Modification date, Creation date, Is a new page, Last editor is, Media type, MIME type,

Restricted properties by default are now '_ASK', '_SOBJ', '_MEDIA', '_MIME' '_ERRP', '_LEDT', '_NEWP', '_CDAT', and '_MDAT'.

mwjames:
A `subobject` with no valid annotation content (property=value declaration) is no longer permitted / created.

kghbln:
This is now also documented in wiki.

kghbln:
> This is now also documented in wiki.

Thanks bro! Where???

krabina:
I think the values "Has processing error", "Has improper value" and "Has processing error text" should be allowed to be set manually. 
If you have property constraints that cannot be set with allows value, you could manually set it and fetch 
the errors alongsind with the real technical ones.

Simple usecase: Property:Persons, Property:Females. Using page forms and parser functions I can very well indicate if the user puts in 100 Persons and 120 Females and ask him to correct that because the number of females must be smaller or equal the number of persons. Therefore I would like to set the above mentioned values manually, in order to display all processing errors even the non-technical ones.

Or imagine asking for a percentage of Males and Females. The sum of both properties must be 100. If it is not, display a message.

mwjames:
> I think the values "Has processing error", "Has improper value" and "Has
> processing error text"

Those properties cannot be used by a user (set manually) especially
`Processing error` (Has processing error) as it is a container type
(\#1293) similar to that of a subobject or query that stores values in
a separate entity and is linked to the main entity.

> should be allowed to be set manually.

Instead you would need to introduce a new parser function (let's
called #set_error or #input_error or something else) that takes in
some strings (e.g. {{#set_error: ... error text ...|for=Foo }}) with
that it builds an error record together with a type (such as `user`,
\#3792) and is later attached to the entity.

Cheers

krabina:
Sorry, I deleted my first comment. This is how I am doing it as a workaround:
I just set the property [[Has processing error:{{FULLPAGENAME}} which then results in a processing error ;-)
This way I can in my template produce a processing error. The text is unusable  (indicating that I cannot set this property), but anyhow I can query for the number of processing errors and display on the page in case there are any...

mwjames:
> I just set the property [[Has processing error:{{FULLPAGENAME}} which then
> results in a processing error ;-)
> This way I can in my template produce a processing error. The text is

Ehh, the is not a workaround this is a hack [0] and shouldn't become
best practice. I suggest you turn my previous comments into a feature
request (state the use case, possible solutions etc.), so other users
can vote on it or comment for further input.

[0] https://www.merriam-webster.com/dictionary/hack

Cheers

krabina:
you are right, there is a better way: I set a user defined attribute: "Error" to a text value, e. g. "Number of females must be smaller then number of people". And set the attribute of type number. This way a processing error is produced.

PR #921, opened 2015-03-25
Fix coding style issues #899

jongfeli:
Some of the left over errors and warnings are probably the way they are because of a reason, they are in the "leave me alone" category. If so they could be marked as such, when possible (like previously noted by @hermannschwaerzlerUIBK) so that they are ignored by composer and they don't pop up anymore. This makes detecting and fixing style issues a bit more efficient in the future.

I am more then happy to do this but am not in a position to determine which errors and warning fit in the "leave me alone" category. I wanted to attach a .txt with the left overs but that does not seem to be possible in a pull request. 

This also means that the "real" left over errors and warnings need to be fixed by a person with a "bit" more php knowledge then me :)


JeroenDeDauw:
@jongfeli thanks a lot again. You can attach files via the web UI

attach

PR #923, opened 2015-03-25
Fix coding style issues #899

jongfeli:
.gitignore and composer.json should be ignored. I did not figure out yet how to exclude composer.json when using git commit on my local repository. Tried to solve this with .gitignore but that is probably not how it works :)


JeroenDeDauw:
Woha! Neat - thanks so much!

JeroenDeDauw:
I'm just going to merge this and then restore `composer.json`.

JeroenDeDauw:
@jongfeli you do not need to modify `composer.json` to run a different command. Simply run the command, not using composer.

So instead of doing the edit and running

    composer cs-standalone

You just run

    vendor/bin/phpcs tests/* --standard=phpcs.xml --extensions=php -spn --sniffs=Squiz.Arrays.ArrayBracketSpacing

jongfeli:
Ok thanks. I will make the commits a little smaller and explain in more detail what has been done. I don't want to be the cause of a bug :). Is there any " easy"  way of reverting a commit(s) or is that up to the person that merges them?

JeroenDeDauw:
If you want to modify something in your latest commit you can make the change and then run `git amend`. There are more advanced approaches as well, and we can modify, remove or add commits before stuff gets merged here. There was no real need for any of that here though.

PR #925, opened 2015-03-26
Fix coding style issues #899

jongfeli:


tests/phpunit/includes/storage/StoreTest.php
```diff
@@ -184,7 +184,7 @@ public function testGetPropertiesSpecial() {
 
 		$this->assertInstanceOf( '\SMW\SQLStore\SimpleListLookup', $result );
 		foreach( $result->fetchResultList() as $row ) {
-			$this->assertEquals( 2, sizeof( $row ) );
```
JeroenDeDauw: /me looks at @mwjames :)

JeroenDeDauw:
Thanks!

tests/phpunit/includes/storage/StoreTest.php
mwjames: Nop, this is not a test I wrote, it just happen to use the new class.

JeroenDeDauw: /me looks at @nischayn22 :)

JeroenDeDauw: Damn, just realized that it's still silly. Fixed in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/944

nischayn22: Don't remember much about this :/ 
Anyway, thanks for fixing it.

PR #945, opened 2015-03-30
Guard against circular reference in a concept query

mwjames:
For example, a concept `Concept:Modification date` with a self-reference to its own concept is just not possible.
```
{{#concept: [[Modification date::+]][[Concept:Modification date]]
 |Modification date
}}
```


mwjames:
@kghbln FYI

Issue #948, opened 2015-03-30
Unable to edit a popular property page


Nikerabbit:
An issue was reported to me that adding a new allowed value to http://tieteentermipankki.fi/wiki/Ominaisuus:Kieli did not work. After investigating I found out that null edit to that page does not complete because it runs out of memory.

I proceeded to do a null edit via command line with unrestricted memory use. First I thought it just gets stuck at something, but once I enabled a debug log I saw what is the issue. The full debug log is 58M, half of it caused by #947. I can provide the full log on request.

Here is what I ran on eval.php:

``` php
$wgShowExceptionDetails = true;
$wgShowSQLErrors        = true;
$wgShowDBErrorBacktrace = true;
$wgDebugComments        = true;
$wgLogQueries           = true;
$wgDebugDumpSql         = true;
$wgDevelopmentWarnings  = true;
$wgDebugTimestamps = true;
$wgShowDebug = true;
$wgDebugLogFile = 'php://stdout';

$title = Title::newFromText( 'Ominaisuus:Kieli' );
$revid = $title->getLatestRevID();
$target = Revision::newFromId( $revid );
$page = new WikiPage( $title );
$page->doEditContent( $target->getContent(), '', EDIT_UPDATE, $target->getId(), User::newFromId( 1 ) );
```

What happened, is that it used about 1.5G of memory, creating around 270k jobs (many of them duplicates and dropped on insert time). That is about twice the amount of pages we have in the wiki. 

Obviously, this does not scale. The new allowed value did appear though, when the script was finished.

mwjames:
> What happened, is that it used about 1.5G of memory, creating around 270k jobs (many of them duplicates and dropped on insert time). That is about twice the amount of pages we have in the wiki.

What is the issue here exactly? What are the steps to replicate this on a different wiki (or be able to write a regression test)?

- Step create page with ... ?
- Step two ...and so forth 

What version of SMW is used (I'm guessing 2.0 due to the #947 reference but that's not obvious). If it is 2.0 can it be observed with 2.1* as well?

mwjames:
@kghbln Any insights on what the issue is here?

Nikerabbit:
The issue is that editing a property page takes too much memory (and too long) if (my assumption) that property is used on many pages. You can try if you see similar behavior on your wiki by replacing the page name in the script with some other popular property.

Observed on 2.0, we haven't yet set a date when to update SMW on this wiki.

[...]

PR #951, opened 2015-03-31
Add UpdateDispatcherJob::createUpdateJobsFromListBySecondaryRun

mwjames:
It changes the way `UpdateJobs` are dispatched:

1. `UpdateDispatcherJob`'s first run is to use the store to create a list of unique subjects with the raw list being send in chunks (of 5000) for the secondary run of the `UpdateDispatcherJob`
2. `UpdateDispatcherJob`'s secondary run (which is scheduled through the job queue) is to use the raw list and create `UpdateJob` instances and push them into the job queue

refs #948

### DIFF
--- a/src/MediaWiki/Jobs/UpdateDispatcherJob.php
+++ b/src/MediaWiki/Jobs/UpdateDispatcherJob.php
@@ -5,16 +5,11 @@ namespace SMW\MediaWiki\Jobs;
 use SMW\ApplicationFactory;
 use SMW\DIProperty;
 use SMW\DIWikiPage;
-use SMW\Profiler;
 use Title;
 
 /**
- * Dispatcher class to invoke UpdateJob's
- *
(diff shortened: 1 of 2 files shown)


mwjames:
@kghbln Made in order to mitigate the issue mentioned in #948. Needs some really testing whether it can diffuse any memory issues while in "on-line" mode (when doing a type/constraint change).

mwjames:
Currently the `CHUNK_SIZE = 5000` (means a list of max 5000 subjects per invoked `UpdateDispatcherJob` on a secondary run) is fixed. I'm not sure whether or not we need an extra customizing so that users can tweak the setting for extreme use cases like #948.

kghbln:
Hope to have time to test during the upcoming easter holidays

mwjames:
> Hope to have time to test during the upcoming easter holidays

@kghbln Any chance you had time to give a test run?

PR #955, opened 2015-04-01
Fix coding style issues #899

jongfeli:


src/SPARQLStore/QueryEngine/QueryEngine.php
```diff
@@ -63,7 +63,8 @@ class QueryEngine {
 	 * @param QueryResultFactory $queryResultFactory
 	 * @param EngineOptions|null $engineOptions
 	 */
-	public function __construct( SparqlDatabase $connection, CompoundConditionBuilder $compoundConditionBuilder, QueryResultFactory $queryResultFactory, EngineOptions $engineOptions = null ) {
+	 // @codingStandardsIgnoreStart phpcs, ignore --sniffs=Generic.Files.LineLength
```
JeroenDeDauw: For most exceptions, this format is nice:

```
// @codingStandardsIgnoreStart
stuff that should be ignored here
// @codingStandardsIgnoreEnd
```

For some you cannot do this (ie unused parameters), though for most, such as line length, it's quite easy.

With how it's done now, I'll be able to see the open tag, but not the closing one, as it's at the end of an already too long line :)

jongfeli:
Ok, no problem. You did [point out](https://github.com/jongfeli/Code-violations-SemanticMediaWiki/commit/f358e9722633a23940007862359eb25403f676bf) some **unused parameters** that could be ignored but I will not touch any new **unused parameters** without consulting a pro first :), this applies to everything I am not sure of. 

You did merge the **@codingStandardsIgnoreEnd**. I can put those on a new line but sometimes it looks funny and it is not exact. Now it ignores exactly what it should ignore. @JeroenDeDauw Do you prefer the new line approach?

JeroenDeDauw:
@jongfeli well, that's what I tried to say - I prefer putting the comments on dedicated lines. It's not hugely important though, which is why I just merged this. Thanks

PR #957, opened 2015-04-02
Fix coding style issues #899 src/ includes/

jongfeli:
I had a go on the **Generic.ControlStructures.InlineControlStructure.NotAllowed** in [includes/](https://github.com/jongfeli/Code-violations-SemanticMediaWiki/blob/master/includes.txt) but I have a another question before I go on. Inline control means it ONLY does stuff that is behind the **if** on the same line and now we should define a beginning and end to the **if** with **{....}**. Is this correct?

This means that this:
```php
	try {
		$jd = floatval( reset( $datecomponents ) );
		if ( $calendarmodel == 'MJD' ) $jd += self::MJD_EPOCH;
		$this->m_dataitem = SMWDITime::newFromJD( $jd, SMWDITime::CM_GREGORIAN, SMWDITime::PREC_YMDT, $this->m_typeid );
	}
```

Becomes this:
```php
	try {
		$jd = floatval( reset( $datecomponents ) );
		if ( $calendarmodel == 'MJD' ) {
		   $jd += self::MJD_EPOCH;
		}
		$this->m_dataitem = SMWDITime::newFromJD( $jd, SMWDITime::CM_GREGORIAN, SMWDITime::PREC_YMDT, $this->m_typeid );
	}
```


joelkp:
Inline control structure means a single thing that ends with a semicolon follows the if (...), while (...), etc., as part of it. It could be on the same line or on a separate line. In either case, your "Becomes this:" example is how it would look after the change.

I don't know whether it occurs on a separate line (with { and } missing) anywhere in SMW code, but in any case I hope that makes it clearer.

JeroenDeDauw:
Indeed, the example posted is correct.

jongfeli:
Ok thanks. It was explained to me earlier but I just wanted to make sure. It is probably the same behavior as **SQL** code, if you don't define a beginning or an end to an **IF** it does the next thing it encounters and only that. The **SELECT  'Some other stuff';** is executed regardless of the outcome of **IF** but written like this makes it "not very clear" what is happening.

```sql
IF @parameter = 1
   
SELECT   '@par = 1';

SELECT  'Some other stuff';
```

You should use:
```sql
IF @parameter = 1
   BEGIN 

         SELECT '@par = 1';

   END; 

SELECT  'Some other stuff';
```

JeroenDeDauw:
Thanks once again for these style improvements!

PR #959, opened 2015-04-03
Add ExpDataSerializer/ExpDataDeserializer

mwjames:


includes/export/SMW_Exp_Data.php
```diff
@@ -68,6 +68,30 @@ public function getDataItem() {
...
 
 	/**
+	 * @since 2.2
+	 *
+	 * @return string
+	 */
+	public function getHash() {
+
+		$hash = array();
+		$hash[] = $this->getSubject()->getHash();
```
JeroenDeDauw: Just access the field directly.

includes/export/SMW_Exp_Data.php
```diff
@@ -68,6 +68,30 @@ public function getDataItem() {
...
+
+		foreach ( $this->getProperties() as $property ) {
```
JeroenDeDauw: Same here

src/Exporter/Element/ExpElement.php
```diff
@@ -76,6 +85,28 @@ public function getSerialization() {
 	}
 
 	/**
+	 * @see ExpElement::newFromSerialization
+	 */
+	protected static function deserialize( $serialization ) {
```
JeroenDeDauw: Not a fan of having such static deserialization functions, same goes for the `ewFromSerialization` below

src/Serializers/ExpDataSerializer.php
```diff
@@ -0,0 +1,75 @@
...
+		if ( !$this->isSerializerFor( $expData ) ) {
+			throw new OutOfBoundsException( 'Object is not supported' );
+		}
+
+		return $this->doSerialize( $expData ) + array( 'serializer' => __CLASS__, 'version' => 0.1 );
+	}
+
+	/**
+	 * @see Serializer::isSerializerFor
+	 *
```
JeroenDeDauw: `isSerializerFor` is not part of `Serializer`, so you do not need to implement it. And if you do implement it, it likely only makes sense for polymorphic usage, in which case you need to implement `DispatchableSerializer`

tests/phpunit/includes/Deserializers/ExpDataDeserializerTest.php
```diff
@@ -0,0 +1,154 @@
...
+			$instance->deserialize( $seralization )
+		);
+
+		$this->assertEquals(
+			$expected->getHash(),
+			$instance->deserialize( $seralization )->getHash()
+		);
+	}
+
+	public function expDataProvider() {
```
JeroenDeDauw: I highly encourage you to create dedicated test methods for different cases being tested. That makes it a lot easier to understand what is being tested.

JeroenDeDauw:
Introducing these dedicated serializers and deserializers seems good to me. What will they be used for though? Will they replace an existing yet inferior implementation?

src/Serializers/ExpDataSerializer.php
mwjames: Noted for a follow-up

tests/phpunit/includes/Deserializers/ExpDataDeserializerTest.php
mwjames: Noted for a follow-up

mwjames:
> Introducing these dedicated serializers and deserializers seems good to me. What will they be used for though? 

You can't send a `ExpData` object into a cache because it could be nested as in `ExpData -> ExpData -> ExpData` etc. and while testing some caching options in the `Exporter` I needed a way the serialize/deserialzer the object from/to cache. The same goes for the hash, now it comes with a stable hasher independent of how child objects where invoked making it actually possible to compare `ExpData` for equality (because the objects are stable and don't rely on any extra `Store` access during a object life-cycle) 

> I'm a bit surprised this has "enhancement" and not "code quality".

Well, it's new and hasn't done before for the `ExpData` object.

PR #965, opened 2015-04-04
SPARQLStore clean-up

mwjames:
- Remove some legacy cruft
- Use similar description matching as in `SQLStore`
- Use of `database` is discontinued and replaced with `Repository` because
the client does more than a simple storing of data it answers queries from a database
and we make mostly use of that capacity
- Some SRP


src/SPARQLStore/Connector/FusekiHttpRepositoryConnector.php
```diff
@@ -0,0 +1,59 @@
...
+
+use SMW\SPARQLStore\Exception\BadHttpDatabaseResponseException;
+use SMW\SPARQLStore\QueryEngine\RepositoryResult;
+use SMW\SPARQLStore\QueryEngine\XmlResponseParser;
+
+/**
+ * @see https://jena.apache.org/documentation/serving_data/index.html
+ *
+ * @license GNU GPL v2+
+ * @since 2.0
```
JeroenDeDauw: Not really

mwjames: Well, only the name changed (git thought it is new instead of moved, go figure why!)

JeroenDeDauw: The tags are here to help people figure out compat stuff, so a new name means it should be changed. You could extend it and have something like `@since 2.2, version with different name since 2.0`. Also, `git blame` is a thing if you want to dig deeper into it anyway

PR #969, opened 2015-04-05
Fix coding style issues #899

jongfeli:
I won't make a habit of creating such "small" pull requests like this morning but after I sat down looked at my screen I realized that the weatherman was wrong, so I "jumped" on my....


JeroenDeDauw:
\o/ Thanks!

PR #989, opened 2015-04-22
Also run tests with PHP 7

JeroenDeDauw:


JeroenDeDauw:
Fails because

> PHP Fatal error:  Argument 1 passed to MWExceptionHandler::rollbackMasterChangesAndLog() must be an instance of Exception, instance of TypeException given, called in /home/travis/build/SemanticMediaWiki/mw/includes/exception/MWExceptionHandler.php on line 141 and defined in /home/travis/build/SemanticMediaWiki/mw/includes/exception/MWExceptionHandler.php on line 114

@mwjames any idea what this `TypeException` thing is? I can't find it's definition anywhere and am wondering if this is a PHP, a MW or an SMW thing...

mwjames:
> 141 and defined in /home/travis/build/SemanticMediaWiki/mw/includes/exception/MWExceptionHandler.php on line 114

We have to remember that above `mw/includes/exception/MWExceptionHandler.php` is coming from the MW installation itself and not from SMW. Maybe it is not a good idea to run PHP7 against a MW 1.22.x release. I'm not sure whether running it against 1.24 would be better (obviously running against 1.25 doesn't make sense as it would fail for already mentioned reason).

JeroenDeDauw:
Tackling the other issues in #1021 and #1022

PR #994, opened 2015-04-23
Performance / Cache the hash

mwjames:
hash-perform

### DIFF
--- a/includes/SemanticData.php
+++ b/includes/SemanticData.php
@@ -235,11 +235,8 @@ class SemanticData {
 	 * hash algorithms that PHP offers.
 	 *
 	 * @note This function may be used to obtain keys for SemanticData
-	 * objects or to do simple equalitiy tests. Equal hashes with very
-	 * high probability indicate equal data. However, the hash is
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Wait... that is actually 18.1s to 1.1s?! First thought 18.1ms since using `,` to separate digit groups is evil.

JeroenDeDauw:
Code looks good \o/

JeroenDeDauw:
Shall I backport this onto `2.2.x`?

mwjames:
> Shall I backport this onto 2.2.x?

:+1:

mwjames:
> Anyway, how much difference (% of SMW cost) would this make on a normal page?

For normal pages with only a few template inclusions not much but those with large inclusions should see an effect (I used the wikiapiary `Template:Extension in use` together with a page as I was looking for something different, the screen-dump is based on that page but it is probably more dramatic because of my old notebook that runs the MW instance).

JeroenDeDauw:
I'll backport and update rel notes

PR #995, opened 2015-04-24
Exporter / Performance / use object for inmemory-cache

mwjames:
Using the object directly pays-off especially for long lists (subject that contains large subobjects) and is crucial for a smooth `SPARQLStore` update performance.

export-perform


mwjames:
Backported to 2.2.x with 9e99166753.

JeroenDeDauw:
The left window is the newer trace right? ;p

PR #999, opened 2015-04-25
Update .travis.yml

paladox:


mwjames:
Thanks but unfortunately this will not work.

There is a reason why we have to run `MW=master@5cc1f1d` and not on master because of some broken stuff in MW 1.25. Also adding more than one `TYPE=coverage` will not add any value and rather prolong the response of Travis when running our tests.

Changing version should not be done lightly.

Issue #467, opened 2014-07-27
SPARQLStore / Property value redirects in queries (bug 40084)


mwjames:
Just encountered [bug 40084](https://bugzilla.wikimedia.org/show_bug.cgi?id=40084) (from 2012) which is still not solved with the current implementation.

lloonngg:
Hi,

I am trying to use 4store 1.1.5 or virtuoso 7.1 as RDF backends for our SMW 2.1 instance.  All of our subobject queries do not work and I assume that it is related to lack of SPARQL 1.1 support (I assume from the RDF backend?).  The subobject queries works on jena-fuseki; however, we have noticed poor performance after migrating over 100K+ pages.  Therefore we prefer to use 4store or virtuoso.  Is there any quick fix to support subobject queries or will this have to wait until SMW 2.2 is released?  Is this something that can be fixed only from the SMW side or SPARQL 1.1 is absolutely required on the RDF backend?

Thanks,

Long

mwjames:
> I am trying to use 4store 1.1.5 or virtuoso 7.1 as RDF backends for our SMW 2.1 instance. All of our subobject queries do not work and I assume that it is related to lack of SPARQL 1.1 support (I assume from the RDF backend?)

No, SPARQL 1.1 has nothing do to with subobject support. SPARQL 1.1 is only relevant to some query features (as mentioned above for example in case of property paths etc.)

> The subobject queries works on jena-fuseki;

There is no discrimination between Fuseki, Sesame, 4store, or Virtuoso when it comes to replicating the data from SMW to the TDB instance because the RDF export (or better the Turtle serialization) is the same for all official listed TDB's that are supported and tested.

Normal annotations (`[[ ... :: ... ]]`) and subobject annotations are tested against [tr.0] involving different use cases [t.1, t.2, t.3] that include subobjects as well. This is to ensure that those objects behave in the same way and are independent from the back-end (`SQLStore` or `SPARQLStore`).

> we have noticed poor performance after migrating over 100K+ pages.

Interesting, can you quantify your findings?

> Therefore we prefer to use 4store or virtuoso. Is there any quick fix to support subobject queries or will this have to wait until SMW 2.2 is released?

The way subobjects are constructed/exported as TurtleTriple have not changed since 2.0 (where it was first supported).

[...]

PR #1001, opened 2015-05-05
SPARQLStore / Add support for prop/val redirects, refs #467

mwjames:
refs #467, Using SPARQL 1.1/inverse property paths [0] to find known redirects in the TDB

`$GLOBALS['smwgSparqlQFeatures'] = SMW_SPARQL_QF_REDI;` is enabled by default as it is expected that any "modern" TDB supports SPARQL 1.1 and for those that don't (Virtuoso 6.1, 4Store don't  fully support SPARQL 1.1) the feature should be disabled by setting `SMW_SPARQL_QF_NONE`.

[0] http://www.w3.org/TR/sparql11-query/#propertypaths

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -97,6 +97,21 @@ $GLOBALS['smwgSparqlDefaultGraph'] = '';
 ##
 $GLOBALS['smwgSparqlDatabaseConnector'] = 'custom';
 
+##
+# Sparql query features that are expected to be supported by the repository:
+#
+# - SMW_SPARQL_QF_NONE does not support any features (as required by SPARQL 1.1)
+# - SMW_SPARQL_QF_REDI to support finding redirects using inverse property paths,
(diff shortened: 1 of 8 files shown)


mwjames:
@kghbln SMW 2.3 / added `$GLOBALS['smwgSparqlQFeatures']` / use cases can be found in `query-06-04-page-property-value-redirects.json`

We can finally close bug 40084 from 2012 and bring the `SPARQLStore` on a par with the `SQLStore`.

Support for hierarchies [1] are done in a follow-up.

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/SemanticMediaWiki.settings.php#L184-L193

[2] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/SemanticMediaWiki.settings.php#L184-L193

kghbln:
This is now [documended on smw.o](https://semantic-mediawiki.org/wiki/Help:$smwgSparqlQFeatures)

PR #1003, opened 2015-05-06
SPARQLStore / Subproperty support

mwjames:
Follow-up on #1001:
 
- `$GLOBALS['smwgSparqlQFeatures'] = SMW_SPARQL_QF_SUBP;` for repositories that fully support SPARQL 1.1 by making use of `rdfs:subPropertyOf*`
- A larger use case with different hierarchy queries can be found in `query-04-02-subproperty-dc-import-marc21.json` including property/value redirects and regex value search


mwjames:
Only missing the remaining `_SUBC` hierarchy support ...

PR #1017, opened 2015-05-14
Move away from deprecated Validator code

### LINKED ISSUE #384: The deprecated ParserHook class is still used
`InfoParserFunction` and `DocumentationParserFunction` still derive from the highly deprecated `ParserHook`. SMW is currently blocking removal of this class from Validator.

We can retain a similar declarative [ParamProcessor](https://github.com/JeroenDeDauw/ParamProcessor) based approach by using the [ParserHooks library](https://github.com/JeroenDeDauw/ParserHooks), which does not suffer from the same design issues.

JeroenDeDauw:
Fixes #384

Might also take care of #880


JeroenDeDauw:
Oh smag you MediaWiki... really? >_>

mwjames:
What about [0] ?

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/SemanticMediaWiki.php#L33-L40

JeroenDeDauw:
That is still working because the latest release of ParserHooks uses Validator rather than ParamProcessor. Which will change. So we should drop that check. I'm wondering if it makes sense to add another one...

JeroenDeDauw:
@mwjames revieeeeeeeeeeeeeeeeeeeeeew! ;p

mwjames:
> revieeeeeeeeeeeeeeeeeeeeeew! ;p

Let me first add a small integration test that calls both parser function so we can ensure that before and after returns the same result.

JeroenDeDauw:
I manually checked that the thing are still working. With this type of change, I'd expect any breakage to be non-subtle.

You are right that tests are rather absent here so far. It should be easier now to write one that gives the code a bunch of input parameters and then checks the output is as expected.

JeroenDeDauw:
@mwjames I might be writing such tests later today but feel this got a bit cookie licked by you now, and want to avoid us doing the same task in parallel. So please clarify if you will do this today or not.

mwjames:
> So please clarify if you will do this today or not.

I'l do the integration test, should be done with it in 10 min or so.

mwjames:
Done with #1019 just having some trouble with #smwdoc to adhere the language specific test setting.

mwjames:
Done with 51bd826

JeroenDeDauw:
Great thanks. This test is now failing after I merged this PR. https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/62772387

Looks like this is caused by the code no longer accessing the global on every run since it's injected at configuration time now. I suspect this is not an issue for the production code and is just something occurring in the test environment. Any idea how to best deal with this?

JeroenDeDauw:
Yeah, so the test is changing the global, which will of course not change the value the code under test has...

mwjames:
> Yeah, so the test is changing the global, which will of course not change the value the code under test has...

Well, yes that's the point as it defines the boundaries the test is supposed to be run. We could kill all hooks before the test invocation and register them again on each run to ensure that all registered hooks do adhere the settings (as in a normal LocalSettings environment).

JeroenDeDauw:
Should be fixed by 54a89b910e5b6a9ce46a5293bf15d68da8e11d6e

JeroenDeDauw:
Well can just sidestep the language variant behaviour. If the thing checks some HTML is output, that is enough of a smoke test. More precise behaviour is better tested with a test that has better isolation.

Issue #682, opened 2014-12-17
SQLStore Id mismatch


mwjames:
The table diff sometimes compares a string Id with an integer Id which will lead to rather odd behaviour especially in connection with redirects.

Screen-dump was produced from `SMWSQLStore3Writers::preparePropertyTableUpdates`

![idmismatch](https://cloud.githubusercontent.com/assets/1245473/5427334/748cb3d8-83db-11e4-8529-f4e6320769e9.png)

With the `di_uri` table containing double entries for above mentioned id's

![doubleentry](https://cloud.githubusercontent.com/assets/1245473/5427358/448d2932-83dc-11e4-8ab1-28ef791c607a.png)

And for yet to determine a reason as to why (`422`/`130`/`localhost:8080/mw124/abc`) does no longer exists (compared with above) after an empty update (no annotation changes).

![deletedentry](https://cloud.githubusercontent.com/assets/1245473/5427386/f9ba6ae4-83dd-11e4-89ca-1e37b4327fc6.png)

__EDIT:__

I spend some time trying to figure out on what is going wrong and at least three issues were apparent:
- `preparePropertyTableInserts` produced float type sortkey which later is compared to a string sortkey in `arrayDeleteMatchingValues`
- ids (`s_id`, `p_id`) were mixed as either string or integer 
- `preparePropertyTableInserts` to build the `newarray` from `SemanticData` could contain double entries, this can happen when a page contains a value assignment to a redirected property and at the same a value assignment to the no-redirected property (as in above case `Has url` and `Has Url` point to the same id `130` but `Has url` redirects to `Has Url` which creates an issue during `arrayDeleteMatchingValues`)

JeroenDeDauw:
Once this is fixed, I'm going to shout some more at @yaronkoren to update Referata :)

mwjames:
It is expected that #1018 addresses the issue as outlined.

PR #1018, opened 2015-05-15
Add PropertyTableRowDiffer, refs #682

mwjames:
- Code responsible for computing row diffs was moved to its own component
- `SMWSql3SmwIds` now always returns `int` for id's to avoid id mismatch as seen in #682

### DIFF
--- a/src/SQLStore/PropertyTableRowDiffer.php
+++ b/src/SQLStore/PropertyTableRowDiffer.php
@@ -1,0 +1,350 @@
+<?php
+
+namespace SMW\SQLStore;
+
+use SMW\MediaWiki\Database;
+use SMW\SemanticData;
+use SMW\Store;
+use SMWDIError as DIError;
+use InvalidArgumentException;
+
+/**
+ * @license GNU GPL v2+
+ * @since 2.3
+ *
+ * @author Markus Krötzsch
+ * @author Nischay Nahata
+ * @author mwjames
+ */
+class PropertyTableRowDiffer {
+
+	/**
+	 * @var Store
+	 */
+	private $store = null;
+
+	/**
+	 * @since 2.3
+	 *
+	 * @param Store $store
+	 */
+	public function __construct( Store $store ) {
+		$this->store = $store;
+	}
+
+	/**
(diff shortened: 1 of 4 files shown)


mwjames:
@JeroenDeDauw @kghbln This PR should improve the reliability of the `SQLStore` (especially in connection with #682) but to avoid a situation where a regression slipped through (well the existing integration tests couldn't catch them) we should be vigilant in the upcoming period of testing when using the master to monitor any unintended behavior. 

We can't really design an integration test that would mimic the observation made in #682 therefore those edge cases need to be filtered manually.

The following counter measures have been added to avoid possible duplicate table rows by:

- Making sure that primitive types are compared to the same type
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1018/files#diff-3a38a0c6ec81c2c749c281edbf906b89R335
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1018/files#diff-6b1b07782b2e982b031ecc568ace2afaR284
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1018/files#diff-6b1b07782b2e982b031ecc568ace2afaR163
- The last line of defense is https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1018/files#diff-6b1b07782b2e982b031ecc568ace2afaR117 (`$hashModifier` is currently unused, only to be used if the above changes still show that hashes are not updated according to the state of the data)

src/SQLStore/PropertyTableRowDiffer.php
```diff
@@ -0,0 +1,350 @@
...
+ * @author Markus Krötzsch
+ * @author Nischay Nahata
+ * @author mwjames
+ */
+class PropertyTableRowDiffer {
+
+	/**
+	 * @var Store
+	 */
+	private $store = null;
```
JeroenDeDauw: No need to set this to null, constructor always sets it

JeroenDeDauw:
Did a superficial review and looks good. Thanks for all the work on this!

mwjames:
Well, nothing is going to change from having it here so let's be bold ...

Issue #975, opened 2015-04-10
Adjust SMW Factbox visibility


hypergrove:
IN MW 1.25 and maybe earlier, there is a Page Information link that displays structured data associated with a page. The SMW Factbox shows on this Page Information page only when it's set to show on the page itself (per a flag). 

It seems a good idea that the SMW Factbox *always* show on a page's Page Information page (where action=info) and to apply the SMW Factbox flag only when action=view.

Misdre:
Apparently SMW isn't using the InfoAction hook to add data to the action=info page... does that mean it's handled by OutputPageParserOutput?

mwjames:
> Apparently SMW isn't using the InfoAction hook to add data to the action=info page

Correct, we don't use the hook because the `Factbox` existed long before `InfoAction` was introduced and I doubt that adding the hook brings any additional insight that is not already provided by the `Factbox` (as in-page inclusion) or by the `Special:Browse`. Unless we receive a PR (with appropriate tests), I can't see supporting the `InfoAction` to be happen in near future. 

> does that mean it's handled by OutputPageParserOutput?

https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/MediaWiki/Hooks/SkinAfterContent.php#L62

hypergrove:
Agreed it's not a new function. This request stems from a desire to 
_conform_ to the addition of the InfoAction -- a new page that displays 
all structured data associated with a page.
SMW should not, in effect, be penalized for having developed its 
software prior to InfoAction being developed by the MW team.
I do request you reconsider your apparently dim view of this change request.
Cheers/john

On 4/29/2015 11:26 AM, mwjames wrote:
>
>     Apparently SMW isn't using the InfoAction hook to add data to the
>     action=info page
>
> Correct, we don't use the hook because the |Factbox| existed long 
> before |InfoAction| was introduced and I doubt that adding the hook 
> brings any additional insight that is not already provided by the 
> |Factbox| (as in-page inclusion) or by the |Special:Browse|. Unless we 
> receive a PR (with appropriate tests), I can't see supporting the 
> |InfoAction| to be happen in near future.
>
>     does that mean it's handled by OutputPageParserOutput?
>
> https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/MediaWiki/Hooks/SkinAfterContent.php#L62
>
> —
> Reply to this email directly or view it on GitHub 
[...]

PR #1030, opened 2015-05-17
Factbox visibility

Misdre:
Added InfoAction hook support + tests.
#975

### DIFF
--- a/src/Factbox/Factbox.php
+++ b/src/Factbox/Factbox.php
@@ -98,7 +98,7 @@ class Factbox {
 	 *
 	 * @since 2.1
 	 *
-	 * @param boolean
+	 * @param boolean $preview
 	 */
 	public function useInPreview( $preview ) {
 		$this->useInPreview = $preview;
@@ -109,7 +109,7 @@ class Factbox {
 	 *
 	 * @since 2.3
 	 *
-	 * @param boolean
+	 * @param boolean $actionInfo
 	 */
 	public function useInActionInfo( $actionInfo ) {
 		$this->useInActionInfo = $actionInfo;
--- a/src/Factbox/Factbox.php
+++ b/src/Factbox/Factbox.php
@@ -74,6 +74,11 @@ class Factbox {
 	private $useInPreview = false;
 
 	/**
+	 * @var boolean
+	 */
+	private $useInActionInfo = false;
+
+	/**
 	 * @since 1.9
 	 *
 	 * @param Store $store
@@ -97,6 +102,17 @@ class Factbox {
 	 */
 	public function useInPreview( $preview ) {
 		$this->useInPreview = $preview;
+	}
+
+	/**
+	 * @note contains information about action=info
+	 *
+	 * @since 2.3
+	 *
+	 * @param boolean
+	 */
+	public function useInActionInfo( $actionInfo ) {
+		$this->useInActionInfo = $actionInfo;
 	}
 
 	/**
@@ -181,6 +197,8 @@ class Factbox {
 			$showfactbox = SMW_FACTBOX_HIDDEN;
 		} elseif ( $this->useInPreview ) {
 			$showfactbox = $settings->get( 'smwgShowFactboxEdit' );
+		} elseif ( $this->useInActionInfo ) {
+			$showfactbox = SMW_FACTBOX_NONEMPTY;
 		} else {
 			$showfactbox = $settings->get( 'smwgShowFactbox' );
 		}
--- a/src/MediaWiki/Hooks/InfoAction.php
+++ b/src/MediaWiki/Hooks/InfoAction.php
@@ -50,7 +50,11 @@ class InfoAction {
 	 * @return true
 	 */
 	public function process() {
-		return $this->canPerformUpdate() ? $this->performUpdate() : true;
+		if ( $this->canPerformUpdate() ) {
+			$this->performUpdate();
+		}
+
+		return true;
 	}
 
 	private function canPerformUpdate() {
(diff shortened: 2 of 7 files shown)


mwjames:
Something with the test seems in vain.

```
1) SMW\Tests\MediaWiki\Hooks\OutputPageParserOutputTest::testProcess with data set #0 (array(array(true), Mock_OutputPage_fdb97c87, ParserOutput), array('SMW\\Tests\\MediaWiki\\Hooks\\OutputPageParserOutputTest::outputDataProvidermock-subject'))
Failed asserting that a string is empty.
```

mwjames:
Try running those test in your local dev environment, see [0] on how to run tests locally.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/master/tests#running-tests

mwjames:
This video [0] might help you in connection with tests and SMW.

[0] https://youtu.be/v6JRfk5ZmsI

Misdre:
I did notice this failure, I just dismissed it too quickly. I will investigate.

JeroenDeDauw:
Thanks a lot for working on this @Misdre!

Can you rebase those commits into a single one so we do not have commits with failing tests in our history after merge?

Misdre:
Done!

Misdre:
What are we supposed to do when Scrutinizer times out?

mwjames:
> What are we supposed to do when Scrutinizer times out?

Nothing, sometimes it just happens that Travis is too slow to send the coverage report to Scrutinizer so that after 35 min. or so Scrutinizer is instructed to fail (otherwise it would wait forever).

> Done!

It is looking good but let me have a proper look once I have a bit more time for it. (Don't be discouraged by not having it merged this week unless @JeroenDeDauw pushes it through)

src/MediaWiki/Hooks/InfoAction.php
```diff
@@ -0,0 +1,85 @@
...
+		$this->outputPage = $outputPage;
+		$this->parserOutput = $parserOutput;
+		$this->pageInfo =& $pageInfo;
+	}
+
+	/**
+	 * @return true
+	 */
+	public function process() {
+		return $this->canPerformUpdate() ? $this->performUpdate() : true;
```
JeroenDeDauw:
```php
if ( $this->canPerformUpdate() ) {
    $this->performUpdate()
}

return true;
```

src/MediaWiki/Hooks/InfoAction.php
```diff
@@ -0,0 +1,85 @@
...
+			$title->isRedirect() ||
+			!$this->isSemanticEnabledNamespace( $title ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	private function performUpdate() {
+		$cachedFactbox = ApplicationFactory::getInstance()->newFactboxFactory()->newCachedFactbox();
```
JeroenDeDauw: `ApplicationFactory::getInstance()` should only be accessed at entry points to SMW, such as the hook handling function you added in `HookRegistry.php`. This is because calling that static method is accessing global state, which causes a whole pile of problems.

src/Factbox/Factbox.php
```diff
@@ -100,6 +105,17 @@ public function useInPreview( $preview ) {
 	}
 
 	/**
+	 * @note contains information about action=info
+	 *
+	 * @since 2.3
+	 *
+	 * @param boolean
```
JeroenDeDauw: `@param boolean $actionInfo`

src/Factbox/Factbox.php
```diff
@@ -100,6 +105,17 @@ public function useInPreview( $preview ) {
...
+	 */
+	public function useInActionInfo( $actionInfo ) {
```
JeroenDeDauw: The format of this method is suspect - it is probably better to have `->foo()` and `->notFoo()` than `->foo( true )` and `->foo( false )`. The naming is also not clear. Without investigating, I cannot tell what this does. Does this mean "is going to be used in 'action info' and should behave as such", "was used in 'action info' and should hold that into account" or "should only be used in 'action info'"?

I see you did not start this pattern though...

tests/phpunit/includes/MediaWiki/Hooks/InfoActionTest.php
```diff
@@ -0,0 +1,338 @@
...
+		$this->applicationFactory->registerObject( 'Settings', $settings );
+	}
+
+	protected function tearDown() {
+		$this->applicationFactory->clear();
+
+		parent::tearDown();
+	}
+
+	public function testCanConstruct() {
```
JeroenDeDauw: This test does not seem to justify itself. It just executes some code without making any assertions that can possibly fail. Furthermore, other tests will already execute this code anyway.

tests/phpunit/includes/MediaWiki/Hooks/InfoActionTest.php
```diff
@@ -0,0 +1,338 @@
...
+		$this->assertInstanceOf(
+			'\SMW\MediaWiki\Hooks\InfoAction',
+			new InfoAction( $outputPage, $parserOutput, $pageInfo )
+		);
+	}
+
+	/**
+	 * @dataProvider outputDataProvider
+	 */
+	public function testProcess( $parameters, $expected ) {
```
JeroenDeDauw: This test, and it's data provider also seem problematic. Both methods are quite big, and the amount of mocking going on is unfortunate. To me this test is hard to understand, and it certainly has a lot of coupling to internal details of the production code it tests, making it hard to make modifications.

JeroenDeDauw:
@Misdre do you need further support before you can continue on this?

Misdre:
I'm fine, I'm going to work on it this week or the next.

Misdre:
too much work left to leave that open.

PR #1031, opened 2015-05-19
CategoryResultPrinter to recognize offset for further results

mwjames:
This isn't really a regression as the 1.8 til 2.2 used `getLink` instead of `getFurtherResultsLink`


mwjames:
@kghbln Backported to 2.2.x with 581a905f17587f354e8ff2afa62623c45f7aa33d

Issue #555, opened 2014-10-02
RFC: Instant query (display updated query results on a wiki page)


mwjames:
The main obstacle for a query to display instantly updated results is MediaWiki's `ParserCache`. A page object will read an article content from cache and avoids expensive text parsing with the limitation that a query most likely doesn't display the current expected results as the `ParserCache` contains an outdated set of results. To make a page object believe that it needs to re-render its content, it is necessary to track query usages and invoke `Title::invalidCache` on pages to stimulate the re-rendering process.

For example, if a page adds/changes/or removes a property/value it is very likely (naturally it depends on the limit set by the query or on other criteria but in a best case scenario it will be part of a potential result pool) that those queries that have references to those items (property, category etc.) will include the newly annotated subject when it is refreshed/updated.

mwjames:
Closed with #1117

PR #1035, opened 2015-05-21
Add IntermediaryBlobValueLookup for fast access to pre-cached lookups

mwjames:
It is suggested to use `redis` but any another BagOStuff provider will also work. The configuration `$GLOBALS['smwgBlobCacheType']` can be used to assign:
- `CACHE_NONE` to disable the lookup and instead use the "standard" non-cached SQLStore/DB access
- `CACHE_ANYTHING` (default) will use the SqlBagOStuff and is the least performing (as it uses the underlying SQL DB) cache layer but doesn't require any external cache provider
- `redis` is the most suitable provider available for MW

```
$wgObjectCaches['redis'] = array(
    'class'                => 'RedisBagOStuff',
    'servers'              => array( '127.0.0.1:6379' ),
);
```

`$GLOBALS['smwgValueLookupCacheLifetime']` is set to `60 * 60 * 24 * 7; // a week` so that left-over items are evicted after the expiry but it can also be used to shorten the life-cycle of a cache item.

While the write process will not much benefit from this, displaying a page should happen a bit faster as we no longer using DB access on values that haven't changed (it does not cache any queries but any call from queries to static values like `_IMPO` or `_TYPE` etc. will be served from cache if available which should reduce the amount of DB queries made to/from the `SQLStore` during the display of a page).

- The blob for a subject will be invalidated on each change, purge, delete, or move action. [blob-store](https://github.com/onoi/blob-store) manages all the operations in regards to the blob-storage and is tested using redis [0] explicitly.
- This PR is building the groundwork for #555 and the support for instant updated queries as demonstrated in [1].
- All existing tests (especially those integration tests) have not been altered indicating that existing behaviour (those for redirects and value lookup) does prevail even with an intermediary present.
- Preliminary benchmarks are indicating that using the `SPARQLStore` with a cache shows the same response time as a `SQLStore` without cache due to certain static lookups being cached during the request lifetime.
- This PR should generally improve performance on pages with long query lists where each `ResultArray` row makes at least one request to `Store::getPropertyValues`
- It will improve performance for extensions that normally make un-cached lookups to `getSemanticData`, `getProperties`, and `getPropertyValues` an avoid a deadlock situation for too many requests that are made within a page where an extension access those methods
- API requests to the mentioned methods will now be redirected to the cache layer to avoid a potential DB deadlock asthe cache can be scaled 

### Performance

Direct comparison of `Special:Browse` using CACHE_NONE (left) vs `redis` (right) on a `SPARQLStore` (Sesame) instance.

image

### Access statistics (view on property page)

```
array(2) {
  ["Onoi\Cache\FixedInMemoryLruCache"]=>
  array(6) {
    ["inserts"]=>
    int(18)
    ["deletes"]=>
    int(0)
    ["max"]=>
    int(500)
    ["count"]=>
    int(18)
    ["hits"]=>
    int(232)
    ["misses"]=>
    int(0)
  }
  ["Onoi\Cache\MediaWikiCache::RedisBagOStuff"]=>
  array(4) {
    ["inserts"]=>
    int(0)
    ["deletes"]=>
    int(0)
    ["hits"]=>
    int(18)
    ["misses"]=>
    int(0)
  }
}
```
[0] https://travis-ci.org/onoi/blob-store/jobs/63284472
[1] https://www.youtube.com/watch?v=vVDOVCnZs_s


mwjames:
@kghbln @thingles @jongfeli For those who are interested in performance.

thingles:
Awesome!

mwjames:
Let's start the field test

kghbln:
Documented:
* [`$smwgValueLookupCacheType`](https://www.semantic-mediawiki.org/wiki/Help:$smwgValueLookupCacheType)
* [`$smwgValueLookupCacheLifetime`](https://www.semantic-mediawiki.org/wiki/Help:$smwgValueLookupCacheLifetime)

Issue #1050, opened 2015-05-29
Concepts / ConceptDescriptionInterpreter fails with undefined method SMW\DIWikiPage::getText


chironimow:
I have upgraded my system from MW1.16 SMW 1.6.2 to MW1.25 and SMW2.2

Since then pages including concepts are failing directly:

( ! ) Fatal error: Call to undefined method SMW\DIWikiPage::getText() in /home/innowiki/wikifarm/latest/extensions/SemanticMediaWiki/src/SQLStore/QueryEngine/Interpreter/ConceptDescriptionInterpreter.php on line 112
Call Stack
#	Time	Memory	Function	Location
1	0.0000	234160	{main}( )	../index.php:0
2	0.0282	4760520	MediaWiki->run( )	../index.php:41
3	0.0282	4762088	MediaWiki->main( )	../MediaWiki.php:414
4	0.0355	5114136	MediaWiki->performRequest( )	../MediaWiki.php:566
5	0.0371	5234672	MediaWiki->performAction( )	../MediaWiki.php:273
6	0.0372	5234904	ViewAction->show( )	../MediaWiki.php:395
7	0.0372	5237264	Article->view( )	../ViewAction.php:44
8	0.0577	6691872	PoolCounterWork->execute( )	../Article.php:674
9	0.0578	6717216	PoolWorkArticleView->doWork( )	../PoolCounterWork.php:123
10	0.0578	6717624	AbstractContent->getParserOutput( )	../PoolWorkArticleView.php:140
11	0.0579	6719136	WikitextContent->fillParserOutput( )	../AbstractContent.php:497
12	0.0579	6721336	Parser->parse( )	../WikitextContent.php:333
13	0.0582	6734656	Parser->internalParse( )	../Parser.php:434
14	0.0583	6738208	Parser->replaceVariables( )	../Parser.php:1231
15	0.0639	6742376	PPFrame_DOM->expand( )	../Parser.php:3322
16	1.5873	9930384	Parser->braceSubstitution( )	../Preprocessor_DOM.php:1169
17	1.5926	9944048	PPFrame_DOM->expand( )	../Parser.php:3647
18	1.5963	9961168	Parser->braceSubstitution( )	../Preprocessor_DOM.php:1169
19	1.5979	9965712	Parser->callParserFunction( )	../Parser.php:3505
20	1.5987	9967888	call_user_func_array ( )	../Parser.php:3771
[...]

PR #1051, opened 2015-05-29
Fix call to DIWikiPage::getText in ConceptDescriptionInterpreter, refs #1050

mwjames:
refs #1050


mwjames:
Only backported the fix to 2.2.x / 85e4b4a.

PR #1053, opened 2015-06-03
Distinguish display of a "normal" link and a subobject

mwjames:
Sometimes it can be hard to distinguish between a standard subject and a subobject when displayed next to each other because the fragment (the part that represents the subobject id) is generally hidden in html. A subject that is identified as subobject now gets a `smw-subobject-entity` class assigned which by default is styled as `italic`.

image

image


mwjames:
@kghbln Do we need or want such feature?

kghbln:
> Do we need or want such feature?

Yes, we do. This will intuitively help users being aware of subobjects.

JeroenDeDauw:
Seems reasonable to me

Issue #895, opened 2015-03-18
Semantic data remains after changing page to redirect


darenwelsh:
Possibly related to [#869] (https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/869) and [#566] (https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/566).

1. Create page using a template that sets semantic properties for that page
 * Browse properties for that page and note semantic data
2. Revise that page so it only has code to redirect to a different page
 * Browse properties for that page again and note that semantic data remains

See [this example] (http://semantic-mediawiki.org/w/index.php?title=User:Darenwelsh/Test_Page_2&redirect=no) and note how [its properties] (http://semantic-mediawiki.org/wiki/Special:Browse/User:Darenwelsh-2FTest_Page_2) remained after I revised it to just redirect to another page.

This also happens on pages that use Semantic Forms, but not on pages where a semantic property is set without use of a template. For example

1. Create page including some kind of property being set with {{#set: myProperty = value }}
 * Browse properties for that page and note semantic data
2. Revise that page so it only has code to redirect to a different page
 * Browse properties for that page again and note that semantic data is gone

See [this example] (http://semantic-mediawiki.org/w/index.php?title=User:Darenwelsh/Test_Page_1&redirect=no) and note how [its properties] (http://semantic-mediawiki.org/wiki/Special:Browse/User:Darenwelsh-2FTest_Page_1) are gone as they should be now that the page is just a redirect to another page.

mwjames:
One could cry, it has to be the redirects that consistently create issues and this time it seems to be connected with template usage.

PS: Currently, I don't feel in the mood for a bug hunt (and its sounds like a real pain bug) the last time it cost me a week to create a proper use case let alone to track down the #347 issue.

mwjames:
Here some hints for people who want to debug this, the move/redirect action is initiated by [0] which eventually calls [1] `SMWSQLStore3Writers ::changeTitle` to reshuffle the internal Id assignments.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/MediaWiki/Hooks/TitleMoveComplete.php

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/storage/SQLStore/SMW_SQLStore3_Writers.php

mwjames:
[...]

PR #1054, opened 2015-06-08
Special:MovePage and #redirect to create consistent data, refs #895

mwjames:
@kghbln @darenwelsh It is expected to fix the observed issue in #895 but required a different approach than outlined in the original analysis because the `parser-07-02-intext-manual-redirect-template.json` test showed different behaviour on 1.24+.

I hope that we now have a consistent behaviour for `Special:MovePage` and `#REDIRECT`.


mwjames:
@kghbln This could be a disruptive change (which I don't expect) therefore user reports about redirects in 2.3 should be followed up in connection with this PR.

PR #1063, opened 2015-06-13
Follow-up on #1035

mwjames:
refs #1035; `$GLOBALS['smwgValueLookupCacheType']` is for opt-in only and it is expected that an optimized cache provider is used. MW's `SqlBagOStuff` pseudo cache or `CACHE_ANYTHING` should not be used.

- Add support for `Store::getPropertySubjects`
- Split some lists as those provide better performance when combined with other objects also use an extra container for `SemanticData`

## Some data 

The following data will not stand a rigour statistical analysis but it should allow to draw some conclusions about the impact of `$GLOBALS['smwgValueLookupCacheType']`.

Measurement based on:
- MediaWiki: 1.23.9 PHP: 5.6.8 with enabled `$wgDebugToolbar = true;` (with only one user at the time of the query execution).
- SPARQLStore Sesame 2.7.14
- SQLStore
- redis 2.8* / Windowsx64

For easy verification, the following data set have been used as base for the sample query (properties are appropriately typed).
```
{{#set_recurring_event:Sample query
 |property=has date
 |Has page=Sample query
 |Has location=London,Berlin,Paris|+sep=,
 |Has area=891.85 km²;64 sqmi;67 sqmi|+sep=;
 |Has wattage=12+33 hp+450 W+15 kW|+sep=+
 |Has text=Simple recurring events text
 |start=June 8, 2010
 |unit=day
 |period=1
 |limit=1000
 |duration=7200
 |include=March 16, 2010;March 23, 2010|+sep=;
 |exclude=March 15, 2010;March 22, 2010|+sep=;
}}
```
Query used via `Special:Ask` to measure data presented below.

```
{{#ask: [[Has page::Sample query]] [[Has area::+]] [[Has wattage::+]]
 |?Has area
 |?Has wattage
 |?Has location
 |?Has date
 |format=broadtable
 |sort=Has date,Has area
}}
```

For `$GLOBALS['smwgValueLookupCacheType'] = CACHE_NONE;` (= legacy behaviour) set

Limit   | SPARQLStore| SQLStore 
------- | -------------------------- | -------------
20      | Q: 218  T: 1.32  M: 4.79 | Q: 199   T: 1.12  M: 4.67 
50      |  Q: 342  T: 2.20  M: 5.14 | Q: 289   T: 1.77  M: 4.98
250     | Q: 1190 T: 6.65  M: 6.18 | Q: 937   T: 6.12  M: 6.35 
500     | Q: 2242 T: 11.95 M: 9.12 |  Q: 1749  T: 11.18 M: 9.78

For `$GLOBALS['smwgValueLookupCacheType'] = redis;`set

Limit   | SPARQLStore | SQLStore                       
------- | ------------------------------ | -------------
20      | Q: 124   T: 1.20  M: 4.86 (*)  |
        | Q: 120   T: 1.20  M: 4.86      | Q: 124  T: 1.06  M: 4.77
50      | Q: 240   T: 2.07  M: 5.56 (*)  |
        | Q: 120   T: 1.80  M: 5.28      | Q: 120  T: 1.42  M: 5.19
250     | Q: 920   T: 7.88  M: 8.94 (*)  |
        | Q: 120   T: __4.70__  M: 7.99      | Q: 120  T: __4.17__  M: 7.88
500     | Q: 1120  T: 12.85 M: 12.49 (*) |
        | Q: 120   T: __7.62__  M: 11.35     | Q: 120  T: __7.51__  M: 11.12

- `Q:` refers to queries, `T:` to time in sec, and `M:` memory in MB displayed by `$wgDebugToolbar`
- (*) refers to items not yet cached / warm-up
- The use of `CACHE_ANYTHING` (MW's internal SqlBagOStuff pseudo cache) did not elevate lookup performance in a way `redis` did therefore it is not recommended to use `CACHE_ANYTHING` as sample data suggest it to be a poor replacement on large lists (250+) for `redis`.
- Lookup is done on an atomic (per item, not per list) level which means once an item is cached it will decrease the amount of queries necessary to lookup an item (see `Q`) 
- If a `#ask` query makes less query requests then it means other user can access those resources more quickly
- After an object is cached response performance for a large list can improve significantly as shown above

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -561,26 +561,22 @@ $GLOBALS['smwgMainCacheType'] = CACHE_ANYTHING; // Isn't used yet
 
 ###
 # Separate cache type to allow for adding a more responsive cache layer
-# (redis, riak) to operations that require it. It is currently used by:
+# (redis, riak) when requesting value lookups from the SQLStore.
 #
-# - ByBlobStoreIntermediaryValueLookup (#1035)
-#
-# CACHE_NONE = disabled for all features that require this cache layer
+# CACHE_NONE = disabled, uses the standard SQLStore DB access for all
+# lookups
 #
 # @since 2.3
 #
-# @default: CACHE_NONE (legacy); users need to actively enable it in order
+# @default: CACHE_NONE, users need to actively enable it in order
 # to make use of it
 ##
-$GLOBALS['smwgBlobCacheType'] = CACHE_NONE;
+$GLOBALS['smwgValueLookupCacheType'] = CACHE_NONE;
 ##
 
 ###
-# While not absolute necessary (because entities will self-invalidate when a change
-# occurs) it is nevertheless possible to set a limit (in seconds) for items handled
-# by `ByBlobStoreIntermediaryValueLookup`.
-#
-# If set to 0 an item is kept until it is replaced, flushed, or dropped.
+# Declares a lifetime of a cached item for `smwgValueLookupCacheType` until it
+# is removed if not invalidated before.
 #
 # @since 2.3
 ##
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
 		"onoi/message-reporter": "~1.0",
 		"onoi/cache": "~1.2",
 		"onoi/event-dispatcher": "~1.0",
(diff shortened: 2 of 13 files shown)


JeroenDeDauw:
@mkroetzsch one of those things you might want to comment on

PR #1065, opened 2015-06-17
Improve MagicWordsFinder

mwjames:
This ensures that the hook `SMW::Parser::BeforeMagicWordsFinder` works as expected when things like [0, 1] are added in combination with other magicwords.

[0] 'SMW::Parser::BeforeMagicWordsFinder'
[1] https://github.com/SemanticMediaWiki/SemanticCite/blob/master/src/HookRegistry.php#L100-L102


mwjames:
I'm certainly get sick an tired for Composer/packagist.org to f... up the CI run.

Issue #1048, opened 2015-05-28
Colons in property values not properly handled in SMW 2.2.0


jaakristioja:
It seems that properties of type Text don't handle well values containing `::`, e.g. `[[Has IPv6::fc00:123:8000::/64]]` where `Property:Has_IPv6` `[[Has type:Text]]`. Any workaround for this?

mwjames:
Interesting case but the regex to filter the `::` pattern hasn't changed (at least not since 1.8).

The current rule in [0] will split your example in 

```
array(3) {
  [0]=>
  string(32) "[[Has IPv6::fc00:123:8000::/64]]"
  [1]=>
  string(23) "Has IPv6::fc00:123:8000"
  [2]=>
  string(3) "/64"
}
```

and then split it further [1] to a property part
```
array(2) {
  [0]=>
  string(8) "Has IPv6"
  [1]=>
  string(13) "fc00:123:8000"
}
```
and a value part
```
string(3) "/64"
```

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/InTextAnnotationParser.php#L161

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/InTextAnnotationParser.php#L273

> Any workaround for this?

Not that I know of, @kghbln @JeroenDeDauw ?

mwjames:
> Any workaround for this?

For the time being you could use `{{#set:Has IPv6=fc00:123:8000::/64}}`

s7eph4n:
Although it is not very helpful in this case, it works as specified (see last row in the table at [1]), so using ``#set`` really is the way to go.

[1] http://semantic-mediawiki.org/wiki/Help:Properties_and_types#Turning_links_into_properties

jaakristioja:
Interesting. Using `{{#set: }}` helps, thanks! But where is the `|template=` syntax for `{{#set: }}` documented? Or is `template` a special property?

mwjames:
> |template= syntax for {{#set: }}

Only to be used if you want to display the value otherwise annotations are hidden by default using `#set`.

> But where is the |template= syntax for {{#set: }} documented?

- 
https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/releasenotes/RELEASE-NOTES-2.2.0.md#new-features
- https://semantic-mediawiki.org/wiki/Help:Setting_values#Template_parameter

[...]

PR #1066, opened 2015-06-18
Fix :::: processing in InTextAnnotationParser, refs 1048, T32603

mwjames:
Add support for `[[Foo::Bar::Foobar]]` and `[[Foo:::0049 30 12345678]]` (other examples include `[[DOI::10.1002/(SICI)1522-2594(199911)42:5<952::AID-MRM16>3.0.CO;2-S]]`, `[[Has text:::0049 30 12345678/::Foo]]`)

According to [0] the issue of `:::` has been open since  "MW 1.17.0 and SMW 1.5.6 or SMW 1.6.1, probably happening to SMW 1.6.0 and SMW <1.5.6 though not tested.".

[0] https://phabricator.wikimedia.org/T32603
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1048


mwjames:
@kghbln As the originator of https://phabricator.wikimedia.org/T32603 it would be nice if you could give this some testing to avoid any regression on behalf of the `::` syntax. I only added a few examles to `parser-04-02-intext-colon-edge-case-parsing.json` from the top of my head but maybe you can think of other combinations that are worthwhile to test.

@jaakristioja If you feel up for the task you might want to test your `[[Has IPv6::fc00:123:8000::/64]]` example with this PR.

mwjames:
@SemanticMediaWiki/testers Any testers for this one because I'd like to close this issue rather quickly.

kghbln:
Just tested this with a couple of data types and SMW is now "behaving" much better. A note though:
20150625_smw_issue 1066
The first value is the directly stored on the second value is the queried value.

* Code is working as expected as is stores and outputs as is.
* Page and Temperature show the leading colon but do not store it. I think it should rather be "Has improper value for" like telephone number and URL, Date and Email
* Telephone number, URL, Date and Email (last two not pictured) are working as expected and give a "Has improper value for"
* Text is working as expected as it is stored and outputted as is. Here we see that MediaWiki takes the colon to indent.

@mwjames What do you think of Page and Temperature?

mwjames:
> Just tested this with a couple of data types

:+1: 

> Page and Temperature show the leading colon but do not store it. I think it should rather be "Has improper value for" like telephone number and URL, Date and Email

The `InTextAnnotationParser` doesn't make any validation of the input/output as this is part of the `DataValue` ( `_wpg`, `_tem` etc.) validation process. Types like URL, Email have specific validations in place (doing a `preg_match` for valid values) and therefore most likely filter those "unfit" values and return with an error /  "Has improper value for".

Page and temperature do not internally check for any specific value limitation therefore a preceding `:`is internally neglected or transformed to a valid (without `:`) value.

If we want have a "Has improper value for" instead for those types then the `DataValue` checks have to be extended such as:

For the Page type (`SMWWikiPageValue`) add something like:

```
	protected function parseUserValue( $value ) {
		...
		if ( $value !== '' ) {

			if ( $value{0} === ':' ) {
				$this->addError( wfMessage( 'smw-datavalue-first-colon-not-permitted', $value )->inContentLanguage()->text() );
				return;
			}
```

For the Number, Temperature type ( `SMWNumberValue`) add something like:
```
	protected function parseUserValue( $value ) {
		// Set caption
		if ( $this->m_caption === false ) {
			$this->m_caption = $value;
		}

		if ( $value !== '' && $value{0} === ':' ) {
			$this->addError( wfMessage( 'smw-datavalue-first-colon-not-permitted', $value )->inContentLanguage()->text() );
			return;
		}
```

PR #1067, opened 2015-06-19
Fix #set to return ('isHTML' => false)

mwjames:
Use case provided by Jethro Waanders / [0] with `parser-02-03-set-subobject-template.json` containing the integration test.

Regression caused by #758, 

[0] http://wikimedia.7.x6.nabble.com/set-breaks-parsing-value-subobject-after-upgrading-to-SMW-2-2-tp5049139.html


mwjames:
Backported with 3491127a1d3 to 2.2.x

PR #1075, opened 2015-06-26
Additional error formatting rules for ":", #1066 + #1074

### LINKED ISSUE #1074: datatype date / error message duplicated
If an improper value for this datatype is stored the error message gets duplicated. See [smw.o page on this datatype for an example.](https://semantic-mediawiki.org/wiki/Help:Type_Date)

mwjames:
- Added some edge cases mentioned in #1066
- Fixed #1074

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -326,6 +326,7 @@
 	"smw-pa-property-predefined_pval": "[https://www.semantic-mediawiki.org/wiki/Help:Special_property_Allows_value\"$1\"] is a predefined property that can define a list of permissible values to restrict value assignments for a property.",
 	"smw-datavalue-property-restricted-use": "Property \"$1\" was marked for restricted use.",
 	"smw-datavalue-restricted-use": "This datavalue was marked for restricted use. $1",
+	"smw-datavalue-invalid-number": "\"$1\" can not be interpret as number.",
 	"smw-query-condition-circular": "A possible circular condition has been detected in \"$1\".",
 	"smw-sp-types-list": "List of datatypes",
 	"smw-sp-types-default": "\"$1\" is a built-in datatype.",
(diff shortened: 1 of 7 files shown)


mwjames:
@kghbln I tried to add some extra cases as mentioned in  #1066 including:
- `[[Has page:::123|abc]] [[Has page:::xyz]]` should show no extra colon while `[[Has page::::123|abc]] [[Has page::::xyz]]` is going to create an error
- `[[Has number:::123]]` creates an error (same as for quantity and temperature)
- `[[Has date:::1 Jan 1970]]` creates an error and only shows one error tooltip (#1074)

Some testing would be neat before pressing the merge button, any other use case that would come to mind?

kghbln:
My field tests were successful. Also tested via the "#set" parser function. One note though:

When adding ``[[Enthält Text:=:fc00:123:8000::/64]]`` one gets an invalid property ":fc00:123:8000" as well as "Enthält Text" with the value "/64". This should probably be a property of type page called "Enthält Text:=:fc00:123:8000" with the value "/64". The second value ":fc00:123:8000::/64" is a second annotation for property "Enthält Text" so this is ok. Using ``:=`` instead of ``::`` came to my mind since this was a pre SMW 1.0 legacy syntax.

20150627_smw_issue 1075

kghbln:
> When adding [[Enthält Text:=:fc00:123:8000::/64]] one gets

Just tested with current master which should be identical in behaviour with SMW 2.2.1 and found this too. So this is definitively no regression. This case is theoretical one too I guess which could probably only occur for old instances not upgraded - better - gardened properly.

kghbln:
I added the "qqq" on tw.n directly so this does not have to be done here. Avoids a second commit check.

mwjames:
@kghbln Thanks for testing.

> This case is a theoretical one too I guess. It could probably only occur for old instances not upgraded - better - gardened properly

As for `:=`, I'd say that editors that still use this form should run the `ReplaceText` extension to substitute `:=` with `::`.

PR #1083, opened 2015-07-07
Add StringValidator::assertThatStringNotContains

mwjames:
A simple check to prove the absence of sth.


mwjames:
Usage example: https://github.com/SemanticMediaWiki/SemanticCite/compare/f3df719b42f6...e4f636848874

Issue #1087, opened 2015-07-13
Value annotation with # character


mwjames:
The hash character (or number sign) `#` has two distinct meanings in MW and SMW. In MW it represents a fragment of a page linking to an internal page anchor as in `FOO#bar` and SMW adds an additional meaning by translating it into an internal representation of a subobject.

The meaning of `#` in MW is rather arbitrary to the point that it refers to the same entity (a page) while in SMW it refers to a distinct entity as in `FOO#bar` or `FOO#baz` with the same subject `FOO`.

A subobject by definition is created through `#subobject`, any annotation (as page type) using `#` as part of its value declaration is currently treated as belonging to a subobject.

To support an annotation similar to  `[[Foo::ABC#bar]]` where the value part is not treated as subobject and rather identified is part of the whole value string (not as a fragment/subobject indicator) requires a change to `SMWWikiPageValue`.

http://wikimedia.7.x6.nabble.com/SMW-2-2-0-LIKE-operator-broken-td5049986.html

mwjames:
Is not going to be supported for the `_wpg` type.

PR #1088, opened 2015-07-13
Add support for # value annotation, refs #1087

mwjames:
refs #1087

### DIFF
--- a/includes/dataitems/SMW_DI_WikiPage.php
+++ b/includes/dataitems/SMW_DI_WikiPage.php
@@ -67,11 +67,6 @@ class DIWikiPage extends SMWDataItem {
 			throw new DataItemException( "Given namespace '$namespace' is not an integer." );
 		}
 
-		// Check for a potential fragment such as Foo#Bar, Bar#_49c8ab
-		if ( strpos( $dbkey, '#' ) !== false ) {
-			list( $dbkey, $subobjectname ) = explode( '#', $dbkey );
-		}
-
 		$this->m_dbkey = $dbkey;
 		$this->m_namespace = (int)$namespace; // really make this an integer
 		$this->m_interwiki = $interwiki;
@@ -195,9 +190,19 @@ class DIWikiPage extends SMWDataItem {
 	 * @return DIWikiPage
 	 */
 	public static function newFromText( $text, $namespace = NS_MAIN ) {
+
+		$subobjectname = '';
+
+		// Check for a potential fragment such as Foo#Bar, Bar#_49c8ab
+		if ( strpos( $text, '#' ) !== false ) {
+			list( $text, $subobjectname ) = explode( '#', $text );
+		}
+
 		return new self(
 			str_replace( ' ', '_', $text ),
-			$namespace
+			$namespace,
+			'',
+			$subobjectname
 		);
 	}
 
--- a/includes/datavalues/SMW_DV_WikiPage.php
+++ b/includes/datavalues/SMW_DV_WikiPage.php
@@ -101,6 +101,19 @@ class SMWWikiPageValue extends SMWDataValue {
 		}
 
 		if ( $value !== '' ) {
+
+			// If a value declaration (e.g Foo#bar) contains a # we expect that
+			// the value is not a subobject (such an entity would be created
+			// through the #subobject parser) therefore the value is perceived
+			// as an entire title of Foo#bar.
+			if ( strpos( $value, '#' ) !== false && $this->getProperty() !== null ) {
+				$this->m_fragment = '';
+				$this->m_prefixedtext = '';
+				$this->m_id = -1; // unset id
+				$this->m_dataitem = new SMWDIWikiPage( $value, $this->m_fixNamespace );
+				return;
+			}
+
 			if ( $value[0] == '#' ) {
 				if ( is_null( $this->m_contextPage ) ) {
 					$this->addError( wfMessage( 'smw_notitle', $value )->inContentLanguage()->text() );
--- a/tests/phpunit/Integration/Query/Fixtures/query-06-05-page-regex-search.json
+++ b/tests/phpunit/Integration/Query/Fixtures/query-06-05-page-regex-search.json
@@ -18,6 +18,18 @@
 		{
 			"name": "Page-3-without-spaces",
 			"contents": "[[Has regexpage::AnnotationTestWithoutSpaces]]"
+		},
+		{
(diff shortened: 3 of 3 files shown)


JeroenDeDauw:
Seems reasonable to me. @mkroetzsch can you have a look?

mkroetzsch:
Looking at the code, I don't quite understand what this does. It mainly seems to disable the ability to use subobjects in property values. Am I missing something?

mwjames:
> , I don't quite understand what this does. It mainly seems to disable the ability to use subobjects in property values. Am I missing something?

No. If you make an annotation such as `Foo::Bar#123` with the current code the value would be turned into a subobject WP value which is wrong because a subobject with its internal dependencies is expected only to be created using `#subobject`.

Instead the code (is a edge case for the `#` handling) allows for `Foo::Bar#123` to be a normal value annotation of value `Bar#123` where `#` does not transpires into a special meaning (such as an indicator of fragment or subobject).

mkroetzsch:
Ok, I think this is not intended then. A MediaWiki title cannot contain #, so it does not make sense to support such strings as title strings in WikiPage items. On the other hand, it might be useful to link to use a subobject as a property value. Note that one can use subobjects that are named like sections of an MW page, and that it would make a lot of sense to use these subobjects/subsections in property values then. There is no contradiction of the use of # in MW and SMW: they both are used to refer to sub-"objects" of a page.

If the usage of a new subobject DI in a property value fails to satisfy certain "internal dependencies" (not sure what you mean), then this would be a bug that should be fixed. Modifying the syntax parsing process would not be a solution for problems that backend might have with valid DIs. Of course, I don't know if there are any such problems.

mwjames:
> I think this is not intended then. A MediaWiki title cannot contain #, so it does not make sense to support such strings as title strings in WikiPage items.

It is intended to be represent a fragment.

> There is no contradiction of the use of # in MW and SMW: they both are used to refer to sub-"objects" of a page.

I'm fairly clear on the subject matter but as I outlined `#` is an internal convention of MW (and for that matter of SMW, HTML also knows it as fragment) which if a real world object is to define a page value in its meaning of `Foo#bar` as a declarative title (in MW terms it constitutes as a title `Foo` + a fragment `Bar`) then we want to preserve that declaration and not assuming it to be an subobject (which it cannot be in connection with a property-value annotation because we don't allow access to a subobject via value declaration of form of `Foo#bar`).

> subobject DI in a property value fails to satisfy certain "internal dependencies" (not sure what you mean)

The processing chain of  `#subobject` (including its object generation and type validations etc.) which is different from that of the normal `_wpg` DI processing.

I don't have any intention to drag this discussion, the issue was outlined in #1087 and in response this economic change was introduced to allow for `#ask` to support `[[Has regexpage::~ABC#*]]`.

mkroetzsch:
I think there are better ways to address this problem without breaking functionality. If nothing else, this discussion might serve to clarify some misunderstandings.

(1) The fragment part of URNs is an integral part of the URN that makes it refer to a different thing. It is not true that a URN of the form http://example.org/page#fragment really refers to http://example.org/page. When using the URN as a URL, it will be resolved against the fragment-less location, but this is part of the mechanism for resolving (http) URLs, not part of the way in which URNs or URIs are to be interpreted in general.

(2) You refer to different processing in the #subobject parser function. It surely is different, but I don't understand what this has to do with this discussion about the DI implementation. Are you saying that there is any problem with the way in which DIs are now creating subobjects? Or is it merely "different" but still working?

(3) The core of this proposed change seems to be the desire to support certain patterns when asking for pages. That's fine, and a worthwhile goal, but it seems strange to me that this should be achieved by changing how user input is parsed. There seems to be a mismatch here between problem (how the database backend processes ~ patterns) and solution (how we parse page titles). A clean solution would be to split patterns at # and turn them into two LIKE conditions. Another option would be to store a combined string with a # in the middle as a label of subobjects, and to run the LIKE query over this column (somehow I would have expected this to be the case, since we already have a label column in the smw_id table, whcih would seem more appropriate for ~ in general ... but I am not sure now).

(4) The proposed change breaks existing functionality. Even if you don't want this functionality to be in SMW, it should be considered carefully if this is used anywhere before doing this. I don't think that the OP had this in mind when asking for LIKE support to be fixed.

mwjames:
> The fragment part of URNs is an integral part of the URN that makes it refer to a different thing. It is not true that a URN of the form http://example.org/page#fragment really refers to http://example.org/page. When using the URN as a URL, it will be resolved against the fragment-less location, but this is part of the mechanism for resolving (http) URLs, not part of the way in which URNs or URIs are to be interpreted in general.

I think we agree that `http://example.org/page#fragment` is not the same as `http://example.org/page` (or better does not described the same thing) independent of how it is technically resolved or presented to the consumer (user).

mwjames:
> (3) The core of this proposed change seems to be the desire to support certain patterns when asking for pages. That's fine, and a worthwhile goal, but it seems strange to me that this should be achieved by changing how user input is parsed.

Well, I was talking about the economic change not a conceptional change (which would require to store the information differently) and it would require more than a three line change to make it work with the `SPARQLStore`. (which I have already have made a peak for)

It is save to assume that non of us has the desire nor the time to make this happen (from an economic point of view) hence  `[[Has regexpage::~ABC#*]]` is not going to be supported for the `_wpg` type in near future.

@kghbln Maybe something to note on smw.org.

Thanks for the input from all involved parties.

mkroetzsch:
I am surprised that you are closing this so quickly. Even after what you said, it seems valid to ask which of the two features is more important now. We could still kill the linking to subobjects if LIKE across subsection/subobject IDs is more important to the community.

I agree that there are some things to do for making this work on all stores without this "economic" change. One would probably have to do it one by one. However, there should not be an SQL error caused, even if the feature is not fully supported in the near future (I think that's what the OP reported, right?).

mwjames:
> I am surprised that you are closing this so quickly. Even after what you said, it seems valid to ask which of the two features is more important now.

Nothing gets done by starring on open issues. In the past I have seen too often things being discussed but never found its ways into actual implementation. If someone wants to tackle the issue in a reasonable amount of time then I'm the first to cheer and vote.

This issue can be revived (as its content is not deleted) if necessary but until then the feature (or bug or edge case) is not going to be supported.

JeroenDeDauw:
If you close it, then it is harder for people to know there is still something that can be done here though. You need to actually read through it before you can determine that this was not closed due to invalidity.

PR #1097, opened 2015-07-29
Redirect property alias to base property

mwjames:
If `Property:Doi` is an alias for `Property:DOI` then this ensures that it always redirects to the base property.

This avoids confusion where `Property:DOI` specifies extras (`_IMPO`) but when displayed as `Property:Doi` the information is not visible due to `Property:Doi` being not `Property:DOI`.

`Geographic coordinate` (deprecated) forwards to `Geographic coordinates`

This only touches the article view for `SMW_NS_PROPERTY`.

### DIFF
--- a/includes/articlepages/SMW_OrderedListPage.php
+++ b/includes/articlepages/SMW_OrderedListPage.php
@@ -1,5 +1,8 @@
 <?php
 
+use SMW\PropertyRegistry;
+use SMW\DIProperty;
+
 /**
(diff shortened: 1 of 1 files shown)


mwjames:
@kghbln Would be neat if you could test this to see whether I missed something.

Just found this by testing SCI where `Property:DOI` has extras such as `[[Imported from::bibo:doi]]` but when I looked at the `Property:Doi` I could only see the subjects but not the definition since the alias `Doi`  was not redirected to its base `DOI`.

PR #1099, opened 2015-08-01
Fix comparison in SMWQueryResult::getQueryLink()

joelkp:
Was '$caption == false' instead of '$caption === false'.  This did not
affect display of further results link in #ask queries. However, some
strings could be ignored if using the method directly, e.g. from another
extension.


JeroenDeDauw:
Thanks a lot for your contribution!

PR #1100, opened 2015-08-01
Add SMW_SQL3_SMWDELETEIW marker

mwjames:
This picks up the [0] discussion about deleted subjects and properties that have a type annotation.

Applies a two-step approach:
- If a subject is deleted then it is cleared of all references and marked
with `SMW_SQL3_SMWDELETEIW` in the object table
- `ByIdDataRebuildDispatcher` will remove those marked with `SMW_SQL3_SMWDELETEIW` and remove any remaining
references from related property tables

[0] http://wikimedia.7.x6.nabble.com/Quick-question-about-property-deletion-tp5050929p5050932.html

### DIFF
--- a/includes/storage/SQLStore/SMW_SQLStore3.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3.php
@@ -26,6 +26,7 @@ define( 'SMW_SQL3_SMWIW_OUTDATED', ':smw' ); // virtual "interwiki prefix" for o
 define( 'SMW_SQL3_SMWREDIIW', ':smw-redi' ); // virtual "interwiki prefix" for SMW objects that are redirected
 define( 'SMW_SQL3_SMWBORDERIW', ':smw-border' ); // virtual "interwiki prefix" separating very important pre-defined properties from the rest
 define( 'SMW_SQL3_SMWINTDEFIW', ':smw-intprop' ); // virtual "interwiki prefix" marking internal (invisible) predefined properties
+define( 'SMW_SQL3_SMWDELETEIW', ':smw-delete' ); // virtual "interwiki prefix" marking a deleted subject, see #1100
 
 /**
  * Storage access class for using the standard MediaWiki SQL database for
--- a/includes/storage/SQLStore/SMW_Sql3SmwIds.php
+++ b/includes/storage/SQLStore/SMW_Sql3SmwIds.php
@@ -598,12 +598,13 @@ class SMWSql3SmwIds {
 	 *
 	 * @param integer $sid
 	 * @param DIWikiPage $subject
+	 * @param integer|string|null $interWiki
 	 */
-	public function updateInterwikiField( $sid, DIWikiPage $subject ) {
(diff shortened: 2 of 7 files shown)


mwjames:
@SemanticMediaWiki/testers This PR requires some regular testing.

- Create a property with type
- Create some annotations with that property
- Check `Special:Properties` (disable the cache) to appear in the list
- Remove the annotations, delete property, `Special:Properties` should not longer list the property.

includes/storage/SQLStore/SMW_Sql3SmwIds.php
```diff
@@ -599,11 +599,11 @@ public function getPropertyInterwiki( SMWDIProperty $property ) {
 	 * @param integer $sid
 	 * @param DIWikiPage $subject
 	 */
-	public function updateInterwikiField( $sid, DIWikiPage $subject ) {
+	public function updateInterwikiField( $sid, DIWikiPage $subject, $iw = null ) {
```
JeroenDeDauw: docblock was not updated

Please also use an understandable var name

includes/storage/SQLStore/SMW_SQLStore3.php
```diff
@@ -26,6 +26,7 @@
 define( 'SMW_SQL3_SMWREDIIW', ':smw-redi' ); // virtual "interwiki prefix" for SMW objects that are redirected
 define( 'SMW_SQL3_SMWBORDERIW', ':smw-border' ); // virtual "interwiki prefix" separating very important pre-defined properties from the rest
 define( 'SMW_SQL3_SMWINTDEFIW', ':smw-intprop' ); // virtual "interwiki prefix" marking internal (invisible) predefined properties
+define( 'SMW_SQL3_SMWDELETEIW', ':smw-delete' ); // virtual "interwiki prefix" marking a deleted subject
```
JeroenDeDauw: I do not understand this comment

JeroenDeDauw:
:+1:

includes/storage/SQLStore/SMW_SQLStore3.php
mwjames: It is just a virtual marker, not sure what explanation you are looking for.

jaideraf:
@mwjames 
Tested.
When ```$smwgCacheType``` is set ```CACHE_NONE```, it works fine to me.

mwjames:
> Tested.

:+1:  As always, the effort and diligence it is much appreciated. 

> When $smwgCacheType is set CACHE_NONE, it works fine to me.

This is just for the test otherwise one would have to wait until the cache expiry to verify the results.

jaideraf:
>As always, the effort and diligence it is much appreciated.

As always, your effort and diligence, contributing to SMW project (because it is more than just software) are much appreciated, thank you. :+1: 

For all software development, automated tests, quality control... for over the years.

SMW is more and more stable, with better performance, and feature rich.
I follow SMW development since version 1.5 (I guess) and I can see the software getting better over the years, it's awesome! So… thank you.

PR #1102, opened 2015-08-05
Add onoi/http-request

mwjames:


JeroenDeDauw:
Any reason why you are not putting this under the SMW vendor? I'm concerned about adding dependencies that are outside of the projects control while not having a solid user and developer base. And yes I know that I'm the first one that went down that road with `mediawiki/parser-hooks` (or rather the Validator extension back then).

JeroenDeDauw:
Don't get me wrong. I do think it's fantastic you are separating out concerns such as http requests and caching. Thanks!

mwjames:
>  I'm concerned about adding dependencies that are outside of the projects control while not having a solid user and developer base. 

Those are well tested, I'm no expecting huge changes to the interface, and because it has no dependency on MW or SMW.

mwjames:
The cache was picked up but another project https://packagist.org/packages/onoi/cache/dependents which is not big but al least it is being used outside of the MW circle.

JeroenDeDauw:
> Those are well tested, I'm no expecting huge changes to the interface, and because it has no dependency on MW or SMW.

That might be, though my concern is not about code quality. It's about liability in general. Even if the code has high quality, using projects that have just a single maintainer is not ideal.

PR #1103, opened 2015-08-05
ParserParameterProcessor::setParameter

mwjames:


src/ParserParameterProcessor.php
```diff
@@ -140,15 +140,20 @@ public function setParameters( array $parameters ) {
...
 	 */
 	public function addParameter( $key, $value ) {
-		if( $key !== '' && $value !== '' ) {
+
+		if ( $key === '' ) {
+			return null;
+		}
+
+		if ( is_array( $value ) ) {
+			$this->parameters[$key] = $value;
```
JeroenDeDauw: In this case you replace any existing values while for the else part you merge the thing in. Is that intended?

mwjames: > In this case you replace any existing values while for the else part you merge the thing in. Is that intended?

It is kind of blurry because it could be seen as a `set` instead of an `add` but I wanted to avoid to add a new method to avoid:

```
foreach ( $values as $value ) {
  $parserParameterProcessor->addParameter( 'foo', $value )
}
```

```
$parserParameterProcessor->addParameter( 'foo', $values )
```

mwjames: If you think it is a dilution then I just use the foreach.

mwjames: The usage surfaced in https://github.com/SemanticMediaWiki/SemanticCite/pull/12/files#diff-8ef6b654a96f2f28377eb95822b0df12R70

JeroenDeDauw: Ah... yeah, I think it's not good to make a method do two things like that. Better to add a new one then.

src/ParserParameterProcessor.php
```diff
@@ -154,6 +152,18 @@ public function addParameter( $key, $value ) {
 	}
 
 	/**
+	 * @since 2.3
+	 *
+	 * @param string $key
+	 * @param array $values
+	 */
+	public function setParameter( $key, array $values ) {
```
JeroenDeDauw: At first I thought this name was wrong. But it's the existing `addParameter` that is wrong and should be something like `addParameterValue` or `addToParameter`

PR #1116, opened 2015-08-16
Add feature flags to CachedValueLookupStore

mwjames:
Allows a more fine-grained setting as to which lookup should be cached or not.

Default setting is `$GLOBALS['smwgValueLookupFeatures'] = SMW_VL_SD | SMW_VL_PL | SMW_VL_PV | SMW_VL_PS;`

refs #1035, #1063

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -601,6 +601,27 @@ $GLOBALS['smwgValueLookupCacheType'] = CACHE_NONE;
 $GLOBALS['smwgValueLookupCacheLifetime'] = 60 * 60 * 24 * 7; // a week
 ##
 
+##
+# Features expected to be enabled in CachedValueLookupStore
+#
+# Flags that declare a enable/disable state of a supported functionality. If a
+# feature is disabled then a connection is always established to the standard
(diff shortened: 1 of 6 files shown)


mwjames:
> After enabling it I think it started to put too much load and data into my “on the limit” redis-server process.

@kghbln @thingles I have now added some flags to determine which lookups can be supported internally. Given that WikiApiary has a constant influx of bot edits, it might be better not the cache the `SemanticData` hence the setting should look like `$GLOBALS['smwgValueLookupFeatures'] = SMW_VL_PL | SMW_VL_PV | SMW_VL_PS;` (removed `SMW_VL_SD` from the list)

`SMW_VL_PV` (PV = PropertyValues) is the flag that should make queries to respond a bit faster and I hope that we somehow can at least facilitate this feature on WikiApiary without reaching the redis limit.

kghbln:
[Finally documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgValueLookupFeatures&type=revision&diff=57646&oldid=57593)

@mwjames I guess the alternative syntax for bitwise operators to change the setting is fine too.

PR #1126, opened 2015-08-28
Skip InTextAnnoationParser when Factbox content is generated

mwjames:
Avoids that annotations are added by the Factbox content when entities contain something like [[ ... ]] in rendered text elements.

```
[[Has url::http://example.org/api.php?action=ask&query=%5B%5BModification%20date::%2B%5D%5D%7C%3FModification%20date%7Csort%3DModification%20date%7Corder%3Ddesc|api.php?action=ask&query=]]
```

### DIFF
--- a/includes/ContentParser.php
+++ b/includes/ContentParser.php
@@ -43,6 +43,11 @@ class ContentParser {
 	private $enabledToUseContentHandler = true;
 
 	/**
+	 * @var boolean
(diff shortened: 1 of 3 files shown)


mwjames:
PR test failed because of a weak Travis worker (`Your test run exceeded 50 minutes.`).

PR #1127, opened 2015-08-28
Add --shallow-update to rebuildData, refs #1100

mwjames:
The `--shallow-update` option allows to only parse those entities that have a different last modified timestamp compared to that of the last revision.

This enables to run `rebuildData` updates on deleted, redirects, and other out of sync entities and reduces the runtime and memory usage considerably.
- If `--shallow-update` is not used then all pages will be parsed (as before) using the `Parser::parse`.
- `smw_iw` field was added as index to `smw_object_ids` in order for the DB select to use an appropriate index selection. To make us of `smw_iw` index, `update.php` is required to be executed,
- option `-t` was removed from `rebuildData`  since `Type` pages are no longer used
refs #1100


mwjames:
@kghbln Added `--shallow-update` option

PR #1129, opened 2015-08-31
Extend ~* search pattern for _ema/_tel

mwjames:
Enables to search for:
- `[[Has telephone number::~*0123*]]`
- `[[Has email::~*123.org]]`

Also expected to solve mentioned issue in [0].

[0] http://wikimedia.7.x6.nabble.com/Encoded-URLs-appear-to-get-decoded-in-properties-td5050062.html

### DIFF
--- a/includes/datavalues/SMW_DV_URI.php
+++ b/includes/datavalues/SMW_DV_URI.php
@@ -35,9 +35,9 @@ class SMWURIValue extends SMWDataValue {
 	private $m_mode;
 
 	/**
-	 * @var UrlEncoder
+	 * @var boolean
 	 */
-	private $urlEncoder;
+	private $showUrlContextInRawFormat = true;
 
 	public function __construct( $typeid ) {
 		parent::__construct( $typeid );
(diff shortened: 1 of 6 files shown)


mwjames:
PR test failure caused by a weak Travis worker.

mwjames:
@kghbln FYI

kghbln:
@mwjames Just to make sure: This also includes "not like" `!~` like it was done also for [datatype date](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1178)

mwjames:
>This also includes "not like" !~ like it was done also for datatype date

I'm guessing yes but since this is from year ago, I had to look at [0] and can only tell that we test for `[[Has email::~*123.org]]` ([0] should be extended to contain a !~ test case).

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/ByJsonScript/Fixtures/q-0702.json

kghbln:
> I'm guessing yes but since this is from year ago,

Yeah,  I know. I just started a summer 2016 documentation campaign and now these undocumented things pop up.

> ([0] should be extended to contain a !~ test case)

Tried to do so and it actually seems to work out allrighty

PR #1146, opened 2015-09-10
Use 'noparse' => false in #set when template is used, refs #1067

mwjames:
Use the parser itself to handle rendering of template supported output.

### DIFF
--- a/includes/ParserData.php
+++ b/includes/ParserData.php
@@ -213,7 +213,7 @@ class ParserData {
 	 */
 	public function pushSemanticDataToParserOutput() {
 
-		$this->parserOutput->setTimestamp( wfTimestampNow() );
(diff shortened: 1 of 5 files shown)


mwjames:
@kghbln Just saw this thread [0], it should be more clear to users that if `[[Has text:: [[Bar]] ]]` fails to render an annotation simply use `{{#set:|Has text= [[Bar]], [[Has url::example.org]]|template=DisplayValue}}` where the template `DisplayValue:<includeonly>{{{value}}}<includeonly>` turns a silent annotation into a visible one by just adding the template parameter. This allows to have annotations within annotations while displaying the rendered output without any max size restrictions (given that `Has text` is of type `Text`).

[0] https://www.mediawiki.org/wiki/Extension_talk:Semantic_Forms#Max_size_for_properties

Issue #1110, opened 2015-08-11
SPARQLStore / Break down triples into groups


mwjames:
Laurent Alquier wrote:

> That allowed me to copy the very long SPARQL query and try it in Virtuoso. The result was an error about exceeding 1000 lines of SQL generated, which is understandable since these are very long pages.

> I ended up patching up the SPARQLStore.php file with the following changes to detect very large queries and break them up into chunks of smaller numbers of triples. As with many thing, this is a hack but I am hope it will send someone more experienced with the internals of the extension in the right direction.

```php
private function doSparqlFlatDataUpdate( SemanticData $semanticData ) {

	$turtleTriplesBuilder = new TurtleTriplesBuilder(
	$semanticData,
	new RedirectLookup( $this->getConnection() )
	);

	if ( !$turtleTriplesBuilder->hasTriplesForUpdate() ) {
	return;
	}

	if ( $semanticData->getSubject()->getSubobjectName() === '' ) {
	$this->doSparqlDataDelete( $semanticData->getSubject() );
	}

	// Get triples first
	// Break down triples into groups
	// Insert data for each group
	$triplesChunkSize = 80;   
	$triplesString = $turtleTriplesBuilder->getTriples();
	$triplesArray = explode(" .", $triplesString);

	if (count($triplesArray) <= $triplesChunkSize) {
	// Only one chunk - preserve old method
	$this->getConnection()->insertData(
	$turtleTriplesBuilder->getTriples(),
	$turtleTriplesBuilder->getPrefixes()
	);
	} else 
	{
	echo("Found  ".count($triplesArray)." triples (limit is ".$triplesChunkSize.") \n");

	$triplesArrayChunks = array_chunk($triplesArray, $triplesChunkSize);
	echo("Breaking up in  ".count($triplesArrayChunks)." chunks\n");

	$chunkCount = count($triplesArrayChunks);
	$chunkIndex = 1;
	foreach($triplesArrayChunks as $triplesChunk ) {
	echo("Processing chunk\n");
	$chunkString = "";

	foreach($triplesChunk as $triple ) {
	$chunkString .= $triple." .";
	}

	// Remove last separator only for last chunk
	if ($chunkIndex == $chunkCount) { 
	$chunkString = rtrim($chunkString, '.');
	}
	
	$chunkString .= "\n";

	$this->getConnection()->insertData(
	$chunkString,
	$turtleTriplesBuilder->getPrefixes()
	);	
	
[...]

PR #1151, opened 2015-09-15
Removes unmatched "ghost" pages from the ID_TABLE, refs #1110

mwjames:
refs #1100

- Matches (and returns) all ids for the DBKey/NS pattern that can include possible ghost pages
- Removes the special pages cache for properties deleted

### DIFF
--- a/includes/storage/SQLStore/SMW_SQLStore3.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3.php
@@ -52,6 +52,7 @@ class SMWSQLStore3 extends SMWStore {
 	 * will appear empty until they are recomputed.
 	 */
 	const CONCEPT_CACHE_TABLE = 'smw_concept_cache';
+	const CONCEPT_TABLE = 'smw_fpt_conc';
 
 	/**
 	 * Name of the table to store the concept cache in.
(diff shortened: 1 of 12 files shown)


mwjames:
When looking at [0] I came across properties being created multiple times (probably during a DB deadlock or a failed rollback), anyway the previous (before this PR) `delete` action would only find one match and left all remaining properties untouched in the ID_TABLE making it impossible to correct the ID_TABLE without direct SQL intervention.

As of this PR, the `delete` action will mark all IDs as `SMW_SQL3_SMWDELETEIW`  that match the same DBKEY/NS. It furthermore invalidates the cache entry for special pages where properties will no longer be available (those marked as deleted) for display without having to alter the `$GLOBALS['smwgCacheUsage']` setting.

@SemanticMediaWiki/testers Anyone care to test?
@kghbln In case someone encounters [0] then the proposed process is to deleted the property and re-created it.

[0] http://www.vanhamel.nl/vhcodecs/index.php?title=Special:Properties&limit=500&offset=500

image

jaideraf:
Tested. It seems fine to me.

mwjames:
> Tested. It seems fine to me.

:+1: Thanks

thingles:
I get these [all over the place on WikiApiary](https://wikiapiary.com/wiki/Special:Properties). To remove the ghosts I need to delete the Property and recreate them? Is that the right way to read the above?

mwjames:
> To remove the ghosts I need to delete the Property and recreate them? Is that the right way to read the above?

This is one way but if you wait a bit #1216 should also make it possible to avoid having them the be deleted /re-created in the first place by using an algo to determine which of the properties are unused and remove them as well.

In combination with `rebuildData.php --shallow-update` it should remove all those entities that are without reference.

PR #1152, opened 2015-09-16
Use canonical identifiers in export

mwjames:
Use canonical identifiers (Category:, Property:) instead of localized names to ensure that RDF/Query statements are language agnostic and do work even after site/content language is changed.

Use `$GLOBALS['smwgExportToUseCanonicalForm'] = false;` to return to the language dependent export.

The canonical form will always resolve to a proper resource but the lang. form doesn't.

Also fixes the assumption on `ucfirst( trim( $propparts[0] )` for UTF-8 which doesn't work, use `mb_convert_case( $printRequestLabel, MB_CASE_TITLE )` instead to make print request labels work correctly for something like:

```
{{#ask: [[カテゴリ:Sample pages]]
 |?Category
 |?カテゴリ
 |limit=5
}}
```
Queries can now use `category:` as stable indentifer to describe a `rdf:type`.

```
PREFIX wiki: <http://example.org/id/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX swivt: <http://semantic-mediawiki.org/swivt/1.0#>
PREFIX property: <http://example.org/id/Property-3A>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
PREFIX category: <http://example.org/id/Category-3A>
SELECT DISTINCT ?result WHERE {
{ ?result rdf:type category:Sample_pages . }
}
ORDER BY ASC(?resultsk)
OFFSET 0
LIMIT 6
```

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -474,6 +474,15 @@ $GLOBALS['smwgExportBacklinks'] = true; // should backlinks be included by defau
 // http://semanticweb.org/id/FOAF
 ##
 
+##
+# Use canonical identifiers (Category:, Property:) instead of localized names
+# to ensure that RDF/Query statements are language agnostic and do work even
+# after the site/content language changes.
+#
+# @since 2.3
+##
+$GLOBALS['smwgExportToUseCanonicalForm'] = true;
+
 ###
 # The maximal number that SMW will normally display without using scientific exp
 # notation. The deafult is rather large since some users have problems understanding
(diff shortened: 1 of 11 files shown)


mwjames:
@kghbln If users encounter any issues in connection with the `SPARQLStore` because of language dependant RDF export of category uri's then setting `$GLOBALS['smwgExportToUseCanonicalForm'] = false;` should be used until the `rebuildData` has been successfully finalized.

PR #1157, opened 2015-09-17
Fix inverse prefix for predefined properties, refs #1096

### LINKED ISSUE #1096: Concept misinterpreting query statement
I may have stumbled on a bug with the use of 'concepts' (MW 1.24.2 + SMW 2.1.3). Today I created a concept containing the following query statement:

   `[[-Subject::<q>[[-Has subobject::Subject:Subject headings]]</q>]]`

or similarly,

   `[[-Subject.-Has subobject::Subject:Subject headings]]`

While an ask query produces results as expected, the concept comes up empty. The problem appears to be that whichever of the two queries above is used, it is internally understood as:

   `[[-Subject.Has subobject::Subject:Subject headings]]`

as is shown by the concept page. Note the absence of a inverse symbol before "Has subobject".

mwjames:
Fixes #1096

### DIFF
--- a/includes/dataitems/SMW_DI_Property.php
+++ b/includes/dataitems/SMW_DI_Property.php
@@ -185,7 +185,7 @@ class DIProperty extends SMWDataItem {
 			return $prefix . str_replace( '_', ' ', $this->m_key );
 		}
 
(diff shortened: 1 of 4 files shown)


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1157

PR #1158, opened 2015-09-17
Simple _geo support in SPARQLStore

mwjames:
This uses a simple aux property to replicate the `_geo` sortkey in order
to support geo queries in SPARQLStore:

```
{{#ask: [[Has coordinates::52°31'N, 13°24'E]]
 |?Has coordinates
}}
```
```
PREFIX wiki: <http://example.org/id/>
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
PREFIX swivt: <http://semantic-mediawiki.org/swivt/1.0#>
PREFIX property: <http://example.org/id/Property-3A>
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
SELECT DISTINCT ?result WHERE {
?result property:Has_coordinates-23aux "52.516666666667,13.4" .

}
ORDER BY ASC(?resultsk)
OFFSET 0
LIMIT 51
```

A http://www.w3.org/wiki/GeoInfo, http://www.w3.org/2003/01/geo/ encoding
is not part of this PR.

### DIFF
--- a/includes/dataitems/SMW_DI_GeoCoord.php
+++ b/includes/dataitems/SMW_DI_GeoCoord.php
@@ -108,8 +108,7 @@ class SMWDIGeoCoord extends SMWDataItem {
 	 * @see SMWDataItem::getSortKey()
 	 */
 	public function getSortKey() {
-		// Maybe also add longitude here? Or is there a more meaningfull value we can return?
-		return $this->latitude;
+		return $this->latitude . ',' . $this->longitude . ( $this->altitude !== null ? ','. $this->altitude : '' );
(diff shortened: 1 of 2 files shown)


mwjames:
Final piece (I hope) to achieve parity between `SPARQLStore` and `SQLStore`.

JeroenDeDauw:
\o/

PR #1159, opened 2015-09-17
Limit property aux use in Exporter

mwjames:
The use of auxiliary properties during the RDF export is now limited to
properties that really require a helper value other such as
(Has subobject etc.) will now use the standard property declaration.

This PR brings the SPARQL query closer to its #ask query.
```
{{#ask: [[Has subobject.Has page::abc]]
 |format=table
}}

```
```
...
?result property:Has_subobject ?v1 .
{ ?v1 property:Has_page wiki:Abc .
}
...
```

Prior to this PR one would have to know assigned aux declaration.
```
...
?result property:Has_subobject-23aux ?v1 .
{ ?v1 property:Has_page wiki:Abc .
}
...
```
To retain backwards compatibility (= not running `rebuildData.php`),
`$GLOBALS['smwgExportBCAuxiliaryUse'] = true;` should be set.

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -906,3 +906,22 @@ $GLOBALS['smwgEnabledQueryDependencyLinksStore'] = false;
 ##
 $GLOBALS['smwgPropertyDependencyDetectionBlacklist'] = array( '_MDAT', '_SOBJ' );
 ##
+
+###
+# The setting is introduced the keep backwards compatibility with existing Rdf/Turtle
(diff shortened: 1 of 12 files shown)


mwjames:
@kghbln For SPARQL repository users that don't want to run a  `rebuildData.php`, `$GLOBALS['smwgExportBCAuxiliaryUse'] = true;` should be set. This BC setting is planned to vanish with 3.x.

mwjames:
Finally getting this resolved (had this in an old 2.0 branch) but it was a rebase nightmare.

PR #1117, opened 2015-08-17
EmbeddedQueryDependencyLinksStore to track/update pages with embedded…

mwjames:
… queries

The biggest obstacle for #ask queries to display refreshed(new) results is the `ParserCache` by which an article decides whether it should be re-parsed or use the data available in cache.

`EmbeddedQueryDependencyLinksStore` tracks known embedded queries and its dependencies. If a dependency (property, subject etc.) has been altered then `ParserCachePurgeJob` will initiate an invalidation of the `ParserCache` for those targets that have an altered dependency.

Once the `ParserCache` for an article has been invalidated the next view to this article will schedule a re-parse including the embedded #ask queries.

Because performance is crucial during an update/edit/delete, only the first 200 (`ParserCachePurgeJob::CHUNK_SIZE`) entities are updated immediately (online) while the rest is added as batch job in order for the job scheduler to run the task on a regular basis.

`ParserCachePurgeJob` is a light job and can (or better should) run at least each 15 min to minimize the amount of possible backlogs.

- `EmbeddedQueryDependencyLinksStore` to track and store dependencies  in the `smw_query_links` table
- `ParserCachePurgeJob` job the handled the invalidation of selected entities
- Changes to queries are detected using the `SMW::SQLStore::AfterDataUpdateComplete`
hook together with the `CompositePropertyTableDiffIterator`.
- Dependencies are resolved for properties, categories, and concepts (non cached)
- Requires to run __`update.php`__

refs #555, #1108 

https://phabricator.wikimedia.org/T109411

smw-query-links-table

PR #1169, opened 2015-09-23
Ren AsyncJobDispatchManager to DeferredRequestDispatchManager, refs #1130

mwjames:
Rename `AsyncJobDispatchManager` to `DeferredRequestDispatchManager` as we might want to do other things besides Jobs, use deferred instead of async to avoid confusion with HHVM async support in `HH\Asio` [0].

Renamed `$GLOBALS['smwgEnabledAsyncJobDispatcher'] = true;` to `$GLOBALS['smwgEnabledHttpDeferredJobRequest'] = true;`

refs #1117, #1130, 8accffd3

[0] http://docs.hhvm.com/manual/en/book.hack.asio.php

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -859,12 +859,14 @@ $GLOBALS['smwgEnabledEditPageHelp'] = true;
 
 ###
 #
-# Improves performance for selected operations that can be executed
-# in asynchronous processing mode.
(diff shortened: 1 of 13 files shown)


kghbln:
Finally [documented!](https://www.semantic-mediawiki.org/wiki/Help:$smwgEnabledHttpDeferredJobRequest)

PR #1171, opened 2015-09-25
Add SELECT Explain to format=debug

mwjames:
Enables a more detailed analysis of the query plan [0] executed by the `#ask` query.

SELECT explain is displayed where possible otherwise it is "Not supported".

image

`#ask` can be  examined by whether the transformed SQL query performs as expected as in case above or does a rather poor job in comparison below due to a full table scan ( "54908 rows / Using index condition; Using temporary; Using filesort") and the use of `filesort`.

image

[0] https://dev.mysql.com/doc/refman/5.0/en/explain-output.html

### DIFF
--- a/src/Query/DebugOutputFormatter.php
+++ b/src/Query/DebugOutputFormatter.php
@@ -29,10 +29,10 @@ class DebugOutputFormatter {
 
 		if ( $query instanceof Query ) {
 			$preEntries = array();
-			$preEntries['Generated Wiki-Query'] = '<pre>' . str_replace( '[', '&#x005B;', $query->getDescription()->getQueryString() ) . '</pre>';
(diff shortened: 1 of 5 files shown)


mwjames:
At the same time improved:
```
80.53% ↗ 84.19% QueryEngine improved
96.82% ↗ 98.69% QuerySegmentListProcessor improved
```

mwjames:
@kghbln FYI; SQLite doesn't create any EXPLAIN reports and I have no idea whether Postgres makes use of it or not.

Interpretation of the output can be found under [0, 1].

[0] https://dev.mysql.com/doc/refman/5.0/en/explain-output.html
[1] https://mariadb.org/explain_analyzer/analyze/

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1171

PR #1172, opened 2015-09-25
Add @category to #subobject

mwjames:
Add `@category` to #subobject

```
{{#subobject:
 |Has page=Example/0301/1
 |@category=ABC;123|+sep=;
}}
```

While subobjects with categories can be queried as any other object, they will not appear in an individual `Category: ...` list as subobject/page as there are a SMW and not a MW concept (which requires a real mw page id to be listed as member of a category when viewed via `Category:...`).

```
{{#ask: [[Category:ABC]][[Has page::Example/0301/1]]
 |?Has page
 |?Category
}}
```

### DIFF
--- a/includes/parserhooks/SubobjectParserFunction.php
+++ b/includes/parserhooks/SubobjectParserFunction.php
@@ -16,9 +16,19 @@ use Parser;
  */
 class SubobjectParserFunction {
 
-	// Fixed identifier that describes the sortkey annotation
-	// parameter
+	/**
+	 * Fixed identifier that describes the sortkey annotation parameter
+	 */
 	const PARAM_SORTKEY = '@sortkey';
+
+	/**
+	 * Fixed identifier that describes the subobject category parameter.
+	 *
+	 * We keep it as a @ fixed parameter since the standard annotation would
+	 * require special attention (Category:;instead of ::) when annotating a
+	 * category
+	 */
+	const PARAM_CATEGORY = '@category';
 
 	/**
 	 * @var ParserData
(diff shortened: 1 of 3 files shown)


mwjames:
@kghbln FYI

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1172

JeroenDeDauw:
Thanks for documenting this in this PR and on the sandbox.

The release notes have this

> Added @category as fixed parameter to #subobject

What is a "fixed parameter"?

mwjames:
> What is a "fixed parameter"?

"fixed parameter" as being fixed in its usage and behaviour. `@category` as parameter has a pre-defined meaning (similar to `@sortkey`) when used in connection with the `#subobject` parser function.

JeroenDeDauw:
Sorry, I don't get it. What is more fixed for this parameter, then say for instance for the limit one? Can you give a concrete example?

mwjames:
> I don't get it. 

Normally when you use `#subobject` and you add a parameter such as `Foo`or `Bar` as FIRST parameter then it is interpret as property with the with intended naming of `Foo` or 'Bar`.

If you use `@category` (or `@sortkey`) as FIRST parameter then this premise (as intended naming) is no longer valid in terms of creating a property with name `@category` but instead `@category` is assigned to property `_INST` (or `@sortkey` -> `_SKEY`) which makes those parameters special or fixed.

```
{{#subobject:
 |Has page=Example/0301/1
 |@category=ABC;123|+sep=;
}}
```

PR #1174, opened 2015-09-26
Add 'SMW::Browse::BeforeIncomingPropertyValuesFurtherLinkCreate' hook

mwjames:
When injecting other incoming properties (those of a text type) via `AfterIncomingPropertiesLookupComplete` but more than the `incomingvaluescount = 21`
are accounted for, the further link to `Special:SearchByProperty` might not be sufficient hence to possibility to redirect via `SMW::Browse::BeforeIncomingPropertyValuesFurtherLinkCreate`
to another endpoint.


mwjames:
@kghbln The also adds the help button (right corner) with target [0] to be available from MW 1.25 onwards.

image

[0] https://semantic-mediawiki.org/wiki/Help:Special:Browse

Issue #1165, opened 2015-09-21
Inline queries not fully supported for PostgreSQL database?


yoonghm:
I have the following software and extension
* **MediaWiki  1.25.2**
* **PHP  5.6.13RC1 (fpm-fcgi)**
* **PostgreSQL 9.4.4**
* **Semantic MediaWiki  2.2.2**

I have the following pages:

**11111111A**
```
* [[Tutorial class is::PA11]]
* [[Advisor is::Teacher_1]]
* [[Category:Student]]
```

**22222222A**
```
* [[Tutorial class is::PA11]]
* [[Advisor is::Teacher_1]]
* [[Category:Student]]
```

**33333333A**
```
* [[Tutorial class is::PA11]]
* [[Advisor is::Teacher_1]]
* [[Category:Student]]
```

**PA11**
```
{{#ask: 
 [[Category:Student]]
 [[Tutorial class is::PA11]]
 |?Advisor is
}}
```

However, page **PA11** is rendered as

                    | Advisor
-----------------|----------------
11111111A  | Teacher_1

Using debug, I can see that the query is
```
SELECT DISTINCT t6.smw_title     AS t,
                t6.smw_namespace AS ns
  FROM smw_object_ids AS t6
 INNER JOIN smw_fpt_inst AS t1
    ON t6.smw_id = t1.s_id 
 INNER JOIN smw_di_wikipage AS t3 
    ON t1.s_id = t3.s_id 
 WHERE (t1.o_id = '53' AND (t3.p_id = '58' AND t3.o_id = '54'))
 ORDER BY t6.smw_sortkey ASC LIMIT 55 OFFSET 0;
```

By looking at the PostgreSQL database, I see that

Table **smw_object_ids**

sms_id  | smw_namespace | smw_title
-----------|-------------------------|--------------
    50     |        0                    |
    51     |        0                    | 11111111A
    52     |        0                    | Teacher_1
    53     |       14                   | Student
    54     |        0                    | PA11
    55     |        0                    | PA11
    56     |        0                    | PA11
    57     |      102                  | Advisor_is
    58     |      102                  | Tutorial_class_is
    59     |        0                    | 22222222A
    60     |        0                    | 33333333A
                                        
Table **smw_fpt_inst**

s_id  | o_id
-------|------
  51  | 53

Table **smw_di_wikipage**

s_id  | p_id  | o_id
-------|--------|-------
  51  | 57     | 52
  51  | 58     | 54

[...]

PR #1060, opened 2015-06-12
Apply changes from refs #19

mwjames:
#19, @vitalif applied the changes and added some integration tests.

```
{{#ask:
 <q>[[Has page-1::Value 1]] OR [[Has page-2::Value 1]]</q> AND
 <q>[[Has text-1::Value 1]] OR [[Has text-2::Value 1]]</q> AND
 <q>[[Has text-1::Value 3]] OR [[Has page-1::Value 3]]</q>
 |format=debug
}}
```
Before the change the above query would have not been resolved while after the change we get something like:
```
Debug Output by SQLStore
Generated Wiki-Query
  [[Has page-1::Value 1]]  OR  [[Has page-2::Value 1]]     [[Has text-1::Value 1]]  OR  [[Has text-2::Value 1]]     [[Has text-1::Value 3]]  OR  [[Has page-1::Value 3]]  
Query Metrics
Query-Size:12
Query-Depth:1
SQL Query
SELECT DISTINCT t22.smw_title AS t,t22.smw_namespace AS ns FROM `abcsmw_object_ids` AS t22 INNER JOIN `abct1` AS t1 ON t22.smw_id=t1.id INNER JOIN `abct8` AS t8 ON t1.id=t8.id INNER JOIN `abct15` AS t15 ON t1.id=t15.id ORDER BY t22.smw_sortkey ASC LIMIT 55 OFFSET 0;
Auxilliary Tables Used
Temporary table t1
  INSERT IGNORE INTO `abct1` SELECT t2.s_id FROM `abcsmw_di_wikipage` AS t2 WHERE t2.p_id='0' AND t2.o_id='2387'
  INSERT IGNORE INTO `abct1` SELECT t5.s_id FROM `abcsmw_di_wikipage` AS t5 WHERE t5.p_id='0' AND t5.o_id='2387'
Temporary table t8
  INSERT IGNORE INTO `abct8` SELECT t9.s_id FROM `abcsmw_di_wikipage` AS t9 WHERE t9.p_id='0' AND t9.o_id='2387'
  INSERT IGNORE INTO `abct8` SELECT t12.s_id FROM `abcsmw_di_wikipage` AS t12 WHERE t12.p_id='0' AND t12.o_id='2387'
Temporary table t15
  INSERT IGNORE INTO `abct15` SELECT t16.s_id FROM `abcsmw_di_wikipage` AS t16 WHERE t16.p_id='0' AND t16.o_id='2391'
[...]

PR #1059, opened 2015-06-12
Additional disjunction tests from T31866

mwjames:
- Incorporate examples from https://phabricator.wikimedia.org/T31866
- Fix usage of `[[Has page::~*a*||~*A*]]` for SPARQLStore

Issue #556, opened 2014-10-02
Missing parentheses in SQL for disjunction when not using temporary tables


mkroetzsch:
SMW supports disjunction in queries using || in property values. If the database supports temporary table creation, this will be used to create a temporary table to join with. If temporary table creation is not supported, then the disjunction is expanded into a WHERE clause using "OR". The bug is that this OR-condition is missing parentheses ( ) around the ORs.

Example (one must use pages and properties that exist on the test wiki):
```
{{#ask: [[SMWCon Spring 2014]] [[Has conference type::SMWCon||Semantic MediaWiki||Another page]] | format=debug}}
```
might show an SQL query like:
```
SELECT DISTINCT t4.smw_title AS t,t4.smw_namespace AS ns
FROM `smw_object_ids` AS t4 INNER JOIN `smw_di_blob` AS t2 ON t4.smw_id=t2.s_id
WHERE
 ( (t2.o_hash='Article') OR (t2.o_hash='Incollection')
   OR (t2.o_hash='Inproceedings') AND t2.p_id='504' AND t2.s_id='1302'
 )
ORDER BY t4.smw_sortkey ASC LIMIT 51 OFFSET 0;
```
Note that the WHERE clause is of the form "a OR b OR c AND d AND e" which means "a OR b OR (c AND d AND e)". The correct form would be "(a OR b OR c) AND d AND e".

This causes unwanted results (since the AND part is only applied to one case, the other cases are weakened). Also, the results will be hard to interpret for the user since they may appear quite random (in the case above, you would get "SMWCon Spring 2014" if it has conference type "Another page" plus all pages that have "SMWCon" or "Semantic MediaWiki" as value of *any* wikipage property).

The fix should be very easy: just add parentheses around the affected code.

The bug was observed in SMW 2.0. The bug might affect people who run into #455 and try to disable temporary tables to avoid this.

mkroetzsch:
Update: This still can be fixed as I described above in SMW 2.1 (the lines are a little further down, that's all). Not sure why the "easy" tag was removed. All that is needed to fix this bug is to copy the exact lines that I have already given above into the file.

mwjames:
> his still can be fixed as I described above in SMW 2.1 (the lines are a little further down, that's all). Not sure why the "easy" tag was removed. All that is needed to fix this bug is to copy the exact lines that I have already given above into the file.

I know, no panic.

[...]

PR #1175, opened 2015-09-26
Use SELECT DISTINCT, refs #1165

mwjames:
Based on the [0] discussion which should avoid duplicate entries and eliminate `ERROR:  duplicate key value violates unique constraint, Key (id)=(146) already exists` in postgres.

Re-enabled skipped tests:
- q-0102 regex blob queries.json
- q-0901 [#1057] common disjunction conjunction queries.json
- q-0902 [#556] text typed disjunctive queries.json
- q-0903 [#1059] regex disjunctive queries.json
- q-0904 [#1060] disjunctive subproperty queries.json

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1165#issuecomment-143191808

### DIFF
--- a/src/SQLStore/QueryEngine/QuerySegmentListProcessor.php
+++ b/src/SQLStore/QueryEngine/QuerySegmentListProcessor.php
@@ -204,7 +204,7 @@ class QuerySegmentListProcessor {
 					if ( $subQuery->joinTable !== '' ) {
 						$sql = 'INSERT ' . 'IGNORE ' . 'INTO ' .
 						       $db->tableName( $query->alias ) .
-							   " SELECT $subQuery->joinfield FROM " . $db->tableName( $subQuery->joinTable ) .
+							   " SELECT DISTINCT $subQuery->joinfield FROM " . $db->tableName( $subQuery->joinTable ) .
 							   " AS $subQuery->alias $subQuery->from" . ( $subQuery->where ? " WHERE $subQuery->where":'' );
 					} elseif ( $subQuery->joinfield !== '' ) {
 						// NOTE: this works only for single "unconditional" values without further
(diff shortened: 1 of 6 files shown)


mwjames:
@yoonghm FYI and as discussed this fixes the `duplicate key value violates unique constraint` on SMW master.

mwjames:
For postgres:

Before
```
OK, but incomplete or skipped tests!
Tests: 3071, Assertions: 9128, Skipped: 37.
```

After

```
OK, but incomplete or skipped tests!
Tests: 3071, Assertions: 9373, Skipped: 32.
```

yoonghm:
It is great! I am glad that can help even I am still not clear on SMW. Not much information from the web except to read the code.

For those test you performed, did you see the problem [0] if your perform Clone Database test?


[0] https://phabricator.wikimedia.org/T113445

mwjames:
> It is great! I am glad that can help even I am still not clear on SMW. Not much information from the web except to read the code.

It is hard for us to present information in a concise way because we don't really know how and what people expect to read about the API or SMW internals.

> For those test you performed, did you see the problem [0] if your perform Clone Database test?
[0] https://phabricator.wikimedia.org/T113445

We are running those postgres tests on 1.19 [1] which doesn't have this problem but [0] is also the reason why we cannot use 1.25 because of the issue you encountered. For now we test postgres on 1.19 until your fix is available.

Would be great if you could figure out a solution for [2, #781].

I'm also not sure why we skipped [3] for postgres but maybe you can remove the skip marker and see what happens.

[1] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/82366187

[2] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1165#issuecomment-143312628

[3] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/MediaWiki/Hooks/TitleMoveCompleteIntegrationTest.php#L102-L103

PR #1178, opened 2015-09-28
~/!~ support for the _dat type

mwjames:
Adds ~/!~ support for the _dat type and removes query related dependencies from the DV components and moves it into a separate `SMW\Deserializers\DVDescriptionDeserializer` NS.

- Match all subjects for Dec 2001

```
{{#ask: [[Has date::~ Dec 2001]]
 |?Has date
}}
```

- Match all subjects that have not been altered at the 15 Jun 2006

```
{{#ask: [[Modification date::!~ 15 Jun 2006]]
 |?Modification date
}}
```

- Match all subjects that have been closed in Dec 2007 (using a record)

```
{{#ask: [[Has status record::~Dec 2007;closed]]
 |?Has status record
}}
```

### DIFF
--- a/includes/datavalues/SMW_DV_Record.php
+++ b/includes/datavalues/SMW_DV_Record.php
@@ -76,6 +76,7 @@ class SMWRecordValue extends SMWDataValue {
 			}
 
 			$values[$valueIndex] = str_replace( "-3B", ";", $values[$valueIndex] );
+			$beforePrepareValue = $values[$valueIndex];
 
 			if ( $queryMode ) { // special handling for supporting query parsing
 				$comparator = SMW_CMP_EQ;
@@ -92,11 +93,7 @@ class SMWRecordValue extends SMWDataValue {
 					if ( $queryMode ) {
 						$subdescriptions[] = new SMWSomeProperty(
 							$diProperty,
-							new SMWValueDescription(
-								$dataValue->getDataItem(),
-								$dataValue->getProperty(),
-								$comparator
-							)
+							$dataValue->getQueryDescription( $beforePrepareValue )
 						);
 					} else {
 						$semanticData->addPropertyObjectValue( $diProperty, $dataValue->getDataItem() );
--- a/includes/datavalues/SMW_DataValue.php
+++ b/includes/datavalues/SMW_DataValue.php
@@ -9,6 +9,7 @@
  */
 use SMW\DataValueFactory;
 use SMW\Query\QueryComparator;
+use SMW\Deserializers\DVDescriptionDeserializerFactory;
 
 /**
  * Objects of this type represent all that is known about a certain user-provided
@@ -415,22 +416,7 @@ abstract class SMWDataValue {
 ///// Query support /////
 
 	/**
-	 * Create an SMWDescription object based on a value string that was entered
-	 * in a query. Turning inputs that a user enters in place of a value within
-	 * a query string into query conditions is often a standard procedure. The
-	 * processing must take comparators like "<" into account, but otherwise
-	 * the normal parsing function can be used. However, there can be datatypes
-	 * where processing is more complicated, e.g. if the input string contains
-	 * more than one value, each of which may have comparators, as in
-	 * SMWRecordValue. In this case, it makes sense to overwrite this method.
-	 * Another reason to do this is to add new forms of comparators or new ways
-	 * of entering query conditions.
-	 *
-	 * The resulting SMWDescription may or may not make use of the datavalue
-	 * object that this function was called on, so it must be ensured that this
-	 * value is not used elsewhere when calling this method. The function can
-	 * return SMWThingDescription to not impose any condition, e.g. if parsing
-	 * failed. Error messages of this SMWDataValue object are propagated.
+	 * @see DataValueDescriptionDeserializer::deserialize
 	 *
 	 * @note Descriptions of values need to know their property to be able to
 	 * create a parsable wikitext version of a query condition again. Thus it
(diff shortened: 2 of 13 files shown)


JeroenDeDauw:
Cool \o/

@mkroetzsch thoughts on this? I'm curious why like was not implemented for dates before - some concern with how the behavior should work or just something that was not added due to lack of time?

src/DataValues/DescriptionEncoders/TimeValueDescriptionEncoder.php
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * @private
+ *
+ * @license GNU GPL v2+
+ * @since 2.3
+ *
+ * @author mwjames
+ */
+class TimeValueDescriptionEncoder extends DescriptionEncoder {
```
JeroenDeDauw: When I read this name I figured I'd do the opposite: construct a string from a Description object. How about something like DescriptionDeserializer?

tests/phpunit/Unit/DataValues/DescriptionEncoders/DvDescriptionEncoderFactoryTest.php
```diff
@@ -0,0 +1,72 @@
...
+			'\SMW\DataValues\DescriptionEncoders\DvDescriptionEncoderFactory',
+			DvDescriptionEncoderFactory::getInstance()
+		);
+	}
+
+	/**
+	 * @dataProvider dataValueProvider
+	 */
+	public function testCanConstructDescriptionEncoder( $dataValue, $expected ) {
+
```
JeroenDeDauw: I think you are better of just creating a method for each of those assertions. There will some some duplication then, though very non problematic one, and less expensive than the complexity and indirection added here.

mkroetzsch:
> @mkroetzsch thoughts on this? I'm curious why like was not implemented for dates before - some concern with how the behavior should work or just something that was not added due to lack of time?

Not just time, no. The ~ functions refer to a string, so the obvious question is: which string? You can enter dates in SMW in many formats, and get them back in many formats. The internal string uses its own format that may partly agree with some formatting options, but I don't know if this is documented anywhere. In fact, it is not even the same across implementations. The SPARQL backend would store dates in a different way, not using strings at all. And if it were strings, they are likely not the same as in the SQL table (at least I don't think I implemented XML Schema serialization for SQL).

I think a more useful feature would be to support query operators for dates to select by month or by year, or even by day of week. This type of information is a clearly defined part of the date type, while the string just is an implementation detail of the current SQL store. Using a suitable string format, many of the date operators could be supported by SQL LIKE functions internally. SPARQL still may not have a good way of doing this, but many (most?) RDF databases have some builtins to access day, month, etc. of a date.

mwjames:
Just to clarify (as given by the attached tests which runs on both SQLStore and SPARQLStore) those do pass as we don't compare strings (on both SQL and SPARQL) but to use the numerical representation in an interval to transform ~Dec 2015 (which is a string by definition for the input) into a lower/upper boundary.  So,  `SQL LIKE` isn't used here at all as this isn't applied on how dates are compared for either the SQLStore or the SPARQLStore.

>I think a more useful feature would be to support query operators for dates to select by month or by year, or even by day of week. 

You do that be just declaring ~ Dec 2010 (all of 2010) or ~ 15 Dec 2010 (all of 15 Dec). ~ becomes a representation for similar (or in natural language like) and with the specified precision, the starting condition lays the groundwork in how narrow or wider it is expected to be similar for a selected date.

In natural language if declaring a ~ 15 Dec 2010 then I'd expected those to match 15 Dec 2010*. Of course you could do this all be hand and specifying an interval (which is what happens in the background) but this is tedious, unnecessary, and should be done by the software not by the user  who just wants to have an issue solved.

As for the RDF (and respectively SPARQL), please remember that we represent a date with two statements and the aux representation is solely used for querying, so comparing is math not a string operation.
```
<swivt:wikiPageModificationDate rdf:datatype="http://www.w3.org/2001/XMLSchema#dateTime">2015-09-28T02:32:10Z</swivt:wikiPageModificationDate>
<property:Modification_date-23aux rdf:datatype="http://www.w3.org/2001/XMLSchema#double">2457293.6056713</property:Modification_date-23aux>
```

mwjames:
There are different opinions about the `~` meaning outside of its mathematical application but in general it indicates an approximated value. In case of dates an interval is chosen to make such approximation to hold true for a given precision. 

[0] https://en.wikipedia.org/wiki/Tilde#Range
[1] https://www.quora.com/What-is-the-proper-use-of-the-tilde-as-punctuation-in-a-sentence
[2] http://encyclopedia2.thefreedictionary.com/tilde

mkroetzsch:
@mwjames Thanks for clarifying. I should read from the beginning before answering comments. Sounds like a very useful feature to have.

JeroenDeDauw:
Thanks for clarifying @mkroetzsch 

I agree that the meaning of `~` in the ask language should not necessarily be bound to its meaning in SQL.

JeroenDeDauw:
@mwjames +2 apart from the inline comments I made

JeroenDeDauw:
@mwjames I'm happy to address my comments myself. Do you want me to do that?

mwjames:
> I'm happy to address my comments myself. Do you want me to do that?

No, I made the changes just haven't pushed it yet.

Just thinking about the NS whether to make it `SMW\Deserializers\DataValueDescriptionDeserializer\DescriptionDeserializerFactory;` or `SMW\Deserializers\DVDescriptionDeserializer\DescriptionDeserializerFactory`;

JeroenDeDauw:
I'd go for `SMW\Deserializers\DescriptionDeserilaizerFactory` and the deserializers at `SMW\Deserializers\DescriptionDeserilaizers\*`. No need to put the factory in the `DescriptionDeserilaizers` namespace, and reason to not do this: its not a DescriptionDeserilaizer. Not sure why you'd include this DataValue part. We refer to these things as descriptions, not data value descriptions.

Actually given this is all already in the Deserializers NS, why not `SMW\Deserializers\Descriptions\*`?

mwjames:
@kghbln FYI

kghbln:
This is not [documented on wiki](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ASelecting_pages&type=revision&diff=45361&oldid=45356).

PR #1182, opened 2015-09-30
Use format as specified by #ask to set further link

mwjames:
```
{{#ask: [[Category:Sample pages]]
 |format=category
}}
```

Points to `format=category` and not to `format=table` (by default).


mwjames:
@kghbln FYI

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1182

PR #1183, opened 2015-09-30
Avoid "Error: 1068 Multiple primary key defined"

mwjames:
This appears when columns with primary keys already specified are tried for an alteration.

### DIFF
--- a/includes/storage/SMW_SQLHelpers.php
+++ b/includes/storage/SMW_SQLHelpers.php
@@ -327,6 +327,11 @@ EOT;
 		} elseif ( $currentFields[$name] != $type ) {
 			self::reportProgress( "   ... changing type of field $name from '$currentFields[$name]' to '$type' ... ", $reportTo );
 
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
How did you find this?

mwjames:
Because I was testing to move the ID limit from INT(8) to INT(14) and when running the update script it just blow up with "Error: 1068 Multiple primary key defined" and after some google time I found that best to avoid that sort of thing is by dropping the PRIMARY first before altering the signature (PRIMARY will be recreated anyway).

```
	case 'id':
	return $wgDBtype == 'postgres' ? 'SERIAL' : ($wgDBtype == 'sqlite' ? 'INTEGER' :'INT(14) UNSIGNED'); // like page_id in MW page table

```

JeroenDeDauw:
Ok, thanks for explaining and fixing!

Odd, we have sqlite and postgres labels, but no mysql one :)

PR #1194, opened 2015-10-07
Use FOO_BAR naming for constant

JeroenDeDauw:
@mwjames if this is used in extensions or something we should probably not do this, or leave an alias


mwjames:
>  if this is used in extensions or something we should probably not do this, or leave an alias

Not that I know of and if so its time we let it crash because those should not be used outside SMW-core. If someone uses it (for whatever reasons) then its time for a public interface but until then ...

Issue #1207, opened 2015-10-12
NotLike (!~) comparator in combination with allowed values is not recognized


mwjames:
[0] shows the issue in combination with `!~*en*` usage.

The issue is in [`SomeValueDescriptionDeserializer`](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Deserializers/DVDescriptionDeserializer/SomeValueDescriptionDeserializer.php#L44-L47)

```php
		if( $comparator == SMW_CMP_LIKE ) {
			// ignore allowed values when the LIKE comparator is used (BUG 21893)
			$this->dataValue->setUserValue( $value, false, true );
		} else {
```

and should be changed to

```php
		if( $comparator == SMW_CMP_LIKE || $comparator == SMW_CMP_NLKE ) {
			// ignore allowed values when the LIKE comparator is used (BUG 21893)
			$this->dataValue->setUserValue( $value, false, true );
		} else {
```

[0] http://sandbox.semantic-mediawiki.org/wiki/NotLikeComparatorWithAllowedProperties

## See also 

https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/3507f84

mwjames:
Also add test cases for (~) as seen in http://sandbox.semantic-mediawiki.org/wiki/LikeComparatorWithAllowedProperties

JeroenDeDauw:
hah, funny bug. Nice catch!

PR #1210, opened 2015-10-14
Fix NLKE for allowed values, refs #1207

mwjames:
refs #1207


JeroenDeDauw:
Don't forget tot update the rel notes ;p

Issue #498, opened 2014-08-17
SQLStore support a way to clean up the `smw_object_ids` table


nischayn22:
There are a lot of ways one can introduce invalid entries in SMW. 

Example: Assign a property-value for a property with no user-defined type. The default type is Page, this will lead to the values assigned being put into the smw_object_ids table. Now change the type to Text. The ids will still be there. 

I think even deleteSubject also currently doesn't support this. The comment in the code says
TODO: Possibly delete ID here...

I am also not sure if doing a database clean + rebuild will fix these issues, otherwise that is a good workaround.

mwjames:
Fixed with #1216

PR #1216, opened 2015-10-16
Add PropertyTableOutdatedReferenceDisposer, refs #498

mwjames:
When a property type is changed, a property is removed, or other object values are deleted chances are that some ID's remain (#498) in the ID_TABLE and to avoid a pile of garbage being collected in the ID_TABLE, `PropertyTableOutdatedReferenceDisposer` will check on ID's whether they can safely be removed or not (during the `rebuidData` run).

For the quick sweep run (without running the memory intensive `Parser::parse`) use `rebuildData.php --shallow-update`.

refs #498, #1100, #1151

Test case 1:

- Create a subject `Foo` with `[[SomeProperty:123]]`
- Delete subject `Foo`
- `Special:Properties` will list `SomeProperty`
- Run `rebuildData.php --shallow-update` and `SomeProperty` should no longer be displayed (== no longer exists in the ID_TABLE) as long it has no other references

Test case 2:

- Create a subject `Foo` with  `[[SomeProperty: ]]` (being empty hence `Has improper value for` annotation is added)
- Remove  `[[SomeProperty: ]]` from the subject
- `Special:Properties` will still list `SomeProperty`
- Run `rebuildData.php --shallow-update` and `SomeProperty` should no longer be displayed as long it has no other references

### DIFF
--- a/src/SQLStore/ByIdDataRebuildDispatcher.php
+++ b/src/SQLStore/ByIdDataRebuildDispatcher.php
@@ -24,9 +24,14 @@ use Hooks;
 class ByIdDataRebuildDispatcher {
 
 	/**
-	 * @var Store
+	 * @var SQLStore
 	 */
 	private $store = null;
+
+	/**
+	 * @var PropertyTableOutdatedReferenceDisposer
+	 */
+	private $propertyTableOutdatedReferenceDisposer = null;
 
 	/**
 	 * @var integer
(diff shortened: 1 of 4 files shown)


mwjames:
@SemanticMediaWiki/testers We just released 2.3 but for the 2.4@alpha season it is already time for testing (see description for the test cases).

The main objective here is that entities (object values, user-defined properties) without reference are removed from the ID_TABLE.

## Question

image

Some properties show 0 usage but are not deleted, why? Because some of them still maintain references to each other!

image

kghbln:
Finally [documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AMaintenance_script_%22rebuildData.php%22&type=revision&diff=45254&oldid=45253)

Issue #1222, opened 2015-10-20
RuntimeException / Unknown column 't1.o_id' in 'where clause'


mwjames:
This just appeared during a different test for when property `HasWorkManifested` was defined as `[[Has type::Text]]` and executed on:

```
[[HasWorkManifested::ABC]]
{{#show:{{FULLPAGENAME}}|?-HasWorkManifested|link=none}}
{{#show:{{FULLPAGENAME}}|?HasWorkManifested|link=none}}
```

`smw_di_blob` doesn't have a `o_id` column so its obvious that the query must fail.

```
Query: SELECT DISTINCT smw_title, smw_namespace, smw_iw, smw_sortkey, smw_subobject FROM `abcsmw_object_ids` INNER JOIN `abccsmw_di_blob` AS t1 ON t1.s_id=smw_id WHERE t1.p_id='10201' AND t1.o_id='10196' 
Function: SMWSQLStore3Readers::getPropertySubjects
Error: 1054 Unknown column 't1.o_id' in 'where clause' (localhost)

#0 ...\includes\db\Database.php(1205): DatabaseBase->reportQueryError('Unknown column ...', 1054, 'SELECT DISTINC...', 'SMWSQLStore3Rea...', false)
#1 ...\includes\db\Database.php(1726): DatabaseBase->query('SELECT DISTINC...', 'SMWSQLStore3Rea...')
#2 ...\extensions\SemanticMediaWiki\src\MediaWiki\Database.php(179): DatabaseBase->select('`abcsmw_object_...', 'DISTINCT smw_ti...', 't1.p_id='10201'...', 'SMWSQLStore3Rea...', Array, Array)
#3 ...\extensions\SemanticMediaWiki\includes\storage\SQLStore\SMW_SQLStore3_Readers.php(461): SMW\MediaWiki\Database->select('`abcsmw_object_...', 'DISTINCT smw_ti...', 't1.p_id='10201'...', 'SMWSQLStore3Rea...', Array)
#4 ...\extensions\SemanticMediaWiki\includes\storage\SQLStore\SMW_SQLStore3_Readers.php(197): SMWSQLStore3Readers->getPropertySubjects(Object(SMW\DIProperty), Object(SMW\DIWikiPage), NULL)
[...]

PR #1224, opened 2015-10-20
Check for DIType in SMWSQLStore3Readers::getPropertySubjects, refs #1222

mwjames:
This is to avoid a DI mismatch.

refs #1222


mwjames:
Uses the test case as described in #1222 to return valid results without causing a `RuntimeException`.

PR #1227, opened 2015-10-21
added support for displaytitle in JavaScript result formats

cicalese:
This patch passes the value of displaytitle to JavaScript-based result formats such as datatables so they can use it for the link text. The DISPLAYTITLE magic word and the SemanticTitle extension both set the displaytitle property in the MediaWiki page_props table. This sets the title displayed on the page itself to alternate text (often a human-friendly title for pages auto-named numerically). SemanticTitle also displays this text as the link text on links to that page from elsewhere in the wiki and self links (a MediaWiki core patch providing this functionality has been proposed: https://gerrit.wikimedia.org/r/#/c/246246/). This allows all PHP-based result formats to display links to pages with the human-friendly link text, but not JavaScript-based result formats. Please consider merging this useful functionality.

Thanks,

Cindy

### DIFF
--- a/res/smw/data/ext.smw.dataItem.wikiPage.js
+++ b/res/smw/data/ext.smw.dataItem.wikiPage.js
@@ -195,7 +195,7 @@
 			if ( linker && this.fullurl !== null ){
 				var attributes = this.exists ? { 'href': this.fullurl } : { 'href': this.fullurl, 'class': 'new' };
 				var displaytitle = this.getDisplayTitle();
-				if (displaytitle === null ) {
+				if ( displaytitle === null ) {
 					displaytitle = this.getText();
 				}
 				return html.element( 'a', attributes , displaytitle );
--- a/res/smw/data/ext.smw.dataItem.wikiPage.js
+++ b/res/smw/data/ext.smw.dataItem.wikiPage.js
@@ -55,11 +55,12 @@
 	 *
 	 * @return {smw.dataItem.wikiPage} this
 	 */
-	var wikiPage = function ( fulltext, fullurl, ns, exists ) {
+	var wikiPage = function ( fulltext, fullurl, ns, exists, displaytitle ) {
 		this.fulltext  = fulltext !== ''&& fulltext !==  undefined ? fulltext : null;
 		this.fullurl   = fullurl !== '' && fullurl !==  undefined ? fullurl : null;
 		this.ns        = ns !==  undefined ? ns : 0;
 		this.exists    = exists !==  undefined ? exists : true;
+		this.displaytitle  = displaytitle !== '' && displaytitle !==  undefined ? displaytitle : null;
 
 		// Get mw.Title inheritance
 		if ( this.fulltext !== null ){
@@ -78,9 +79,9 @@
 	 * @class
 	 * @constructor
 	 */
-	smw.dataItem.wikiPage = function( fulltext, fullurl, ns, exists ) {
+	smw.dataItem.wikiPage = function( fulltext, fullurl, ns, exists, displaytitle ) {
 		if ( $.type( fulltext ) === 'string' && $.type( fullurl ) === 'string' ) {
-			this.constructor( fulltext, fullurl, ns, exists );
+			this.constructor( fulltext, fullurl, ns, exists, displaytitle );
 		} else {
 			throw new Error( 'smw.dataItem.wikiPage: fulltext, fullurl must be a string' );
 		}
@@ -149,6 +150,17 @@
 		},
 
 		/**
+		 * Returns display title
+		 *
+		 * @since  2.3
+		 *
+		 * @return {string}
+		 */
+		getDisplayTitle: function() {
+			return this.displaytitle;
+		},
+
+		/**
 		 * Returns if the wikiPage is a known entity or not
 		 *
 		 * @since  1.9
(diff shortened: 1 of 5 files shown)


res/smw/data/ext.smw.dataItem.wikiPage.js
```diff
@@ -182,7 +194,11 @@
 		getHtml: function( linker ) {
 			if ( linker && this.fullurl !== null ){
 				var attributes = this.exists ? { 'href': this.fullurl } : { 'href': this.fullurl, 'class': 'new' };
-				return html.element( 'a', attributes , this.getText() );
+				var displaytitle = this.getDisplayTitle();
+				if (displaytitle === null ) {
```
JeroenDeDauw: missing space

src/Serializers/QueryResultSerializer.php
```diff
@@ -74,6 +107,10 @@ public static function getSerialization( DataItem $dataItem, $printRequest = nul
 					'namespace' => $title->getNamespace(),
 					'exists' => $title->isKnown()
 				);
+				$displayTitle = self::getDisplayTitle ($title->getArticleID() );
```
JeroenDeDauw: missing space

src/Serializers/QueryResultSerializer.php
```diff
@@ -74,6 +107,10 @@ public static function getSerialization( DataItem $dataItem, $printRequest = nul
...
+				if ( $displayTitle ) {
+					$result['displaytitle'] = $displayTitle;
+				}
```
mwjames: The serialization format is changing, so the version needs to go up 0.5 -> 0.6.

src/Serializers/QueryResultSerializer.php
```diff
@@ -56,6 +56,39 @@ public function isSerializerFor( $queryResult ) {
 	}
 
 	/**
+	 * Get display title of a page.
+	 *
+	 * @param int $pageId page ID of page being queried
+	 *
+	 * @return string|bool display title value or false if not found
+	 */
+	public static function getDisplayTitle( $pageId ) {
```
JeroenDeDauw: should not be public

It being static also causes global state usage (`wfGetDB`). Better use dependency injection. Then again, the functionality added here seems to be out of scope for this class. It is fetching of additional data, not just "serializing" the object that was provided. I can't check right now where this new code would fit unfortunately.

src/Serializers/QueryResultSerializer.php
```diff
@@ -56,6 +56,39 @@ public function isSerializerFor( $queryResult ) {
...
+
+		if ( ! is_numeric( $pageId ) || $pageId < 1 ) {
+
+			return false;
+
+		}
+
+		$dbr = wfGetDB( DB_SLAVE );
```
mwjames: I'm not so happy about the DB access in the QueryResultSerializer (which is untested) and should probably be handled by a service lookup class. In general I would see the Title object to actually have to information available since you rely on a convention introduced in [0] to provide/store the information.

Also, `QueryResultSerializer` is one of the performance drainer doing an uncached DB read doesn't make this class faster.

[0] https://gerrit.wikimedia.org/r/#/c/246246/

cicalese: The PageProps class at [0] encapsulates the DB interactions and adds caching. Would it suffice to add that class (or a modification thereof) to SMW at this point, since the earliest it would be included in MW core would be MW 1.27? I also plan to pursue the discussion of whether the display title functionality in core would be better served being part of a core class such as Article or WikiPage rather than as a page property. So, the mechanism for accessing the display title *may* require a change when MW 1.27 is released, but there would not be a change from the perspective of the user, and it would be highly beneficial to have this capability sooner rather than later.

[0] https://gerrit.wikimedia.org/r/#/c/246246/

Fannon:
Today I've stumbled over this problem, as (1) the filtered result format has issues with Semantic Titles and (2) the JSON result format does not provide the displaytitle in the results. Both would be fixed by this patch (I've written a similar one on my own).

But I agree that including a query at this point is not a very clean approach and may come with some performance costs. It would be much nicer if the MW Title class would support this natively.

Looking really forward to see better semantic title support coming! Thanks for your effort, @cicalese !

mwjames:
> But I agree that including a query at this point is not a very clean approach and may come with some performance costs. It would be much nicer if the MW Title class would support this natively.

How to move forward?

- [ ] Add `SMW\MediaWiki\TitleLookup::getDisplayTitleFor` to encapsulate the DB access (+ appropriate caching)
- [ ] Add a unit test to `SMW\MediaWiki\TitleLookup` to verify the behaviour

You can have access to a connection in the `QueryResultSerializer` via `$queryResult->getStore()->getConnection( 'mw.db' ) );` ( `$titleLookup = new TitleLookup( $queryResult->getStore()->getConnection( 'mw.db' ) );`.

Please don't just drop a `$dbr = wfGetDB( DB_SLAVE );` it is the death of any test environment because it would always seek an active connection especially when called from a static method.

cicalese:
I just last week got a change merged into MediaWiki core (https://gerrit.wikimedia.org/r/#/c/263581/) to provide cached, bulk access to page properties (including displaytitle), but it will not be available until MW 1.27. I plan to revisit this patch in light of that change as soon as I get a chance. However that would introduce a version dependency. Thoughts?

cicalese:
@Fannon, I'm hoping that @mwjames' new patch, #1410, can be the basis of this functionality. I tested with my JavaScipt changes from this patch on top of that one, and it makes the datatables result format show pages' displaytitles as their link text as desired. I have not tested with the filtered result format, but that would be great if it works for that as well. We used to use that result format quite a bit, but had to work around the semantic title issue.

cicalese:
This has been replaced by #1423.

PR #1231, opened 2015-10-22
Do not doubly define abstract methods

JeroenDeDauw:


JeroenDeDauw:
Perhaps we should move the coverage build to the allowed failures section? Then we don't need to wait 40 mins for the CI to report back here.

mwjames:
> Perhaps we should move the coverage build to the allowed failures section? 

Yes but then scrutinizer needs to be removed or changed otherwise it will fail as well because the coverage will be delayed beyond the current 30 min waiting period.

JeroenDeDauw:
When does this period start? When the first job finishes? If it's on build start, then it should make no difference right?

PR #1240, opened 2015-11-01
Fix coding style issues #899

jongfeli:
Concerning the the easy stuff, just one little thingy to fix this time... Because it is "hard" not to generate a lot of "easy to fix" code errors you all get a :cookie:.


mwjames:
:+1:

PR #1242, opened 2015-11-01
Bail out early for queries with limit < 1

mwjames:
Instead of trying to resolve a query (and its segments) before returning a link to `Special:Ask`, create the link early in the process without the need to use a `DB` or `Respository` connection (which improves performance of pages that contain a lot of those "search on request" queries).


mwjames:
See [0] which contains several of these "search on request" queries.

[0] http://sandbox.semantic-mediawiki.org/wiki/Malignant_melanoma

PR #1243, opened 2015-11-01
Make failed queries discoverable

mwjames:
Add a `_ERRP` annotation for queries with errors to make them discoverable
using the "[[Has improper value for::Has query]]" property list.

refs #893

### DIFF
--- a/includes/parserhooks/AskParserFunction.php
+++ b/includes/parserhooks/AskParserFunction.php
@@ -199,6 +199,15 @@ class AskParserFunction {
 
 	private function createQueryProfile( $query, $format, $duration ) {
 
+		// In case of an query error add a marker to the subject for
+		// discoverability of a failed query
+		if ( $query->getErrors() !== array() ) {
+			$this->parserData->getSemanticData()->addPropertyObjectValue(
+				new DIProperty( '_ERRP' ),
+				DIProperty::newFromUserLabel( '_ASK' )->getDiWikiPage()
+			);
+		}
+
 		$queryProfileAnnotatorFactory = $this->applicationFactory->newQueryProfileAnnotatorFactory();
 
 		$jointProfileAnnotator = $queryProfileAnnotatorFactory->newJointProfileAnnotator(
--- a/tests/phpunit/includes/parserhooks/AskParserFunctionTest.php
+++ b/tests/phpunit/includes/parserhooks/AskParserFunctionTest.php
@@ -287,6 +287,45 @@ class AskParserFunctionTest extends \PHPUnit_Framework_TestCase {
 		}
 	}
 
+	public function testEmbeddedQueryWithError() {
+
+		$params = array(
+			'[[--ABC·|DEF::123]]',
+			'format=table'
+		);
+
+		$expected = array(
+			'propertyCount'  => 2,
+			'propertyKeys'   => array( '_ASK', '_ERRP' ),
+		);
+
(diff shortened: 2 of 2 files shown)


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Isssue/1243_%28Make_failed_queries_discoverable%29

mwjames:
@kghbln @thingles For the curious mind, this is why wikiapiary's error rate went up (3786).

kghbln:
Oops, did not realise this yet. Error count still on the rise. Now I have to figure out which one of the numerous queries is causing this. :| Only wiki-site wikis seem to be concerned ...

thingles:
And continues to, wow…

screenshot

I made a [failed queries](https://wikiapiary.com/wiki/User:Thingles/Failed_queries) page but I don't think I understand how to use this feature. When I click on the pages referenced and inspect the queries in the property browser I don't see any errors highlighted.

mwjames:
> When I click on the pages referenced and inspect the queries in the property browser I don't see any errors highlighted.

If you pick one 

image

then one query, contains an error

image

It doesn't tell which query because it could be one in the page or an imported one from a template/transclusion.

All we know is that a query failed.

thingles:
@mwjames Got it. Thanks. I was thinking I would then see some sort of an error message attached to one of those subobjects (which would be incredible if it did btw. :-P)

Sure wish there was a "group by" and "count" method for queries so you could show the pages that have the most errors.

WikiApiary has a couple of parser function errors that throw thousands of SMW errors. Notably the one that does database version parsing. But this is showing some additional errors I wasn't aware of. Probably time to spend hours in a  text editor matching curly's. :-(

mwjames:
> of an error message attached to one of those subobjects (which would be incredible if it did btw. :-P)

You should see something like [0].

[0] http://sandbox.semantic-mediawiki.org/wiki/Issue/1243_%28Make_failed_queries_discoverable%29

thingles:
@mwjames That is what I have at https://wikiapiary.com/wiki/User:Thingles/Failed_queries.

thingles:
It's interesting to me that the failed queries that have a property other than "Has query" when I look at the page I can find the yellow "!" notice that is associated with the error. The vast majority of these though just say "Has query" and when I look at the page there is no user visible error message.

mwjames:
> The vast majority of these though just say "Has query" and when I look at the page there is no user visible error message.

The error property is just a "plain" page property therefore it can only link to the property where the condition of an error has been identified. Of course you can ignore it but the system just makes things visible that otherwise stay hidden.

With the statistics above, I'm guessing it is one particular query that comes with a template and is embedded in all pages (and slowly grows as the bot makes it way though the pages).

thingles:
@mwjames fwiw, and maybe it isn't possible, but if you attached the 'text' property for the error message it would be huge huge huge. Possible?

mwjames:
> nd maybe it isn't possible, but if you attached the 'text' property for the error message it would be huge huge huge. Possible?

Not in the way the `_ERRP` (== `Has improper value for`, it is just a `_wpg` object) is specified, you have to turn this into a subobject itself in order to attach additional information (conceptional it is a different entity representing different information relating to a subject)

> Possible?

You mean something like #1293?

image

thingles:
@mwjames that is incredible! Seriously! Huge huge. Actually big enough for me to risk the composer update. :-)

PR #1253, opened 2015-11-07
Create SMW_LanguageHu.php

hangya:
Hungarian translation

### DIFF
--- a/languages/SMW_LanguageHu.php
+++ b/languages/SMW_LanguageHu.php
@@ -1,0 +1,105 @@
+<?php
+/**
+ * @file SMW_LanguageHu.php
+ * @ingroup SMWLanguage
+ */
+
+/**
+ * Protect against register_globals vulnerabilities.
+ * This line must be present before any global variable is referenced.
+ */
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Thanks a lot!

kghbln:
Thanks a ton! Do not forget [SMW_Aliases.php](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/languages/SMW_Aliases.php) and [SMW_Magic.php](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/languages/SMW_Magic.php) though the latter may very well be broken. Still ...

JeroenDeDauw:
> though the latter may very well be broken.

What?? Do we have an issue for that?

kghbln:
> What?? Do we have an issue for that?

Oops, just tested this and the localized parserfunctions and behaviour switches indeed do work. This was not the case last time I tested this with presumably SMW 1.5.2 or so. Great, so @hangya go for the localization! Thanks!

JeroenDeDauw:
> SMW 1.5.2

2010 called, it wants its SMW version back!

Issue #1252, opened 2015-11-06
Annotation of multiple properties broken since SMW 2.3


krabina:
The annotation of multiple properties described in [1] ("assigning one value to multipel properties") is broken since SMW 2.3. It can be seen here: [2].

The annotation ``[[Testproperty1::Testproperty2::200]]`` does not work: Testproperty1 is set to "Testproperty2::200" instead of Testproperty1 and Testproperty2 both set to "200".

[1] https://semantic-mediawiki.org/wiki/Help:In-text_annotation
[2] http://sandbox.semantic-mediawiki.org/wiki/Testproperty

mwjames:
> he annotation [[Testproperty1::Testproperty2::200]] 

@kghbln Do we really support this?? This is __not__ tested at all because we explicitly fixed `::` annotation use (#1066, #1075) so we know that the first part is a property and the rest belongs to a value `[[DOI::10.1002/123::abc]]` or `[[Foo:::123]]`.

We have a test case `[[Has page:::Page/::Foo:Bar]] [[Has text:::Text/::Foo:Bar]] [[Has url::http://example.org/::Foo:Bar]]` where we specifically test that values contain ` "Page/::Foo:Bar", ":Text/::Foo:Bar", "http://example.org/::Foo:Bar"`

If this is really a supported use case then we need to add a config setting for `[[property::property::property::value]]` to explicitly enable this as we have no way of telling what is or isn't a property. I also would make this setting an exception rather then a default because `[[ ... :: ... ]]]` is to declare one triple. (subject -> property -> value).

krabina:
It worked in the previous version and it is documented in the user manual...

kghbln:
This possibility to assign one value to multiple properties was there from the/my start so SMW 1.4 minimum so this is indeed a regression. Personally I have never used this so it did not come to my mind when you added the new features but it may very well be out there and we cannot blame them to have used it. I'd opt for a configuration setting which defaults to "false". "true" would allow to migrate or willingly use it as an exception.

krabina:
My current use case is to use it in an arraymap:
{{#arraymap:{{{Benutzer|}}}|,|x|[[Benutzer::Assigned to::Benutzer:x]]}}
This is so I can set a property of my choice alongside the property assigned to that is defined by Semantic Tasks. But I just tested it, and 
[...]

PR #1254, opened 2015-11-07
Add InTextAnnotationParserStrictMode flag, refs #1252

mwjames:
By default (as of 2.3) a more strict parsing process is enforced to minimize annotation interpretation issues.

`$GLOBALS['smwgEnabledInTextAnnotationParserStrictMode']` can be set `false` to loosen the strictness which may result in certain features not being supported or can cause unanticipated misinterpretation in case of additional colons.

refs #1252

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -929,3 +929,18 @@ $GLOBALS['smwgExportBCAuxiliaryUse'] = false;
 # @since 2.3
 ##
 $GLOBALS['smwgExportBCNonCanonicalFormUse'] = false;
+
+##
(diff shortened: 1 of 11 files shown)


mwjames:
https://semantic-mediawiki.org/wiki/Help:$smwgEnabledInTextAnnotationParserStrictMode

mwjames:
Needs to be backported to the 2.3.x branch so it can be shipped with the next 2.3.1 release.

mwjames:
@kghbln Any comments?

kghbln:
Yes, a backport makes very much sense here!

mwjames:
Backported to 2.3.x with dd6cbbd.

PR #1256, opened 2015-11-07
EmbeddedQueryDependencyLinksStore to create object ID if not availabl…

mwjames:
…e, refs #1117

For example if a new query is created with entities that have not yet been used before (== new) then the SQLStore does not have nor will it create an ID for something that is not an annotation (a query description is not an annotation).

While this is an edge case (where a query is created before any annotation is available) nevertheless we start to track its dependency early.


mwjames:
[0] showed the symptom when the query was created before an actual annotation was available.

[0] http://sandbox.semantic-mediawiki.org/wiki/QueryUpdateByCategory

mwjames:
@kghbln As explained per email this should solve the issue.

Backported to 2.3.x with 68b7ec5.

Issue #1255, opened 2015-11-07
Ask with format=template breaks HTML output 


jongfeli:
The move from **SMW 2.2.3** to **SMW 2.3** seems to break HTML in a ```{{#ask:...``` query with ```format=template```. We have some buttons on which users can press which automatically update properties or values on a page but they are not working anymore. Example of the issue on: ~~**http://sandbox.semantic-mediawiki.org/wiki/User:Jongfeli**~~

http://sandbox.semantic-mediawiki.org/wiki/Issue/1255

I tested this on **MW 1.24.1** and SMW **2.1**, **2.2**, **2.2.3** & **2.3.0** but it is only happening in **SMW 2.3.0**

Below how it looks in **SMW 2.2.3** (I noticed it shows **2.2.2 (D540302)** on the MW version page).

![screenshot from 2015-11-07 09 25 34](https://cloud.githubusercontent.com/assets/6680027/11014337/923be194-8531-11e5-8193-db05ce46ebaa.png)

mwjames:
#1068 added option `$smwgEnabledResultFormatsWithRecursiveAnnotationSupport` to allow listed formats to use a "full parse" instead of a `'[[SMW::off]]' . $wgParser->replaceVariables( $result ) . '[[SMW::on]]'`. If you remove `template` from that list or set `$smwgEnabledResultFormatsWithRecursiveAnnotationSupport = array()` then it should return to your expected form.

Test cases that describe behaviour when activated 
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/fd777541ce0f455f6ae5c7774a66850fd1dcbcd9/tests/phpunit/Integration/ByJsonScript/Fixtures/p-0406%20%5B%231055%5D%20unrestricted%20intext%20annotation%20via%20template%20ask%20parse.json

- https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/fd777541ce0f455f6ae5c7774a66850fd1dcbcd9/tests/phpunit/Integration/ByJsonScript/Fixtures/p-0701%20%5B%23711%5D%20ask%20template%20invert%20annotation.json

- https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/fd777541ce0f455f6ae5c7774a66850fd1dcbcd9/tests/phpunit/Integration/ByJsonScript/Fixtures/p-0205%20%5B%231055%5D%20set%20ask%20parser%20recursive%20annotation%20via%20template.json

[...]

PR #1068, opened 2015-06-19
Add setting for unrestricted template parse, refs #1055

mwjames:
Allow to copy annotation data from a recursive parse, refs #1055, #1076. Unrestricted import annotations from a recursive parse (#711) for default formats:
```
$GLOBALS['smwgEnabledResultFormatsWithRecursiveAnnotationSupport'] = array(
	'template',
	'list'
);
```

This setting can be used to mitigate template parsing in connection
with the VisualEditor (1055).

[0] https://phabricator.wikimedia.org/T15336#179238

PR #1257, opened 2015-11-07
Support recursive annotation per query, refs #1255, #1068

mwjames:
This removes `smwgEnabledResultFormatsWithRecursiveAnnotationSupport` (was only added with 2.3) setting and instead handles support for recursive annotation on an individual query level.

By default `|recursive-parse` is set `false`.

refs #1255, #1068

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -421,24 +421,6 @@ $GLOBALS['smwgResultFormats'] = array(
 $GLOBALS['smwgResultAliases'] = array( 'feed' => array( 'rss' ) );
 ##
 
-###
-#
-# Enables specific formats to import its annotation data into the
-# subject that resolves the result query.
-#
-# E.g. using ask query to search/set an invert property value
-#
-# This setting can be used to mitigate template parsing in connection with the
-# VisualEditor (#1055).
-#
-# @since 2.3
-##
-$GLOBALS['smwgEnabledResultFormatsWithRecursiveAnnotationSupport'] = array(
-	'template',
-	'list'
-);
-##
-
 ### Predefined sources for queries
 # Array of available sources for answering queries. Can be redefined in
 # the settings to register new sources (usually an extension will do so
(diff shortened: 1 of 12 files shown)


mwjames:
Backported to 2.3.x with ec9d82d.

s7eph4n:
@mwjames , I am not quite sure what this parameter does. Is it just that it allows annotations in templates? Why is it limited to the `template` format?
Also, to reduce coupling between the ResultPrinters and the rest of SMW, would it make sense to register the parameter not in `ResultPrinter::getParamDefinitions()`, but in `SMWQueryProcessor::getParameters()` when a method like `ResultPrinter::supportsRecursiveAnnotation()` returns `true`?

Loads of questions. Might be that I am missing the point completely. :smile:

mwjames:
> Loads of questions. Might be that I am missing the point completely.

I'll try to use a simple example:

Page A

- Adds [[Foo::bar]]

Template:Foobar

- Does some formatting and
- Contains some extra annotation like [[SomeProperty:SomeValue]]

Page Q1:

- {{#ask: [[Foo::bar]] |template=Foobar }} finds Page A and does nothing else

Page Q2

- {{#ask: [[Foo::bar]] |template=Foobar |import-annotation=true}}
finds Page A and imports the annotation from the template hence
[[SomeProperty:SomeValue]] appears on Page Q2

> allows annotations in templates? Why is it limited to the `template`
> format?

Adding of annotations as demonstrated above can only happen when the
parser is executed to substitute the template code (and we have a
ParserOutput object).

> Also, to reduce coupling between the ResultPrinters and the rest of SMW,
> would it make sense to register the parameter not in
> `ResultPrinter::getParamDefinitions()`, but in
> `SMWQueryProcessor::getParameters()` when a method like
> `ResultPrinter::supportsRecursiveAnnotation()` returns `true`?

+1

There is another issue but I had no time to deal with this where the
ResultPrinter uses $GLOBALS['wgParser'] to parse the template code.

What I want is to pass the parser instance used by the
AskParserFunction to the actual printer so that `$GLOBALS['wgParser']
` is removed from the ResultPrinter instance.

A bunch of code that deals with the template parsing should be
isolated into something like `RecursiveTextProcessor` which is
injected and passed around.

```
...
$printer->setRecursiveTextProcessor( new RecursiveTextProcessor() );
...
```

kghbln:
[Example on sandbox](https://sandbox.semantic-mediawiki.org/wiki/Issue/1257)

kghbln:
[Documented](https://www.semantic-mediawiki.org/wiki/Help:Plainlist_format/parameter_import-annotation)

PR #1264, opened 2015-11-09
Removed pg_unescape_bytea-ing the result

hangya:
Since PostgreSQL 9.0 (earlier version are at end-of-life) the [default binary format is "hex"](http://www.postgresql.org/docs/9.0/static/datatype-binary.html). Unescaping bytea strings were only necessary in earlier PostgreSQL versions, which used "escape" format - forcing escapeing causes issues in rendering (see [here](https://github.com/SemanticMediaWiki/SemanticMaps/pull/47))

### DIFF
--- a/includes/queryprinters/ResultPrinter.php
+++ b/includes/queryprinters/ResultPrinter.php
@@ -253,10 +253,6 @@ abstract class ResultPrinter extends \ContextSource implements QueryResultPrinte
 			$result = $this->handleNonFileResult( $result, $results, $outputMode );
 		}
 
-		if ( $GLOBALS['wgDBtype'] == 'postgres' ) {
-			$result = pg_unescape_bytea( $result );
-		}
-
 		return $result;
 	}
 


mwjames:
> Since PostgreSQL 9.0 (earlier version are at end-of-life)

Interesting another user that uses PostgreSQL in recent times.

> Unescaping bytea strings were only necessary in earlier PostgreSQL versions

Would this also appear in the normal table format? (I'm looking for a small tests case where this issue appeared besides SM and is fixed with above change)

We have the `q-0103 ...` test disabled for postgres, could this be resolved with this change?

```
SMW\Tests\Integration\ByJsonScript\ByJsonScriptFixtureTestCaseRunnerTest::executeTestCasesFor with data set "q-0103 regex blob escape character queries.json" ('/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/ByJsonScript/Fixtures/q-0103 regex blob escape character queries.json')
Requires special \ escape sequence on postgres
```

JeroenDeDauw:
Thanks a lot for the pull request!

PR #1270, opened 2015-11-13
Display correct property header for an inverse printout

mwjames:
Discovered by Cheifgeek157 [0].

[0] https://semantic-mediawiki.org/wiki/Thread:Help_talk:Inline_queries/Header_name_for_inverse_properties_in_ask_queries


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1270

PR #1273, opened 2015-11-14
Fix coding style issues #899

jongfeli:
One small fix and updated https://github.com/jongfeli/Code-violations-SemanticMediaWiki/blob/master/all


JeroenDeDauw:
Thanks!

PR #1276, opened 2015-11-15
Exported API time format is insufficient

mwjames:
The simple `getMwTimestamp` output (e.g. `-18208281600` for `1 Jan 1393`) is insufficient for a general workable API solution in order for a consumer to create a meaningful interpretation.

The output was extended to carry two representations, the previous timestamp and now the raw format (e.g. `2/1393/1/1`) which can can be used to deserialize the data into its original value depiction (with the intended precision).

`smw.dataItem.data` as been adjusted to find the correct timestamp element.

Consumers that use JS/`smw.dataItem.data` can work with both API results (< 0.7 >).

```
"Has date": [
    {
        "timestamp": "-2000101000000",
        "raw": "2/-200"
    },
    {
        "timestamp": "-18208281600",
        "raw": "2/1393/1/1"
    }
```

### DIFF
--- a/res/smw/data/ext.smw.data.js
+++ b/res/smw/data/ext.smw.data.js
@@ -130,7 +130,7 @@
 							break;
 						case '_dat':
 							$.map( value, function( t ) {
-								factoredValue.push( new smw.dataItem.time( t ) );
+								factoredValue.push( new smw.dataItem.time( ( t.hasOwnProperty( 'timestamp' ) ? t.timestamp : t ) ) );
 							} );
 							break;
 						case '_num':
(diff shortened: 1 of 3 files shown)


mwjames:
Worst, if you try to recover something from `-2000101000000` with `SMWDITime::newFromTimestamp( -2000101000000 )` it just returns "Part of the date is out of bounds", so the value as API result is meaningless if you instead use `\SMWDITime::doUnserialize( 2/-200 );` you will be able to get the original `200 BC`,

JeroenDeDauw:
+

PR #1282, opened 2015-11-19
Add extra search column to support case-insensitive regex searches

mwjames:
Instead of using the sortkey to match a text string, any regex search (if enabled) will use the normalized search index to allow for case-insensitive result matches.

A query condition such as `[[Has text::~lorem*]]` will return the same results as for a `[[Has text::~LOREM*]]` (or `[[Has text::~lOrEm*]]`) condition.

Test cases include:

- `[[Has text::~lorem*]]` and `[[Has text::~*REM*]]`
- `[[Has text::!~*REM*]]`
- `[[Has url::~*rem*]]` and `[[Has url::~*REm*]]`
- `[[~*/6/lore*]]` and `[[~*/6/lORe*]]`
- `[[Has text::~*æå*]]`

## Features and limitations 

- ~~`$GLOBALS['smwgEnabledCaseInsensitiveRegExMatchSearch']`~~ `$GLOBALS['smwgEnabledEnhancedRegExMatchSearch']` requires to be set true
- `update.php` and `rebuildData.php` both required to be run after the setting was enabled
- Any non regex search will use the standard sortkey to return correct matches of things like `=`, `>`, `<` 
- Something like `LIKE 'whatever%'` (as `[[Has text::~whatever*]]`) will use the extra index whereby `LIKE '%whatever%'` (as `[[Has text::~*whatever*]]`) will not (a limitation set by the SQL engine)

refs http://wikimedia.7.x6.nabble.com/semediawikiuser-Case-insensitive-search-still-not-at-option-td5055261.html

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -926,3 +926,16 @@ $GLOBALS['smwgExportBCNonCanonicalFormUse'] = false;
 # @default true
 ##
 $GLOBALS['smwgEnabledInTextAnnotationParserStrictMode'] = true;
+
+##
+# This setting enables RegEx (like [~*...*], not like [!~...]) searches to operate
+# equally on all input strings without differentiation on lower or CAPITAl letters.
+#
+# It is required that `update.php` is run, followed by `rebuildData.php` because this
+# feature creates an additional search column that expects data to be populated
+# before any RegEx query can successfully return match results.
+#
+# @since 2.4
+# @default false
+##
+$GLOBALS['smwgEnabledEnhancedRegExMatchSearch'] = false;
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -139,7 +139,8 @@ class Settings extends SimpleDictionary {
 			'smwgPropertyDependencyDetectionBlacklist' => $GLOBALS['smwgPropertyDependencyDetectionBlacklist'],
 			'smwgExportBCNonCanonicalFormUse' => $GLOBALS['smwgExportBCNonCanonicalFormUse'],
 			'smwgExportBCAuxiliaryUse' => $GLOBALS['smwgExportBCAuxiliaryUse'],
-			'smwgEnabledInTextAnnotationParserStrictMode' => $GLOBALS['smwgEnabledInTextAnnotationParserStrictMode']
+			'smwgEnabledInTextAnnotationParserStrictMode' => $GLOBALS['smwgEnabledInTextAnnotationParserStrictMode'],
+			'smwgEnabledEnhancedRegExMatchSearch' => $GLOBALS['smwgEnabledEnhancedRegExMatchSearch']
 		);
 
 		$settings = $settings + array(
--- a/includes/storage/SQLStore/SMW_DIHandler_Blob.php
+++ b/includes/storage/SQLStore/SMW_DIHandler_Blob.php
@@ -1,4 +1,7 @@
 <?php
+
+use SMW\SearchField;
+
 /**
  * @ingroup SMWDataItemsHandlers
  */
@@ -39,7 +42,16 @@ class SMWDIHandlerBlob extends SMWDataItemHandler {
 	 * @return array
 	 */
 	public function getTableFields() {
-		return array( 'o_blob' => 'l', 'o_hash' => 't' );
+		return array( 'o_blob' => 'l', 'o_hash' => 't', 'o_search' => 't' );
+	}
+
+	/**
+	 * @see SMWDataItemHandler::getTableIndexes()
+	 * @since 1.8
+	 * @return array
+	 */
+	public function getTableIndexes() {
+		return array( 'o_hash', 'o_search' );
 	}
 
 	/**
@@ -58,7 +70,10 @@ class SMWDIHandlerBlob extends SMWDataItemHandler {
 	 * @return array
 	 */
 	public function getWhereConds( SMWDataItem $dataItem ) {
-		return array( 'o_hash' => self::makeHash( $dataItem->getString() ) );
+		return array(
+			'o_hash'   => self::makeHash( $dataItem->getString() ),
+			'o_search' => SearchField::getSearchStringFrom( $dataItem->getString() )
+		);
 	}
 
 	/**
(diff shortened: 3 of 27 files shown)


mwjames:
Of course, this doesn't replace any real text indexing engine (EL, Lucene etc.) but it provides an out of the box (without any external dependency) solution and enables to recover result matches based on a 255 char text search field that should be enough for most common applications (subject match, object value search) when the identify of a value string is arbitrary and `~/!~` is used as operator.

The storage __doesn't__ involve any elaborate text indexing mechanism  such as tokenising, double metaphones, or stemming.

mwjames:
> It's always strange to tell users, that e.g. search fields in Semantic
Query Forms are case sensitive. So, I think this is an important issue.

@formwandler If you feel up for the task, you could test this.

@SemanticMediaWiki/testers Any user with interest in this topic is encouraged to give it a try as well.

jaideraf:
@mwjames I just need to say: I love you. :heart:

This is so, so great! SMW, to me, now has passed from 2.3 to 4.0, not 2.4, nor 3.0.
This is such a long waited feature... of course it is not our dreamed search/indexing engine (ElasticSearch, Solr, Lucene, etc.), but it is, at least, competitive to other CMSs search.

IMHO, this should be the default behaviour of SWM.
So thank you! Really thank you. :clap:

No more things like: 
```[[property::~*{{#replace:{{{searchString|}}}| |*}}*||~*{{ucfirst:{{#replace:{{{searchString|}}}| |*}}}}*||~*{{lc:{{#replace:{{{searchString|}}}| |*}}}}*||~*{{ucfirst:{{lc:{{#replace:{{{searchString|}}}| |*}}}}}}*]]```

**Comment by thingles** [comment]
Whoa! This is great to see! On WikiApiary I have a couple of properties that I set thousands of instances of just to work around this limitations. This is great @mwjames!

**Comment by mwjames** [comment]
> This is so, so great! SMW, to me, now has passed from 2.3 to 4.0, not 2.4, nor 3.0.
> This is great to see! On WikiApiary I have a couple of properties that I set thousands of instances 

Wow, didn't expect that __BUT__ this needs __rigorous testing__ before it can be merged since we are about to create an extra column to keep the search term apart from any other data.

Currently, I'm also looking for a cheep tokenizer that ensures that terms stored with the search column being removed from stop words and reduced in size (more on this later).

Preliminary tests showed a slightly better performance on the index search but I can't predict the overall perf impact for the DB with more than 30K entries in the ID_TABLE.

> IMHO, this should be the default behaviour of SWM.

The column will have its own index and is only updated when any other part of the `SemanticData` is altered. To populate the search term you need to run `-f`otherwise the field is empty for existing rows and the search will return no results (that's why it cannot be a default setting because you have to migrate/populate data before you can use the search index).

Of course, one could write a migration script (to avoid running `rebuildData.php -f`) to fill/update the column but I'd rather focus on the something else (if someone wants to write a script I'm more than willing to review.)

**Comment by mwjames** [comment]
> Of course, one could write a migration script (to avoid running rebuildData.php -f)

As I suspected that it was fairly unlikely that one would come forward and write a script and now that I needed something for testing (running the `rebuildData.php -f` each time I changed the indexer rule wasn't feasible) I hereby present the `rebuildSearchFieldContent.php` script.

It is a "brute" force script that selects tables that carry an additional search column and then updates the search field with the new index according to the rules set forth by the `SearchField` class.

`update.php` (or `setupStore.php`) still has to be run but now instead of scrapping the whole DB (via `rebuildData.php -f`) just run `rebuildSearchFieldContent.php` and the search field should hold updated values in a swift.

```
$ php rebuildSearchFieldContent.php

The script updates the index field for tables that support regex
value searches. Any change of the index rules (altered stopword
list, new stemmer etc.) and/or a newly added or altered table
requires to run this script (or `rebuildData.php`) to ensure that
the index complies to the rules set forth by the `SearchField`
class.

Depending on the size of the tables selected, it may take a moment
before the update is completed.
---

Starting the property tables update ...

- smw_di_blob        100% (63485/63485)
- smw_di_uri         100% (183147/183147)
- smw_fpt_type       100% (55/55)
- smw_fpt_unit       100% (2/2)
- smw_fpt_conv       100% (16/16)
- smw_fpt_pval       100% (7/7)
- smw_fpt_list       100% (6/6)

- smw_fpt_askfo      100% (3563/3563)
- smw_fpt_askst      100% (3563/3563)
- smw_fpt_impo       100% (4/4)
- smw_fpt_uri        100% (5/5)
- smw_fpt_media      100% (6/6)
- smw_fpt_mime       100% (6/6)

Starting the ID table update ...

- smw_object_ids     100% (59658/59658)
```

I renamed the setting `smwgEnabledCaseInsensitiveRegExMatchSearch` to `smwgEnabledEnhancedRegExMatchSearch` because I'm experimenting with stopword filtering (the search field is 255 char so we want to make optimal use the size given) where case-insensitivity is just a by-product of a more versatile regex match search.

**Comment by krabina** [comment]
I also think this feature is really great! Thank you!

**Comment by Joshkking** [comment]
Just wanted to drop a line here as I was heavily interested in this coming to be. Although my wiki is not really developed enough for testing, keep us posted as I imagine with a little instructions a few would enjoy testing it out (and myself whenever I can).

**Comment by thingles** [comment]
Since I did a fresh pull I was hoping to turn this on but my composer stuff is all messed up.

```
MediaWiki 1.25.3 Updater

wikimedia/composer-merge-plugin: 1.3.0 installed, ^1.3 required.
mediawiki/semantic-media-wiki: dev-master installed, ^2.3 required.
Error: your composer.lock file is not up to date, run "composer update" to install newer dependencies
```

And currently `composer update` fails to run because it has some issue it can't satisfy. (insert emoticon of me pounding head on wall). I need to do a fresh MediaWiki install on my farm. Don't suppose there is anyway to get `update.php` to work without fixing my composer issues?

mwjames:
> Since I did a fresh pull I was hoping to turn this on 

Hold your horse for a bit, I need to sync with my repo first that includes https://github.com/onoi/tesa.  I reworked the way blob tables are selected (because not all need an extra index).

I will give some instructions for when the first version can be tested.

> And currently composer update fails to run because it has some issue it can't satisfy. (insert emoticon of me pounding head on wall). I need to do a fresh MediaWiki install on my farm. Don't suppose there is anyway to get update.php to work without fixing my composer issues?

How about `composer update --with-dependencies`?

mwjames:
This is now superseded by #1481 (using a better approach).

PR #1283, opened 2015-11-23
Update en.json

kghbln:
* Switch consistently to www.semantic-mediawiki.org (predominantly used)
* Typos


JeroenDeDauw:
But but but! All that wasted bandwidth and disk space... You don't need those w's!

^ please ignore the above

kghbln:
;)

PR #1291, opened 2015-12-05
Isolate part of the SMWQueryParser

mwjames:
Move some parts from `SMWQueryParser` into `SMW\Query\Parser\DescriptionProcessor` to refactor the `QueryParser` to only handle string manipulation and syntax logic.

Add support for range queries as in `[[>Example/0608-A]] [[<Example/0608-C]]` (or
`[>Example/0608-AA]] [[<Example/0608-CCC]]`).

```json
{
	"about": "#0 range >A <C (non strict comparator)",
	"condition": "[[Category:Q0610]] [[>Example/Q0610-A]] [[<Example/Q0610-C]]",
	"printouts" : [],
	"parameters" : {
		"limit" : 10
	},
	"queryresult": {
		"count": 3,
		"results": [
			"Example/Q0610-A#0##",
			"Example/Q0610-AAA#0##",
			"Example/Q0610-BBB#0##"
		]
	}
},
{
	"about": "#1 range >AA <CCC (non strict comparator)",
	"condition": "[[Category:Q0610]] [[>Example/Q0610-AA]] [[<Example/Q0610-CCC]]",
	"printouts" : [],
	"parameters" : {
		"limit" : 10
	},
	"queryresult": {
		"count": 3,
		"results": [
			"Example/Q0610-AAA#0##",
			"Example/Q0610-BBB#0##",
			"Example/Q0610-CCC#0##"
		]
	}
}
```

refs #1246

PR #1293, opened 2015-12-07
Add _ERRC (container to store erroneous annotation), refs #1243

mwjames:
Having only a link to a misguided property annotation might not always be enough to resolve the issue `_ERRC` is added as a container and can now store additional information about a problem.

- `_ERRP` remains a `_wpg` type to point to the erroneous property referenced
- `_ERRC` represents the error container and id internally hashed (i.e `_ERR6fd39e1d9c9c36a9cf917bdb3d80e0fb`)
- `_ERRT` contains a textual description of an identified issue

image

https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1243#issuecomment-162519453

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -321,7 +321,7 @@
 	"smw-datavalue-import-invalidvalue":"\"$1\" is not a valid format and is expected to consists of \"namespace\":\"identifier\" (e.g. \"foaf:name\").",
 	"smw-pa-property-predefined_impo": "\"$1\" is a predefined property that describes a relation to an [https://www.semantic-mediawiki.org/wiki/Help:Import_vocabulary imported vocabulary].",
 	"smw-pa-property-predefined_type": "\"$1\" is a predefined property that describes the [[Special:Types|datatype]] of a property.",
-	"smw-pa-property-predefined_sobj": "\"$1\" is a predefined property representing a container construct that allows to accumulate property-value assignments similar to that of a normal wiki page.",
+	"smw-pa-property-predefined_sobj": "\"$1\" is a predefined property representing a [https://www.semantic-mediawiki.org/wiki/Help:Container container] construct that allows to accumulate property-value assignments similar to that of a normal wiki page.",
 	"smw-pa-property-predefined_errp": "\"$1\" is a predefined property to track input errors for irregular value annotations that was likely caused by type or [[Property:Allows value|allowed value]] restrictions.",
 	"smw-pa-property-predefined_pval": "[https://www.semantic-mediawiki.org/wiki/Help:Special_property_Allows_value \"$1\"] is a predefined property that can define a list of permissible values to restrict value assignments for a property.",
 	"smw-datavalue-property-restricted-use": "Property \"$1\" was marked for restricted use.",
@@ -338,5 +338,7 @@
 	"smw-sp-types_tel": "\"$1\" is a special datatype to describe international telephone numbers according to RFC 3966.",
 	"smw-sp-types_txt": "\"$1\" is a primitive datatype to describe strings of arbitrary length.",
 	"smw-sp-types_dat": "\"$1\" is a datatype to represent points in time in a unified format.",
-	"smw-specials-browse-helplink": "https://www.semantic-mediawiki.org/wiki/Help:Special:Browse"
+	"smw-specials-browse-helplink": "https://www.semantic-mediawiki.org/wiki/Help:Special:Browse",
+	"smw-pa-property-predefined_errc": "\"$1\" is a predefined property representing a [https://www.semantic-mediawiki.org/wiki/Help:Container container] for errors that appeared in connection with improper value annotations or input processing.",
+	"smw-pa-property-predefined_errt": "\"$1\" is a predefined property containing a textual description of an error."
 }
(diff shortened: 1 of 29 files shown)


mwjames:
For now, only `#ask`/`#show` will use the new container to store errors all other reporters will use the simple `_ERRP` instead (but can be converted over time to use the new container).

Error counting is not expected to be influenced since `_ERRP` is referenced in the error container.

mwjames:
@kghbln This comes with two new pre-defined properties, `_ERRC` (Has error) and `_ERRT` (Has error text).

If you think `Has error` as textual descriptor is to ambiguous, feel free to change it.

kghbln:
I think it actually is. I suggest "Has query error" and "Has query error text". This would also be more consistent with the other query-related special properties.

mwjames:
> I think it actually is. I suggest "Has query error" and "Has query error text". This would also be more consistent with the other query-related special properties.

No, it is a general purpose error container and doesn't not only have to contain query errors (for now it only contains query errors since I don't have time to convert all other `_ERRP` callers).

mwjames:
A possible name change can be done in a follow-up.

kghbln:
Ah, ok, what about "Has semantic error" etc.?

mwjames:
> Ah, ok, what about "Has semantic error" etc.?

What is a semantic error? What about `Has processing error`?

The description reads
- "smw-pa-property-predefined_errc": "\"$1\" is a predefined property representing a container for errors that appeared in connection with improper value annotations or input processing.",
- "smw-pa-property-predefined_errt": "\"$1\" is a predefined property containing a textual description of an error.""

kghbln:
An error in connection with using SMW, however this was indeed not an ideal suggestion since it suggest that actual semantics are bellies up. Yeah, "Has processing error" sounds great.

kghbln:
Cannot do it right now since I am trying to do it via git instead of browser edit. Well, it does not pick up the branch I created remotely and also tells me that everything is up to date though the files I added are changed. Tough luck. I expected GitHub to be easier than Git/Gerrit. So far lit looks like the same kind of mess just different. I'll dig myself though the docu which I admittedly have not so far.

mwjames:
> So far lit looks like the same kind of mess just different. I'll dig myself though the docu which I admittedly have not so far.

I merged the branch which means just rebase against master, create a new branch, and add the changes.

kghbln:
@JeroenDeDauw Pls do not sync until this is sorted out

thingles:
I have deployed this in WikiApiary FWIW. Still waiting to see errors getting populated.

mwjames:
> I have deployed this in WikiApiary FWIW. Still waiting to see errors getting populated.

I'm quite certain it is the array query but the thing is that I'm no longer able to catch the query issue on wikiapiary as error (I have a hunch as to why). 

@thingles @kghbln Could you figure out which query would generated the output below (and post-it in its as raw form)?

image

An empty query like `{{#ask: |mainlabel=-|limit=0 |format=array}}` will definitely get an error.

image

image

mwjames:
> Could you figure out which query would generated the output below (and post-it in its as raw form)?

I figured out the query `<includeonly>{{#ask: [[Farm:{{{Farm|}}}]] | ?Is collecting | format=array | mainlabel=-}}</includeonly>`, I have to investigate the missing error report (but I know what is causing this.)

mwjames:
>  I have to investigate the missing error report (but I know what is causing this.)

#1291 added range query support but one old `if condition` was processed too early to catch a possible error and therefore gone silent.

#1297 should allow to resurface the error with `{{{Farm|}}}` being empty and then something like `"[[Farm:]] ("Farm:" cannot be used as a page name in this wiki.)"` will be reported.

I added a test for a similar query to verify that an error is reported.

> https://wikiapiary.com/wiki/Property:Has_error_text

@kghbln merged #1296, `Has error text` became `Has processing error text`.

@thingles Thanks for the early deployment (made it easier to catch this oversight), the only thing left is a new Composer update.

mwjames:
@thingles @kghbln Queries like `{{#ask: [[Has text::abc_[123]]] }}` will cause an issue, try to separate the brackets `{{#ask: [[Has text::abc_[123] ]] }}`.

image

thingles:
@mwjames I don't want to overstate it but these changes you've put in to make errors visible is such a huge help for anyone running a large SMW installation I just can't thank you enough. Seriously, catching errors in SMW is super hard and on any non-trivial wiki nearly impossible. Putting the error text into a property and allowing people to query it is a huge help. Before, I wouldn't have even attempted to bring the error rate on WikiApiary down. Now I can see how I can do that!

Thank you!

kghbln:
Indeed another leap when it gets to [semantic gardening](https://semantic-mediawiki.org/wiki/Semantic_gardening)! :+1:

thingles:
@mwjames I want to share with you what this change allowed me to do on WikiApiary. Look at the error count.

screenshot

Also, this is now a great place for admins to see how healthy the wiki is

https://wikiapiary.com/wiki/WikiApiary:Processing_errors

Big win!

thingles:
@mwjames Inspired by the way that you aggregated error data using this property I've now got WikiApiary starting to aggregate error data across the wikiverse. Mostly simple stuff like extensions that have no name, or have a URL associated that is not valid, etc. But, can extend further over time.

https://wikiapiary.com/wiki/Property:Has_remote_error

kghbln:
Now documented [here](https://www.semantic-mediawiki.org/wiki/Help:Special_property_Has_processing_error_text) and [there](https://www.semantic-mediawiki.org/wiki/Help:Special_property_Has_processing_error)

PR #1294, opened 2015-12-08
Extends _ERRC usage, refs #1293

mwjames:
Objects reporting an error will now automatically use the new `_ERRC` container to track errors.

refs #1293


mwjames:
@kghbln An example for a non annotation related error.
```
{{#set: 
 |_FooBar=abc
}}
```

image

```
{{#subobject:
 |Modification date=1 Jan 1970
}}
```

image

```
{{#subobject:
 |Bar=foo Bar
 |-Foo=foo Bar
}}
```

image

PR #1296, opened 2015-12-08
Ren `Has error` to `Has processing error` (_ERRC, _ERRT), refs #1293

mwjames:
#1293

### DIFF
--- a/languages/SMW_Language.php
+++ b/languages/SMW_Language.php
@@ -94,8 +94,8 @@ abstract class SMWLanguage {
 		'Has query duration' => '_ASKDU',
 		'Has media type'     => '_MEDIA',
 		'Has mime type'      => '_MIME',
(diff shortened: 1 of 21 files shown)


mwjames:
@kghbln In case you are happy with the naming, just press the merge button otherwise amend to  your liking.

kghbln:
@mwjames I was already happy with the new name. I just failed to get GitHub accept changes when I try to push them. I just received bla, bla instead as indicated in the other post.

mwjames:
> I just failed to get GitHub accept changes when I try to push them. I just received bla, bla instead as indicated in the other post.

If you override an existing branch content then you need 

image

kghbln:
> If you override an existing branch content then you need 

@mwjames As I realised I tried to push master instead of the local branch, so I was destined to fail. Now things are in fluff. See other issue comment.

PR #1299, opened 2015-12-09
Check subobject named identifier

mwjames:
Named identifiers containing a dot are reserved for extension use.

```
{{#subobject:foo.bar
 |foo=bar
}}
```

### DIFF
--- a/src/Error.php
+++ b/src/Error.php
@@ -51,14 +51,21 @@ class Error {
 			$property = new DIProperty( $property->getKey() );
 		}
 
-		$subWikiPage = new DIWikiPage(
+		$errorMsg = is_array( $errorMsg ) ? implode( ' ', $errorMsg ) : $errorMsg;
+
+		$subject = new DIWikiPage(
 			$this->subject->getDBkey(),
 			$this->subject->getNamespace(),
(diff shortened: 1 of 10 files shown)


src/Error.php
```diff
@@ -51,14 +51,21 @@ public function getContainerFor( DIProperty $property = null, $errorMsg = '' ) {
...
 			$this->subject->getNamespace(),
 			$this->subject->getInterwiki(),
-			'_ERR' . md5( $property !== null ? $property->getKey() : 'UNKNOWN' )
+			'_ERR' . md5( ( $property !== null ? $property->getKey() : 'UNKNOWN' ) . $errorMsg )
 		);
 
-		$containerSemanticData = new ContainerSemanticData( $subWikiPage );
+		// Encode brackets to avoid an annotion is created/included
+		return $this->newDiContainer( $subject, $property, str_replace( '[', '&#x005B;', $errorMsg ) );
+	}
```
mwjames: @kghbln This fixes [0] where a queried error list copied `[[ ... ]]` annotations from the query result into a page.

[0] https://wikiapiary.com/wiki/Special:Browse/User:Thingles-2FFailed_queries

PR #1302, opened 2015-12-09
Less restrictive rule for sobj named identifiers, refs #1299

mwjames:
This is a bit less restrictive (#1299 banned all "." usage) but leaves enough room for extension usage.

```
{{#subobject:あいうえ.willNotWork
 |foo=bar
}}
```

```
{{#subobject:あいうえお.willWork
 |foo=bar
}}
```
```
{{#subobject:abc.willNotWork
 |foo=bar
}}
```

```
{{#subobject:abcde.willWork
 |foo=bar
}}
```

Named subobjects containing a "." in the first five characters are reserved to be used by extensions only in order to separate them from user land and avoid having them accidentally to refer to the same named ID.


mwjames:
SIL already uses `ill.`+ language code [0] and `iwl.` + language code [1]

[0] https://github.com/SemanticMediaWiki/SemanticInterlanguageLinks/blob/master/src/InterlanguageLink.php#L66

[1] https://github.com/SemanticMediaWiki/SemanticInterlanguageLinks/blob/master/src/InterwikiLanguageLink.php#L62

mwjames:
@kghbln @thingles I saw [0] therefore I made the rule added with #1299 less restrictive.

[0] https://wikiapiary.com/wiki/Special:Browse/Youbianku-23Interwiki_zh.wikipedia

PR #1307, opened 2015-12-13
Update composer.json

paladox:


composer.json
```diff
@@ -49,7 +49,7 @@
 		"onoi/callback-container": "~1.0"
 	},
 	"require-dev": {
-		"squizlabs/php_codesniffer": "~2.1",
+		"mediawiki/mediawiki-codesniffer": "0.5.0",
 		"phpmd/phpmd": "~2.1"
```
mwjames: @JeroenDeDauw ?

JeroenDeDauw:
We already have our own CS specification and might not agree with everything that is in the MW one. So I prefer to keep this as is.

PR #1315, opened 2015-12-15
Feed printer (rss/atom) to parse media files

mwjames:
Based on the question raised by [0], it is to allow for a feed to display things like:

image

[0] http://wikimedia.7.x6.nabble.com/display-specific-datas-in-RSS-feed-td5053425.html

### DIFF
--- a/src/Query/ResultPrinters/FeedExportPrinter.php
+++ b/src/Query/ResultPrinters/FeedExportPrinter.php
@@ -275,7 +275,14 @@ final class FeedExportPrinter extends ResultPrinter implements ExportPrinter {
 			// Loop over all values for the property.
 			while ( ( $dataValue = $field->getNextDataValue() ) !== false ) {
 				if ( $dataValue->getDataItem() instanceof SMWDIWikipage ) {
-					$itemSegments[] = Sanitizer::decodeCharReferences( $dataValue->getLongWikiText() );
+
+					$linker = null;
+
+					if ( $dataValue->getDataItem()->getSubobjectName() === '' && $this->params['link'] !== 'none' ) {
+						$linker = smwfGetLinker();
+					}
+
+					$itemSegments[] = Sanitizer::decodeCharReferences( $dataValue->getLongWikiText( $linker ) );
 				} else {
 					$itemSegments[] = Sanitizer::decodeCharReferences( $dataValue->getWikiValue() );
 				}
--- a/tests/phpunit/Integration/JSONScript/TestCases/s-0027.json
+++ b/tests/phpunit/Integration/JSONScript/TestCases/s-0027.json
@@ -1,5 +1,5 @@
 {
-	"description": "Test `format=feed` output via `Special:Ask`",
+	"description": "Test `format=feed` output via `Special:Ask` (`wgEnableUploads`, `wgFileExtensions`)",
 	"setup": [
 		{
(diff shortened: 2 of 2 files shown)


mwjames:
I don't have any tests for this so this needs manual testing.

includes/queryprinters/FeedResultPrinter.php
```diff
@@ -205,15 +205,17 @@ protected function feedItem( array $row ) {
 			// Loop over all values for the property.
 			while ( ( $dataValue = $field->getNextDataValue() ) !== false ) {
 				if ( $dataValue->getDataItem() instanceof SMWDIWikipage ) {
-					$itemSegments[] = Sanitizer::decodeCharReferences( $dataValue->getLongHTMLText() );
+					$itemSegments[] = Sanitizer::decodeCharReferences(
+						$dataValue->getLongWikiText( $this->params['link'] !== 'none' ? smwfGetLinker() : null )
+					);
```
JeroenDeDauw: You could drop the negation

includes/queryprinters/FeedResultPrinter.php
```diff
@@ -245,31 +247,36 @@ protected function feedItem( array $row ) {
...
-			$parserOptions = new ParserOptions();
-			$parserOptions->setEditSection( false );
-
-			if ( $this->params['page'] === 'abstract' ) {
-				// Abstract of the first 30 words
-				// preg_match( '/^([^.!?\s]*[\.!?\s]+){0,30}/', $wikiPage->getText(), $abstract );
-				// $text = $abstract[0] . ' ...';
+		if ( !in_array( $this->params['page'], array( 'abstract', 'full' ) ) ) {
+			return '';
+		}
```
JeroenDeDauw: Yay for guard clause

JeroenDeDauw:
Ping @SemanticMediaWiki/testers

JeroenDeDauw:
Did a quick review and looks good. Then again, I'm not even sure which part is required for the media file parsing

kghbln:
@mwjames Is this one dead? Looks like a nice enhancement.

mwjames:
> @mwjames Is this one dead? Looks like a nice enhancement.

Mmm ... I think there was an edge case but I cannot recall what exactly was the issue. Pushing this to the end of the 3.0 timeline.

kghbln:
> Mmm ... I think there was an edge case but I cannot recall what exactly was the issue. Pushing this to the end of the 3.0 timeline.

No worries. To me it looked like an interesting enhancement without the need to do too much voodoo. If there is not enough time we can still move this to 3.1

mwjames:
> No worries. To me it looked like an interesting enhancement without the need to do too much voodoo. If there is not enough time we can still move this to 3.1
> was an edge case but I cannot recall 

I still cannot remember the edge case but waiting on a divine intervention will only increase the likelihood of code rot so instead `s-0027.json` adds an integration test and the PR now becomes part of 3.0

kghbln:
> I still cannot remember the edge case

I am sure we will find out sooner than later since this is a supposedly more heavily used feature. Even better if not at all.

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AFeed_format&type=revision&diff=66531&oldid=66517)

Issue #1314, opened 2015-12-15
Avoid text with [[ ]] to copy value annotations when queried


mwjames:
As demonstrated in [0], value annotations are regularly parsed when displayed using a query (where in fact it is not the desired outcome).

[0] http://sandbox.semantic-mediawiki.org/wiki/Text_that_includes_other_annotations/Query

PR #1320, opened 2015-12-17
Add obscureAnnotation / removeAnnotation to InTextAnnotationParser, r…

mwjames:
…efs #1314

- `InTextAnnotationParser::removeAnnotation` to entirely remove traces of any annotation within the text
- `InTextAnnotationParser::obscureAnnotation` to encode part of the annotation so that
the `InTextAnnotationParser::parse` is not able to process that part of a text

### DIFF
--- a/src/InTextAnnotationParser.php
+++ b/src/InTextAnnotationParser.php
@@ -141,6 +141,67 @@ class InTextAnnotationParser {
 	}
 
 	/**
+	 * @since 2.4
+	 *
+	 * @param string $text
+	 *
+	 * @return text
+	 */
+	public static function decodeSquareBracket( $text ) {
+		return str_replace( array( '%5B', '%5D' ), array( '[', ']' ), $text );
+	}
+
+	/**
+	 * @since 2.4
+	 *
+	 * @param string $text
+	 *
+	 * @return text
+	 */
+	public static function obscureAnnotation( $text ) {
+		return preg_replace_callback(
+			self::getRegexpPattern( false ),
+			function( array $matches ) {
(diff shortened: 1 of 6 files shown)


mwjames:
`InTextAnnotationParser::obscureAnnotation` to produce

image

`InTextAnnotationParser::removeAnnotation` to produce

image

src/InTextAnnotationParser.php
```diff
@@ -141,6 +141,64 @@ public function parse( &$text ) {
 	}
 
 	/**
+	 * @since 2.4
+	 *
+	 * @param string $text
+	 *
+	 * @return text
+	 */
+	public static function obscureAnnotation( $text ) {
```
JeroenDeDauw: 3326873_orig

Production usage of this new method is not in this PR, so it's hard to say much more about this

src/InTextAnnotationParser.php
```diff
@@ -141,6 +141,64 @@ public function parse( &$text ) {
...
+
+		// Convert possible internal URL encoding
+		$text = str_replace( array( '%5B', '%5D' ), array( '[', ']' ), $text );
```
JeroenDeDauw: You don't need this local var, or the override the previous value of the parameter. You can just inline this below. At the same time it seems like a good candidate for extraction into its own method, due to the comment and the presence of low level details.

src/InTextAnnotationParser.php
mwjames: Replaces [0].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Error.php#L63-L64

Issue #1306, opened 2015-12-11
HttpDatabaseConnectionException  reports curl error: 56 (Problem (2) in the Chunked-Encoded data)


mwjames:
Reported by [0]
> (6/157 4%) Processing page Property:Bibtex:number ... 
Exception encountered, of type "SMW\SPARQLStore\Exception\HttpDatabaseConnectionException" 
[a5b39356] [no req]   SMW\SPARQLStore\Exception\HttpDatabaseConnectionException from line 63 of /home/jethro/mediawiki/extensions/SemanticMediaWiki/src/SPARQLStore/BadHttpResponseMapper.php: Failed to communicate to Endpoint: http://localhost:3030/test/update
due to curl error: 56 (Problem (2) in the Chunked-Encoded data). 

Users that encountered `curl error: 56 (Problem (2) in the Chunked-Encoded data)` in a non-SMW related context [1, 2, 3] suggested either to upgrade the cURL PHP library or to set `curl_setopt($ch, CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_0);` which are all unrelated to SMW.

If you want to try the second suggestion, you may change [4]:

```php
+		$curlRequest = new CurlRequest( curl_init() );
+		$curlRequest->setOption( CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_0 );
+
 		$connection = new $repositoryConnector(
 			new RepositoryClient( $this->defaultGraph, $this->queryEndpoint, $this->updateEndpoint, $this->dataEndpoint ),
-			new CurlRequest( curl_init() )
+			$curlRequest
 		);
```

[0] http://wikimedia.7.x6.nabble.com/SPARQLStore-Exception-rebuildData-php-tp5056274p5056324.html
[1] http://stackoverflow.com/questions/29146515/curl-problem-2-in-the-chunked-encoded-data
[2] https://github.com/jackalope/jackalope-jackrabbit/issues/89
[3] http://stackoverflow.com/questions/18766946/php-curl-response-differs-based-on-network-encoding-issue
[4] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/SPARQLStore/RepositoryConnectionProvider.php#L135

mwjames:
This not really a bug that can be solved by SMW. The suggestions is made based on the input from the sources cited if someone encounters a similar problem.

[...]

PR #1321, opened 2015-12-17
Setting to force HttpVersion for RepositoryConnector, refs #1306

mwjames:
Default setting is `$GLOBALS['smwgSparqlRepositoryConnectorForcedHttpVersion'] = false;`
and should not be changed unless error #1306 was encountered.


mwjames:
Backported to 2.3.x with  04e6dce.

PR #1329, opened 2015-12-23
Add positional preference for units

mwjames:
The unit position declared in `[[Corresponds to::€ 1]]` vs `[[Corresponds to::1.06 USD, US$, $]]`
determines the positional preference (prefix, suffix).

https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/135#issuecomment-159918758

http://sandbox.semantic-mediawiki.org/wiki/Issue/1329_%28Positional_unit_preference%29

### DIFF
--- a/includes/datavalues/SMW_DV_Number.php
+++ b/includes/datavalues/SMW_DV_Number.php
@@ -46,6 +46,13 @@ class SMWNumberValue extends SMWDataValue {
 	protected $m_unitvalues;
 
 	/**
+	 * Whether the unit is preferred as prefix or not
+	 *
+	 * @var array
+	 */
+	protected $prefixalUnitPreference = array();
+
+	/**
 	 * Canonical identifier for the unit that the user gave as input. Used
 	 * to avoid printing this in conversion tooltips again. If the
 	 * outputformat was set to show another unit, then the values of
@@ -66,7 +73,7 @@ class SMWNumberValue extends SMWDataValue {
 	 * @return integer 0 (no errors), 1 (no number found at all), 2 (number
 	 * too large for this platform)
 	 */
-	static protected function parseNumberValue( $value, &$number, &$unit ) {
+	static protected function parseNumberValue( $value, &$number, &$unit, &$asPrefix = false ) {
 		// Parse to find $number and (possibly) $unit
 		$decseparator = NumberFormatter::getInstance()->getDecimalSeparatorForContentLanguage();
 		$kiloseparator = NumberFormatter::getInstance()->getThousandsSeparatorForContentLanguage();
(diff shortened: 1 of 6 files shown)


mwjames:
@jongfeli Would you be so kind and test this (it is based on your comments in SemanticMediaWiki/SemanticResultFormats#135 (comment))

@SemanticMediaWiki/testers are invited as well

includes/datavalues/SMW_DV_Number.php
```diff
@@ -66,7 +73,7 @@ class SMWNumberValue extends SMWDataValue {
 	 * @return integer 0 (no errors), 1 (no number found at all), 2 (number
 	 * too large for this platform)
 	 */
-	static protected function parseNumberValue( $value, &$number, &$unit ) {
+	static protected function parseNumberValue( $value, &$number, &$unit, &$asPrefix = false ) {
```
JeroenDeDauw: #naaaaw

mwjames: Well, the whole thing needs a general overhaul but I'm not in a mood to bite the bullet yet (for once I want to add some new things not re-factor legacy stuff).

jongfeli:
My apologies for the delay @mwjames. I tested this with [0] and it works BUT @JeroenDeDauw's ```#naaaaw``` and your reaction to that probably means that the solutions is not what you want :)

Al-tough it is great to be able to define the unit position for a value, the way it is now and for units in front of the value this only makes sense when used inline or with ```{{#show:...```. When the results are in a table then the position of the unit should depend on how big the biggest number in the result is, otherwise the unit or the number is all over the place (depending on alignment). This basically means that the ```[[Has type::Quantity]]``` datatype should give the user more control on how the numbers are presented.

You asked me to write a **RFC**, I will do this (after the holidays).

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/pos-unit

JeroenDeDauw:
@jongfeli my previous comment is a code quality nitpick. It is not about the soundness of the overall approach.

mwjames:
> . I tested this with [0] and it works

Thanks

> #naaaaw and your reaction to that probably means that the solutions is not what you want :)

The overall quality of the DV classes are hard to extend hence my comment.

> When the results are in a table then the position of the unit should depend on how big the biggest number in the result is, otherwise the unit or the number is all over the place (depending on alignment).

"Prefexial dependency" and output alignment are distinct issues that should not be mixed with each other. The alignment of  numbers within a column table to display an ordered (in terms of space) position has to be done in a result printer which generates the HTML/wikitext.

This PR is about "positional preference" which is part of the DV (DataValue) itself to determine the relative correctness (correct in terms of what a user has defined to be either right or left) and keeps the information of the definition close that what the user has specified.

A single DV has no (and can not have) knowledge about a possible competing DV that may (or may not) appear in a printout and therefore cannot be made dependent on such at the time the DV is created. 

Of course, you could have language dependant positional preference [0] but that cannot be solved with the current way general purpose quantity properties are defined.You would need to extend the Quantity DV and account for language specific output with the help of the i18n system.

[0] http://english.stackexchange.com/questions/11326/what-is-the-difference-between-20-and-20

mwjames:
... and for the due diligence fraction, I added some extra tests to cover things like `[[Currency::<¥300]] OR [[Currency::≥1400 JPY]] OR [[Currency::< € .8]]`

mwjames:
@kghbln FYI

PR #1330, opened 2015-12-24
Schema definition

mwjames:
`MediaWiki:smw-schema-definition` introduced as feature to provide information about user-defined properties and categories in a JSON format so that entity descriptions can be maintained more easily.

Creating a list of properties or categories in a consistent and maintainable manner has been a difficult endeavour from its outset and while the occasional user may not long for such feature, more professional oriented users may have an interest in a more convenient deployment method.

MW 1.24 comes with an in-build JSON parser to display `CONTENT_MODEL_JSON` content in a more human readable form. 

image

Properties and categories part of the `smw-schema-definition` are being blocked from manual editing to keep it consistent with the schema. Once an entity is removed from the schema it can be freely edited or deleted as any other user-defined property.

image

https://www.semantic-mediawiki.org/wiki/Help:Schema_definition


JeroenDeDauw:
Very interesting. This is similar to the [Page Schemas extension](https://www.mediawiki.org/wiki/Extension:Page_Schemas) by @yaronkoren. I'm quite interested in hearing Yarons opinion on this approach as he has put quite some though and effort into this kind of stuff before. Beyond that, it's probably also wise to consult other people that have experiences with this, such as users of the PS extension (for instance by writing the user list).

I realize this thing is not all that big in LOC, however it is not clear to me it's better to include this in SMW core rather than in a dedicated extension. (And once its shipped with SMW itself, it'll be difficult to move it out.)

mwjames:
> This is similar to the Page Schemas extension by @yaronkoren. (Have you seen that?)

I have seen it.

> I realize this thing is not all that big in LOC, however it is not clear to me it's better to include this in SMW core rather than in a dedicated extension. (And once its shipped with SMW itself, it'll be difficult to move it out.)

I thought about it and this time (for this PR) I decided it would be better to make it a resistant core-feature instead of placing it into an extension so that users can take immediate control of those definitions without the need for an extra extension as those definitions (for those who want a more restricted control) should be managed by core itself.

The logic applied by the `SchemaManager` (incl. `SchemaReader`and `SchemaUpdateJob`) is fairly lean and with the support of the `JsonContentHandler` users can easily manipulate the schema via an API using standard MW methods. Further more, all descriptions that are part of a schema are prohibited from manual intervention (editing, deletion, or moving) to ensure that definitions stay in sync with what the schema dictates.

Of course if you remove a definition from the schema, you can do "whatever" you want with it (as it is not longer controlled by a schema definition).

PS: Individual descriptions are not yet part of it as it requires the new monolingual text datatype to set language dependant descriptions in support for things like`[[Has property description::Property that does something@en]]` `[[Has property description::このプロパティーは何かする@ja]]`.

mwjames:
> for instance by writing the user list

Done with [0].

[0] http://wikimedia.7.x6.nabble.com/A-more-centralized-way-to-manage-property-definitions-td5056786.html

yaronkoren:
This seems like a good idea to me, actually - the more SMW knows about the wiki's data structure, the better. I don't see this really conflicting with Page Schemas anyway - PS uses a schema basically as a way to create templates, form etc., while SMW uses it for (I guess) enabling better display of information - whatever the exact uses, they're different from what PS does. Not that it would necessarily be bad for SMW to offer the exact functionality of Page Schemas - I'm generally in favor of centralizing functionality, so that there are fewer "moving parts", like was done with Semantic Internal Objects.

In the future, PS could potentially use this SMW schema information to create an initial default schema for any category.

Fannon:
Hi,

this looks really good!

Have you taken a look at JSON Schema? It's a widely used specification for schema documentation in http://json-schema.org/). By itself it would be insufficient to describe what your example describes, but it's no problem to extend it for those custom properties.

Going with existing standards would have some benefits, as it would be possible to re-use existing libraries/documentation, etc..

I can also see a big benefit in using the schema definition of a SMW property for real-time validation in Semantic Forms and more flexible back-end validation in SMW.

Merry Chrismas :)
Simon

mwjames:
> Have you taken a look at JSON Schema? It's a widely used specification for schema documentation in http://json-schema.org/)

You can use [0] to externally validate our smw-schema [1] otherwise the json-schema form is not really suitable for a "normal" application without the help of a tool (and we want to make it simple not more complicated).

Using [1] with [0] and [2] should give you a pass. [0] could be either hosted on smw.org or github but ought to be accessible to ensure that new types or item identifiers can be added in a swift.

Currently the MW `JsonContentHandler` doesn't support json-schema validation but [3, 4] demonstrates how the handler can be extended. It is beyond the objective of this PR and it is better placed with MW-core instead of trying to re-replicated schema validation with SMW.

I really want to keep this lean and clean [2] without too much overhead so that users unfamiliar with the format are nevertheless able to edit it if the occasion arises.

[0] https://gist.github.com/mwjames/67d385f0f2d486b57a5e
[1] http://jsonschemalint.com/draft4/
[2] https://www.semantic-mediawiki.org/wiki/Help:Schema_definition#Example

[3] https://github.com/wikimedia/mediawiki-extensions-EventLogging/blob/master/includes/JsonSchema.php
[4] https://github.com/wikimedia/mediawiki-extensions-EventLogging/blob/master/includes/JsonSchemaContent.php

WolfgangFahl:
@mwjames  welcome to the club!
This is another approach to get the schema definitions right into SMW. I love to see this discussion to integrate things in to the SMW Core. At SWM Con Fall 2015 Barcelona I had proposed to do a workshop between the people that have an interest in this to come up with a common understanding instead of currently somewhat "competing" approaches. Currently the different approaches have a tendency to confuse people as to e.g. what the naming for schema elements should be. 

Currently there seem to be the approaches (this list is probably still incomplete):
- https://www.mediawiki.org/wiki/Extension:Page_Schemas (PageSchema)
- https://github.com/Fannon/mobo (mobo)
- http://profiwiki.bitplan.com/index.php/MetaModel (smartMediaWiki/profiWiki)
- http://www.w3.org/wiki/SemanticWebTools (native RDF tools e.g. Protége)
- the proposal by John Mc Clure from 2015-05

In the closed wiki partner.bitplan.com there has been an ongoing discussion during 2015 and most people from the SMW community that where interested got an account for that wiki. I'd be willing to put the more interesting stuff on the semantic-mediawiki.org wiki to get the discussion going. To do so I'd need the plantUML and graphviz extensions up and running there that have also been used on the open http://profiwiki.bitplan.com site. 

For a start 
http://profiwiki.bitplan.com/index.php/Concept:Property
shows the properties a property can have.

It would be great if we could agree on the naming and come up with a common metamodel schema or at least a proper mapping between the different metamodel schema proposals

Please note that the profiWiki MetaModel calls for "Action" as a new Metamodel element. For me this is the most interesting addition to the show since it will allow a page to have services/operations/functions added to it making such a Wiki a full fledged Application environment. This has e.g. already been done by http://platform.xwiki.org/xwiki/bin/view/Features/Applications. But obviously this would be another thread of discussion made possible in the future.

WolfgangFahl:
I have looked at the PR and my main comment is that we should allow the schema to be edited with SMW itself as it is already done in the http://profiwiki.bitplan.com/index.php/MetaModel approach. This was original based on a proposal by by Pierre Racine and its implementation is possible with standard SMW mechanisms. I think the core functionality should only have a very limited set of mandatory things. 
Please note that the original idea was to separte some of the concerns e.g. there should be "Property" and "PropertyConfiguration" as separate things. Property would have the mandatory stuff e.g. name and type while PropertyConfiguration can be quite implementation specific e.g. There might be an RDFPropertyConfigurationan an SQLStorePropertyConfiguration an SMWFormsPropertyConfiguration and so on. Hardcoding theses things in PHP would IMHO complicate things.

WolfgangFahl:
It would be great if the PlantUML extension would be installed and I could continue explaining ...

mwjames:
Actually this will be superseded by #2784 in future.

Issue #1125, opened 2015-08-25
Precision of Type:Number cannot be specified on output


dbolser:
Following up from https://phabricator.wikimedia.org/T22217, which I couldn't find here...

I'd suggest that the properties default output format could be specified on the property page, i.e. [[Has output format::.2f]],  etc. This way format would work like units, and could be overridden for 'raw' output in the usual way.

kthxbi.

mwjames:
Merged with 7cfa279 and expected to be deployed with 2.4. For examples, please have a look at [0].

[0] http://sandbox.semantic-mediawiki.org/wiki/Issue/1335_%28Display_precision%29

PR #1335, opened 2015-12-30
Add possibility to specify precision for a numeric datatype, refs 1125

mwjames:
`[[Display precision of::<int>]]` can declare a scale for decimal precision to a numeric property type so that display output can be restricted to something like 2 digits as in `€ 42.45, € 17.14, € 60.00`.

`Display precision of` only restricts output representation and does not interfere with the internal "full" precision.

It is expected that values are transformed using:

`[[Display precision of::2]]`

- 0.001 -> 0.00
- 0.005 -> 0.01
- € 60  -> € 60.00
- 30.445 € -> € 30.45 (3.00 ¥, 32.27 $, 21.31 £)
- 100 °F -> 310.93 K (37.78 °C, 100.00 °F, 559.67 °R)

`[[Display precision of::4]]`

- 100 °F -> 310.9278 K (37.7778 °C, 100.0000 °F, 559.6700 °R)
- 30.445 € -> € 30.4450 (3.0000 ¥, 32.2717 $, 21.3115 £)

refs #1125, https://phabricator.wikimedia.org/T46910

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -352,5 +352,6 @@
 	"smw-pa-property-predefined_media": "\"$1\" is a predefined property that describes the media type of an uploaded file and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki].",
 	"smw-pa-property-predefined_askfo": "\"$1\" is a predefined property that holds the name of the result format used in a query and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki].",
 	"smw-pa-property-predefined_askst": "\"$1\" is a predefined property that describes the conditions of the query as a string and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki].",
-	"smw-pa-property-predefined_askdu": "\"$1\" is a predefined property containing a time value (in seconds) the query had required to complete its execution and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki]."
+	"smw-pa-property-predefined_askdu": "\"$1\" is a predefined property containing a time value (in seconds) the query had required to complete its execution and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki].",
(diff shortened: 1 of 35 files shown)


mwjames:
@SemanticMediaWiki/testers Would be nice to have some people test this. 

```
Property:Has temperature

* Has type: [[Has type::Temperature]]
* Display precision of: [[Display precision of::4]]
```

```
Property:Currency

* Has type: [[Has type::Quantity]]
* Display units: [[Display units::€, ¥, $, £]]
* Corresponds to: [[Corresponds to:: € 1.00]]
* Corresponds to: [[Corresponds to::1.06 US, USD, $]]
* Corresponds to: [[Corresponds to::0.70 British Pound, GBP, pounds sterling, £]]
* Corresponds to: [[Corresponds to::7.7742 HKD]]
* Corresponds to: [[Corresponds to::¥,JPY,Japanese Yen 114.2121]]
* Display precision of: [[Display precision of::2]]
```

```
Property:Has number

* Has type: [[Has type::Number]]
* Display precision of: [[Display precision of::2]]
```

mwjames:
Requires to run `php update.php` because `_PREC` is a fixed property and is stored in its own table to avoid having the lookup to request data from the general `smw_di_number` table.

The printout option `-p<number of digit>` can now be used to specify a precision on-the-fly during a query output and will override a possible predetermined precision (set by `Display precision of`) of a property.

```
{{#ask: [[Has area::+]]
 |?Has area#m²-p3=(#m²-p3)
 |?Has area#-p2=(#-p2)
 |?Has area#-n-p2=(#-n-p2)
 |limit=3
}}
```

http://sandbox.semantic-mediawiki.org/wiki/Issue/1335_%28Display_precision%29 (which will only display expected values after the merge)

mwjames:
The funny thing is that 2/3 of the LOC belongs to `p-0410.json`, so creating the test took longer than the actual implementation.

mwjames:
@kghbln FYI

mwjames:
https://www.semantic-mediawiki.org/wiki/Help:Special_property_Display_precision_of

kghbln:
This is now live on sandbox.smw.o!

mwjames:
> This is now live on sandbox.smw.o!

Given the fact that the original inquiry [0] is from ~ 08.2009, I'd say we are pretty quick.

[0] https://phabricator.wikimedia.org/T22217

kghbln:
:+1:

Issue #594, opened 2014-10-26
Multilingual Semantic MediaWiki


mwjames:
[0] pointed towards [Multilingual Semantic MediaWiki](https://www.mediawiki.org/wiki/Multilingual_Semantic_MediaWiki) (a.k.a MSMW) as reference on this topic.

@nemobis @kghbln Not sure SMW-core actually needs to do something here.

[0] https://www.semantic-mediawiki.org/w/index.php?title=Talk:AIFB_Web_Portal&offset=20141026172749&lqt_mustshow=1491#Translations_1491

hypergrove:
Managing language-specific property values is problematic in a semantic wiki that has pages in more than one language; there is no good support in SMW today, a situation that should be remedied.  MediaWiki provides Special:MyLanguage/somepagename to display (a subpage of) somepagename in a manner that matches the user's language (per personal preferences) eg Main Page/es would be the Spanish version of Main Page in a default English wiki -- this is fine for content-only wikis, not so great for semantic wikis.  There are two strategies I think of:

1. Place English-relevant annotations on the English page and Spanish-relevant annotations on its Spanish page
2. Place all annotations on the English (ie default) page, flagging language-specific values within the annotations themselves

SMW Ontologies (SMWO) takes the latter approach due primarily to the belief it is quite alot easier to input, translate, query and maintain annotations expressed in multiple languages if they are on the same default-language page.  Thus pieces of a 'multilingual annotation strategy' are as follows.

1 property-specific templates (or parser functions) that store language-tagged values in subobjects named per the property
2 #set and in-line annotation syntax modified to let users specify a language relevant to an annotation
    property-specific templates (or parser functions) that output language-tagged values
3 #ask and #show modified to give users the ability to specify a base retrieval language, with others as backups
4 #ask and #show modified to optionally return Special:MyLanguage invocations; to count attribute values sensibly; etc
5 Special:Browse modified to display language-qualified values for a property
6 page-specific templates to format a page's display based on the page's display language
7 SemForms extension modified for multi-lingual fields is a possibility
8 Special:PageAssembly to display a page's "structure" among its transclusions, language subpages & templates
9 Translate extension modified to create {{semantic}} template work-units

nemobis:
You can now read a preprint of "Multilingual Semantic MediaWiki for 
Finno-Ugric dictionaries" at 
[...]

PR #1344, opened 2016-01-02
Add MonolingualTextValue

mwjames:
This feature allows to associate a language code to a text component with something like:

```
Property:Has text with language

[[Has type::Monolingual text]]
```

```
[[Has text with language::example one@en]]

{{#subobject:
 |Has text with language=例の一@ja
}}
```

```
{{#ask: [[Has text with language::?@ja]]
 |?Has text with language
}}
```

This is prerequisite to support [0] multi-lingual property labels as demonstrated in [0].

[0] https://www.youtube.com/watch?v=jcau-ZA_uYc

refs #594

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -959,6 +959,9 @@ $GLOBALS['smwgEnabledInTextAnnotationParserStrictMode'] = true;
 # behaviour is not expected to be altered, nevertheless it is recommended that
 # the setting is enabled to improve user friendliness in terms of query execution.
 #
+# - SMW_DV_MLTV_LCODE (MonolingualTextValue) is to require a language code in order
+# for a DV to be completed otherwise a MLTV can operate without a language code
+#
 # @since 2.4
 ##
-$GLOBALS['smwgDVFeatures'] = SMW_DV_PROV_REDI;
+$GLOBALS['smwgDVFeatures'] = SMW_DV_PROV_REDI | SMW_DV_MLTV_LCODE;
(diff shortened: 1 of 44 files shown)


mwjames:
https://www.semantic-mediawiki.org/wiki/Help:Type_Monolingual_text

mwjames:
@kghbln @JeroenDeDauw Increasing the level of awsomeness.

kghbln:
:+1: WOW I actually have two wikis in the pipeline which could use this right away

mwjames:
It is always nice to see Travis turn green, can't imaging writing code without it.

mwjames:
>  WOW I actually have two wikis in the pipeline which could use this right away

... is merged now, so knock yourself out (remember to run `update.php`). `p-0411.json` contains various tests and queries.

mwjames:
## Notes

- `[[Has text with language::some text@en]]` === `[[Has text with language::some text@EN]]` because language codes are transformed to a IETF language tag / BCP47 (`Localizer::asBCP47FormattedLanguageCode`)
- Property "Language code" is now a fixed property `_LCODE` which has its own `LanguageCodeValue` DV to ensure that whenever a "Language code" property is used whether in a MLT context or as single annotation property it conforms with BCP47 and is checked against MW's `Language::isSupportedLanguage`
- Internally each set (text + language code) is identified as separate entity addressable using `_ML` + content hash

image

`$GLOBALS['smwgDVFeatures']` (as internal setting) contains `SMW_DV_MLTV_LCODE` which means that a language code is required. If the flag is not set then even though a property is typed with `Monolingual text` it is allowed to be stored without a language code (acts as semi standard text value internally it is still stored as MLT container).

Of course in the strict sense of a `Monolingual text` type, a language code is required (hence the default setting `SMW_DV_MLTV_LCODE` == `MonolingualTextValue::needsLanguageCode`)

Special:Types will display the requirement for the type.

image

vs.

image

kghbln:
This is now live on [sandbox.smw.o](http://sandbox.semantic-mediawiki.org/wiki/Special:Types/Monolingual_text) though this may not come as news for everybody on this pull. :) Great, kudos!

mwjames:
@kghbln To clarify "... so one should use the lower cases for the language code." whether you use EN, eN, or en it doesn't matter because the `LanguageCodeValue` (which is assigned to `_LCODE` == `Language code`) transforms any of the mentioned strings to en. For example, zh-hans it transformed to zh-Hans (BCP47 standard, `Localizer::asBCP47FormattedLanguageCode`).

kghbln:
Oops, the comparator is quite clear on this. Yeah, I messed this up. Thanks for the pointer!

krabina:
A great feature. But the problem is that in ask queries you ONLY find (and therefore can display) content in your userlanguage (using {{USERLANGUAGECODE}} )).

MediaWiki in general has the fallback language English for system messages. In a similar way it would be great if in an ask query a fallback language could be defined, maybe with an additional language code addes like so:
`{{#ask:
 [[Category:Test one]]
 [[Has monolingual text::?@{{USERLANGUAGECODE}}@en]]
 |?Has monolingual text|+index=1
 |mainlabel=-
 |headers=hide
}}`
So if the there is no content in the user's language, the @en version will be shown.

PR #1357, opened 2016-01-08
Fix usage of `|+align=...` in `format=table`

mwjames:
```
{{#ask: [[Currency::+]]
 |?Currency|+align=right|+limit=2
 |format=table
}}
```

https://phabricator.wikimedia.org/T18571

### DIFF
--- a/includes/queryprinters/TableResultPrinter.php
+++ b/includes/queryprinters/TableResultPrinter.php
@@ -154,7 +154,7 @@ class TableResultPrinter extends ResultPrinter {
 			$alignment = trim( $resultArray->getPrintRequest()->getParameter( 'align' ) );
 
 			if ( in_array( $alignment, array( 'right', 'left', 'center' ) ) ) {
-				$attributes['style'] = "text-align:' . $alignment . ';";
+				$attributes['style'] = "text-align:$alignment;";
 			}
 			$attributes['class'] = $columnClass . ( $dataValueType !== '' ? ' smwtype' . $dataValueType : '' );
 
--- a/tests/phpunit/Integration/ByJsonScript/Fixtures/f-0205.json
+++ b/tests/phpunit/Integration/ByJsonScript/Fixtures/f-0205.json
@@ -1,0 +1,62 @@
+{
+	"description": "Test `format=table` on `|+align=`/`|+limit`/`|+order` extra printout parameters (T18571, en)",
+	"properties": [
+		{
+			"name": "Has number",
+			"contents": "[[Has type::Number]]"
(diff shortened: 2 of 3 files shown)


mwjames:
@kghbln It seems no one really missed this feature `|+align=right` otherwise someone would have complained. At least now it is tested.

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1357

mwjames:
[0] this one needs more love.

[0] https://www.semantic-mediawiki.org/wiki/Help:List_of_printout_parameters

kghbln:
> `|+align=right` per ?-row

Indeed, so far we only seem to have had a global `|align=right` as in

```
...
|?...
|format=table
|align=right
...
```
So this is not just a bug fix but also an enhancement?!

mwjames:
> So this is not just a bug fix but also an enhancement?!

According to [0] NO, it states "This has now been implemented: additional parameters to individual ?printouts can be given by writing "|+ parameter=value" after the printout. Currently, there are parameters for limiting and ordering results ("limit=number" and "order=ascending/descending"), and special parameters for some printouts (e.g. format=table supports "align=right" and "align=center").

The bug remains open since these parameters do not work properly with Special:Ask yet: they are simply lost when clicking on "further results"."

Of course, the "...do not work properly with Special:Ask yet: they are simply lost when clicking on "further results"" remains valid but that's for a follow-up.

[0] https://phabricator.wikimedia.org/T18571

kghbln:
> According to [0] NO

Ah, true so the docu of this on smw.o was messed up from the start since it states it should be used globally for a query. I guess this has not been used wide spread since it is pretty useless. However, now there is more light on this. Cool this makes sense.

Since this is only format specific for table and broadtable it should probably appear in `#smwdoc` for these two?

mwjames:
> Since this is only format specific for table and broadtable it should probably appear in #smwdoc for these two?

#smwdoc deals with enabled parameters while `|+align=` (or in case of the record type `|+index=`) are printout specific parameters and not bound to an "official" parameter declaration.

kghbln:
> #smwdoc deals with enabled parameters

Yeah, still a bit edgy. However, it makes sense not to do exceptions.

mwjames:
> Of course, the "...do not work properly with Special:Ask yet: they are simply lost when clicking on "further results"" remains valid but that's for a follow-up.

Just as a side note, trying to use something like `?Has number|+align=right|+limit=2|+order=asc` in the `Special:Ask` printout text area will fail because of the `|` usage.

Issue #832, opened 2015-02-19
Last execution time for "rebuildData.php" and "rebuildPropertyStatistics.php"


kghbln:
In case one does not control the server on which the cron is setup for these two scripts there is no easy way to determine if they have actually been run successfully. Thus it will be nice to have date and time information somewhere on the wiki, perhaps on "Sepecial:SMWAdmin" saying e.g. "Last time successfully run on February 19, 2015, 05:15 am".

We have this for "rebuildConceptCache.php" on the respective concept pages and this is really useful.

mwjames:
OK this can be done failry easy all you need is something like:

- Add `'maintenance-script-execution` to the bottom of settings.php
```php
$GLOBALS['smwgLogEventTypes'] = array(
	'sqlstore-query-execution' => false,
	'maintenance-script-execution' => true
```
- Add a simple logger statement such as after `$maintenanceHelper->reset();`
```php
$store->getLogger()->logToTable(
	'maintenance-script-execution', // smw event type
	'smw-maintenance-runner', // user
	'rebuildData', // target
	'rebuildData.php execution' // message in the log
);
```
and `Special:Log` (or `index.php?title=Special:Log&type=smw`) should show a log entry together with time/date of the execution. If you want the content of the options to appear in the log entry than you have do a little do more work and extend the `rebuildData.php execution'` message string.

Maybe if you only want to log specific runs, add a `--with-log` option to the script.

That's all you need

kghbln:
This turned into a disaster. However `smwgLogEventTypes` introduced with SMW 2.1 seems to have been removed in the meantime.

mwjames:
> This turned into a disaster. However smwgLogEventTypes introduced with SMW 2.1 seems to have been removed in the meantime so this may have something to do with it.

I kept the main class [0] `ManualEntryLogger` but I wasn't happy with the `$store->getLogger()->logToTable` thing so I removed it for now since this wasn't really helpful.

The maintenance script need to do something like:

```php
use SMW\MediaWiki\ManualEntryLogger;

	public function execute() {
		...

		if ( $this->hasOption( 'with-log' ) ) {
			$manualEntryLogger = new ManualEntryLogger ();
			$manualEntryLogger->registerLoggableEventTypes( array( 'maintenance-script' ) );

			$manualEntryLogger->log( 'maintenance-script', 'UserX', 'rebuildData', 'Some nifty comment about ...' );
		}

```

[...]

PR #1361, opened 2016-01-09
Add MaintenanceLogger, refs #832

mwjames:
image

refs #832


mwjames:
@kghbln Currently outputs what is normally seen when using `--report-runtime`, feel free to add whatever information you want to have logged.

kghbln:
Wow, that's great already. Probably 'maintenance' should be localizable and changed to 'data maintenance script' or perhaps even better 'is a software script that performed data maintenance for Semantic MediaWiki' to have it fit into the general mw log message scheme. Is this the "logentry-smw-maintenance" system message piping into this position though I do not really see where it is used.

What does the $1 mean in "log-show-hide-smw"? I'd say that this is the show/hide for selecting this log to be shown or not.

I could not have made something like this possible since I just did not have enough exposure time to php I'm afraid.

mwjames:
> Probably 'maintenance' should be localizable and changed to 'data maintenance script' or perhaps even better 'is a software script that performed data maintenance for Semantic MediaWiki'

'maintenance' (could be "maintenance-script" in case this fits better) is an event type correlating with the recommendation found in [0]. 

```
{
	"log-name-foo": "Foo log",
	"log-description-foo": "These events track when Foo events happen in the system.",
	"logentry-foo-bar": "$1 {{GENDER:$2|did bar}} to page $3"
}
```

[1] sets the type which could be anything (non localized) and is only bound to the `MaintenanceLogger`.

>  What does the $1 mean in "log-show-hide-smw"? I'd say that this is the show/hide for selecting this log to be shown or not.
 
Relates to `wgFilterLogTypes` and here I'm just following [2].

> I could not have made something like this possible since I just did not have enough exposure time to php I'm afraid.

This I understand hence the helping hand. Most maintenance scripts had been re-factored before allowing us to add something without making the code worse.

MediaWiki dependency is isolated to the `ManualEntryLogger` and the `MaintenanceLogger` is a wrapper reducing the interface and some noise by keeping `ManualEntryLogger` from directly interacting with other parts of the code so that we now simply do:

```php
if ( $this->hasOption( 'with-log' ) ) {
	$maintenanceLogger = $maintenanceFactory->newMaintenanceLogger( 'RebuildConceptCache' );
	$maintenanceLogger->log( $maintenanceHelper->transformRuntimeValuesForOutput() );
}
```

[0] https://www.mediawiki.org/wiki/Manual:Logging_to_Special:Log

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1361/files#diff-160abe824e4fea8250d6af4f4857896eR34

[2] https://github.com/wikimedia/mediawiki/blob/master/languages/i18n/en.json#L2731-L2732

kghbln:
First I have to apologize fro having kept you waiting so long.

Admittedly I do not know how to transform the linked instructions on the mw help page into usable code. Even after two hours I have no idea what to make out of these instructions. Currently the system messages "log-show-hide-smw" and "logentry-smw-maintenance" are dead code.

Is it just a matter of replacing 'maintenance' by the system message "logentry-smw-maintenance"? so we get:

"17:49, 5 February 2016 RebuildData (Talk | block) Maintenance related events emitted by Semantic MediaWiki (Memory used: 32060080 (b: 17812552, a: 49872632) with a runtime of 47.34 sec)"

instead of currently:

"17:49, 5 February 2016 RebuildData (Talk | block) maintenance (Memory used: 32060080 (b: 17812552, a: 49872632) with a runtime of 47.34 sec)"

Another thing that should probably be done is that the users "RebuildData" and "RebuildConceptCache" cannot be registered to avoid piggybacking, but this is an extra. The [NewUserMessage](https://git.wikimedia.org/blob/mediawiki%2Fextensions%2FNewUserMessage/master/NewUserMessage.class.php) extension does this. Perhaps this is useful code.

mwjames:
Example usage:

* `php maintenance/rebuildPropertyStatistics.php --with-maintenance-log`
* `php maintenance/rebuildData.php --with-maintenance-log`
* `php maintenance/rebuildConceptCache.php --with-maintenance-log`

mwjames:
@kghbln `log-show-hide-smw` follows the same pattern as in [0].

[0] https://github.com/wikimedia/mediawiki/blob/master/languages/i18n/en.json#L2829-L2830

kghbln:
@mwjames [This is really cool!](http://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial%3AJournal&type=smw&user=&page=&year=&month=-1&tagfilter=&hide_thanks_log=1&hide_patrol_log=1&hide_tag_log=1&hide_smw_log=1)

PR #1381, opened 2016-01-24
Add localizable context help for properties, refs #1344

mwjames:
"Has property description" is a predefined property and can be used to describe a local property in context of a language.

If a user switches its language preference (`wgLang`) and a property with description matches the language then it will be used during the highlight display.

If a predefined property contains a local "Has property description" then this information will be displayed instead of the standard message.

```
Property: SomeProperty

[[Has property description::Is a property that ...@en]]
[[Has property description::このプロパティは。。。@ja]]

```
The `browsebyproperty` API (#1267) has been amended to contain description information
such as:

```
api.php?action=browsebyproperty&property=SomeProperty&lang=en

{
    "query": {
        "SomeProperty": {
            "label": "SomeProperty",
            "key": "SomeProperty",
            "isUserDefined": "",
            "usageCount": 5343,
            "description": {
                "en": "Is a property that ..."
            }
        }
    },
    "version": 0.2,
    "query-continue-offset": 0,
    "meta": {
        "limit": 50,
        "count": 1
    }
}
```

refs #1344, #1267

### DIFF
--- a/SemanticMediaWiki.php
+++ b/SemanticMediaWiki.php
@@ -151,6 +151,10 @@ $GLOBALS['wgExtensionMessagesFiles']['SemanticMediaWikiMagic'] = $GLOBALS['smwgI
  */
 $GLOBALS['wgExtensionFunctions'][] = function() {
 
+	// 3.x reverse the order to ensure that smwgMainCacheType is used
+	// as main and smwgCacheType being deprecated with 3.x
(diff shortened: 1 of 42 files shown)


mwjames:
https://www.semantic-mediawiki.org/wiki/Help:Special_property_Has_property_description

mwjames:
@kghbln FYI Uses the `MonolingualText` type to support localizable context help.

PR #1384, opened 2016-01-24
Add limitreport to ParserOutput

mwjames:
image


mwjames:
@kghbln FYI

kghbln:
Nice! Just added the qqq messages. They will roll in with the translations.

PR #1382, opened 2016-01-24
Add DispatchingDataValueFormatter

mwjames:
Allow to split formatter related code from a `DataValue` into a registered separate module using the `ValueFormatterRegistry::registerDataValueFormatter`:

- Interface `ValueFormatter`
- Abstract base class `DataValueFormatter` which defines `WIKI_SHORT`, `WIKI_LONG`, `HTML_SHORT`, `HTML_LONG`, and `VALUE`
- `DispatchingDataValueFormatter` will return a formatter based on the `isFormatterFor`
- Moved `MonolingualTextValue` (#1344) formatter related part to `MonolingualTextValueFormatter`

PR #1385, opened 2016-01-24
Add StringValueFormatter + CodeStringValueFormatter, refs #1382

mwjames:
Moving code from `SMWStringValue` to `StringValueFormatter` and `CodeStringValueFormatter`.

refs #1382


mwjames:
@kghbln For some reason we get some text output regression reports relating to 2.4 (I added extra tests but you never know) then this is the first place to look at.

PR #1389, opened 2016-01-25
Add free date/time formatting support

mwjames:
- `-F[ ... ]` denotes the free format in form of
 `|?Date#-F[D M j G:i:s T Y]` or `|?Date#-F[l jS \of F Y h:i:s A]`
- Formatting options encapsulate by `-F[ ... ]` expected to adhere [0] rules
- Rules described by `-F[ ... ]` will use the `IntlTimeFormatter` to replace textual representation for available localizable text elements (e.g. Monday -> 日 or 日曜日)
- PHP's `DateTime` object handles the actual formatting hence [0]

## Example output

image

## Features and limitations 

- Dates like `10000 BC` (prehistroy) are not supported by the formatter (the ISO date `--10000-01-01` is returned instead) other BC(E) dates like `2 Feb 900 BC` will be formatted as `-0900/01/25` or `-0900/02/02 JL`
- Dates stored with a JD(JL) are not automatically transformed and instead marked with `JL` to indicate its model

refs #1388

[0] http://php.net/manual/en/function.date.php

### DIFF
--- a/includes/dataitems/SMW_DI_Time.php
+++ b/includes/dataitems/SMW_DI_Time.php
@@ -35,6 +35,8 @@ class SMWDITime extends SMWDataItem {
 	const PREC_YMD  = 2;
 	const PREC_YMDT = 3;
 
+	const PREHISTORY = -10000;
+
 	/**
 	 * Maximal number of days in a given month.
 	 * @var array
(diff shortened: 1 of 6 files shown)


mwjames:
@SemanticMediaWiki/testers Would be nice if someone finds to time to play around with this (for those using a fresh 2.4-alpha, please remember to run `update.php` or `setupStore.php`)

jaideraf:
I am looking for a time to test this, but I just want to say this seems awesome!

jaideraf:
I tested: it's awesome indeed. :+1:
I did not find any problem, though I tested without SRF.

mwjames:
> I tested: it's awesome indeed

Thanks for testing.

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1389_%28DateFreeFormat%29

Issue #1348, opened 2016-01-05
years appearing as "1 January 1970 + timestamp"


Contrafibularity:
This is an old one, first reported in 2011 I think: years before 1100(?) appearing as "1 January 1970 + timestamp", as demonstrated here: http://sandbox.semantic-mediawiki.org/wiki/User:Cavila/test   

See https://phabricator.wikimedia.org/T56151, which was closed as being a duplicate of another report, which for some reason I cannot find.

mwjames:
>  which was closed as being a duplicate of another report, which for some reason I cannot find.

https://phabricator.wikimedia.org/T49259

## Solution

After a "very" quick look, the following change to `SMWTimeValue` should allow for a correct interpretation of a year annotation.

```php
-		// Check if it's parseable by wfTimestamp when it's not a year (which is wrongly interpreted).
-		if ( strlen( $value ) != 4 && wfTimestamp( TS_MW, $value ) !== false ) {
+		// If no space is present in the user value we assume it is a year
+		// value without any further information
+		if ( strpos( $value, ' ' ) === false ) {
+			$this->m_dataitem = new SMWDITime( $this->getCalendarModel( null, $value, null, null ), $value );
+		} elseif ( strlen( $value ) != 4 && wfTimestamp( TS_MW, $value ) !== false ) {
+			// Check if it's parseable by wfTimestamp when it's not a year (which is wrongly interpreted).
```
```
[[Date::-1000]], [[Date::-100000]]

[[Date::910]], [[Date::910 AD]], [[Date::910 BC]]

[[Date::1582]], [[Date::1582 AD]], [[Date::1582 BC]]

[[Date::1910]], [[Date::1910 AD]], [[Date::1910 BC]]

[[Date::11910 ]], [[Date::11910 AD]], [[Date::11910 BC]]
```

## Tasks

- [ ] Do some more manual testing (in case I missed something)
- [ ] Turn this into a PR
- [ ] Write some tests (for this class I can live without an specific unit test BUT I'd like to see some [0] integration tests so that user input/output is covered in case the class gets an overhaul in future)

[0] https://www.semantic-mediawiki.org/wiki/Help:Integration_tests

mwjames:
On a side note with [0] noting that:

[...]

PR #1395, opened 2016-02-03
Fix single year detection, refs #1348

mwjames:
For example, year `1902` is being interpret as `1902` and not as "1 January 1970".

refs #1348, https://phabricator.wikimedia.org/T56151, https://phabricator.wikimedia.org/T49259


mwjames:
@kghbln FYI

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1395_%28YearOnly%29

PR #1396, opened 2016-02-03
Serialize positive era information, refs #1348

mwjames:
Keep positive era information serialized so that the display can account for what it was intended.

[0] noted "... is displayed without the 'AD' -- '910 AD' becomes '910' ... should not change the meaning of the date"

- If an era for a positive date was added then AD is placed in front indepedant of the actual annotated value (e.g. "AD 1902", "1902 AD").
- Internally, `-` indicates that it is a `BC(E)` era, we only mark a positive number with a `+` in case a positive year has an explicitly annotated `AD/CE` era and in all other cases the serialized number remains plain.
- `[[Date::-901 AD]]` is not allowed and has a test case

refs #1348

[0] https://phabricator.wikimedia.org/T49259

### DIFF
--- a/includes/dataitems/SMW_DI_Time.php
+++ b/includes/dataitems/SMW_DI_Time.php
@@ -91,6 +91,11 @@ class SMWDITime extends SMWDataItem {
 	protected $timezone;
 
 	/**
+	 * @var integer|null
+	 */
+	protected $era = null;
+
+	/**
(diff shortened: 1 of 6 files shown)


mwjames:
@kghbln FYI

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1396_%28Era%29

PR #1398, opened 2016-02-03
Semantics opt-out setting (a.k.a. disableSemantics)

mwjames:
refs [0]

`disableSemantics()` as a non-disruptive setting that can be added to the `LocalSettings` to temporary disable the storage and query capabilities of SMW for a selected instance.

- In-text annotation parsing remains active to avoid having `::` annotation clutter the display
- DB update for any extension that rely on SMW is disabled (== no DB activity)
- `#ask`/`#show` are disabled
- All SMW related special pages are disabled
- Special:Version will display the extension (as it is installed)
- Namespace Property/Concept remains available but is disabled
- API modules are disabled

[0] http://wikimedia.7.x6.nabble.com/Problems-installing-SMW-into-wiki-family-td5058854.html

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -36,6 +36,20 @@ $GLOBALS['smwgExtraneousLanguageFileDir'] = __DIR__ . '/languages';
 ##
 
 ###
+# Semantic MediaWiki's operational state
+##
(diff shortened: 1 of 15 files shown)


mwjames:
@kghbln To counter to whole, "I hate Composer because I cannot disable an individual instance" discussion.

PR #1399, opened 2016-02-03
Isolate JD computation

mwjames:
- A JL date is now explicitly marked as such
- Allows to annotate `[[Date::2488346.0804977 MJD]]` or `[[Date::2488346.0804977 JD]]`.

While the switch over year for the GR calendar is 1582/10/4 and the assumption that everybody knows the difference between JL/GR maybe valid, it still can create confusion as the what date is being used/displayed.

image

### DIFF
--- a/includes/dataitems/SMW_DI_Time.php
+++ b/includes/dataitems/SMW_DI_Time.php
@@ -1,6 +1,7 @@
 <?php
 
 use SMW\DataItemException;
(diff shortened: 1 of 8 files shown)


mwjames:
@kghbln FYI

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1399_%28JulianDay%29

PR #1400, opened 2016-02-03
Clean-up TemperatureValue

mwjames:
Enables to set unit preference via `Display unit`.


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1400_%28TemperatureValue%29

Issue #541, opened 2014-09-16
Result column duplicaton on "further results ..."


roland2025:
Problem looks like one from a few years ago.

https://bugzilla.wikimedia.org/show_bug.cgi?id=33473

{{#ask: [[SOME QUERY]]|?Property|mainlabel=some text|limit=0}}

when mainlabel is defined, query page will show duplicate columns.

Contrafibularity:
Is this perhaps in any way related to #481 and #249? It's similar at least in the sense that syntax is added to the original query after "further results" has been clicked.

JeroenDeDauw:
@kghbln so this is a usability issue and not a bug? (then the label should be enhancement)

If it's a UX issue, I'm also gonna guess it's still present in 2.1?

kghbln:
Yeah, this on is still present. This started with SMW 1.6 and is present ever since. At that time four years ago I labelled this as a bug since it appeared to be not very intuitive to suppress the main result column and than adding it again at the same time. However, this may indeed better be called an enhancement now.

mwjames:
Expected to be solved by #1407.

kghbln:
@mwjames Wow, great, this was a personal long time wish. Works perfect as the [field test](http://sandbox.semantic-mediawiki.org/wiki/Issue/541) proves. Things are now much more intuitive! Thanks a ton!

kghbln:
:+1:  This fix is eligible for the kudo of the year award. :)

PR #1407, opened 2016-02-16
Add extra download links to Special:Ask

mwjames:
- Results displayed can now be downloaded immediately in format JSON | CSV | RSS | RDF
- Tidy the search navigation for a better interaction experience
- #541 expected to be solved
- Search appropriately preserves the input such as `Has property=Foo|+index=1`

image

image


mwjames:
@kghbln While I was only interested in the quick download links, I nevertheless added some visual honey for the common user.

kghbln:
@mwjames This is really a cool idea! Just saw this is action and indeed improves what is generally referred to as "user experience". Thanks for this!

PR #1354, opened 2016-01-07
Added usage count information to property pages

kghbln:
References issue https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1313

Code changes mainly provided by James Hong Kong. Thanks!

PR #1409, opened 2016-02-16
Follow-up on #1354 (property usage count display)

mwjames:
refs #1354

image

image

image


mwjames:
@kghbln I rearranged the display since the display of "dynamic" count information mingled with static text didn't really clicked with me.

The top indicator was introduced with MW 1.25, anyone else with lower MW release will see nothing. Users you want to have this information should upgrade.

For the property page, if you hover over the number then you get time and date (I didn't want to clutter the display, the concept page is a bit different as you have to run a job to update the cache)

Issue #1295, opened 2015-12-08
1.27 issues


mwjames:
## Known (reported) issues
- [0] / TitleMoveComplete-closure has invalid call signature has been reported  at 15 Nov with no response from WMF, possible being connected to [2]
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1332
## Potential issues

- [1] is a possible candidate for problems (Migrate WikiPage::doEditContent to using startAtomic/endAtomic)

[0] https://phabricator.wikimedia.org/T118683
[1] https://phabricator.wikimedia.org/T120718 
[2] https://phabricator.wikimedia.org/T120218

mwjames:
... and master is broken again!

Some merges to master between `2015-12-13 08:36`and `2015-12-09 10:45` caused the test failures (and for the record, we have not merged any changes functional or test related between the mentioned time frame).

[0] master failed `2015-12-13 08:36`

```
Semantic MediaWiki: 2.4-alpha (SMWSQLStore3, sqlite)
MediaWiki:          1.27alpha (Extension vendor autoloader)
Execution date:     2015-12-13 08:36
PHPUnit 4.6.10 by Sebastian Bergmann and contributors.
Configuration read from /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/phpunit.xml.dist
.............................................................   61 / 3213 (  1%)
.............................................................  122 / 3213 (  3%)
.............................................................  183 / 3213 (  5%)
.............................................................  244 / 3213 (  7%)
.............................................................  305 / 3213 (  9%)
.............................................................  366 / 3213 ( 11%)
.............................................................  427 / 3213 ( 13%)
.............................................................  488 / 3213 ( 15%)
.............................................................  549 / 3213 ( 17%)
.............................................................  610 / 3213 ( 18%)
....................R.............R..........................  671 / 3213 ( 20%)
.............................................................  732 / 3213 ( 22%)
.............................................................  793 / 3213 ( 24%)
.............................................................  854 / 3213 ( 26%)
.............................................................  915 / 3213 ( 28%)
.............................................................  976 / 3213 ( 30%)
............................................................. 1037 / 3213 ( 32%)
............................................................. 1098 / 3213 ( 34%)
............................................................. 1159 / 3213 ( 36%)
............................................................. 1220 / 3213 ( 37%)
............................................................. 1281 / 3213 ( 39%)
[...]

Issue #1186, opened 2015-10-01
Issues with "sandbox.semantic-mediawiki.org"


kghbln:
This tracking issue is intended for reporting issues with the above mentioned wiki, e.g. it is down etc.

kghbln:
I am closing this for now. If there is an issue this one should be reopened.

kghbln:
Scheduled downtime tomorrow 2015-10-21 for about 10 minutes between 10 and 11 pm CET

kghbln:
There is currently 2015-12-07 an issue with the server hosting sandbox which started at 15:15 UTC. This is being worked on and hopefully the machine is back up running soon.

kghbln:
@mwjames Please ping me here as soon as [this](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1335#issuecomment-169837796) was merged for me to do the "update.php" run.

kghbln:
Semantic Form is currently broken, thus the site is down: [T123308](https://phabricator.wikimedia.org/T123308)

mwjames:
@kghbln As for the Chameleon-skin, maybe copy `/skins/chameleon/layouts/fixedhead.xml` to a local folder so that:

```
$egChameleonLayoutFile = __DIR__ . `myLocation/sandbox-fixedhead.xml';
```

and after [0] modify the content of `sandbox-fixedhead.xml` so that the language bar turns up in the NavMenu.

```
@@ -24,11 +24,11 @@ with this program. If not, see <http://www.gnu.org/licenses/>.
 -->
 <structure>
 
 	<component type="NavbarHorizontal" fixed="1">
 		<component type="Logo" position="head"></component>
-		<component type="NavMenu" flatten="navigation" showTools="no" showLanguages="no"></component>
+		<component type="NavMenu" flatten="navigation" showTools="yes" showLanguages="yes"></component>
 		<component type="PageTools" position="right" hideSelectedNameSpace="yes"></component>
 		<component type="SearchBar" position="right"></component>
 		<component type="PersonalTools" position="right"></component>
 	</component>
 
@@ -45,11 +45,11 @@ with this program. If not, see <http://www.gnu.org/licenses/>.
 			</cell>
 		</row>
 
 		<row>
 			<cell span="12">
-				<component type="ToolbarHorizontal" hideTools="no" hideLanguages="no"></component>
[...]

PR #1410, opened 2016-02-17
Support for DISPLAYTITLE caption

mwjames:
Adds general support for subjects that contain a `{{DISPLAYTITLE:Foo}}`. The pre-defined property `_DTITLE` is unrestricted therefore manually annotations via `Display title of` [0] are allowed.

Settings in connection with `{{DISPLAYTITLE:Foo}}`:

```
$wgAllowDisplayTitle = true;
$wgRestrictDisplayTitle = false;
```

[0] https://www.semantic-mediawiki.org/wiki/Help:Special_property_Display_title_of

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -964,6 +964,17 @@ $GLOBALS['smwgEnabledInTextAnnotationParserStrictMode'] = true;
 # - SMW_DV_PVAP (Allows pattern) to allow regular expression pattern matching
 # when `Allows pattern` property is assigned to user-defined property
 #
+# - SMW_DV_WPV_DTITLE (WikiPageValue) is to allow requesting a lookup for a display
+# title and if present will be used as caption for the invoked subject
+#
+# - SMW_DV_PROV_DTITLE (PropertyValue) in combination with SMW_DV_WPV_DTITLE; If
+# enabled it will attempt to resolve a property label by matching it against a
+# possible assigned property "Display title of" value. For example, property
+# "Foo" has "Display title of" "hasFoolishFoo" where "hasFoolishFoo" is being
+# resolved as "Foo" when creating annotations. Currently, this uses an
+# uncached lookup and therefore is disabled by default to avoid a possible
+# performance impact (which has not been established or analyzed).
+#
 # @since 2.4
 ##
-$GLOBALS['smwgDVFeatures'] = SMW_DV_PROV_REDI | SMW_DV_MLTV_LCODE | SMW_DV_PVAP;
+$GLOBALS['smwgDVFeatures'] = SMW_DV_PROV_REDI | SMW_DV_MLTV_LCODE | SMW_DV_PVAP | SMW_DV_WPV_DTITLE;
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -382,5 +382,6 @@
 	"smw-datavalue-allows-pattern-mismatch": "\"$1\" was classified as invalid by the \"$2\" regular expression.",
 	"smw-datavalue-allows-pattern-reference-unknown": "The \"$1\" pattern reference could not be match to an entry in [[MediaWiki:Smw allows pattern]].",
 	"smw-datavalue-feature-not-supported": "The \"$1\" feature is not supported or was disabled on this wiki.",
-	"smw-pa-property-predefined_pvap": "\"$1\" is a predefined property that can specify a [[MediaWiki:Smw allows pattern|pattern reference]] to apply [https://en.wikipedia.org/wiki/Regular_expression regular expression] matching and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki]."
+	"smw-pa-property-predefined_pvap": "\"$1\" is a predefined property that can specify a [[MediaWiki:Smw allows pattern|pattern reference]] to apply [https://en.wikipedia.org/wiki/Regular_expression regular expression] matching and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki].",
+	"smw-pa-property-predefined_dtitle": "\"$1\" is a predefined property that can assign a distinct display title to an entity and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki]."
 }
--- a/includes/Defines.php
+++ b/includes/Defines.php
@@ -142,4 +142,6 @@ define( 'SMW_DV_NONE', 0 );
 define( 'SMW_DV_PROV_REDI', 2 );  // PropertyValue to follow a property redirect target
 define( 'SMW_DV_MLTV_LCODE', 4 );  // MonolingualTextValue requires language code
 define( 'SMW_DV_PVAP', 16 );  // Allows pattern
+define( 'SMW_DV_WPV_DTITLE', 32 );  // WikiPageValue to use an explicit display title
+define( 'SMW_DV_PROV_DTITLE', 64 );  // PropertyValue allow to find a property using the display title
 /**@}*/
--- a/includes/SMW_Infolink.php
+++ b/includes/SMW_Infolink.php
@@ -461,14 +461,7 @@ class SMWInfolink {
 
 			foreach ( $ps as $p ) {
 				if ( $p !== '' ) {
-					$p = rawurldecode( str_replace( '-', '%', $p ) );
-					$parts = explode( '=', $p, 2 );
-
-					if ( count( $parts ) > 1 ) {
-						$result[$parts[0]] = $parts[1];
-					} else {
-						$result[] = $p;
-					}
+					$result[] = rawurldecode( str_replace( '-', '%', $p ) );
 				}
 			}
 		}
--- a/includes/articlepages/SMW_PropertyPage.php
+++ b/includes/articlepages/SMW_PropertyPage.php
@@ -153,10 +153,15 @@ class SMWPropertyPage extends SMWOrderedListPage {
 
 		if ( count( $diWikiPages ) > 0 ) {
 			$pageLister = new SMWPageLister( $diWikiPages, null, $this->limit, $this->from, $this->until );
+
 			$this->mTitle->setFragment( '#SMWResults' ); // Make navigation point to the result list.
 			$navigation = $pageLister->getNavigationLinks( $this->mTitle );
 
-			$titleText = htmlspecialchars( $this->mTitle->getText() );
+			$dvWikiPage = DataValueFactory::getInstance()->newDataItemValue(
+				$this->mProperty
+			);
+
+			$titleText = htmlspecialchars( $dvWikiPage->getWikiValue() );
 			$resultNumber = min( $this->limit, count( $diWikiPages ) );
 
 			$result .= "<a name=\"SMWResults\"></a><div id=\"mw-pages\">\n" .
@@ -210,7 +215,7 @@ class SMWPropertyPage extends SMWOrderedListPage {
 			}
 
 			// Property name
-			$searchlink = SMWInfolink::newBrowsingLink( '+', $dvWikiPage->getShortHTMLText() );
+			$searchlink = SMWInfolink::newBrowsingLink( '+', $dvWikiPage->getWikiValue() );
 			$r .= '<tr><td class="smwpropname">' . $dvWikiPage->getShortHTMLText( smwfGetLinker() ) .
 			      '&#160;' . $searchlink->getHTML( smwfGetLinker() ) . '</td><td class="smwprops">';
 
--- a/includes/datavalues/SMW_DV_Property.php
+++ b/includes/datavalues/SMW_DV_Property.php
@@ -58,15 +58,6 @@ class SMWPropertyValue extends SMWDataValue {
 	}
 
 	/**
-	 * @since 2.4
-	 *
-	 * @return boolean
-	 */
-	public function isToFindPropertyRedirect() {
-		return ( $this->getOptionValueFor( 'smwgDVFeatures' ) & SMW_DV_PROV_REDI ) != 0;
-	}
-
-	/**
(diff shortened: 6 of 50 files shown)


mwjames:
@kghbln @cicalese Since we support `DEFAULTSORT`, I now made this a bit more round in the sense that if a subject contains a `DISPLAYTITLE` then a `Display title of` property will be set and wikipage captions are to recognize the explicit title and modify the output in queries etc. through the wikipage dataValue object.

mwjames:
@cicalese I made a comment in the serializer [0] on how to access the display title content but I'd like to see some performance tests whether this is elevating the runtime or not before making this available.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1410/files#diff-45f5fa9530bfcefe0517e74d8acf8a3eR102

mwjames:
@SemanticMediaWiki/testers Anyone care to test?

cicalese:
Thank you, @mwjames!

Coincidentally, I was working on my older patch tonight, but this is obviously a more full featured solution. I will test this, with the QueryResultSerializer bit uncommented plus my suggested changes to res/smw/data/ext.smw.data.js and res/smw/data/ext.smw.dataItem.wikiPage.js to make this useful for JavaScript result formats.

cicalese:
I'm having the following issue while testing. I create a page with DISPLAYTITLE set.

screen shot 2016-02-17 at 7 53 50 am

I save the page and immediately browse the properties for the page and see the "Display title of" property:

screen shot 2016-02-17 at 7 54 02 am

But, when I refresh the page, it disappears:

screen shot 2016-02-17 at 7 54 18 am

This is probably not significant, but if I refresh the page again, "Modification date" changes from text to a link:

screen shot 2016-02-17 at 7 54 31 am

mwjames:
I tried to follow your example but could not confirm the observation. Did you run a `php update.php` (2.4 requires to run it)?

Tested with MediaWiki	1.25.1
PHP	5.6.8 (apache2handler)
MySQL	5.6.24

cicalese:
Yes, I ran update. My environment:

MediaWiki	1.27.0-alpha (86c08b2) 18:39, 16 February 2016
HHVM	3.6.5 (srv)
MySQL	5.5.47-0ubuntu0.14.04.1

I'm running under MediaWiki-vagrant.

mwjames:
@kghbln In case you have a minute maybe you could give it a run.

> MediaWiki 1.27.0-alpha (86c08b2)

We can't use 1.27 since all our integration tests are broken after 2016-01-24 (caused by some MW-core change) which also includes #1295.

cicalese:
OK, I will back off to an earlier MW version.

mwjames:
> ... "Modification date" changes from text to a link:

I've never seen this behaviour and our sandbox [0] also doesn't indicate that such phenomenon (aside from the current PR) is observable.

[0] http://sandbox.semantic-mediawiki.org/wiki/Main_Page

mwjames:
> But, when I refresh the page, it disappears:

That shouldn't happen because `_DYTLE` (Display title of) is set by (or better in) the same hook `NewRevisionFromEditComplete` as `_MDAT` (Modification date) which is only executed after an edit.

kghbln:
> In case you have a minute maybe you could give it a run.

I currently cannot go 1.27 since this version now requires PHP 5.5.9+ or HHVM 3.1+ which is probably the reason for the tests to fail. I will see to it that the sandbox server (https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1186) gets an upgrade and go 1.27-alpha thereafter.

mwjames:
> now requires PHP 5.5.9+ or HHVM 3.1+ which is probably the reason for the tests to fail.

The test failures have nothing to do with the PHP version, they pass on MW 1.26/PHP 5.6.

kghbln:
> The test failures have nothing to do with the PHP version, they pass on MW 1.26/PHP 5.6.

Ouch, so this is worse than I had hoped. Sometimes dreams die a sudden death I suppose. :(

cicalese:
I have now tested under:

MediaWiki 1.26.2 (c2d388d) 
PHP 5.5.32 (apache2handler) 
MySQL 5.5.48 

The patch is working better in that environment. When I create a page, the "Display title of" property is getting set. However, when I run the refreshLinks/runJobs maintenance scripts, the property goes away. I am suspicious that no jobs are being put on the job queue by refreshLinks. I also tried running the SMW rebuildData script, but it did not help. If I edit the page and re-save it, the property comes back.

As an additional test, I added in the JavaScript changes from my earlier pull request and uncommented the line in the serialization code. With those modifications, the datatables result format beautifully renders the displaytitle as the link text. This is HUGE progress, and I do hope you will consider incorporating these additions to SMW.

Thank you.

mwjames:
> I have now tested under:

Thanks for testing.

> I also tried running the SMW rebuildData script, but it did not help. If I edit the page and re-save it, the property comes back.

Since we rely on `...->getProperty( 'displaytitle' )` I moved from the `NewRevisionFromEditComplete` to the `ParserAfterTidy` hook which allows to fetch the 'displaytitle' on a regular purge or update instead of only on an edit which should also be sufficient for the `rebuildData` script.

cicalese:
@mwjames, I don't see a new commit to test that the refreshLinks bug is fixed. Is there anything else I can do to help move this forward? Have you had a chance to assess any performance impact (hopefully minimal) to uncommenting the display title query in the serialization code? If you are satisfied that it does not have a significant impact, I would like to propose adding the JavaScript modifications from my earlier patch to this or a following patch. Thank you.

mwjames:
> I don't see a new commit to test that the refreshLinks bug is fixed. Is there anything else I can do to help move this forward?

There you go.

>  Have you had a chance to assess any performance impact (hopefully minimal) to uncommenting the display title query in the serialization code?

Indeed I did (always a daunting task) and after some back and forth where the `InMemoryCache` did cache repeated requests for the session I wasn't satisfied that it would still engage in store requests. Since I can not assume that every user has enabled the `CachedValueLookupStore`, I decided to introduce `CachedPropertyValuesPrefetcher` so that preselected property lookups (mostly for high frequent usages) can divert to a persistent cache (using `redis`) without compromising the overall data integrity where each time a subject is changed the cache for the entity is evicted as well.

The `CachedPropertyValuesPrefetcher` also solves my concerns over repeated requests from the API (each subject/subobject are to be checked at least once during a request to see whether a display title was set or not).

On a different note, if a subobject does not assign a `Display title of` property as part of its declaration, it will inherit the display title of its parent (the entity that embeds the subobject).

Now that subobjects are considered as well, I've decided to place values from `_DTITLE` in an extra `fpt` table therefore running `update.php` or `setupStore.php` is required.

`SMW_DV_WPV_DTITLE` has been added as feature flag to `$GLOBALS['smwgDVFeatures']` in order to disable the caption usage for wikipage dataValues and restore the current behaviour by removing the entry from the setting.

`SMW_DV_PROV_DTITLE` has been added as option but is disabled by default (see the settings comment).

>  impact (hopefully minimal) to uncommenting the display title query in the serialization code?

I leave the code commented in the serializer so that is unrelated to the current PR (and can be enabled by your PR together with bumping the version number).

mwjames:
> It is based upon the enhancements in the dtitle branch. I have tested it until MediaWiki 1.26.2, and it is functioning properly.

@cicalese My I go right in my assumption that testing was successful, given the implicit confirmation in #1423?

@kghbln The sandbox would need a nudge (either `update.php` or `setupStore.php`) after the merge.

@thingles Just some heads up, we make intensive use of the cache layer to ensure we don't flood the `Store` with lookups and based on your cache settings this might occupy some space given the frequent updates by bots, if you encounter an issue just remove the `SMW_DV_WPV_DTITLE` from the settings as outlined in [0].

@Fannon Does this solve the https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/144 issue?

[0] https://www.semantic-mediawiki.org/wiki/Help:$smwgDVFeatures#Changing_the_default_setting

cicalese:
@mwjames Yes, my testing of this patch under MW 1.26.2, augmented by #1423, was successful both for the datatables and filtered result formats.

My testing uncovered an unrelated bug in the SMW ask API that I will be posting about separately.

mwjames:
> My testing uncovered an unrelated bug in the SMW ask API that I will be posting about separately.

I for once hoped I can focus on something new instead of fixing bugs. (would appreciate a good analysis)

cicalese:
Yes, that's why I haven't posted it yet. I'm trying to narrow it down so you don't have to hunt (too much) for it.

Fannon:
@mwjames I haven't tested your changes, but I think the issue you've mentioned is more related to #1423. Semantic Title support still wouldn't come automatically by your changes, but would allow for a cleaner implementation than my current one.

kghbln:
@mwjames Thanks for crunching this issue and the note on updating! I have just manually updated sandbox, so we are fine here.

mwjames:
@kghbln I think we need the `$wgAllowDisplayTitle = true;` and `$wgRestrictDisplayTitle = false;` settings on the sandbox.

@cicalese FYI

cicalese:
I just this very moment discovered the same thing and was planning to write to @kghbln about it.

kghbln:
Done now. I got a bit distracted by my German IBAN regex. All cool now.

cicalese:
Thank you, @kghbln!

All, you can now see a working test of this functionality at [0].

The first tab, table, is a baseline which was already working before this patch.

@Fannon, does the example in the filtered tab address your original concern? This fix did not require #1423 since the filtered result format does not use JavaScript to render the filters. But, maybe there was an additional issue with the filtered result format.

When the sandbox is updated to include the recently merged #1423, the display titles of the pages on the datatables tab will render correctly.

@mwjames, you can also see the other unrelated bug in action on the datatables tab by refreshing the table with the button in the top right. All of the links become red. I will add a separate bug report for that linking to the example in the sandbox.

[0] http://sandbox.semantic-mediawiki.org/wiki/Display_Title_Test

mwjames:
> can also see the other unrelated bug in action on the datatables tab by refreshing the table with the button in the top right.

I think this worked in previous releases (not sure though) but since the [0] JSON format change the following "...Modules that were previously outputting raw booleans in JSON may now output those properties using the convention that has always been the standard (for version 1): empty-string for true and absent key for false...." may pushed this over the cliff.

[0] https://www.mediawiki.org/wiki/API:JSON_version_2

cicalese:
I just opened #1428 for that unrelated bug. Hopefully with the diagnosis, the fix will be straightforward.

kriegfrj:
How does this feature addition tie in with the Semantic Title extension (https://www.mediawiki.org/wiki/Extension:Semantic_Title)? Does it make it redundant or partially redundant, or is it complementary?

mwjames:
> How does this feature addition tie in with the Semantic Title extension (https://www.mediawiki.org/wiki/Extension:Semantic_Title)? Does it make it redundant or partially redundant, or is it complementary?

I don't know but maybe @cicalese @Fannon can answer this question otherwise I'd suggest you make an inquiry on the extension page. 

If someone has an answer to this question it may prove helpful to extend [0] as orientation for other users.

[0] https://www.semantic-mediawiki.org/wiki/Help:Special_property_Display_title_of

cicalese:
This feature,(replacing the page title with the display title, if present, in links to a page in query results and adding a Semantic MediaWiki property holding the display title)  is complementary to the functionality provided by the SemanticTitle extension.

SemanticTitle performs several important functions in addition to this feature:

- replaces the page title with the display title, if present, in wikitext links to a page as long as alternate link text is not provided (both links from other pages as well as self links, the latter of which appear as bold text on the page)
- adds a parser function to return the display title of a page (works in the absence of Semantic MediaWiki, which provides the same data through a query)
- optionally adds a subtitle containing the page title to pages with a display title

In addition, for legacy reasons, SemanticTitle supports setting the display title of a page from the value of a Semantic MediaWiki property or Cargo field. This was rendered unnecessary in version 3.0 of SemanticTitle, which retrieves the display title from the same mechanism used by the MediaWiki DISPLAYTITLE magic word (the displaytitle value in the page_props database table). Setting the display title using the DISPLAYTITLE magic word is both more efficient and behaves better with respect to caching than using the legacy mechanisms, which will be removed in a future version of SemanticTitle.

SemanticTitle no longer requires Semantic MediaWiki - although they are compatible and, as described above, complementary. It is possible that SemanticTitle will be renamed DisplayTitle in the future, to more closely align it with its relationship to DISPLAYTITLE and to make it clear that it does not require Semantic MediaWiki. This would be done in conjunction with removing the legacy mechanisms described above. It is also possible that we might try to have the three SemanticTitle features listed above accepted as optional capabilities in MediaWiki core.

kriegfrj:
Thanks for that summary, @cicalese. I managed to find out that they were compatible - I simply changed the SemanticTitle configuration to use the `Display title of` built-in property rather than my own, and then the `#semantic-title` parser function worked. I agree that this parser function it would make sense for it (or an equivalent) to be somewhere in the MW core - it is often handy to find out what the display title is of another page.

Interesting comment about using the `DISPLAYTITLE` tag directly rather than via a property as being more efficient - is this true also of the `Display title of` property (ie, is it more efficient to use `DISPLAYTITLE` than to use `Display title of` property)?

mwjames:
> Interesting comment about using the DISPLAYTITLE tag directly rather than via a property as being more efficient - is this true also of the Display title of property (ie, is it more efficient to use DISPLAYTITLE than to use Display title of property)?

To clarify further on this issue and as noted in [0], this feature is about using DISPLAYTITLE as method to influence the display of a page entity. The `Display title of` property is to indicate to other SMW related processes to use its value as preferred caption. Using `Display title of` itself may make sense in some situations [1] but best practice is to rely on DISPLAYTITLE to modify the display characteristics as we do in a similar case with DEFAULTSORT to alter the sort attribute.

[0] https://www.semantic-mediawiki.org/wiki/Display_title
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1410#issuecomment-188066483

PR #1417, opened 2016-02-24
Define value constraint using regular expressions

mwjames:
Similar to `Allows value`, the `Allows pattern` property is to specify a regular expression to match permissible values for an assigned property. If a value cannot be matched then a general purpose message is being displayed.

To avoid any issues with expressions that include | or [] it is suggested to use #set (`|+pipe` indicator ensures that MW doesn't interfere with the expression):

```
Whitelist

{{#set:
 |Allows pattern=^(Foo|Foo bar|Bar)$|+pipe
}}
```
```
IPv4

{{#set:
 |Allows pattern=(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$|+pipe
}}
```
- `$GLOBALS['smwgDVFeatures'] = SMW_DV_PAVP` is enabled by default but if removed will disable the functionality
- https://www.semantic-mediawiki.org/wiki/Help:Special_property_Allows_pattern
- http://sandbox.semantic-mediawiki.org/wiki/Issue/1417_%28Allows_pattern%29


mwjames:
@SemanticMediaWiki/testers I'd appreciate if this finds some love amongst testers to give it a run down.

jaideraf:
Cool!
@mwjames This feature applies to all datatypes (Text, Page, URL, etc.)?

jaideraf:
>I'd appreciate if this finds some love amongst testers to give it a run down.

Tested. It seems fine to me. :+1:

JeroenDeDauw:
https://twitter.com/benestar_wm/status/702803623378591744

mwjames:
> https://twitter.com/benestar_wm/status/702803623378591744

> Bene* ‏@benestar_wm 14h14 hours ago
> JeroenDeDauw SemanticMW github I remember we had some concerns when considering this for wikidata because of security issues

As listed in [0], " The regex modifier `/e` is not permitted and removed from any expression".

If users feel more comfortable we can ship this as an opt-in rather than an opt-out by removing the `SMW_DV_PAVP` flag from `$GLOBALS['smwgDVFeatures']` in the initial deploy setting.

[0] https://www.semantic-mediawiki.org/wiki/Help:Special_property_Allows_pattern

JeroenDeDauw:
I don't remember the details of the concerns @Benestar mentioned, though have a vague notation that it was DoS by means of resource consumption.

Benestar:
Related ticket can be found here: https://phabricator.wikimedia.org/T105126

Benestar:
Also relevant and linked from the ticket: http://www.regular-expressions.info/catastrophic.html

mwjames:
Thanks

> Related ticket can be found here: https://phabricator.wikimedia.org/T105126

I did a quick read through and with "... about this for months now and it seems there is no solution. Either have regular expressions or not.", I concur that I want regex support but I also recognize the potential impact therefore I've decided (just made a rough prototype) to add an administrative level.

Instead of allowing to add a pattern directly, the property expects a reference (as shown below) with the actual regex being maintained in `MediaWiki:Smw allows pattern` in order to have more control over who can edit the page [0].

image

[0] https://www.mediawiki.org/wiki/Manual:$wgNamespaceProtection

mwjames:
Over the weekend, I was elaborating on the approach and as indicated above, regular expression declaration has been moved to one dedicated place `MediaWiki:Smw allows pattern`.

Not enough that the property `Allows pattern` now only excepts a reference, `MediaWiki:Smw allows pattern` page is also not editable by anyone, users are required to have the `smw-patternedit` right (see the newly created `smwcurator` group). By default `smw-patternedit` is not assigned to any user (not even the sysop) and therefore has to be given explicitly.

@kghbln FYI, PR comes with a new group/right (`smwcurator`/`smw-patternedit`) [0]

[0] https://www.semantic-mediawiki.org/wiki/Help:Permissions_and_user_rights

mwjames:
Speaking of permission, @kghbln would you mind assigning the curator group to me so I can create an example.

image

kghbln:
@mwjames Not at all and this is done now. Great to see that my setup works indented.

kghbln:
@JeroenDeDauw So far you do not have an account on sandbox. Shame on you and currently not extra rights! ;)

mwjames:
> Not at all and this is done now. Great to see that my setup works indented.

:+1: 

* http://sandbox.semantic-mediawiki.org/wiki/Issue/1417_%28ISBN-10%29
* http://sandbox.semantic-mediawiki.org/wiki/Issue/1417_%28Allows_pattern%29

kghbln:
> http://sandbox.semantic-mediawiki.org/wiki/Issue/1417_%28ISBN-10%29
> http://sandbox.semantic-mediawiki.org/wiki/Issue/1417_%28Allows_pattern%29

Cool, will try to add patterns for IBAN DE, AT and CH

kghbln:
[All one needs for IBANs](http://sandbox.semantic-mediawiki.org/wiki/Issue/1417_%28IBAN-DE%29) including ref!

kghbln:
Are there more than one "Allows pattern" assignments possible per property? I would guess yes.

mwjames:
> Are there more than one "Allows pattern" assignments possible per property? I would guess yes.

Yes BUT only the last will win [0].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/PropertySpecificationLookup.php#L183

mwjames:
> BUT only the last will win [0].

This saves a lot of headaches as we don't have to make assumptions. Regular expressions are very "stretchy", so for those who want to combine different patterns, create a new reference, merge them in a coherent way leaving SMW to not second guess the intend.

kghbln:
> Yes BUT only the last will win [0].

Thanks for the info. Indeed this makes much sense. This is now documented.

PR #1421, opened 2016-02-26
Add DeferredDependencyLinksUpdater

mwjames:
Add an `QUERY_LINKS_TABLE` updater with a `DeferrableUpdate` signature to avoid complains from the `TransactionProfiler` which starts to flood messages in 1.26+ on updates that appear during a page view action.

### DIFF
--- a/includes/storage/SQLStore/SMW_SQLStore3.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3.php
@@ -372,6 +372,7 @@ class SMWSQLStore3 extends SMWStore {
 			$result = $this->fetchQueryResult( $query );
 		}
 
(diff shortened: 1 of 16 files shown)


mwjames:
This was changed in #1435 where `DeferredCallableUpdate` is used instead so that this class becomes a simple `DependencyLinksTableUpdater` without having to know anything about the transaction mode.

```php
	$dependencyLinksTableUpdater->addUpdateList( $sid, $dependencyList );

	$deferredCallableUpdate = ApplicationFactory::getInstance()->newDeferredCallableUpdate( function() use( $dependencyLinksTableUpdater ) {
		$dependencyLinksTableUpdater->doUpdate();
	} );

	$deferredCallableUpdate->enabledDeferredUpdate( true );
	$deferredCallableUpdate->pushToDeferredUpdateList();
```

PR #1425, opened 2016-02-29
Add some responsive css

mwjames:
https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/

In combination with a mobile friendly skin such as `Chameleon` (Vector should work to a certain degree as well), wpg/uri values adapt and are displayed within the border of the display.

i-Pad

image

image

Galaxy-S4

image

image

image

MobileFrontend (i-Phone 6)

image

### DIFF
--- a/res/smw/ext.smw.css
+++ b/res/smw/ext.smw.css
@@ -461,6 +461,15 @@ label.smw-form-checkbox {
 	font-weight: 500;
 }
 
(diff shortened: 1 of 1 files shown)


kghbln:
@mwjames Cool! I just installed MobileFrontend and Foreground on sandbox, Chameleon to follow after the upgrade of the server to Debian 8 and of MW to 1.27-alpha, hopefully the upcoming weekend.

Issue #1327, opened 2015-12-19
rebuildData.php / --ignore-exceptions


kghbln:
In some cases one may just get to many exceptions which interrupt the execution of the "rebuildData.php" script. I some cases this makes even script sitting a painful task. Thus it will be nice to have an option which allows to ignore these exceptions and let the script run till the end without the need for special attention.

Such a request was raised on the mailing list a couple of times and I personally like it.

mwjames:
> have an option which allows to ignore these exceptions and let the script run till the end without the need for special attention.

What happens with those exceptions when ignored? The exception are raised for a reason so someone has to work on them sooner or later.

kghbln:
> What happens with those exceptions when ignored?

If possible they should be logged or a not should be issued that there were worries. Indeed these exception should be dealt with.

After the 100+ exception an 60 minutes of sitting I gave up running the script for a wiki we all know. Will create issues for them soon.

mwjames:
> If possible they should be logged or a not should be issued that there were worries.

The questions how to log them and I'm not sure a simple

```
try {
    $byIdDataRebuildDispatcher->dispatchRebuildFor( $id );
} catch (Exception $e) {
    echo 'Karsten went berserk because of: ',  $e->getMessage(), "\n";
}
```

at [0] is sufficient.

> After the 100+ exception and 60 minutes of sitting I today gave up running the script for a wiki we all know. Will create issues for them soon

マジ

I thought for once I could work on some exciting stuff and not sitting through a bunch of issues :-(

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Maintenance/DataRebuilder.php#L247

kghbln:
> I thought for once I could work on some exciting stuff and not sitting through a bunch of issues :-(

I know exactly how you feel. Please do not ask why it took a week to actually move the wiki. :(

jamesmontalvo3:
I also would love a way to handle exceptions in a method other than blindly quitting `rebuildData.php`. When I run `rebuildData.php` on hundreds of thousands of IDs before going to bed, knowing it'll take hours, it's unfortunate to find in the morning that it stopped processing IDs in the first thousand. It would be great if it skipped the failures, and at the end said "231262 IDs attempted with 6 failures" and either printed the failures to the terminal or dumped them into a log.

mwjames:
Done with #1433.

mwjames:
[...]

PR #1433, opened 2016-03-04
rebuildData to output more verbose info, add `--ignore-exceptions`, r…

mwjames:
…efs #1327

Output more verbose information about what entity is being rebuild.

- Add `--ignore-exceptions` and `--redirects` option
- Option `--categories` is `-c`

image

ID's marked with * are in reference to the MW `page` table where the same ID does not necessarily represent the same entity (see 30).

image

### DIFF
--- a/includes/formatters/MessageFormatter.php
+++ b/includes/formatters/MessageFormatter.php
@@ -206,17 +206,18 @@ class MessageFormatter {
 		foreach ( $messages as $msg ) {
 
 			if ( $msg instanceof \Message ) {
-				$newArray[] = $msg->inLanguage( $this->language )->text();
+				$text = $msg->inLanguage( $this->language )->text();
+				$newArray[md5( $text )] = $text;
 			} elseif ( (array)$msg === $msg ) {
 				foreach ( $this->doFormat( $msg ) as $m ) {
-					$newArray[] = $m;
+					$newArray[md5( $m )] = $m;
 				}
 			} elseif ( (string)$msg === $msg ) {
-				$newArray[] = $msg;
+				$newArray[md5( $msg )] = $msg;
 			}
 		}
 
-		return array_unique( $newArray );
+		return $newArray;
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


mwjames:
@kghbln OK, I didn't plan on this but `--ignore-exceptions` was just added (in case of an exception, the file `exception-....log` [0] will contain that information).

Happy testing!

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1433/files#diff-05d44ea7d2b4980df07a3b7396a34eeeR142

jamesmontalvo3:
This change is phenomenal and I can't thank you enough for it. Looking forward to testing it out. If I'm reading it right it puts the log file in the `extensions` directory. Any reason that directory was chosen? Mind if I take a stab at changing the `--ignore-exceptions` option to allow specifying a log file? Would it be better to have yet another option like `--exception-log=/some/path` or maybe make just the one option?

kghbln:
This is utterly great. Thanks a ton! It does not relieve admins from tracking down the issues and reporting them but it helps actually getting the task of rebuilding data done as well.

Having the log file in the extensions directory is not ideal especially in farmed environments, so being able to specify a path will be preferable. I do not think an extra option is necessary, so I can imagine

``--ignore-exeptions`` would write to the standard path specified by the script and

``--ignore-exeptions=/some/path`` would allow to specify an alternate path for the log

Not logging exceptions does not make much sense I believe so this option appears not necessary to me.

mwjames:
> Having the log file in the extensions directory is not ideal especially in farmed environments, so being able to specify a path will be preferable. 

I'd go for the `--exception-log=/some/path` option, as the connotation of `--ignore-exeptions` is a either or decision (== boolean) which in case of a `true` could imply an additional information but I'd rather split them.

kghbln:
> I'd go for the --exception-log=/some/path option ...  but I'd rather split them.

You have a point here. That's fine with me too.

mwjames:
> I'd go for the --exception-log=/some/path option

... and so it is done `php maintenance/rebuildData.php -v --ignore-exceptions --exception-log="/D/xampp/tmp/logs/"
`

kimcollin:
Glad u liked it

On Friday, 4 March 2016, James Montalvo <notifications@github.com> wrote:

> This change is phenomenal and I can't thank you enough for it. Looking
> forward to testing it out. If I'm reading it right it puts the log file in
> the extensions directory. Any reason that directory was chosen? Mind if I
> take a stab at changing the --ignore-exceptions option to allow
> specifying a log file? Would it be better to have yet another option like
> --exception-log=/some/path or maybe make just the one option?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1433#issuecomment-192537146>
> .
>


-- 
Kimberly collins

PR #1435, opened 2016-03-05
Support for DeferrableUpdate

mwjames:
MW 1.26/1.27 started to operate most updates via `DeferrableUpdate`.

Use `DeferredCallableUpdate` to ensure that updates are queued and really executed after `WikiPage::doEditUpdates`.

`$GLOBALS['smwgEnabledDeferredUpdate'] = false;` the switch off and restore existing behaviour.


mwjames:
@kghbln `$GLOBALS['smwgEnabledDeferredUpdate'] = true;` is the default setting and it should not influence MW installations that are lower than MW 1.26 therefore the sandbox should be fine as well. In case some unexpected behaviour is being observed the setting should be reverted.

This should allow for a quicker front-end response when doing a page purge or when moving/redirecting pages.

mwjames:
https://www.semantic-mediawiki.org/wiki/Deferred_updates

PR #1436, opened 2016-03-05
Follow-up on #1409 (property usage count display)

mwjames:
Instead of text, it is now visually distinct from the rest of the text information.

image

refs #1409


mwjames:
@kghbln One can think about whether to change the colour or not if the count reaches a specific threshold. (e.g. 5000+ yellowish, 50.000+ reddish)

kghbln:
> One can think about whether to change the colour or not if the count reaches a specific threshold. (e.g. 5000+ yellowish, 50.000+ reddish)

Sorry, this took me a while. Indeed, though this should probably only come if an additional  explanatory text via a popup and such is provided. Otherwise people will not know how to interpret this.

mwjames:
> . Otherwise people will not know how to interpret this. 25.000+ for reddish.

Done with #1777.

PR #1438, opened 2016-03-06
Skip test for DataTime with ms on PHP 5.6.19

mwjames:


mwjames:
What a pain ...

mwjames:
Probably related to http://www.php.net/ChangeLog-5.php#5.6.19 / Fixed bug #68078 (Datetime comparisons ignore microseconds).

PR #1439, opened 2016-03-08
Strip tags from displayTitle, refs #1410

mwjames:
refs #1410


mwjames:
@kghbln  :+1:  I saw this on the sandbox.

DISPLAYTILE contained `<span style="position: absolute; clip: rect(1px 1px 1px 1px); clip: rect(1px, 1px, 1px, 1px);">FOO</span>'` and is now stripped.

mwjames:
@cicalese FYI

Issue #1424, opened 2016-02-29
Property namespace disappears (1.27)


mwjames:
https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1410 noted that "But, when I refresh the page, it disappears:", so I tried to replicate the issue on MediaWiki 1.27.0-alpha (c2f1c25) with the following steps:

refs #1295

### Create a property

![image](https://cloud.githubusercontent.com/assets/1245473/13380280/2a516c22-de3f-11e5-838e-cf9a7506b5f3.png)

which shows an empty page after the save.

![image](https://cloud.githubusercontent.com/assets/1245473/13380294/8895dd22-de3f-11e5-9152-8f53a2a2e72d.png)

### Special:RecentChanges

Switching to `Special:RecentChanges` one can see that the `Property` NS has disappeared.

![image](https://cloud.githubusercontent.com/assets/1245473/13380298/b43aaaac-de3f-11e5-9676-33e0dee0f72a.png)

Doing a normal refresh on `Special:RecentChanges` page causes the NS to reappear.

![image](https://cloud.githubusercontent.com/assets/1245473/13380309/d71da34e-de3f-11e5-905e-d7348b1b8d6c.png)

If we now move back to the `Property:Foo` page then it shows the expected content without any additional edit or refresh.

![image](https://cloud.githubusercontent.com/assets/1245473/13380310/eb1ae028-de3f-11e5-929b-3ac225897654.png)

mwjames:
@cicalese "This is probably not significant, but if I refresh the page again, "Modification date" changes from text to a link" @kghbln "Ouch, so this is worse than I had hoped.", exactly I don't know what changed in 1.27 or why this happens. Maybe some static doesn't get cleared or contains uninitialized settings, who knows but it is clear this only happens with 1.27.

If I add something to our `NamespaceManager` the issue seems to go away but I have no idea as to why 1.27 would need this. Given this issue and the failing integrations test, my confidence in the upcoming 1.27 release is becoming less and less favourable.

```
@@ -109,10 +109,12 @@ class NamespaceManager {
 		foreach ( $instance->buildNamespaceIndex( $globalVars['smwgNamespaceIndex'] ) as $ns => $index ) {
 			if ( !$instance->isDefinedConstant( $ns ) ) {
 				define( $ns, $index );
 			};
 		}
+
[...]

Issue #1406, opened 2016-02-16
ConfirmEdit blocks SMW


stefahn:
I have a fresh install of both MediaWiki and SMW. SMW was installed by composer without any issues. 

After adding the first property I wanted to check Special:Browse, but the Special page wouldn't load. I also saw that SMW namespaces didn't exist.
I disabled the other extensions and found that ConfirmEdit is causing the trouble. If I disable ConfirmEdit, SMW runs fine (namespaces exist and Special pages too).

I tested this both with ConfirmEdit 1.4.0 (bundled with MW 1.26.2) and the master version of ConfirmEdit. Both seem to block SMW.

Versions:
Semantic MediaWiki	2.3.1
MediaWiki 	1.26.2
PHP 	5.4.45-0+deb7u2 (cgi-fcgi)
MySQL 	5.5.47-0+deb7u1
ICU 	4.8.1.1
More version info see http://awesomescreenshot.com/0515mx7i91

JeroenDeDauw:
Thanks for reporting the issue!

mwjames:
I do think this issue has been discussed before on the mailing list back in 2014 [0].

[0] http://wikimedia.7.x6.nabble.com/SMW-Namespaces-seem-to-be-blocked-td5032732.html

stefahn:
@mwjames : Good find. Funny that back then it was me too, who encountered that problem. But this time it's a brand new install. 
Hope someone can reproduce the issue.

mwjames:
Unless someone finds the reasons as to what `ConfirmEdit` does so that both cannot be used together, it is unlikely that SMW core can provide any assistance in solving this issue.

Feel free to re-open this issue in case some actionable tasks can be assigned to it.

kghbln:
@stefahn FYI https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1686

PR #1440, opened 2016-03-08
Changes to accommodate 1.27

mwjames:
- Various language changes because trying to get early access
(during `Setup`) to `Language::getCode` violates MW's setup policy
and is marked as such during the not completed `wgFullyInitialised` state.
- Ensure that `DeferredUpdates` are finished before validating possible
assertions (#1435)
- Ensures `wgExtraNamespaces` is set to at least the canonical form
- Ensures that a `ExpResource` is using the canonical form

refs #1424, #1295

### DIFF
--- a/includes/NamespaceManager.php
+++ b/includes/NamespaceManager.php
@@ -44,6 +44,19 @@ class NamespaceManager {
 	}
 
 	/**
+	 * @since 2.4
+	 *
+	 * @param string $languageCode
+	 *
+	 * @return array
+	 */
+	public static function getNamespacesByLanguageCode( $languageCode ) {
+		$instance = new self( $GLOBALS, $GLOBALS['smwgIP']  );
+		$instance->initContentLanguage( $languageCode );
+		return $GLOBALS['smwgContLang']->getNamespaces();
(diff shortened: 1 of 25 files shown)


mwjames:
> Is there something in particular to test?

@kghbln To answer the question:

- Property NS (its canonical form) should be set no matter what and is expected to solve #1424 and most likely #1406
- Switching a user language should work now like any other content in MW
- If a predefined property (goes for core and extensions) doesn't have a localized label, the canonical (== English) label is supposed to be used
- RDF will always use the canonical form (to my surprise encountered that my SPARQL RDF/Turtle would use the localized property labels and queries stopped working after I switched the user language ... what a drag ...)

PS: This would have all been easier if one could use `Language::getCode` early on but someone decide that for 1.27 (it worked before) you can't do that which brought all kinds of ... Anyway (for the predefined properties at least) switching languages is as easy as pressing the ULS button and labels are expected to adjust.

Namespace is a matter of the site language (not the user language). I was unable to find a way to extend `wgNamespaceAliases` with NS aliases at the time the user language changes.

image

image

Issue #817, opened 2015-02-14
Concept dataitem export in RDF/SMWExporter


mwjames:
refs https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/814#issuecomment-74350427

JeroenDeDauw:
I do not understand what this issue is about, or how exactly we can close it

mwjames:
> I do not understand what this issue is about, or how exactly we can close it

Just needs to be added, and this issue is for tracking the [0] TODO.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/export/SMW_Exporter.php#L696

mwjames:
Fixed with #1450.

PR #1450, opened 2016-03-12
Export concepts, refs #814, #817

mwjames:
refs #814, #817


mwjames:
Integration test queries on Concepts are skipped for `Virtuoso` because of  "Virtuoso reports 'SPARQL compiler: Blank node ... is not allowed in a constant clause" which is caused by https://github.com/openlink/virtuoso-opensource/issues/126.

PR #1451, opened 2016-03-12
Add ExtraneousLanguage

mwjames:
Preparation for converting `php` lang files into json files.

If an extension has registered its predefined properties with a msgKey then we are able to interpret things like `[[A le type@fr::+]]` as `[[Has type::+]]` at the time a user switches languages.


mwjames:
@kghbln Next part of the "SMW goes Multilingual".

mwjames:
This PR only deals with pre-defined properties. Support for multilingual alternative user-defined labels is a whole other beast to tame.

Issue #1076, opened 2015-06-26
"Embedded" format also transcludes properties and subobjects


D-Groenewegen:
http://www.semantic-mediawiki.org/wiki/Help:Embedded_format claims that "SMW will take care that embedded articles do not import their semantic annotations, so these need not be treated specifically"

Yet this is precisely what's happening on http://www.vanhamel.nl/codecs/Rennes,_Biblioth%C3%A8que_de_Rennes_M%C3%A9tropole,_MS_598/ff._90-125
The "Embedded" format is used here to embed three tables of contents, whose properties (e.g. Semantic Form's "Has default form") and subobjects are inherited by the page. 

For the factbox, see http://www.vanhamel.nl/codecs/Special:Browse/Rennes,_Biblioth%C3%A8que_de_Rennes_M%C3%A9tropole,_MS_598-2Fff._90-2D125

mwjames:
#1068 solves the issue by excluding "embedded" from the recursive annotation support. `parser-08-01-embedded-ask.json` tests that no annotations are imported for the default setting.

D-Groenewegen:
Thanks for this as well as for that issue with the Pagewidget format.

D-Groenewegen:
Just a question. I'd like to test
$smwgEnabledResultFormatsWithRecursiveAnnotationSupport but I'm not sure
how to use it in order for the "embedded" or the "pagewidget" result format
to exclude any semantic annotations from the source page? In SMW 2.4alpha,
import of annotations seems to be true by default (at least that's what
still happens on my installation), so how does the 'switch' work?

Best,
Dennis




2015-07-07 8:42 GMT+02:00 mwjames <notifications@github.com>:

mwjames:
> n SMW 2.4alpha, import of annotations seems to be true by default

No, it should not (I think we have some tests that explicit test this).

`smwgEnabledResultFormatsWithRecursiveAnnotationSupport` was actually removed with 2.3.1 as it was creating more trouble (#1255) then anticipated and instead of a global setting, importing has being made available only on a per query basis when the `import-annotation` parameter is set `true` (which by default is set false).

#1257 restored the previous behaviour.

mwjames:
Feel free to create an example at [0] (which hosts the most recent development incl. SMW) to demonstrate your observation.

[0] http://sandbox.semantic-mediawiki.org

D-Groenewegen:
[...]

PR #1455, opened 2016-03-13
Allow to block recursive annotation during an internal parse, refs 1076

mwjames:
refs #1076

Use case: The `embedded` format no longer adds `#set`/`#subobject` annotations that are imported via an embedded page using `#ask`.

Categories remain special [0] as there are parsed and handled by MW.

[0] https://www.semantic-mediawiki.org/wiki/Help:Embedded_format#Limitations

### DIFF
--- a/includes/ParserData.php
+++ b/includes/ParserData.php
@@ -98,6 +98,24 @@ class ParserData {
 	}
 
 	/**
+	 * @since 2.4
+	 *
+	 * @return boolean
+	 */
+	public function canModifySemanticData() {
+
+		// getExtensionData returns null if no value was set for this key
+		if (
+			$this->hasExtensionData() &&
+			$this->parserOutput->getExtensionData( 'smw-blockannotation' ) !== null &&
+			$this->parserOutput->getExtensionData( 'smw-blockannotation' ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
 	 * @since 2.1
 	 *
 	 * @return boolean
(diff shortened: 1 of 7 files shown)


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Lorem_ipsum_3

mwjames:
@kghbln @D-Groenewegen This should fix the issue where `#set`/`#subobject` annotations are copied/imported when `embedded` is used.

D-Groenewegen:
Not sure why the 'facts about' box does not appear, but our example at http://sandbox.semantic-mediawiki.org/wiki/Lorem_ipsum_3 now demonstrates that contents from a different page can be embedded without importing its annotations. 

Thanks for this @mwjames!

mwjames:
>Not sure why the 'facts about' box does not appear, but our example at http://sandbox.semantic-mediawiki.org/wiki/Lorem_ipsum_3

Because your example page only has non-visible property values where the Factbox will not appear.

[0] shows that we have to lift the block after the recursive parse has finished (which was initiated by the result printer) to ensure that `[[ ... :: ...]]` (`#set` are isolated and run before the `InTextAnnotationParser`) can be processed afterwards.

```
@@ -346,10 +346,15 @@ abstract class ResultPrinter extends \ContextSource implements QueryResultPrinte
 			}
 
 			self::$mRecursionDepth--;
 		}
 
+		$wgParser->getOutput()->setExtensionData(
+			'smw-blockannotation',
+			false
+		);
```

[0] http://sandbox.semantic-mediawiki.org/wiki/Issue/1455_%28with_Factbox%29

kghbln:
Finally [documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AEmbedded_format&type=revision&diff=48924&oldid=48919).

PR #1460, opened 2016-03-17
composer.json: Require PHP extension mbstring

s7eph4n:
RPM based systems (RedHat and friends) do not have mbstring installed together with PHP out of the box [1].
This change will make SMW depend on the mbstring extension and fail installing if the extension is not available [2].

[1] https://phabricator.wikimedia.org/T129435
[2] https://getcomposer.org/doc/02-libraries.md#platform-packages

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,7 @@
 	},
 	"require": {
 		"php": ">=5.3.0",
+		"ext-mbstring": "*",
 		"composer/installers": "1.*,>=1.0.1",
 		"mediawiki/parser-hooks": "~1.4",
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
I just noticed that I also did not have this installed on my PHP7. Guess we can't easily ditch usage of this extension tho

mwjames:
> Guess we can't easily ditch usage of this extension tho

`mb_string` has been a requirement since (who knows ...) [0] and MW 1.28+ now requires it as well.

[0] https://www.semantic-mediawiki.org/wiki/Help:Installation#Environment

PR #1462, opened 2016-03-19
Add SubobjectParserFunction::doPrepareParameters

mwjames:


mwjames:
This is needed to make sure we can produce stable ID's (and is added before so we know/test that current generated ID's remain the same) when working with lang. annotated properties like `Foo@en`.

Issue #1402, opened 2016-02-04
How to delete orphaned properties?


WolfgangFahl:
see http://stackoverflow.com/questions/28253718/how-to-delete-orphaned-properties

At Special:Properties redlinks show up for properties that are referenced by deleted pages. Basically the question is the same as in: http://sourceforge.net/p/semediawiki/mailman/message/31901925/

How do i  get rid of such useless properties that make no sense since they are only referenced in deleted pages ...?

mwjames:
In 2.4, running `rebuildData.php` will check with the `PropertyTableOutdatedReferenceDisposer` (#1216) whether an entity is without reference (including user-defined properties) and mark them as disposable hence should no longer appear in queries or lists. If the `EmbeddedQueryDependencyLinksStore` is enabled then properties remain available (as we need to keep the ID reference to those components) when referenced in a query condition or as printout parameter (#1323) .

> How do i get rid of such useless properties that make no sense since they are only referenced in deleted pages ...?

Not sure what this means.

mwjames:
In case the `rebuildData.php` is not able to remove outdated properties [0] then [1] might come to rescue.

[0] https://www.semantic-mediawiki.org/wiki/Help:Outdated_properties
[1] https://www.semantic-mediawiki.org/wiki/Help:Object_ID_disposal

PR #1474, opened 2016-03-25
Add search link for zero properties

mwjames:
#1216 disposes outdated entities that have no remaining references but it is not always easy to find some "hidden" references that prevents the removal when the property has no obvious value annotation.

`PropertyTableIdReferenceFinder` tries to scan related tables to find at least "one" reference that prevents the `PropertyTableIdReferenceDisposer` from exercising the removal.

`Special:Properties` is extended to actual display a link to ease the search.


mwjames:
image

image

image

mwjames:
Somewhat tangential to the  "How to delete orphaned properties?" (#1402) question because when running `rebuildData.php`, the script will try to identify outdated properties (== with zero references) and remove them (this only goes for user-defined properties, pre-defined properties are excluded) but sometimes the "zero" reference can be deceiving which is also why those property remain in the list.

PR #1477, opened 2016-03-25
Filter "swivt:redirectsTo" in CompoundConditionBuilder

mwjames:
- Remove entities that contain a "swivt:redirectsTo" predicate to avoid results sets contain duplicate entries
- `TitleLookup` to actual return the redirect source not the targets (only relevent for `--redirects` option of `rebuildData.php`)


mwjames:
`p-0904.json` contains the use case that would return different result sets for `SQLStore` / `SPARQLStore` if the PR is not applied.

PR #1483, opened 2016-03-30
Add statistics on outdated entities, refs #1100

mwjames:
Outdated entities are not removed immediately from the `ID_TABLE` and in instead "marked for removal" using the `smw:delete` indicator.

An administrator or curator (SMW) should ensure that this number is kept low by running the `rebuildData.php` script on a regular basis.

It is also an indicator (besides the error count) whether an installation is maintained routinately or not.

refs #1100


mwjames:
https://www.semantic-mediawiki.org/wiki/Help:Outdated_entities

mwjames:
Special:Statistics

image

`api.php?action=smwinfo&info=proppagecount|deletecount`

image

PR #1490, opened 2016-04-02
Removed some usage of a class alias

JeroenDeDauw:


JeroenDeDauw:
Not done this pretty (using appropriate imports), just a search and replace using FQN

PR #1498, opened 2016-04-07
Fix some test failures (appeared with PHPUnit 5.3+)

mwjames:


mwjames:
Still something with PHPUnit 5.3+ is off, see https://github.com/sebastianbergmann/phpunit/issues/2132.

tests/phpunit/Unit/Factbox/CachedFactboxTest.php
```diff
@@ -475,6 +483,14 @@ public function outputDataProvider() {
 			->method( 'getContext' )
 			->will( $this->returnValue( new \RequestContext() ) );
 
+		$store = $this->getMockBuilder( '\SMW\Store' )
```
JeroenDeDauw: I think it's real icky to reuse variables like this, or have such a big scope

JeroenDeDauw:
Yeah right, you can no longer reuse the same instance of a mock. This is good :)

tests/phpunit/Unit/Factbox/CachedFactboxTest.php
mwjames: Right, but this PR just aims at getting it run with 5.3 (otherwise it would just stop in the middle of a test drive). Those issues should be addressed in follow-up, someday.

I'm not claiming that we do much better but I hope we are not as close as [0].

[0] https://gerrit.wikimedia.org/r/#/c/195297/113/tests/phpunit/includes/auth/AuthManagerTest.php

JeroenDeDauw: > [0] https://gerrit.wikimedia.org/r/#/c/195297/113/tests/phpunit/includes/auth/AuthManagerTest.php

siriasly

PR #1504, opened 2016-04-09
Removed dead local vars

JeroenDeDauw:


src/Exporter/DataItemByExpElementMatchFinder.php
```diff
@@ -109,7 +109,6 @@ private function tryToMatchDataItemForWikiNamespaceUri( $uri ) {
 	private function tryToMatchNamespaceName( $name ) {
 		// try the by far most common cases directly before using Title
 		$namespaceName = str_replace( '_', ' ', $name );
-		$namespaceId = -1;
```
mwjames: This one is suspicious, I have to have a close look at this one.

JeroenDeDauw: The next line reassigns the value. So given that the initial assign did not cause side effects, it is dead code. IDE agrees

PR #1510, opened 2016-04-09
Grammar correction in smw-pa-property-predefined-long_pvuc

amire80:
The sentence talks about values, so I suppose that it should be "their" and not "its". If I am misunderstanding something, please close this pull request and explain it [in the message's qqq](https://translatewiki.net/wiki/MediaWiki:Smw-pa-property-predefined-long_pvuc/qqq).


kghbln:
Thanks for spotting and amending this!

Issue #1378, opened 2016-01-18
PHP7 test failures


mwjames:
Already caught and fixed 16ddfd8.

```
Semantic MediaWiki: 2.4-alpha (SMWSQLStore3, mysql)
MediaWiki:          1.26.2 (MediaWiki vendor autoloader)

Execution date:     2016-01-18 00:00

PHPUnit 4.8.21 by Sebastian Bergmann and contributors.

Runtime:        PHP 7.0.2
```

remaining issues are:

```
There were 2 failures:

1) SMW\Tests\MediaWiki\Api\ApiQueryResultFormatterTest::testSetIndexedTagNameException
Failed asserting that exception of type "ReflectionException" matches expected exception "InvalidArgumentException". Message
was: "Invocation of method SMW\MediaWiki\Api\ApiQueryResultFormatter::setIndexedTagName() failed" at
#0 ...\extensions\SemanticMediaWiki\tests\phpunit\Unit\MediaWiki\Api\ApiQueryResultFormatterTest.php(49)
: ReflectionMethod->invoke(Object(SMW\MediaWiki\Api\ApiQueryResultFormatter), Array, NULL)
#1 [internal function]: SMW\Tests\MediaWiki\Api\ApiQueryResultFormatterTest->testSetIndexedTagNameException()
#2 ...\vendor\phpunit\phpunit\src\Framework\TestCase.php(908): ReflectionMethod->invokeArgs(Object(SMW\T
ests\MediaWiki\Api\ApiQueryResultFormatterTest), Array)
#3 ...\vendor\phpunit\phpunit\src\Framework\TestCase.php(768): PHPUnit_Framework_TestCase->runTest()
#4 ...\vendor\phpunit\phpunit\src\Framework\TestResult.php(612): PHPUnit_Framework_TestCase->runBare()
#5 ...\vendor\phpunit\phpunit\src\Framework\TestCase.php(724): PHPUnit_Framework_TestResult->run(Object(
SMW\Tests\MediaWiki\Api\ApiQueryResultFormatterTest))
#6 ...\vendor\phpunit\phpunit\src\Framework\TestSuite.php(747): PHPUnit_Framework_TestCase->run(Object(P
HPUnit_Framework_TestResult))
#7 ...\vendor\phpunit\phpunit\src\Framework\TestSuite.php(747): PHPUnit_Framework_TestSuite->run(Object(
PHPUnit_Framework_TestResult))
[...]

PR #1512, opened 2016-04-13
Fix test failure on PHP7, refs #1378

mwjames:
refs #1378

### DIFF
--- a/src/MediaWiki/Api/ApiQueryResultFormatter.php
+++ b/src/MediaWiki/Api/ApiQueryResultFormatter.php
@@ -195,7 +195,7 @@ class ApiQueryResultFormatter {
 	 * @param array &$arr
 	 * @param string $tag
 	 */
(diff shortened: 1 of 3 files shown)


mwjames:
The code coverage run didn't finish lately because of the 50 min Travis time limit where PHP 5.6 was unable to cope with the requirements of Xdebug + PHPUnit and our test suite but now that PHP7 runs without failure, the switch to PHP7 allows it to finish within ~25 min.

src/MediaWiki/Api/ApiQueryResultFormatter.php
```diff
@@ -195,7 +195,7 @@ protected function formatErrors( array $errors ) {
 	 * @param array &$arr
 	 * @param string $tag
```
JeroenDeDauw: string|null

PR #1515, opened 2016-04-13
Fix zone offset in XsdValueMapper

### LINKED ISSUE #1503: Weird XsdValueMapper assignment
Probably a bug

probbug

mwjames:
refs #1503


JeroenDeDauw:
Solves the mentioned issue, though idk if the behaviour change is correct

mwjames:
> Solves the mentioned issue, though idk if the behaviour change is correct (which does not mean anything really)

I read through [0] and having the year (2002Z) or (2002-01Z) marked with a time zone is not required nor specified therefore only something like 2002-01-12Z (date + time) gets a zone offset.

[0] https://www.w3.org/TR/2005/NOTE-timezone-20051013/

PR #1517, opened 2016-04-13
Add `--with-property-statistics` to rebuildData

mwjames:
https://github.com/onoi/message-reporter/commit/939bc1b78dec647d6ee2fa97c22d7068548f68dd

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,7 +68,8 @@
 	},
 	"autoload": {
 		"psr-4": {
-			"SMW\\": "src/"
+			"SMW\\": "src/",
+			"SMW\\Maintenance\\": "maintenance/"
 		},
 		"files" : [
 			"SemanticMediaWiki.php"
--- a/maintenance/rebuildConceptCache.php
+++ b/maintenance/rebuildConceptCache.php
@@ -2,8 +2,6 @@
 
 namespace SMW\Maintenance;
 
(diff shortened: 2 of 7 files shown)


mwjames:
@kghbln I reckoned that most users will forget to run `rebuildPropertyStatistics.php` in the process, so this just adds the option `--with-property-statistics` to run it immediately after.

mwjames:
I'd like to merge this today, any opposition in regards to the option name?

kghbln:
Well this is a convenience option. I personally have no objection against the name. I would only object if this was enabled by default which it is not. So ... :)

I myself set up a daily cron for every wiki to get "rebuildPropertyStatistics.php" rolling which I still see as the recommended way.

mwjames:
> I myself set up a daily cron for every wiki to get "rebuildPropertyStatistics.php" rolling which I still see as the recommended way.

Yes.

> Well this is a convenience option.

I think when you would ask users on how many of them run `rebuildData.php` let alone `rebuildPropertyStatistics.php`then my guess is there aren't much and this option should lower the barrier for the lazy ones myself included.

`--with-property-statistics` does not add any new code path it just reuses what is already available in the `MaintenanceFactory`.

kghbln:
> my guess is there aren't much and this option should lower the barrier for the lazy ones myself included.

Fair enough. I also believe that this gets pretty close to reality.

PR #1521, opened 2016-04-21
Mark duplicate properties, refs #1100, #1216

mwjames:
refs #1100, #1216, #191, #1474

### DIFF
--- a/src/SQLStore/ByIdDataRebuildDispatcher.php
+++ b/src/SQLStore/ByIdDataRebuildDispatcher.php
@@ -234,7 +234,7 @@ class ByIdDataRebuildDispatcher {
 
 		$res = $db->select(
 			\SMWSql3SmwIds::TABLE_NAME,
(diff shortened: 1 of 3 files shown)


mwjames:
Hopefully this will help to clean up the `smw_object_ids` table [0] when running `rebuildData`.

image

image

image

@D-Groenewegen @kghbln I don't know how/why those properties are duplicate but this PR should find them and add a `SMW_SQL3_SMWDELETEIW ` marker so they can be removed on the succeeding `rebuildData` run. Since I cannot identify the cause, I'm unable to tests this thoroughly therefore the field test must do.

 `SMW_SQL3_SMWDELETEIW ` marker as in something like when looking at the `smw_object_ids` table.

image

[0] http://www.vanhamel.nl/vhcodecs/index.php?title=Special:Properties&limit=500&offset=0

kghbln:
@thingles FYI

PR #1523, opened 2016-04-21
Set a language code for the DV formatter

mwjames:
This allows to set a language code based on context (storage vs. display).

refs #1388


mwjames:
image

image

PR #1525, opened 2016-04-21
Rename property-page-indicator to page-indicator

mwjames:


mwjames:
Changed to allow `smw-page-indicator` to be re-used.

image


```
{{#tag:indicator
 |<span class="smw-page-indicator" style="width:95px;" title="Foo">1.0.0 - 1.7.0</span>
 |name="foo"
}} 
```

PR #1526, opened 2016-04-21
Fix `_` in Special:SearchByProperty

mwjames:


mwjames:
Based on [0] where david (aka  Dvdgmz) wrote:

> Yes, the problem it is the same in the Sandbox wiki:
> http://sandbox.semantic-mediawiki.org/wiki/Test-keywordQ
> And if the value on {{{1}}} has white spaces its are passed as an
> underscore _ and the query don't work properly.

[0] http://wikimedia.7.x6.nabble.com/Problem-with-SearchByProperty-link-and-white-spaces-td5062711.html

PR #1527, opened 2016-04-21
Concept cache display, refs #1409

mwjames:


mwjames:
Somehow the top indicator (#1409) didn't "shine" through instead the information is now put in the centre.

image

image

PR #1530, opened 2016-04-28
Add `data-row-number` to table output

mwjames:


mwjames:
`data-row-number` is a hidden attribute that can be accessed by JS components but more importantly the number can be used to match sorting positions during tests that assert string/output equality.

PR #1533, opened 2016-04-28
Improve `Special:Browse` output handling, refs #1523

mwjames:
refs #1523, #1545

### DIFF
--- a/includes/parserhooks/AskParserFunction.php
+++ b/includes/parserhooks/AskParserFunction.php
@@ -116,10 +116,12 @@ class AskParserFunction {
 
 		$this->parserData->pushSemanticDataToParserOutput();
 
-		// Allow to reparse in case the user language changed
-		// 1.23+
+		// 1.23+ add options so changes are recognized in case of:
+		// - 'userlang' will trigger a cache fragmentation by user language
+		// - 'dateformat'  will trigger a cache fragmentation by date preference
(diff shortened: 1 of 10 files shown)


mwjames:
@kghbln FYI

```
 * By rule means that
 *
 * 1. If the page content language is different from the content language then
 * use the page content language (as it is clear that the page content was
 * intended to be in a specific language)
 * 2. If the page content language and content language are indifferent then
 * use the user language
 * 3. If the page content language and user language could not be determined then
 * use the content language
 *
 * General rules:
 * - Special pages are in the user language
 * - Display of values (DV) should follow rules outlined above
 * - Storage of values (DI) should always use the content language
 *
 * Notes:
 * - The page content language is the language in which the content of a page is
 * written in wikitext
```

The only known extension that actively changes the `page content language` is SIL where content is to be represented in a denoted language and therefore the formatting of content is expected to reflect that intention by using the `page content language` instead of the `user language` if `#MEDIAWIKI` was selected as output formatter.

refs http://wikimedia.7.x6.nabble.com/Use-of-MEDIAWIKI-to-format-dates-in-query-results-td5062876.html

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1533

mwjames:
Special:Browse will use `MEDIAWIKI` as output formatting option to adjust the display if the users chooses a different language.

image

image

image

image

PR #1534, opened 2016-04-28
`Display title of` to augment sortkey, refs #1410

mwjames:
If the `Display title of` property is annotated then also augment the sortkey unless given.

refs #1410

### DIFF
--- a/includes/SemanticData.php
+++ b/includes/SemanticData.php
@@ -199,6 +199,17 @@ class SemanticData {
 	}
 
 	/**
+	 * @since 2.4
+	 *
+	 * @param DIProperty $property
+	 *
+	 * @return boolean
+	 */
+	public function hasProperty( DIProperty $property ) {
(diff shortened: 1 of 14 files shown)


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1534

mwjames:
Not directly related but nevertheless is a more practical example (taken from [0]) where an external ID 2706490 is used as subject entity that denotes `{{DISPLAYTITLE:Agrostis capillaris L.}}` [1] to identify itself as `Agrostis capillaris L.` subject.

Searches will most likely involve using `[[~Agrostis*]]` unless someone knows the external ID to match `[[GBIFID:2706490]]` directly as subject entity.

[0] http://www.gbif.org/species/2706490
[1] http://sandbox.semantic-mediawiki.org/wiki/GBIFID:2706490

mwjames:
@kghbln @cicalese FYI

PR #1542, opened 2016-04-29
Simpler regex for condition that contains `=`, refs #710, #640

mwjames:
This should allow for object values `[[This has a = in it]]` to be querable as well.

refs #710, #640

### DIFF
--- a/includes/query/SMW_QueryProcessor.php
+++ b/includes/query/SMW_QueryProcessor.php
@@ -296,7 +296,7 @@ class SMWQueryProcessor {
 			// Bug 32955 / #640
 			// Modify (e.g. replace `=`) a condition string only if enclosed by [[ ... ]]
 			$rawParam = preg_replace_callback(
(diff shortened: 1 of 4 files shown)


mwjames:
In response to [0].

@kghbln FYI

I would expect people interested in having this issue fixed to test this PR more rigorously as the change itself is small but the consequence of an oversight would be rather disastrous.

[0] http://wikimedia.7.x6.nabble.com/Query-problems-due-to-equal-signs-in-page-name-td5062965.html

mwjames:
@SemanticMediaWiki/testers Before merging I'd like to seek confirmation that we actually improved the process and have not introduced some bad behaviour (e.g with/without templates, other parser functions to set the condition etc.)

mwjames:
This PR has not received any feedback (given the moratorium of two weeks) and I'm going to assume that prolonging this will not change this fact therefore merging it to avoid code rot.

Issue #1326, opened 2015-12-19
Unable to page through property values when many subobjects


thingles:
Wasn't sure how to title this. You can see this issue at this URL

https://wikiapiary.com/w/index.php?title=Property:Has_remote_error&from=BruceTambling.com#SMWResults

The "from" value works, but this query returns more than 25 results. When you do "(next 25)" since it starts at the top you just get the same page back and you can’t page any further.

mwjames:
> The "from" value works, but this query returns more than 25 results. When you do "(next 25)" since it starts at the top you just get the same page back and you can’t page any further.

I have seen in the past, it is an old issue (pre 1.8) and the problem sparks from how the page name is set and used for the next iteration.

For example, the first 25 are starting with "BruceTambling.com" so the next starts with "BruceTambling.com" as well which makes it a loop in loop and the search request goes in circles for "from=BruceTambling.com#SMWResults".

MW's webrequest object cuts the # subobject marker from the url request so the name will always be "BruceTambling.com" and not "BruceTambling.com#Extension_MediaPreview".

thingles:
Yeah, figured that was the issue. Would there be a hack-y workaround to use some special character to pass the "#' into the URL and keep MediaWiki from slicing it off. Maybe even just URL encode the #? Or send something silly like "BruceTambling.com^Extension_MediaPreview" and mangle it on either end?

mwjames:
> pass the "#' into the URL and keep MediaWiki from slicing it off. Maybe even just URL encode the #? Or send something silly like "BruceTambling.com^Extension_MediaPreview" and mangle it on either end?

OK, this is some old code but what if you try to replace part in `SMWPageLister` with 

```
-				$last = \SMW\StoreFactory::getStore()->getWikiPageSortKey( $this->mDiWikiPages[$resultCount - 1] );
+				$last = $this->mDiWikiPages[$resultCount - 1];
```

The `getWikiPageSortKey` returns a text without a subobject and quick test shows we encode `#` which should allow for the webrequest to fetch the whole string.

thingles:
Well, that is better but it still isn't right. I put that change in my Semantic MediaWiki code locally so you can see it if you go to the URL and page around.

It does indeed allow the "(next 25)" links to work and you can page through it. However, it skips now. I guess I would take it skipping some over just refusing to page forward.

[...]

PR #1543, opened 2016-04-29
Improve navigation on property/concept page, refs #1326

mwjames:
- Concept columns made responsive
- Solved 1326 for both the concept and propery page
- `property-page-results` class can be used to adjust the display

refs #1326


mwjames:
image

image

PR #1545, opened 2016-05-01
Add #LOCL as TimeValue output format

mwjames:
#MEDIAWIKI has some restrictions (seconds are no displayed and is supposed to be influenced by the user selected language).

#LOCL support indiviual preferences to match SMW internal time precision while text elements are localized

- `SMW_DV_TIMEV_CM` as feature to hint the calendar model
- `TimeValueFormatter::getCaptionFromDataItem` will use the content language
which is either the default content language or if available the page content language
- `ExtraneousLanguage::getPreferredDateFormatByPrecision` determines
a PHP DateTime format string as preference for a specific language in
combination with a precision
- Each language can maintain a preference for a precision (as default)
 - 'SMW_PREC_Y'    => 'Y',
 - 'SMW_PREC_YM'   => 'F Y',
 - 'SMW_PREC_YMD'  => 'F j, Y',
 - 'SMW_PREC_YMDT' => 'H:i:s, j F Y'
- Output format `#LOCL` will use the `IntlTimeFormatter::getLocalizedFormat` and
 `ExtraneousLanguage::getPreferredDateFormatByPrecision` to dermine a
localized date/time output

### DIFF
--- a/SemanticMediaWiki.settings.php
+++ b/SemanticMediaWiki.settings.php
@@ -960,6 +960,9 @@ $GLOBALS['smwgEnabledInTextAnnotationParserStrictMode'] = true;
 # uniqueness for a value is established by the fact that it is assigned before
 # any other value of the same representation to a property).
 #
+# - SMW_DV_TIMEV_CM (TimeValue) to indicate the CalendarModel if is not a
+# CM_GREGORIAN
+#
 # @since 2.4
 ##
-$GLOBALS['smwgDVFeatures'] = SMW_DV_PROV_REDI | SMW_DV_MLTV_LCODE | SMW_DV_PVAP | SMW_DV_WPV_DTITLE;
+$GLOBALS['smwgDVFeatures'] = SMW_DV_PROV_REDI | SMW_DV_MLTV_LCODE | SMW_DV_PVAP | SMW_DV_WPV_DTITLE | SMW_DV_TIMEV_CM;
--- a/includes/Defines.php
+++ b/includes/Defines.php
@@ -110,6 +110,15 @@ define( 'SMW_DAY_YEAR', 3 ); // an entered digit can be either a month or a year
 /**@}*/
 
 /**@{
+ * Constants for date/time precision
+ */
+define( 'SMW_PREC_Y', 0 );
+define( 'SMW_PREC_YM', 1 );
+define( 'SMW_PREC_YMD', 2 );
+define( 'SMW_PREC_YMDT', 3 );
+/**@}*/
+
+/**@{
  * Constants for SPARQL supported features (mostly SPARQL 1.1) because we are unable
  * to verify against the REST API whether a feature is supported or not
  */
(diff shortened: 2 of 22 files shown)


mwjames:
@kghbln FYI, I  only added the preference for a precision on en, es, and ja.

kghbln:
Wow, did not see this. Now we also get de.

kghbln:
@mwjames What is the sytax for this? `|?Date#LOCL[Y]=date`

kghbln:
Ah, this is a "come as you are" formatter, right? So if `[[Has date:2016/08/17]]` one gets 17. August 2016  or with `[[Has date::Aug 2016]]` one gets August 2016 etc. when doing `|?Has date#LOCL=date`. It is just depending what data value arrives at the formatter via the query.

mwjames:
> What is the syntax for this? |?Date#LOCL[Y]=date

`#LOCL` acts upon on the codified formats (as above) and switches based on a user language (LOCL=== localized) [0].

For example, `#LOCL` is also set on `Special:Browse` and `Property:...` page as standard formatter to ensure that when a user switches its language, it shows as localized output. This is why numbers , boolean and date output changes with a user language (i.e. this is not changing the internal representation which depends on the content language).

`#MEDIAWIKI` is limited (I think this is also noted on the help page) hence we use a formatter we control and effectively can influence.

If you want a specific format then `#F[...]` is your friend.

[0] http://sandbox.semantic-mediawiki.org/wiki/Issue/1533

I remember now, for the Number/Boolean type `#LOCL` is extended to do things like `#LOCL@en` allowing it to be selective about a preferred language output as seen in [1, 2]. This specific feature is not implemented for the Date type since I was too busy to get this overall concept implemented and never bothered afterwards.

Hopefully, [2] can be of assistance. 

[1] http://sandbox.semantic-mediawiki.org/wiki/Issue/1591
[2] http://sandbox.semantic-mediawiki.org/wiki/Issue/1545_%28LOCL%29

mwjames:
> This specific feature is not supported for the Date type

#1793 adds `#LOCL@...` to work with dates as well.

gesinn-it:
Is there a way, to take the user's timezone into account? Like dates being stored in UTC but rendered as CET with LOCL@de formatting?

kghbln:
Isn't this a matter of configuring [timezones?](https://www.mediawiki.org/wiki/Manual:Timezone)

gesinn-it:
Yes, but if you configure the timezone for your user, SMW does not use this timezone for rendering DateTimes.

mwjames:
> SMW does not use this timezone for rendering DateTimes.

A while ago I added a TZ support in #1838 but as the comment [0] points out, I'm reluctant to support user preferences for dates in smw-core.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/b8df038174bfb3309efca38a2bba724b68fbf3a6/src/IntlTimeFormatter.php#L196-L204

kghbln:
> A while ago I added a TZ support in #1838 but as the comment [0] points out.

Thanks for clarifying!

PR #1559, opened 2016-05-07
Recognize QuerySource in QueryLinker

mwjames:


mwjames:
Just ensures that results retrieved from an external source continue to do so in `Special:Ask` when referenced through a `further results` link.

image

PR #1561, opened 2016-05-07
ExtraneousLanguageFileHandler to add zh-cn as fallback for zh*

mwjames:


mwjames:
image

which ensures that we find a corresponding match for zh* when switching the user language.

image

PR #1564, opened 2016-05-07
Use the canonical form in a printrequest, refs #1554

mwjames:
refs #1554


mwjames:
Ensures that when switching the user language it will provide the correct reference for the underlying printrequest.

image

image

Issue #1567, opened 2016-05-12
`#` and tags in property names produce data junk


mwjames:
### Setup and configuration

- SMW version: any release
- MW version: N/A
- DB (MySQL etc.):

### Issue

Property names in its current form can contain `[[&#124; &#93;&#93;&lt;/td&gt;&lt;::123]]` [0] or `[[<code>Has property a</code>::ABC]]` which allows to produce data junk (or better nonsense).

![image](https://cloud.githubusercontent.com/assets/1245473/15153077/2ff2dba8-16d8-11e6-95e8-7828da57e994.png)

![image](https://cloud.githubusercontent.com/assets/1245473/15153238/c0e3be20-16d8-11e6-89d1-69cbb5007d4d.png)

[0] http://species-id.net/openmedia/Special:Properties

PR #1570, opened 2016-05-12
Add "Object ID disposal" to Special:SMWAdmin, refs #1567

mwjames:
refs #1567


mwjames:
https://www.semantic-mediawiki.org/wiki/Help:Object_ID_disposal

mwjames:
@kghbln FYI

PR #1577, opened 2016-05-12
Position in PLURAL

mwjames:
Follow-up on https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/c2e5daa762373a2c6255b3ace454d78377f0ad3c#commitcomment-17451900

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -204,7 +204,7 @@
 	"smw-ask-sorting": "Sorting",
 	"smw-ask-search": "Search",
 	"smw-ask-format-selection-help": "For a detailed description, please visit the $1 help page.",
-	"smw-ask-query-search-info":"The query <code><nowiki>$1</nowiki></code> was answered by the {{PLURAL:$3|<code>$2</code> (from cache)|<code>$2</code>}} in $4 {{PLURAL:$4|second|seconds}}.",
+	"smw-ask-query-search-info":"The query <code><nowiki>$1</nowiki></code> was answered by the {{PLURAL:$3|1=<code>$2</code> (from cache)|<code>$2</code>}} in $4 {{PLURAL:$4|second|seconds}}.",
 	"searchbyproperty": "Search by property",
 	"smw_sbv_docu": "Search for all pages that have a given property and value.",
 	"smw_sbv_novalue": "Enter a valid value for the property, or view all property values for \"$1\".",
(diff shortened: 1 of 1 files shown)


mwjames:
@kghbln I'm entrusting you with the merge.

i18n/en.json
```diff
@@ -204,7 +204,7 @@
 	"smw-ask-sorting": "Sorting",
 	"smw-ask-search": "Search",
 	"smw-ask-format-selection-help": "For a detailed description, please visit the $1 help page.",
-	"smw-ask-query-search-info":"The query <code><nowiki>$1</nowiki></code> was answered by the {{PLURAL:$3|<code>$2</code> (from cache)|<code>$2</code>}} in $4 {{PLURAL:$4|second|seconds}}.",
+	"smw-ask-query-search-info":"The query <code><nowiki>$1</nowiki></code> was answered by the {{PLURAL:$3|<code>$2</code>|<code>$2</code> (from cache)}} in $4 {{PLURAL:$4|second|seconds}}.",
```
Nikerabbit: I don't see this doing what I asked. I am expecting to see `{{PLURAL:$3|1=<code>$2</code>|<code>$2</code> (from cache)}}` or something similar.

mwjames: > I don't see this doing what I asked

Not sure I understand the expectation. So you want a number as in `1=<code>$2</code> (from cache)` and `0=<code>$2</code>`?

Nikerabbit: If you have `{{PLURAL:$1|1=something|other}}` it basically means that language specific plural rules are ignored: first parameter is only used for number 1 (and not whatever the language considers singular), and the other parameter for everything else. I don't mind which part is `something` and which part `other`.

kghbln: From what I have read this should be it (assuming that true is singular = from cache):

```The query <code><nowiki>$1</nowiki></code> was answered by the {{PLURAL:$3|1=<code>$2</code> (from cache)|<code>$2</code> (from cache)|<code>$2</code>}} in $4 {{PLURAL:$4|second|seconds}}```

@Nikerabbit If you could send an ok and I will add this ...

Nikerabbit:
@kghbln Can I assume you figured out your question yourself, or is more clarification needed?

kghbln:
@Nikerabbit I will take this as a yes. Thanks!
> (assuming that true is singular = from cache)

Was not to be checked by you in connection with this. :)

PR #1579, opened 2016-05-12
Enable to alter SQLStore::FIXED_PROPERTY_ID_UPPERBOUND

mwjames:
`FIXED_PROPERTY_ID_UPPERBOUND = 50` can become an issue in future, the PR  ensures that the border can be altered without running a full rebuild (by moving/re-assigning any entity that is not a fixed property within the border to a new ID).


mwjames:
The `FIXED_PROPERTY_ID_UPPERBOUND` ID was not adjusted therefore behaviour is expected to remain the same.

PR #1587, opened 2016-05-15
Support LOCL output in Factbox and SearchByProperty

mwjames:
https://www.semantic-mediawiki.org/wiki/Localization


mwjames:
Just so that those DV that support localized display can make use of it.

image

image

PR #1580, opened 2016-05-12
Tidy BooleanValue and add support for localization

mwjames:
- Add unit test
- Add output option LOCL (and LOCL@... to indicate a specifc language)

PR #1591, opened 2016-05-16
Add support for LOCL in IntlNumberFormatter

mwjames:


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1591

Number output formatting dependants on the user language (switching the user lang will also adjust the output) or can be specified using `LOCL@...`.

If the [page content language](https://www.semantic-mediawiki.org/wiki/Page_content_language) is available (as for [0, 1]) then parsing of values will take that into account locale-specific rules as in "number 12345.67 is "12,345.67" in the US, "12 345,67" in France and "12.345,67" in Germany".

image

image

image

[0] http://sandbox.semantic-mediawiki.org/wiki/Issue/1591/FR
[1] http://sandbox.semantic-mediawiki.org/wiki/Issue/1591/EN

mwjames:
This should be the final piece (after `Date`(#1545) and `Boolean` (#1580)) to bring general [localization](https://www.semantic-mediawiki.org/wiki/Localization) support to SMW users. This means that when switching the user language (e.g. via ULS) content is expected to be displayed with a more localized flavour (under the constraint that localization rules for a specific data type are available).

mwjames:
@SemanticMediaWiki/testers It would be nice if some people find the time to fetch the new 2.4-alpha master and do some test rounds.

jaideraf:
@mwjames awesome!
I did'nt test very much the number datatype (because I don't use it in my wiki), but as far I can see it is all working right. :+1: 
This, however, reminds me this old feature request:
https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/30
Maybe it can interest you. It's sad that some result formats don't behave like the "standard" or "default" result formats (like broadtable).

PR #1601, opened 2016-05-20
Add InfoLinksProvider

mwjames:
It moves code responsible for info/service links generation from `DataValue` into its own `InfoLinksProvider` class. It also tackles the issue for links containing a `::`.


mwjames:
> It also tackles the issue for links containing a `::`

http://sandbox.semantic-mediawiki.org/wiki/Issue/1601

PR #1499, opened 2016-04-07
Silence the TransactionProfiler in 1.27+, refs #1295

mwjames:
refs #1295 

```
[DBPerformance] Expectation (writes <= 0) by MediaWiki::main not met:
query-m: CREATE TEMPORARY TABLE `t3`( id INT UNSIGNED KEY ) ENGINE=MEMORY [TRX#a85b3b08e008]
TransactionProfiler.php line 311 calls wfBacktrace()
TransactionProfiler.php line 200 calls TransactionProfiler->reportExpectationViolated()
Database.php line 856 calls TransactionProfiler->recordQueryCompletion()
Database.php line 242 calls DatabaseBase->query()
QuerySegmentListProcessor.php line 303 calls SMW\MediaWiki\Database->query()
QuerySegmentListProcessor.php line 242 calls SMW\SQLStore\QueryEngine\QuerySegmentListProcessor->resolveHierarchyForSegment()
QuerySegmentListProcessor.php line 123 calls SMW\SQLStore\QueryEngine\QuerySegmentListProcessor->doResolveBySegment()
QuerySegmentListProcessor.php line 123 calls SMW\SQLStore\QueryEngine\QuerySegmentListProcessor->doResolveBySegment()
QuerySegmentListProcessor.php line 181 calls SMW\SQLStore\QueryEngine\QuerySegmentListProcessor->doResolveBySegment()
QuerySegmentListProcessor.php line 123 calls SMW\SQLStore\QueryEngine\QuerySegmentListProcessor->doResolveBySegment()
QuerySegmentListProcessor.php line 112 calls SMW\SQLStore\QueryEngine\QuerySegmentListProcessor->doResolveBySegment()
QueryEngine.php line 215 calls SMW\SQLStore\QueryEngine\QuerySegmentListProcessor->doExecuteSubqueryJoinDependenciesFor()
SMW_SQLStore3.php line 387 calls SMW\SQLStore\QueryEngine\QueryEngine->getQueryResult()
SMW_SQLStore3.php line 377 calls SMWSQLStore3->fetchQueryResult()
SMW_QueryProcessor.php line 514 calls SMWSQLStore3->getQueryResult()
AskParserFunction.php line 189 calls SMWQueryProcessor::getResultFromQuery()
AskParserFunction.php line 115 calls SMW\AskParserFunction->doFetchResultsForRawParameters()
ParserFunctionFactory.php line 248 calls SMW\AskParserFunction->parse()
```

PR #1603, opened 2016-05-22
Add resetTransactionProfiler to DatabaseConnectionProvider, refs #1499

mwjames:
refs #1499

### DIFF
--- a/includes/Setup.php
+++ b/includes/Setup.php
@@ -113,9 +113,12 @@ final class Setup {
 		);
 
 		// Connection can be used to redirect queries to another DB cluster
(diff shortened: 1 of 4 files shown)


mwjames:
Instead of a global `\Profiler::instance()->getTransactionProfiler()->resetExpectations();` we reset the `TransactionProfiler` for a specific connection (`QueryEgnine` only) as seen in [0].

[0] https://github.com/wikimedia/mediawiki/blob/d5c44513800ed85b49c5a7c5c1bb930d8787a6ef/includes/objectcache/SqlBagOStuff.php#L156

PR #1610, opened 2016-05-24
Fix PostgreSQL support

vitalif:
Without this commit, SMW with pgsql results in:

Notice: pg_send_query(): There are results on this connection. Call pg_get_result() until it returns FALSE in ./includes/db/DatabasePostgres.php on line 490

and then

RuntimeException from line 181 of ./extensions/SemanticMediaWiki/src/MediaWiki/Database.php: Unable to post new query to PostgreSQL

Also using anonymous DO block does not pollute pg_temp with stored procedures and is shorter.


JeroenDeDauw:
Thank you for contributing

vitalif:
Thanks :)

Issue #1611, opened 2016-05-24
Native displaytitle rendering has issues with ampersand character


Fannon:
### Setup and configuration
- SMW version: 2.4
- MW version: 1.25
- DB MySQL

### Issue
When using an Ampersand (&) as part of the Displaytitle, the resulting link title will contain the corresponding HTML entity `&amp;` 

### Steps to reproduce the observation:
* Create a new page "Web_-26_Wine", containing:
```
{{DISPLAYTITLE:Web & Wine}}
```
* Go To Browse Properties. The title will be: `Web &amp; Wine`
* Do an ASK Query: [[Web & Wine]] - the title will be `Web &amp; Wine`

This seems to be caused, because SMW created the display title itself. If I create a link to the page somewhere else, the Semantic Title Extension will rewrite the title in a correct way. `Web & Wine`

### Probably related:
* https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1410
* https://phabricator.wikimedia.org/T136092

gesinn-it:
To reproduce, create a page "Web & Wine" including the ampersand, e.g. via Redlink. MW will automatically create the urlencoded (%26) page as mentioned by @Fannon .

JeroenDeDauw:
Thank you both for reporting this issue

mwjames:
Are we clear on this one?

Fannon:
Yes, this issue should be resolved!

2016-05-26 12:36 GMT+02:00 mwjames <notifications@github.com>:

PR #1613, opened 2016-05-25
removing html special char escaping from link title of wikipages

Fannon:
This fixes my Issue #1611 
I'm not quite sure that htmlspecialchar escaping is needed in this place, as the title should have been properly escaped before?

To me it looks like the title string is encoded twice.


mwjames:
Thanks.

> I'm not quite sure that htmlspecialchar escaping is needed in this place, as the title should have been properly escaped before?
> To me it looks like the title string is encoded twice.

This class hasn't seen any proper review and I'd like to avoid to open a can or worms especially in terms of escaping which has its own huge problems in MW.

While it maybe that we have some sort of excessive escaping in this class, I think #1614 addresses the issue at its root and therefore is a better approach to solve #1611.

Fannon:
I agree. Your solution is the safer way to solve this problem. Thanks!

PR #1614, opened 2016-05-25
Display title to decode html entities, refs #1611

mwjames:
The display title retrieved from the parser is encoded hence `DisplayTitlePropertyAnnotator` need to decode the string.

refs #1611


Fannon:
Thanks!

PR #1620, opened 2016-05-26
patch for avoiding call sparql endpoint during execution of semantic …

gcornacchia:
…queries.

This is the pull request for what discussed  [here](http://wikimedia.7.x6.nabble.com/Performance-issues-with-external-triple-store-td5053674.html).
In our experience this patch improves significantly the performance of displaying a page that contains a lot of sparql queries.

### DIFF
--- a/src/SPARQLStore/SPARQLStore.php
+++ b/src/SPARQLStore/SPARQLStore.php
@@ -270,14 +270,13 @@ class SPARQLStore extends Store {
 	 * @since 1.6
 	 */
 	public function getQueryResult( Query $query ) {
-		$callEndpoint = $GLOBALS['smwgCallSparqlEndpointInQuery'];
 		$result = null;
 
 		if ( \Hooks::run( 'SMW::Store::BeforeQueryResultLookupComplete', array( $this, $query, &$result ) ) ) {
-			if ($callEndpoint == true) {
+			if ($GLOBALS['smwgCallSparqlEndpointInQuery']) {
 				$result = $this->fetchQueryResult( $query );
 			} else {
-				$result = $this->baseStore->getQueryResult($query);
+				$result = $this->baseStore->getQueryResult( $query );
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


src/SPARQLStore/SPARQLStore.php
```diff
@@ -270,11 +270,15 @@ public function doSparqlDataDelete( DataItem $dataItem ) {
 	 * @since 1.6
 	 */
 	public function getQueryResult( Query $query ) {
-
+		$callEndpoint = $GLOBALS['smwgCallSparqlEndpointInQuery'];
 		$result = null;
 
 		if ( \Hooks::run( 'SMW::Store::BeforeQueryResultLookupComplete', array( $this, $query, &$result ) ) ) {
-			$result = $this->fetchQueryResult( $query );
+			if ($callEndpoint == true) {
```
JeroenDeDauw: `if ( $GLOBALS['smwgCallSparqlEndpointInQuery'] ) {`

src/SPARQLStore/SPARQLStore.php
```diff
@@ -270,11 +270,15 @@ public function doSparqlDataDelete( DataItem $dataItem ) {
...
+				$result = $this->fetchQueryResult( $query );
+			} else {
+				$result = $this->baseStore->getQueryResult($query);
```
JeroenDeDauw: `( $query )`

JeroenDeDauw:
Thanks for contributing!

mwjames:
> In our experience this patch improves significantly the performance of displaying a page that contains a lot of sparql queries.

While I'm not opposed to this change, I'm __burning to see__ an example (on the sandbox which can be anonymized) to see the effect and possible ramifications.

The sandbox itself doesn't use the `SPARQLStore` but I can use the content on my local instance (which uses Sesame) to run some tests.

JeroenDeDauw:
Sandbox: http://sandbox.semantic-mediawiki.org

gcornacchia:
Any news?

mwjames:
> Any news?

I'm still waiting on an example to evaluate the influence and have a reference in terms of possible improvements in future.

I'm also wondering about introducing a new setting, what about using `smwgSparqlQueryEndpoint = false` instead to make a switch.

mwjames:
>I'm also wondering about introducing a new setting, what about using smwgSparqlQueryEndpoint = false instead to make a switch.

Superseded by #2159.

Issue #1590, opened 2016-05-16
1.28.0 / Linker::link() rewrite


mwjames:
[0] wrote:

> For the most part, you'd use it in similar ways:
 Linker::link( $title, $html, $attribs, $query );
is now:
 $linkRenderer = MediaWikiServices::getInstance()
   ->getHtmlPageLinkRenderer();
 $linkRenderer->makeLink( $title, $html, $attribs, $query );

[0] https://lists.wikimedia.org/pipermail/wikitech-l/2016-May/085634.html

mwjames:
@legoktm We have at least two access points where `Linker::link` (off the top of my head) are used and I would highly appreciate that we avoid any situation where this gets broken. 

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/datavalues/SMW_DV_WikiPage.php#L359

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/datavalues/SMW_DV_WikiPage.php#L364

legoktm:
I think the best way we can do that is to make sure there are test cases in MediaWiki core (<https://phabricator.wikimedia.org/diffusion/MW/browse/master/tests/phpunit/includes/LinkerTest.php>) that cover the functionality that SMW is using to generate links - I can watch out and make sure I don't break anything, but tests are going to be the best way to make sure no one else unintentionally does as well.

@mwjames: Is writing additional test cases for core something you can work on?

legoktm:
This landed in master about a week ago - was there any breakage?

mwjames:
> This landed in master about a week ago - was there any breakage?

Thanks for caring.

I run our integration tests locally (unit tests wouldn't help much in this case when run as mocked instance) plus \[0] and so far it looks good (no reported errors that could be traced to this change).

[0] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/134658789

kghbln:
Actionable tasks to be assessed for SMW 3.1.0

kghbln:
[...]

PR #1622, opened 2016-05-30
Stop using DummyLinker

legoktm:
In MW 1.28, we are working on deprecating DummyLinker, so always use
Linker instead, since the static methods can still be called as instance
methods without any issues.

### DIFF
--- a/includes/GlobalFunctions.php
+++ b/includes/GlobalFunctions.php
@@ -171,18 +171,19 @@ function &smwfGetSparqlDatabase() {
  * Compatibility helper for using Linker methods.
  * MW 1.16 has a Linker with non-static methods,
  * where in MW 1.19 they are static, and a DummyLinker
- * class is introduced, which can be instantaited for
- * compat reasons.
+ * class is introduced, which can be instantiated for
+ * compat reasons. As of MW 1.28, DummyLinker is being
+ * deprecated, so always use Linker.
  *
  * @since 1.6
  *
- * @return Linker or DummyLinker
+ * @return Linker
  */
 function smwfGetLinker() {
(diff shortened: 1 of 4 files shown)


includes/GlobalFunctions.php
```diff
@@ -171,18 +171,19 @@ function &smwfGetSparqlDatabase() {
  * Compatibility helper for using Linker methods.
  * MW 1.16 has a Linker with non-static methods,
  * where in MW 1.19 they are static, and a DummyLinker
- * class is introduced, which can be instantaited for
- * compat reasons.
+ * class is introduced, which can be instantiated for
+ * compat reasons. As of MW 1.28, DummyLinker is being
+ * deprecated, so always use Linker.
```
JeroenDeDauw: So now the comment says we're doing A because foo but actually do B because bar?

JeroenDeDauw:
Thanks!

JeroenDeDauw:
Does this fix https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1590?

includes/GlobalFunctions.php
legoktm: Basically....I'm not far how back MW versions you want this doc comment to be about. Since it used to return DummyLinker, I added a comment explaining why it no longer does.

legoktm:
> Does this fix #1590?

No, this just switches from `DummyLinker` (effectively deprecated since Linker replaced it in 1.19), to `Linker`, which is going to become deprecated in 1.28. AIUI #1590 is about switching to `LinkRenderer`, which is the new-new system.

mwjames:
Thanks

PR #1627, opened 2016-06-02
Add TemporaryEntityListAccumulator

mwjames:
Allows to track entities created by/during QueryResult/ResultArray instantiation and avoids a possible performance penalty when trying to record them independently.


mwjames:
image

mwjames:
This all just means that those subjects in a column are now traceable without a performance penalty (and before this PR those in a column were not tracked except for subjects in column 0).

image

PR #1629, opened 2016-06-03
fixing one translation; and proposed enhancements

agharbeia:
Fixing messed up translation in "PageProperty";
Unifying on "خصيصة" instead of "خاصية" for "Properties" and "مفاهيم" instead of "مبادئ" for "Concepts";
Acronyms in Latin script rather than Arabic script for better recognizability.

### DIFF
--- a/languages/SMW_Aliases.php
+++ b/languages/SMW_Aliases.php
@@ -2,6 +2,7 @@
 /**
  * Aliases for special pages
  *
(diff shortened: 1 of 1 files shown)


mwjames:
Thanks

I'll leave this to @kghbln for review and merge.

kghbln:
Thanks for the improvements. There is always the possibility to extend the array like at "SMWAdmin"

PR #1630, opened 2016-06-03
Arabic / new translations; a couple of fixes; enhancements

agharbeia:
Abstracted names of Datatype lables for better suitability as data type descriptions and fixed one messed up translation;
changed several names of SpecialProperties for better suitability as descriptions of relationships;
trasnlated several untranslated strings;
used acronym in Latin script for better recognizability;
Fixed messed up one SpecialPropertyAliases translation;
Unified on "خصيصة" for "property" instead of "خاصية" and "مفهوم" for "concept" instead of "مبدأ";
Three-letter abbreviations for month names;

### DIFF
--- a/languages/SMW_LanguageAr.php
+++ b/languages/SMW_LanguageAr.php
@@ -1,6 +1,5 @@
 <?php
 /**
- * @author أحمد غربية <ahmad@arabdigitalexpression.org>
  * @file
  * @ingroup SMWLanguage
  */
@@ -20,7 +19,7 @@ include_once ( $smwgIP . 'languages/SMW_Language.php' );
  *
  * @author Mahmoud Zouari  mahmoudzouari@yahoo.fr http://www.cri.ensmp.fr
  * @author Meno25
- * @auhtor Ahmad Gharbeia أحمد غربية <ahmad@arabdigitalexpression.org>
+ * @author Ahmad Gharbeia أحمد غربية <ahmad@arabdigitalexpression.org>
  * @ingroup SMWLanguage
  * @ingroup Language
  */
--- a/languages/SMW_LanguageAr.php
+++ b/languages/SMW_LanguageAr.php
@@ -1,5 +1,6 @@
 <?php
 /**
+ * @author أحمد غربية <ahmad@arabdigitalexpression.org>
  * @file
  * @ingroup SMWLanguage
  */
@@ -19,27 +20,28 @@ include_once ( $smwgIP . 'languages/SMW_Language.php' );
  *
  * @author Mahmoud Zouari  mahmoudzouari@yahoo.fr http://www.cri.ensmp.fr
  * @author Meno25
+ * @auhtor Ahmad Gharbeia أحمد غربية <ahmad@arabdigitalexpression.org>
  * @ingroup SMWLanguage
  * @ingroup Language
  */
 class SMWLanguageAr extends SMWLanguage {
 
 	protected $m_DatatypeLabels = array(
-		'_wpg' => 'الصفحة', // name of page datatype
+		'_wpg' => 'صفحة', // name of page datatype
 		'_txt' => 'نص',  // name of the text type
 		'_cod' => 'كود',  // name of the (source) code type
 		'_boo' => 'منطقي',  // name of the boolean type
 		'_num' => 'عدد',  // name for the datatype of numbers
-		'_geo' => 'الإحداثيات الجغرافية', // name of the geocoord type
-		'_tem' => 'الحرارة',  // name of the temperature type
-		'_dat' => 'التاريخ',  // name of the datetime (calendar) type
-		'_ema' => 'البريد الإلكتروني',  // name of the email type
-		'_uri' => 'مسار',  // name of the URL type
-		'_anu' => 'التعليق علي معرف الموارد الموحد',  // name of the annotation URI type (OWL annotation property)
-		'_tel' => 'رقم الهاتف',  // name of the telephone (URI) type
-		'_rec' => 'تسجيل', // name of record data type
-		'_qty' => 'Quantity', // name of the number type with units of measurement //TODO: translate
-		'_mlt_rec' => 'Monolingual text',
+		'_geo' => 'إحداثيات جغرافية', // name of the geocoord type
+		'_tem' => 'درجة حرارة',  // name of the temperature type
+		'_dat' => 'تاريخ',  // name of the datetime (calendar) type
+		'_ema' => 'عنوان بريد إلكتروني',  // name of the email type
+		'_uri' => 'URL',  // name of the URL type
+		'_anu' => 'URL حاشية',  // name of the annotation URI type (OWL annotation property)
+		'_tel' => 'رقم هاتفي',  // name of the telephone (URI) type
+		'_rec' => 'ّسجل', // name of record data type
+		'_qty' => 'كميّة', // name of the number type with units of measurement
+		'_mlt_rec' => 'نص أحادي اللغة',
 	);
 
 	protected $m_DatatypeAliases = array(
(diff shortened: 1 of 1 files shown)


languages/SMW_LanguageAr.php
```diff
@@ -1,5 +1,6 @@
 <?php
 /**
+ * @author أحمد غربية <ahmad@arabdigitalexpression.org>
```
kghbln: This line can be removed since you are already mentioned in the file.

languages/SMW_LanguageAr.php
```diff
@@ -19,27 +20,28 @@
  *
  * @author Mahmoud Zouari  mahmoudzouari@yahoo.fr http://www.cri.ensmp.fr
  * @author Meno25
+ * @auhtor Ahmad Gharbeia أحمد غربية <ahmad@arabdigitalexpression.org>
```
kghbln: typo auhtor -> author

languages/SMW_LanguageAr.php
```diff
@@ -52,62 +54,58 @@ class SMWLanguageAr extends SMWLanguage {
...
-		'_LCODE' => 'Language code',
-		'_TEXT'  => 'Text',
-		'_PDESC' => 'Has property description',
-		'_PVAP'  => 'Allows pattern',
-		'_DTITLE' => 'Display title of',
-		'_PVUC' => 'Has uniqueness constraint',
+		'_CDAT' => 'تاريخ الإنشاء',
+		'_NEWP' => 'هي صفحة جديدة',
+		'_LEDT' => 'آخر مَن حرّرها',
+		'_ERRP' => 'فيها قيمة غير صحيحة في',//الخصيصة
```
kghbln: This does not look ok to me.

kghbln:
We have to be aware here that this commit results into a breaking change for all wikis using "ar" as the instance language, especially the new names for the special properties which may also require changing wikitext on wiki pages.

languages/SMW_LanguageAr.php
agharbeia: As far as I can see the diff is misaligned for some reason but all of the strings are there.

agharbeia:
Yes. We recognise it is a breaking change. We hesitated for long before we decided to instate it.

However it is based on real-life experimentation with SMW in which we found the original translations to be immature as to be confusing.

Some strings are outright wrong, for example "display unit" is translated into Arabic as to mean "display of unit" rather than "unit of display". This is attributed both to the ambiguity in English in this kind of compound nouns, and to a common mistake that amateur Arabic translators from English exhibit because of the discrepancy of the order of noun and adjective between the two languages.

Other translations suffer different shortcomings that I can present my rational for changing each, if needed.

The Arabic translation of MW itself suffers similar shortcomings and errors, but fixing that is much more difficult due to its ties with the Wikipedia community (which I'm a founding member of, but still doesn't make it easier). We use our own translations or MW in our installations for that reason.

I thought I'd fix Arabic SMW before this becomes more difficult.

I believe the couple of SMW installations I'm currently maintaining are the only life  examples of SMW usage in Arabic, and I'm working on a couple more.

kghbln:
Oops, I nearly "forgot" about this one. :( Well thank you for your elaboration. From my experience as a translator and what I have read I believe that the proposed changes will bring more consistency in naming and actually improvements as for the actual meaning (true semantics) of the translations - even without me understanding Arabic at all. **We just have to be aware of one thing.** Once this is merged there is no way back, i.e. further changes are indeed not acceptable. So it will probably be an idea to have a second opinion about these translations. Is there a way to organize this to be utterly sure about the translations?

languages/SMW_LanguageAr.php
kghbln: > are you concerned with the inline comment causing the BiDi mess

Yeah I am, but on the other hand the automated tests would ring alarm bells if this was breaking PHP and so. Since this is not the case...

agharbeia:
Sure. Another eye is always good. I always regret it when I don't.
I don't want to recommend people. We can ask Meno25 since his involvement comes naturally. I would have suggested to put it out for the crowd at http://arabeyes.net to dissect it. Some of my close allies and arch-rivals used to collaborate on arabizing FOSS there but I'm not sure anyone active there now.

kghbln:
I am not sure if you did so I asked [Meno25](https://translatewiki.net/wiki/Thread:User_talk:Meno25/Help_requested_for_Arabic_translations) and [محمد أحمد عبد الفتاح](https://translatewiki.net/wiki/Thread:User_talk:%D9%85%D8%AD%D9%85%D8%AF_%D8%A3%D8%AD%D9%85%D8%AF_%D8%B9%D8%A8%D8%AF_%D8%A7%D9%84%D9%81%D8%AA%D8%A7%D8%AD/Help_requested_for_Arabic_translations) for their input.

languages/SMW_LanguageAr.php
Meno25: This is a bug in how GitHub viewer handles BiDi text. The patch itself is correct. Regarding the translations, I have 2 objections:
1-The change in line 108 (protected $m_monthsshort): Gharbeia, you know well that those short names are simply not used in the Arab world (or at least not widely used outside the Arab Eyes people). It doesn't make sense to add short names by yourself if most native speakers don't know them. I advise against merging that specific line.
2-I am opposed to using tashkil (تشكيل or Arabic diacritics) in the translations especially in things that can't be changed directly using the user interface such as the case here. Tashkil looks bad on screens with low resolutions and it makes the "Find" or "Search" function in the web browser useless. Also, some diacritics are hard to be found when using an Arabic keyboard.
I made a new patch based on Gharbeia's one but without line 108 and tashkil.

[mypatch.txt](https://github.com/SemanticMediaWiki/SemanticMediaWiki/files/350362/mypatch.txt)

kghbln: @Meno25 Thanks a lot for your elaborate input!

So to sum up as someone who neither speaks nor understands Arabic: Apart from the strong concerns mentioned in 1 the translations as such are ok except for the way these are actually written. I assume that MediaWiki is not using diacritics either so I makes sense to use your patch. To clarify 1 should be easy too since MediaWiki also uses short names. For consistency these should be used here too.

Meno25: @kghbln Thank you for your comment. Yes. The issue of using diacritics is rather controversial on Arabic Wikipedia. Some users support it while others don't. Anyway, the Arabic MediaWiki translations file mostly doesn't contain diacritics. However, If you need a wider discussion by community about this issue, then I can leave a note on the Arabic Wikipedia (where I am admin and checkuser) Village Pump.

kghbln: > The issue of using diacritics is rather controversial on Arabic Wikipedia.

I do not think that we should try to solve a problem here which exists on Arabic Wikipedia for 15 years - obviously without conclusion.

> Anyway, the Arabic MediaWiki translations file mostly doesn't contain diacritics.

So that's what we should do here, too. Note at all to avoid misunderstandings: I am not judging whether diacritics are better to use or not.

@Meno25 The reason I asked for a second opinion was just to make sure that the translations as such are sensible and understandable in the light of their context. This seems to be the case.

agharbeia: Thanks for your review and feed back.

Let me first comment about the rationale: You probably know, Meno, that many norms and widely accepted conventions of localisation have been changed and re-thought during the past ten years of FOSS localisation. Including the introduction of idioms that hadn't existed before, and bringing others under critique. Whether those changes are for the better or for the worse. This will remain controversial. I personally think that much remains in need of rethinking. I also know that the Arabic Wikipedia community has become increasingly more conservative; not only in regard to localisation.

Regarding the short months names case: I know they are used in Gnome, a major FOSS package, as well as the Mozilla suite.

Language - and modes thereof - is just another technology of communication. It has phases. ICT has introduced modes and idioms that hadn't existed before, and I don't see why we should insist to sticking to what has been known before in this specific point, i.e. month names. For me if is is useful, it should be allowed; people should be given the chance to try them. After all in all applications I know of, dates formats are selectable. No imposing of anything is taking place here.

As for tashkeel: You probably know that it is an essential part of the Arabic writing system, without which it is not considered complete. The absence of tashkeel from computing has been traditionally - in my opinion - due to the difficulties of presentation; i.e fonts, resolutions, encoding. This is all mostly of the past now. What remains of these difficulties should be tackled technically, if possible, rather than imposing restrictions in how the language is used. The language is and will continue to change because of the effects of technology. I just don't see this change and rational as valid or necessary. An example is the search function in browsers, which should implement switches to ignore tashkeel, as what popular search engines are doing. in turn search engines SHOULD implement switches to not ignore tashkeel because this can make a hell lot of a difference in search results in some cases, such as classical texts. The Arabic keyboard is evil, but that's another discussion.

Most importantly, the absence of tashkeel causes ambiguity and/or prevents the use of concise expressions that are arguably more suitable in certain cases, for fear of ambiguity.

I personally think that since SMW is not yet an integral part of Wikipedia, and that SMW is almost not used in Arabic mediawikis at all, it should not be strictly governed by WP-ar conventions. At least not without a space for reconsidering solutions.

In all cases, for the sake of the projects I'm working on, I will have to maintain a different translation for this and some other modules, as we do for our base MW installations, and hope that new insights in the discussions will bring them closer in the future. Even though I doubt this will be possible, if the rationale is basically "we need to do it this way because this is how it has been done so far in those other places" due of the exponential difficulty of effecting change in the body of localisation texts, while in fact fixing a small part and rarely used module such as SMW should be independently possible; but alas.

Thank you all, and regards.

Meno25: > Whether those changes are for the better or for the worse. This will remain controversial.

I agree here.
> I also know that the Arabic Wikipedia community has become increasingly more conservative; not only in regard to localisation.

Maybe you are right. However, Arabic Wikipedia is the largest Arabic language wiki on the Internet. So, the opinion of the Arabic Wikipedia community should be taken into consideration here. After all, Semantic MediaWiki is a MediaWiki extension and it was we (Arabic Wikipedia users) who started the Arabic MediaWiki translation in the first place.
>I know they are used in Gnome, a major FOSS package, as well as the Mozilla suite.

Both were translated by Arab Eyes people if I was not mistaken.
>The Arabic keyboard is evil, but that's another discussion.

I agree.
>For me if is is useful, it should be allowed; people should be given the chance to try them.

I respect your opinion but we disagree here. We simply don't "invent" language. In Wikipedia, we use only words and phrases commonly used in Arabic sources/literature.
>  that it is an essential part of the Arabic writing system

Yes, I know. Both of us have more than 10 years experience with Wikipedia. However, you only make a few edits every month while I do make hundreds of edits every week. That's why we disagree here. The issue of diacritics is very annoying when are trying to find a certain word in 150 kilobytes article/webpage and you don't have time to read all the article/webpage. I agree that web browsers and search engines code should be modified as you described. However until this happens, let's not make things harder for very active users (who make most edits in any wiki as you know).

Another note: You know well that most Arabic websites don't use diacritics for the reasons you and I stated here (due to the difficulties of presentation; i.e fonts, resolutions, encoding).
> This is all mostly of the past now.

Not in my opinion.
> prevents the use of concise expressions that are arguably more suitable in certain cases, for fear of ambiguity.

True. However, the problem of using Arabic diacritics has no ideal solution. We are just weighting pros and cons.

Bottom line: I have stated my opinion here and I have nothing further to add. I have already submitted a patch and the decision of whether to merge your patch or mine is up to kghbln. I was happy to have this conversation with you and thank you.

agharbeia: >     I also know that the Arabic Wikipedia community has become
>     increasingly more conservative; not only in regard to localisation.
> 
> Maybe you are right. However, Arabic Wikipedia is the largest Arabic
> language wiki on the Internet. So, the opinion of the Arabic Wikipedia
> community should be taken into consideration here. After all, Semantic
> MediaWiki is a MediaWiki extension and it was we (Arabic Wikipedia
> users) who started the Arabic MediaWiki translation in the first place.

In my opinion it doesn't matter who *started* the translation in the
first place. Many first translations of software packages are worthless.
What I mean here is that first instance in itself should not be a
determinant.


>     I know they are used in Gnome, a major FOSS package, as well as the
>     Mozilla suite.
> 
> Both were translated by Arab Eyes people if I was not mistaken.

If this is so, does it automatically mean they are not orthodox?
I hope you're not governing on the work just by the identity of the
maker. If the point your putting forward is existing usability, then
existence in such popular software packages could mean these proposed
short-forms are probably used by many, regardless of how they got there.
Just like new words that pop-up everyday in all spoken languages. I
don't claim to have numbers, though.

> 
>     For me if is is useful, it should be allowed; people should be given
>     the chance to try them.
> 
> I respect your opinion but we disagree here. We simply don't "invent"
> language. In Wikipedia, we use only words and phrases commonly used in
> Arabic sources/literature.

I think you're carrying the "no original work" rule from Wikipedia
beyond its applicability here. All localisation of software and
technology entails original creation. Technology has already brought new
modes of language into usage. The Wikipedia jargon is itself an example
of that.


>     that it is an essential part of the Arabic writing system
> 
> Yes, I know. Both of us have more than 10 years experience with
> Wikipedia. However, you only make a few edits every month while I do
> make hundreds of edits every week. That's why we disagree here. The
> issue of diacritics is very annoying when are trying to find a certain
> word in 150 kilobytes article/webpage and you don't have time to read
> all the article/webpage. I agree that web browsers and search engines
> code should be modified as you described. However until this happens,
> let's not make things harder for very active users (who make most edits
> in any wiki as you know).

I think it's a matter of egg and chicken. If we put rules in place to
inhibit use because of limitations of technology, then it is less likely
for technology to solve.

Moreover, this difficulty could be valid in body text of articles. But
not in UI, right?

Regards,
Ahmad

kghbln: Thanks for all your input. However that's exactly the kind of discussion I wanted to avoid since I believe that it should be done on the MediaWiki level. I had hoped that my https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1630#discussion_r71041858 was unambiguous in this respect.

Here we should stick to what MediaWiki is doing and not what Gnome, KDE or Cinnamon are doing. This may very well be disappointing from one point or view or the other but I believe it is preferable to have consistency within the MediaWiki environment.

You can have issues in a lot of languages, e.g. in German translations MediaWiki writes "du" instead of "Du" as salutation which I feel is impolite or even rude. Both versions are formally possible and "du" was chosen by the community. So it stick to it for the sake of consistency which I very much prefer.

mwjames:
I wonder if we can wrap up this discussion here?

kghbln:
I am inclined to not just cancel the pull request but to also remove the file. Thus every user can add a file according the her or his personal requirements.

PR #1631, opened 2016-06-03
Arabic enhancements

agharbeia:
Unified on "مفهوم" instead of "مبدأ" for "Concept";
"وثائق" instead of "توثيق" for "documentation" as the latter is used for "certification";
Clearer Arabic wording for factbox magic;

### DIFF
--- a/languages/SMW_Magic.php
+++ b/languages/SMW_Magic.php
@@ -3,6 +3,7 @@
 /**
  * Magic words
  *
+ * @author أحمد غربية <ahmad@arabdigitalexpression.org>
  * @file
  * @ingroup Extensions
  * @ingroup SMWLanguage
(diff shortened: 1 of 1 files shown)


kghbln:
This is also a breaking change for wikis in "ar" using these magic words and behaviour switches. Thus I suggest to add the former translations to the end of the respective arrays to allow a transition period.

agharbeia:
I agree. I'll do it.
I just don't know how this is supposed to be done.
Should the index value be increased from 0 to 1 for each proposed change where the larger index is of the original translation, like:
```
…
'SMW_SHOWFACTBOX' => array( 0, '__عرض_صندوق_الحقائق__' ),
'SMW_SHOWFACTBOX' => array( 1, '__عرض_صندوق_الحقيقة__' ),
…
```
I'll appreciate it if you explain this to me.

kghbln:
The French translations could serve as an [example](https://github.com/agharbeia/SemanticMediaWiki/blob/33224924c2826bf2455a514a0bd4cd7522fe5cb9/languages/SMW_Magic.php#L147) so this is e.g. the solution:

 ```'SMW_SHOWFACTBOX' => array( 0, '__عرض_صندوق_الحقائق__' ,'__عرض_صندوق_الحقيقة__' ),```

agharbeia:
OK. Done.
Noting that:
- My motivation for the change pertaining to SMW_NOFACTBOX and SMW_SHOWFACTBOX is my claim that the semantics of the usage of حقيقية vs حقائق are like those of "truth" vs. "facts", even though they are merely different in grammatical number, i.e singular vs. plural respectively.
- I exclude "concept" because I find مفهوم to be radically different in meaning from مبدأ which was used before my proposed change. The latter meaning "principle". I find it misleading and unconductive to use that for a _concept_ so pivotal for understanding the model of SMW.
- I think the same, to a lesser extent, of "declare".
- I added the correction of تعيين instead of مجموعة for "set" since the latter means the mathematical "Menge", also "set" in English.

Q: Where is "set" ever used alone? I haven't encountered it in my use so far.

I think that hints to translators about usage in UI, and whether a word is used as a verb or noun are useful. Especially for those translating blindly, like via TW.net. I added some of those in other places in comments to avoid unsubstantiated reverts from future translators.

kghbln:
> Noting that:

I perfectly understand and it is good to read that you have put a lot of thought into these changes. I do translations to German and face similar issues. Having the "inappropriate" translation in there is to make this change less breaking. Perhaps with SMW 2.6 or so we can drop them.

> Q: Where is "set" ever used alone? I haven't encountered it in my use so far.

See the docu [#set](https://www.semantic-mediawiki.org/wiki/Help:Setting_values). I guess it is particularly nice for multiple value assignments and other cool stuff

PR #1643, opened 2016-06-08
Enforce enableSemantics

mwjames:
`enableSemantics`[0] is required to make sure NS are correctly set and avoid issues as described in #1424.

refs #1398 

[0] https://www.semantic-mediawiki.org/wiki/Help:EnableSemantics


mwjames:
@kghbln Given the issues with 1,27+, I now decided that we enforce `enableSemantics` otherwise SMW is disabled by default.

 I left the installation routine untouched in a sense that `update.php` can still run before `enableSemantics` is added.

```
4. Run the setup script from the base directory[3] of your MediaWiki installation: 
5. Add a call to enableSemantics() to ...
```

PR #1645, opened 2016-06-10
#show not be disturbed by `=`, refs #1542

mwjames:
refs #1542


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1542_%28Equals_sign%29

PR #1661, opened 2016-06-17
Fixed ref to MonolingualTextValue

JeroenDeDauw:
NOTE: this code is clearly not tested, as execution of the part
that refs MonolingualTextValue should result in an error before
this change.

### DIFF
--- a/tests/phpunit/Unit/Query/DescriptionFactoryTest.php
+++ b/tests/phpunit/Unit/Query/DescriptionFactoryTest.php
@@ -225,4 +225,35 @@ class DescriptionFactoryTest extends \PHPUnit_Framework_TestCase {
 		);
 	}
 
+	public function testCanConstructDescriptionFromMonolingualTextValue() {
+
+		$containerSemanticData = $this->getMockBuilder( '\SMWContainerSemanticData' )
+			->disableOriginalConstructor()
+			->getMock();
+
+		$containerSemanticData->expects( $this->atLeastOnce() )
+			->method( 'getPropertyValues' )
+			->will( $this->returnValue( array( $this->dataItemFactory->newDIBlob( 'Bar' ) ) ) );
+
+		$dataValue = $this->getMockBuilder( '\SMW\DataValues\MonolingualTextValue' )
+			->disableOriginalConstructor()
+			->setMethods( array( 'isValid', 'getProperty', 'getDataItem' ) )
+			->getMock();
+
+		$dataValue->expects( $this->atLeastOnce() )
+			->method( 'isValid' )
+			->will( $this->returnValue( true ) );
+
+		$dataValue->expects( $this->atLeastOnce() )
+			->method( 'getDataItem' )
+			->will( $this->returnValue( $this->dataItemFactory->newDIContainer( $containerSemanticData ) ) );
+
+		$instance = new DescriptionFactory();
+
+		$this->assertInstanceOf(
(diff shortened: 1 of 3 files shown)


mwjames:
> this code is clearly not tested, as execution of the part
that refs MonolingualTextValue should result in an error before
this change.

Having a peak now ..

tests/phpunit/Unit/Query/DescriptionFactoryTest.php
```diff
@@ -225,4 +225,35 @@ public function testCanConstructDescriptionFromValidDataValue() {
...
+
+		$containerSemanticData = $this->getMockBuilder( '\SMWContainerSemanticData' )
+			->disableOriginalConstructor()
+			->getMock();
+
+		$containerSemanticData->expects( $this->atLeastOnce() )
+			->method( 'getPropertyValues' )
+			->will( $this->returnValue( array( $this->dataItemFactory->newDIBlob( 'Bar' ) ) ) );
+
+		$dataValue = $this->getMockBuilder( '\SMW\DataValues\MonolingualTextValue' )
```
JeroenDeDauw: Why are you mocking a "value object"?

tests/phpunit/Unit/Query/DescriptionFactoryTest.php
```diff
@@ -225,4 +225,35 @@ public function testCanConstructDescriptionFromValidDataValue() {
...
+			->method( 'getDataItem' )
+			->will( $this->returnValue( $this->dataItemFactory->newDIContainer( $containerSemanticData ) ) );
+
+		$instance = new DescriptionFactory();
+
+		$this->assertInstanceOf(
+			'SMW\Query\Language\Conjunction',
+			$instance->newFromDataValue( $dataValue )
+		);
+	}
```
JeroenDeDauw: This test has 9 string references to code and binds to implementation details.

Some notes here https://www.entropywins.wtf/blog/2016/05/13/5-ways-to-write-better-mocks/

tests/phpunit/Unit/Query/DescriptionFactoryTest.php
mwjames: Because I know what I want to test and setting up a real object is a pain and doesn't help here as I want a specific condition.

tests/phpunit/Unit/Query/DescriptionFactoryTest.php
mwjames: > This test has 9 string references to code and binds to implementation details.

Yes, that's purpose and if someone breaks the interface the test will fail and that is what I'm counting on. Writing a "real" test double is just not worth it, for this you have the mock object.

tests/phpunit/Unit/Query/DescriptionFactoryTest.php
JeroenDeDauw: Creating a `MonolingualTextValue` is a pain? wtf?

tests/phpunit/Unit/Query/DescriptionFactoryTest.php
JeroenDeDauw: It's pretty much the same amount of effort two write one using mock API magic, and using a normal class. The difference in effort is that the later is significantly less of a maintenance hassle.

I'm not asking you to change this here, I think one instance does not really matter. I'm providing the info in the hope you find something you can learn in it.

PR #1677, opened 2016-06-17
Add missing `var`

PeterTheOne:


PeterTheOne:
Ah, this might be intended `/*global console:true message:true */`

PR #1679, opened 2016-06-18
Be strict about a possible duplicate match, refs #1521

mwjames:
refs #1521


mwjames:
This just came to light during SNO reporting an edge case (different property but same sortkey because they used the same display title).

image

Issue #1367, opened 2016-01-13
Provide support for extension.json?


mwjames:
I do not champion the use of `extension.json` [0] and I personally will not dedicate any time in supporting it. I'm not opposed to a PR as long as it does **not**
- interfere with our current development practice
- break our tests
- increase maintenance work for a developer
- increase the infrastructural maintenance burden (e.g. test setup, Travis etc.)

In case someone, intends to provide a patch to add  `extension.json` support the following elements are expected to be manage by PHP (and not by  `extension.json`) as those are individually tested.
-  Hooks
-  autoloading
-  SpecialPages, Actions, JobClasses, APIModules

PS: My position on this extends to each repository found in https://github.com/SemanticMediaWiki.

[0] https://www.semantic-mediawiki.org/wiki/Dependencies,_Composer,_and_more

JeroenDeDauw:
I think it'd be great if someone volunteered to adding support for this thing, creating the relevant documentation for it on wiki, and maintain both. Such a contribution would be very welcome.

This is talking about autoloading via extension.json. Personally I think defining hooks in a custom on-PHP format is big fallacy (which was pointed out on the extension.json RFC and not replied to at all).

I can understand James complete lack of enthusiasm towards doing this though, and share it to a good extend.

mwjames:
@s7eph4n did a nice job with the `extension.json` conversion in `SematicGlossary` (aka SG) where autoloading remains the domain of Composer via PSR-4, hooks are under the control of the extension, and tests do work as before without any changes. 

Yet, the problem lies behind that in how to manage its dependencies. SG relies on Lingo (it also relies on SMW but that's another story) and in order to ensure that nothing falls through the cracks, SG somehow has to ensure Lingo is available before SG is loaded (where Stephan came up with a neat hack).

```
	// must NOT use ExtensionRegistry::getInstance() to avoid recursion!
	$registry = new ExtensionRegistry();
	$registry->load( $GLOBALS[ 'wgExtensionDirectory' ] . '/Lingo/extension.json' );
```

Imaging that each SMW related extension has to add one or more of those lines before the invoker.

Of course, right now we check on `SMW_VERSION` in some of the extensions but this is just a constant which leaves knowledge about `extension.json` or a directory where some of its dependencies are placed (e.g. `$GLOBALS[ 'wgExtensionDirectory' ] . '/Lingo`) invisible and unnecessary for the requester to know about it.
 
[...]

PR #1680, opened 2016-06-18
Move class_alias to its own file

mwjames:
Refs https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1367

### DIFF
--- a/SemanticMediaWiki.php
+++ b/SemanticMediaWiki.php
@@ -24,144 +24,129 @@ if ( defined( 'SMW_VERSION' ) ) {
 	return 1;
 }
 
-define( 'SMW_VERSION', '2.4-alpha' );
+define( 'SMW_VERSION', '2.4.0-alpha' );
 
 if ( version_compare( $GLOBALS['wgVersion'], '1.19c', '<' ) ) {
 	die( '<b>Error:</b> This version of Semantic MediaWiki requires MediaWiki 1.19 or above; use SMW 1.8.x for MediaWiki 1.18.x or 1.17.x.' );
 }
 
-if ( is_readable( __DIR__ . '/vendor/autoload.php' ) ) {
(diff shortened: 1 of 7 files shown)


mwjames:
This also __prepares__ the init in case someone wants to play with the `extension.json` stuff (would require 1.27+ and a functional requires section for extensions [0] though):

```
{
	"name": "Semantic MediaWiki",
	"version": "...",
	"author": [	],
	"url": "https://github.com/SemanticMediaWiki/SemanticMediaWiki/",
	"descriptionmsg": "...",
	"license-name": "GPL-2.0+",
	"type": "semantic",
	"requires": {
		"MediaWiki": ">= 1.27"
	},
	"MessagesDirs": {
		"SemanticMediaWiki": [
			"i18n"
		]
	},
	"AutoloadClasses": {
		"SemanticMediaWiki": "SemanticMediaWiki.php"
	},
	"callback": "SemanticMediaWiki::initExtension",
	"ExtensionFunctions": [
		"SemanticMediaWiki::onExtensionFunction"
	],
	"load_composer_autoloader":true,
	"manifest_version": 1
}
```

[0] "The key of the requires object is the name of the dependency (currently only MediaWiki supported), the value is a valid version constraint (the format has to match the one used by composer)." https://www.mediawiki.org/wiki/Manual:Extension_registration#Requirements_.28Dependencies.29

JeroenDeDauw:
+1, though I'd go with simply Aliases.php, and perhaps not put it in the root

mwjames:
> simply Aliases.php, and perhaps not put it in the root

There is one called Aliases.php in the includes DIR for PHPStorm. Would that interfere?

JeroenDeDauw:
You could rename that one, it's name does not really matter. It's not included by anything after all.

Issue #399, opened 2014-07-09
Remove support for the "type" namespace entirely


kghbln:
This namespace was depreciated and its usage discouraged with SMW 1.6.0 three years ago at end of July 2011. Perhaps SMW 2.0 is an opportunity to remove the support entirely together with the `$smwgHistoricTypeNamespace` setting.

JeroenDeDauw:
@mwjames might be a fun thing to kill :)

kghbln:
We also need to add a depreciation notice for this [setting](https://www.semantic-mediawiki.org/wiki/Help:$smwgHistoricTypeNamespace)

kghbln:
> @mwjames might be a fun thing to kill :)

Since this seems to be a namespace day today: Perhaps this is a good chance to get namespace related stuff for 3.0.0 out of the way for good. :)

PR #1693, opened 2016-06-26
Respect $smwgHistoricTypeNamespace (removal of the now-unused Type: n…

Seb35:
…amespace)

This was introduced in dea5dd0d (2011-05-22), but the namespace Type: still appears,
due to the fact it is registered in the hook CanonicalNamespaces.

This commit removes the last place where this namespace is made visible for the end
user: now users can no more access to it, but content already saved is kept and they
can made it visible by adding:

    $wgExtraNamespaces[SMW_NS_TYPE] = 'Type';
    $wgExtraNamespaces[SMW_NS_TYPE_TALK] = 'Type_talk';

### DIFF
--- a/includes/NamespaceManager.php
+++ b/includes/NamespaceManager.php
@@ -64,8 +64,6 @@ class NamespaceManager {
 	 */
 	public static function getCanonicalNames() {
 
-		global $smwgHistoricTypeNamespace;
-
 		$canonicalNames = array(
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
+1

mwjames:
Thanks

mwjames:
refs #399

PR #1694, opened 2016-06-26
Pull up getQueryMode from FileExportPrinters to parent class

s7eph4n:
GetQueryMode has the exact same content for all FileExportPrinters, so I pulled
it up to the parent class.

(This also fixes a bug in the SRF\Excel result format that did not have this
method and produced empty result sets.)


JeroenDeDauw:
Awesome, thanks

Issue #169, opened 2014-02-07
Analyse _NEWP annotation regression


mwjames:
Was found during #164 testing.
### Cause

it should be called `$this->revision->getParentId() === null` and not `$this->revision->getParentId() !== ''` because `Revision` returns

```
     * @return Integer|null
     */
    public function getParentId() {
        return $this->mParentId;
    }
```
## Solution

```
public function isNewPage() {

    if ( $this->isFilePage() ) {
        return isset( $this->auxiliary->reupload ) ? !$this->auxiliary->reupload : false;
    }

    return $this->revision ? $this->revision->getParentId() === null : false;
}
```

As for the `File` object, the `reupload` needs to be used in order to decide whether it is new or not.

PR #1700, opened 2016-06-30
Fix "Cannot redeclare class" when using the tarball install, refs #169…

mwjames:
…9, #1680


SemanticMediaWiki.php
```diff
@@ -24,129 +20,64 @@
...
-	 * @see https://www.mediawiki.org/wiki/Manual:$wgExtensionFunctions
-	 *
-	 * @since  1.9
-	 */
-	public static function onExtensionFunction() {
-
-		// 3.x reverse the order to ensure that smwgMainCacheType is used
-		// as main and smwgCacheType being deprecated with 3.x
-		$GLOBALS['smwgMainCacheType'] = $GLOBALS['smwgCacheType'];
+require_once __DIR__ . "/load.php";
```
JeroenDeDauw: You sure the issue is fixed like this? That file still defines symbols and has side effects

DefaultSettings.php
```diff
@@ -1,26 +1,17 @@
 <?php
 
-namespace SMW;
-
 /**
- * SemanticMediaWiki default settings
+ * DO NOT EDIT!
```
JeroenDeDauw: The rename of this file needs to be reflected in the docs. I'm thinking this is better done after the next release, rather than right before when an RC already happened.

SemanticMediaWiki.php
mwjames: I tested by hand.

DefaultSettings.php
mwjames: Well, I didn't want to touch this at all before 2.4 but now that my hand was forced I did it in one flush.

JeroenDeDauw: Forced how? What does not work if you keep the old name?

PR #1246, opened 2015-11-02
Add support for ~/!~ on single value queries

mwjames:
Allows to query things like `{{#ask: [[~Foo/*]] }}`

PR #1701, opened 2016-07-05
Keep upper/lower case on single values for approximate search, refs #1246

mwjames:
`[[~foo*]]` is not the same as `[[~Foo*]]` therefore ensure that the `Title` doesn't transform the value literal in context of an approximate search.

refs #1246, #1410, #1534


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1701

PR #1704, opened 2016-07-07
Added PHP 7.1 to TravisCI

JeroenDeDauw:


mwjames:
See https://github.com/travis-ci/travis-ci/issues/6148

mwjames:
Maybe `PHPUNIT=4.8.*` (5.* will most likely not work).

JeroenDeDauw:
Ok, so not 7.1 yet :(

I don't understand what you are saying about PHPUnit. Why would we need an older version?

mwjames:
> I don't understand what you are saying about PHPUnit. Why would we need an older version?

I think you have this backwards, by default 3.8.* would be installed and I suggested you use 4.8.* instead since running 5.4.* will break the job.

JeroenDeDauw:
Oh, the default version of PHPUnit does not work with PHP 7?

mwjames:
> Oh, the default version of PHPUnit does not work with PHP 7?

 As far as I recall, the mock objects went berserk the last time I tried PHPUnit 3.8* with SMW/MW/PHP 7.

PR #1705, opened 2016-07-07
Update composer.json

kghbln:
Go www


JeroenDeDauw:
But but but... all those bytes wasted on not needed www!

PR #1706, opened 2016-07-07
PatternConstraint to escape early, refs #1417

mwjames:
refs #1417


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Attribut:NDL_ID
http://sandbox.semantic-mediawiki.org/wiki/Q1339

Issue #1529, opened 2016-04-24
Broken Quantity type after upgrade to 2.3


alex-mashin:
After upgrade to SMW 2.3 under MW 1.26.2, MySQL storage (MariaDB) my currency properties stopped working.

I have several of them, i.e. https://traditio.wiki/%D0%A1%D0%B2%D0%BE%D0%B9%D1%81%D1%82%D0%B2%D0%BE:%D0%92%D1%8B%D1%80%D1%83%D1%87%D0%BA%D0%B0 (in Russian, "Property:Revenue"). Their definitions include the template https://traditio.wiki/%D0%A8%D0%B0%D0%B1%D0%BB%D0%BE%D0%BD:%D0%98%D0%B7%D0%BC%D0%B5%D1%80%D1%8F%D0%B5%D1%82%D1%81%D1%8F_%D0%B4%D0%B5%D0%BD%D1%8C%D0%B3%D0%B0%D0%BC%D0%B8 ("Template:Measured as money"), which defines units of measurement for several currencies (with exchange rates as of January 2016) and also thousands, millions and billions of them.

It worked fine in January 2016, under SMW 2.2 until I decided to upgrade.

Now it doesn't. Pages with semantic annotations with currency properties and units of measurement that I have defined for them now appear to have improper values. Example: https://traditio.wiki/%D0%A2%D1%80%D0%B0%D0%B4%D0%B8%D1%86%D0%B8%D1%8F:%D0%A2%D0%B5%D1%81%D1%822. Units that have 1:1 ratio (руб., ₽) are recognised; others aren't ($, USD, тыс. руб.).

It seems that you have broken the special property "Corresponds to".

I have also noticed that spaces are now missing from units of measurement set with "Display units" property (the same example; "млн" and "руб." should have ". " between them but don't.

mwjames:
> It worked fine in January 2016, under SMW 2.2 until I decided to upgrade

There have been no changes to either Quantity or Unit (\[0]) processing (as far as I remember) between 2.1 and 2.3.1 [0] therefore I cannot validate the claim which rests on the 2.2 vs 2.3 argument. We do have changes that were made in the upcoming 2.4 release but those were not deployed with 2.3*.

> It seems that you have broken the special property "Corresponds to".

We have different regression tests [1, 2, 3, 4, 5] that include different conversion rules in connection with currency and "Corresponds to" and they did not indicate malfunction or deviation from the expected output.

[...]

PR #1718, opened 2016-07-10
Preserve spaces in unit labels, refs #1529

mwjames:
Adds the `SMW_DV_NUMV_USPACE` flag for preserving spaces in unit labels which up to now has been squashed. For example, `sq mi` (official notation) is converted to `sqmi` without the flag.

In order to allow for different spellings a user is expected to maintain a list of permitted notations such as `[[Corresponds to::1 km², km ², k m ²]]` with the first declaration to mark the "standard" output norm.

To enable this feature `smwgDVFeatures` requires the `SMW_DV_NUMV_USPACE` flag (disabled by default to avoid BC as it would require to list all norms explicitly).

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -993,6 +993,9 @@ $GLOBALS['smwgEnabledInTextAnnotationParserStrictMode'] = true;
 # - SMW_DV_TIMEV_CM (TimeValue) to indicate the CalendarModel if is not a
 # CM_GREGORIAN
 #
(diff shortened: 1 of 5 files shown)


mwjames:
## Default output
image

## Output with spaces
image

mwjames:
@kghbln FYI

kghbln:
[Documented](https://www.semantic-mediawiki.org/wiki/Help:Custom_units#Unit_labels)

PR #1719, opened 2016-07-10
Add query profile to #concept

mwjames:
https://github.com/SemanticMediaWiki/SemanticNotifications/pull/10


JeroenDeDauw:
What is a query profile?

mwjames:
> What is a query profile?

Something like:

http://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Browse/Berlin-23_QUERY414f168636444d974d7576148e15cffa

image

JeroenDeDauw:
So a query profile is the meta data we store about it since some versions back?

mwjames:
> So a query profile is the meta data we store about it since some versions back?

Yes.

SemanticMediaWiki/SemanticNotifications#10 requires the information to be available in the `CompositePropertyTableDiffIterator` so that any change to its query can be watched and eventually trigger a notification.

JeroenDeDauw:
+1

Issue #698, opened 2015-01-04
Drop PHP 5.3 and 5.4 support


JeroenDeDauw:
I suggest we drop PHP 5.3 support in the next big release. This version is not getting security fixes anymore since August 2014, so supporting users in continuing to use it is bad.

There was some talk about dropping PHP 5.3 AND 5.4 support for MediaWiki itself (since WMF finally moved away from 5.3 with HHVM). If that happens, we can do that as well.

See also:

* [Why to not support 5.3 and 5.4](http://blog.ircmaxell.com/2014/12/on-php-version-requirements.html)
* [MW version lifecycle](https://www.mediawiki.org/wiki/Version_lifecycle)
* [MW PHP compatibility](https://www.mediawiki.org/wiki/Compatibility#Software_required_to_run_MediaWiki)

mwjames:
I hope we can agree on dropping support for MW 1.19-1.22 as well by the time PHP 5.3 is excluded.

JeroenDeDauw:
Sounds good to me. Presumably by then they are all no longer supported (except maybe 1.19, but then again, there already is a new LTS (1.23) people can use)

kghbln:
Continuing to support MW 1.23 would mean to continue supporting PHP 5.3 until May 2017. However, I would love to see support go for MW 1.19-1.22 too.

JeroenDeDauw:
> Continuing to support MW 1.23 would mean to continue supporting PHP 5.3 until May 2017.

We can support MW 1.23 without supporting PHP 5.3. We've been in this situation before, since we dropped 5.2 support before MediaWiki did. (At least I think we did, not sure now...)

JeroenDeDauw:
I wonder if dropping PHP and MW support should not be in the compat docs somewhere... We can have this tracked all nicely in our issue tracker here, but then basically none of our users are going to see the compat drop coming.

JeroenDeDauw:
I've now added this info by inserting a 3.x row into the compat table: 74299937d7b13f841d4679ec3ee7a1d4ba0dcadf

kghbln:
I believe that June 2015 is too early for SMW 3.x Debian 8 will ship with PHP 5.6 if I am not mistaken but will probably not be release before the third quarter 2015. After that we should probably allow another tree months to allows server admins their act together. So SWM 3.x will be a thing of the first quarter 2016?

mwjames:
[...]

PR #1720, opened 2016-07-10
Drop PHP 5.3 and 5.4 support (TravisCI), refs #698

mwjames:
refs #698

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,35 +3,35 @@ language: php
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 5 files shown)


mwjames:
`composer.json` also requires a change `"php": ">=5.3.0"`

JeroenDeDauw:
Funny how we do this on the day PHP drops support for 5.5. We where SOOOOO close to not supporting any obsolete PHP versions at a given time here :)

Issue #1364, opened 2016-01-12
2.4 release


JeroenDeDauw:
Tracking issue for the 2.3 release.

* [Release notes](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/RELEASE-NOTES.md)
* [Open issues for 2.4](https://github.com/SemanticMediaWiki/SemanticMediaWiki/milestones/SMW%202.4)
* [2.4 issues already closed](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues?q=milestone%3A%22SMW+2.4%22+is%3Aclosed)
* [Changes since 2.3](https://github.com/SemanticMediaWiki/SemanticMediaWiki/compare/2.3.0...master)

See also: [3.0 release](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1365)

--

RC1 target: June 20th

* [x] Finalize rel notes
* [x] Update version number (2.4.0-rc.1) (see https://github.com/dbrock/semver-howto)
 * [x] SemanticMediaWiki.php
 * [x] COMPATIBILITY.md
* [x] Create tag
* [x] Create tarballs
* [x] Run composer update for SMW sandbox
* [x] Announce via mail
* [x] Do a release tweet on Twitter

--

Release target: July 9th

* [x] Finalize rel notes (include contributors (`git shortlog -sn 2.3.0...2.4.0`))
* [x] Update version number  (2.4.0)
 * [x] SemanticMediaWiki.php
 * [x] README.md
 * [x] COMPATIBILITY.md
* [x] Check Special:Version
* [x] Create branch
* [x] Create tag
* [x] Create tarballs
* [x] Update release template
* [x] Run `composer update` for SMW wiki
* [ ] Announce on wiki --> via https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1716
* [x] Update version on wikidata, en, de, mw and wikiapiary
* [x] Touch LocalSettings.php
* [ ] Announce via mail --> via https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1716
* [ ] Do a release tweet on Twitter --> via https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1716

mwjames:
## Preliminary notes

- 2.4 requires to run`update.php` because of changes in #1335, #1369

JeroenDeDauw:
[...]

PR #1722, opened 2016-07-11
Update build_tarballs.sh

kghbln:
Cater for GitHub's file nameing conventions. Refs https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1364#issuecomment-231562343 f.


kghbln:
@JeroenDeDauw FYI and action

PR #1731, opened 2016-07-12
SkinAfterContent avoid error on null

mwjames:
Uncovered by https://github.com/Wikia/app/pull/10803/commits/5b0d34da602d0f2e49be88c1e35549461b49ddaa.

### DIFF
--- a/src/MediaWiki/Hooks/SkinAfterContent.php
+++ b/src/MediaWiki/Hooks/SkinAfterContent.php
@@ -50,6 +50,10 @@ class SkinAfterContent {
 
 	private function canPerformUpdate() {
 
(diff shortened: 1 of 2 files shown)


src/MediaWiki/Hooks/SkinAfterContent.php
```diff
@@ -50,6 +50,10 @@ public function process() {
 
 	private function canPerformUpdate() {
 
+		if ( !$this->skin instanceof Skin ) {
```
JeroenDeDauw: `! ( $this->skin instanceof Skin )`

I think that form is nicer, since otherwise the ! and the first part look like they are grouped.

PR #1732, opened 2016-07-12
Add support for extension.json, refs #1367

mwjames:
It is expected that:

- Composer + `extension.json` is working in different environments and does not cause issues like https://phabricator.wikimedia.org/T139853 (`wgExtensionDirectory`)
- The tarball installation is working as well
- Local vendor installation is working
- https://phabricator.wikimedia.org/T151136 (Using extension.json and callback)
- https://phabricator.wikimedia.org/T152263 (Loading dependency from within another extension)

## Expected

`extension.json` should only be introduced by the time:

- A major SMW release change (e.g. 3.x) occurs
- `requires` section in extension.json is supported for extensions
- MW 1.27 to be a minimum requirement

## Requires

- to update installation documentation
- to update depending extension and CI environments

## Usage

```
enableSemantics( 'example.org' );
```
Because `wfLoadExtension` only processes extension definition after the loading of the `LocalSettings.php` file, definitions such as `SMW_NS_PROPERTY` are not yet available by the time the `LocalSettings.php` makes a reference (as it depends on `GLOBALS['smwgNamespaceIndex']` setting).

You need add `enableSemantics( 'example.org' );` (as it contains a shortcut to have `SMW_NS_PROPERTY` defined as soon as it is included) to avoid the `Use of undefined constant SMW_NS_PROPERTY - assumed 'SMW_NS_PROPERTY' in ...LocalSettings.php` notice.

Since we require more control over the NS setup [0], `enableSemantics` is responsible to load `wfLoadExtension( 'SemanticMediaWiki' );` leaving all setups as-is but it also means users should __not__ add `wfLoadExtension( 'SemanticMediaWiki' );` themselves and instead make sure that `enableSemantics` is added or removed.

[0] https://phabricator.wikimedia.org/T152582


mwjames:
If users are really interested to get this working then it is expected that:

- Composer + `extension.json` is working in different environment alongside and does not cause issues like https://phabricator.wikimedia.org/T139853 (`wgExtensionDirectory`)
- The tarball installation is working as well
- Local vendor installation is working

## Expected

`extension.json` should only be introduced by the time:

- A major SMW release change (e.g. 3.x) occurs
- `requires` section in extension.json is supported for extensions
- MW 1.27 to be a minimum requirement

## Requires

- to update installation documentation
- to update depending extension and CI environments

## Notes

If you need to use `SMW_NS_PROPERTY` as part of the customising, it is required to add:

```
wfLoadExtension( 'SemanticMediaWiki' );
enableSemantics( 'example.org' );
```

Because `wfLoadExtension` only processes extension definition after the loading of the `LocalSettings.php` file, definitions such as `SMW_NS_PROPERTY` are not yet available by the time the `LocalSettings.php` makes a reference (as it depends on `GLOBALS['smwgNamespaceIndex']` setting).

You need add `enableSemantics( 'example.org' );` (as it contains a shortcut to have `SMW_NS_PROPERTY` defined as soon as it is included) to avoid the `Use of undefined constant SMW_NS_PROPERTY - assumed 'SMW_NS_PROPERTY' in ...LocalSettings.php` notice.

Nikerabbit:
How would even enableSemantics be available before the extension is loaded?

mwjames:
This is the follow-up to #1680 and my final involvement on the #1367 issue. I will only agree on a merge after it is clarified:

- How dependent extension are to use the required section of the `extension.json` and it is ensured that extensions load `wfLoadExtension( 'SemanticMediaWiki')` on its own as part of the dependency 
- Extensive testing by users to avoid issues like https://phabricator.wikimedia.org/T139853
- All tests pass
- On a major SMW release
- MW 1.27 is to be a minimum requirement

PS: I will not except (and it is meant in its strong connotation) that conceptional issues are to be blamed on using Composer where in fact it is due to design issues in  `ExtensionRegistry`.

extension.json
```diff
@@ -0,0 +1,31 @@
+{
+	"name": "SemanticMediaWiki",
+	"version": "3.0.0-alpha",
+	"author": [
+        "[https://www.mediawiki.org/wiki/User:Jeroen_De_Dauw Jeroen De Dauw]",
+        "James Hong Kong",
+        "[http://korrekt.org Markus Krötzsch]",
```
JeroenDeDauw: I am more happy with the old ordering. Markus definitely did more than me so I prefer to be listed after him.

mwjames:
> How would even enableSemantics be available before the extension is loaded?

Autoloading is done by Composer with `GlobalFunctions.php`  containing `enableSemantics` and is therefore available outside of the `wfLoadExtension` invocation.

extension.json
mwjames: Feel free to change the order, I just wanted to get this out of my git box so others can play with it.

mwjames:
```
There were 2 failures:
1) SMW\Tests\System\InstallationGlobalsProviderIntegrityTest::testNamespacesWithSemanticLinksOnTravisCustomNamespace with data set #0 ('GLOBALS', array(true, false, true, false, true, false, true, false, true, false, true, false, true, false, false, false, false, false, true, false, true, false))
Asserts that smwgNamespacesWithSemanticLinks retrieved from GLOBALS contains the expected 998 NS
Failed asserting that false is true.
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/System/InstallationConfigurationIntegrityTest.php:97

2) SMW\Tests\System\InstallationGlobalsProviderIntegrityTest::testNamespacesWithSemanticLinksOnTravisCustomNamespace with data set #1 ('Settings', array(true, false, true, false, true, false, true, false, true, false, true, false, true, false, false, false, false, false, true, false, true, false))
Asserts that smwgNamespacesWithSemanticLinks retrieved from Settings contains the expected 998 NS
Failed asserting that false is true.
```

mwjames:
To avoid any further misunderstandings from people resulting in comments like "... surely they're doing it wrong then? ..." [0] about the support of `extension.json`, I explained the requirements in length [1] on what the project requires so that `extension.json` support can become a reality.

This PR has been open for a year, and it will not be merged if the stated requirements aren't met to avoid jeopardizing the software or its stability.

[0] https://phabricator.wikimedia.org/T61872
[1] https://phabricator.wikimedia.org/T61872#3681772

kghbln:
If I understood correctly [T173799](https://phabricator.wikimedia.org/T173799) was a major blocker for this too. This will now ship with MW 1.31+

mwjames:
I decided to merge this PR despite unresolved issues in order to tick it off my agenda and no longer occupies an argumentative position from certain user groups that push for its support.

@kghbln For the end-user nothing will change since `enableSemantics` remains the function that needs to be added via `LocalSettings.php` to ensure that:

- `$smwgNamespaceIndex` is usable to modify the NS index before they are defined
- Define NS numbers before any reference happens in `LocalSettings.php` 
- Defines the Export URL

## What about?

```
// Meaningless because enableSemantics calls wfLoadExtension
wfLoadExtension( 'SemanticMediaWiki' );
enableSemantics( 'http://example.org/id/', true );
```
```
enableSemantics( 'http://example.org/id/', true );
// Meaningless because enableSemantics calls wfLoadExtension
wfLoadExtension( 'SemanticMediaWiki' );
```
```
wfLoadExtension( 'SemanticMediaWiki' );
$smwgNamespacesWithSemanticLinks[SMW_NS_PROPERTY] = true;

// no enableSemantics 
// Will cause "Use of undefined constant SMW_NS_PROPERTY - assumed 'SMW_NS_PROPERTY' in ...LocalSettings.php"
```

kghbln:
I really like this approach. a) existing installation do not need to touch "LocalSettings.php" and b) the configuration parameters still sit in a dedicated file. I have seen extension registration files appearing much beyond any decent readability while trying to put everything in "extension.json".

mwjames:
@s7eph4n @kghbln @oetterer FYI

One drawback (among others) is that those extension that require SMW now need to load SMW explicitly as part of the test setup (see https://github.com/SemanticMediaWiki/SemanticCite/commit/4d3b2bc43dc060e6bf8488ae73bdb285d6f2e74b and https://github.com/SemanticMediaWiki/SemanticResultFormats/commit/829bb2a444402988ca880443827bb0109058f7e6).

kghbln:
> One drawback (among others) is that those extension that require SMW now need to load SMW explicitly as part of the test setup

@mwjames I could add this to the other extensions?!

mwjames:
> @mwjames I could add this to the other extensions?!

That would be nice.

cicalese:
Is this a solution for the problem of Semantic MediaWiki being automatically enabled? In the absence of enableSemantics, would Semantic MediaWiki classes still be autoloaded? This would indeed be a welcome new feature if it would prevent any Semantic MediaWiki code from being autoloaded or executed if enableSemantics() is not present.

kghbln:
As I have understood it:

> Is this a solution for the problem of Semantic MediaWiki being automatically enabled?

Yes, and moreover it introduces the extension registration system of MediaWiki without the need to adapt your "LocalSettings.php" if one is upgrading.

> In the absence of enableSemantics, would Semantic MediaWiki classes still be autoloaded?

No, in this case the code would sit on the server but SMW will be unavailable for the respective wiki.

cicalese:
Excellent!! Thank you @mwjames for adding this functionality! Thank you @kghbln for your response!

Will the other Semantic* extensions be following suit? Unfortunately, since they do not already have a function akin to enableSemantics included in LocalSettings.php, changing them will require a change to LocalSettings.php to add wfLoadExtension.

kghbln:
> Will the other Semantic* extensions be following suit?

I believe so. Some already do, however only SESP, SG and SR as part of the most popular ones. You can always look at the [sandbox setup](https://sandbox.semantic-mediawiki.org/wiki/SMW_Sandbox:Setup_information#Semantic_extensions) to see which extensions already use extension registration, at least on master.

> Unfortunately, since they do not already have a function akin to enableSemantics included in LocalSettings.php, changing them will require a change to LocalSettings.php to add wfLoadExtension.

This is indeed true.

mwjames:
> Already documented.

One last thing I'd like to have confirmed before we declare victory, could we create a tarball release for the current master and see if it still works. I hate to be surprised by any issue late in the release process.

kghbln:
> One last thing I'd like to have confirmed before we declare victory, could we create a tarball release for the current master and see if it still works.

Looks good to me: https://gist.github.com/kghbln/382eb01dcf4eef0e564660612077b457

mwjames:
> Looks good to me: https://gist.github.com/kghbln/382eb01dcf4eef0e564660612077b457

So, when following the tarball installation and activation instructions it still works?

kghbln:
OMG. Extracting to `/extensions/` and adding `enableSemantics( 'example.org' );` results in `PHP Fatal error:  Uncaught Error: Call to undefined function enableSemantics() in /../w/LocalSettings.php` I guess this is what you have to do. Apparently I only have 5 minutes for this today so I may have done something wrong. Hmm, I'd rather drop support for tarballs than reverting.

PR #1735, opened 2016-07-12
JobBase.php: PHP docs for batchInsert method

macbre:


JeroenDeDauw:
Thanks. There is no need to use a fully qualified name here; `JobBase` will do. However given you are inside of this class, using `self` is nicer.

macbre:
@JeroenDeDauw, good point - changed in eee96b3

PR #1740, opened 2016-07-12
SMWSQLStore3::getConnection - improve PHP docs

macbre:
`SMW\MediaWiki\Database` is returned here (with `DatabaseBase`-compatible API)

### DIFF
--- a/includes/storage/SQLStore/SMW_SQLStore3.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3.php
@@ -702,7 +702,7 @@ class SMWSQLStore3 extends SMWStore {
 	 *
 	 * @param string $connectionTypeId
 	 *
-	 * @return mixed
+	 * @return SMW\MediaWiki\Database
 	 */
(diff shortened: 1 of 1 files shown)


includes/storage/SQLStore/SMW_SQLStore3.php
```diff
@@ -702,7 +702,7 @@ public function clear() {
 	 *
 	 * @param string $connectionTypeId
 	 *
-	 * @return mixed
+	 * @return SMW\MediaWiki\Database
```
JeroenDeDauw: This uses relative path. The doc will break when the class is put in a NS. So it's better to use the FQN or do an import.

JeroenDeDauw:
Thanks again

includes/storage/SQLStore/SMW_SQLStore3.php
macbre: Done in 9213859

Issue #1511, opened 2016-04-09
Remove I18n shim for php-files


kghbln:
This is possible as soon as support for < MW 1.23 is dropped, currently planned for SMW 2.5.0

PR #1742, opened 2016-07-13
Remove I18n shim

kghbln:
* Originally required to support MediaWiki 1.17.x to 1.22.x

Refs issue https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1511


Nikerabbit:
Yay +1

JeroenDeDauw:
+1

load.php
```diff
@@ -70,7 +70,6 @@ public static function initExtension() {
 
 		// Because of MW 1.19 we need to register message files here
 		$GLOBALS['wgMessagesDirs']['SemanticMediaWiki'] = $GLOBALS['smwgIP'] . 'i18n';
-		$GLOBALS['wgExtensionMessagesFiles']['SemanticMediaWiki'] = $GLOBALS['smwgIP'] . 'languages/SMW_Messages.php';
```
mwjames:
```
There was 1 failure:
1) SMW\Tests\SetupTest::testRegisterMessageFiles with data set #0 ('SemanticMediaWiki', array(array('')))
Failed asserting that false is true.
```

This line is checked in [0], so removing `SemanticMediaWiki` from [0] should make the test pass.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/includes/SetupTest.php#L279

kghbln: Done! Makes sense. :) Thanks.

PR #1745, opened 2016-07-26
Translation of special pages’ names in French

Seb35:
Most follow what is already in the French translation of the title of these pages (5/10). I tried to be as short and understandable as possible for others, as in English.

@NicolasNallet: what do you think about these translations?


NicolasNallet:
it seems good

kghbln:
Thanks for providing the translations!

Issue #1748, opened 2016-08-01
SMW disables itself when Cargo is installed


JeroenDeDauw:
This behaviour is present since SMW 2.4 and was intentionally added. My view is that this is an anti-feature that is not in the interest of most users and should be removed.

See also these threads on the mailing list:

* [Semantic MediaWiki and Cargo/WikiBase](https://sourceforge.net/p/semediawiki/mailman/semediawiki-user/thread/1466658604630-5064994.post%40n6.nabble.com/#msg35177423)
* [all smw functionality missing after upgrade to SNPedia.com](https://sourceforge.net/p/semediawiki/mailman/semediawiki-user/thread/CAGdMhTa8v-n-GBMkSq5iCU%2BaX5Go2r%2B%3D_1bLAmK%2BtBjMT%2BGmHg%40mail.gmail.com/#msg35253456)

JeroenDeDauw:
If you are running into this, untill it is fixed, you can get around it by setting this in your LocalSettings.php:

```php
$GLOBALS['smwgEnabledCompatibilityMode'] = true;
```

mwjames:
All I have to say in this topic, I did so with [0].

If a users wishes to run the software then s(he) has to make the appropriate settings __BUT__ it will no longer be expected to work specified or tested (which nullifies any possible volunteer support time).

[0] http://wikimedia.7.x6.nabble.com/Semantic-MediaWiki-and-Cargo-WikiBase-tp5064994p5065021.html

ckoerner:
@mwjames, I understand your frustration and appreciate all the time you have dedicated to this project. There have been a few long-time SMW folks getting bitten by this change. Not replying to the questions posed by others in the mailing list and repudiating further discussion doesn't seem like the right approach. Please reconsider.

Again, i know this bugs you, but may I make a suggestion? A compromise of sorts? Perhaps instead of silently disabling SMW we should instead give the user a notification of some sort. Something that can't be missed and could alert the user to concerns over compatibility.

This address, in part, your concerns of "Hey, don't bug me when things break with XYZ Extension!" and lets folks know what's going on rather than spending their time digging around trying to figure out what is 'broken'.

PR #1749, opened 2016-08-01
Drop silent disabling of SMW when Cargo or WB are installed

JeroenDeDauw:
Fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1748

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -38,15 +38,6 @@ $GLOBALS['smwgSemanticsEnabled'] = false;
 ##
 
 ###
-# CompatibilityMode is to force SMW to work with other extensions that may impact
-# performance in an unanticipated way or may contain potential incompatibilities.
-#
-# @since 2.4
-##
-$GLOBALS['smwgEnabledCompatibilityMode'] = false;
-##
-
-###
 # Use another storage backend for Semantic MediaWiki. The default is suitable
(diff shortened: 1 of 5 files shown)


mwjames:
Do you think I will +1 this? I have given my reason on this topic on mailing list as to why this was introduced.

PS: Forcing me to comment right after my vacation, isn't really appreciated.

JeroenDeDauw:
> Do you think I will +1 this? I have given my reason on this topic on mailing list as to why this was introduced.

You have indeed outlined your reasoning. Thanks for that. Since I outlined my reasoning afterwards, and neither of us have changed position, it is not reasonable to expect that discussion involving the same argumentation will be constructive.

If we accept that there is strong disagreement about having this functionality between main stakeholders, then I hope we can agree it's not a path we want to go down and retain the state without the controversial change. If I want to make a change that I think is real cool, with which you disagree strongly, then I'll certainly respect that.

While I'm not expecting anything in particular, I definitely hope you agree with doing this revert, even if you disagree the feature itself is a bad idea.

-- 

> PS: Forcing me to comment right after my vacation, isn't really appreciated. 

I am not forcing you to do anything. If you feel compelled to discuss this further, that is your own compulsion. While it might not be your intention to do so, I do not appreciate being accused like this.

PR #1750, opened 2016-08-02
Add RdbmsTableBuilder to replace SMWSQLHelpers

mwjames:
- RDBMS specific code has been moved into corresponding subclasses with `RdbmsTableBuilder` itself implementing the `TableBuilder` interface
- Add unit/integration tests to verify that tables/indicies are created as specified

### DIFF
--- a/includes/storage/SQLStore/SMW_SQLStore3_SetupHandlers.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3_SetupHandlers.php
@@ -2,6 +2,8 @@
 
 use Onoi\MessageReporter\MessageReporter;
 use SMW\CompatibilityMode;
+use SMW\SQLStore\SQLStoreFactory;
+use Onoi\MessageReporter\MessageReporterFactory;
 
 /**
  * @ingroup SMWStore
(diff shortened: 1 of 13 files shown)


mwjames:
@kghbln I tried to be extra observant (and added an extra integration test) to avoid any breakage but this PR can potentially contain a regression during the table setup/update therefore I'm given a prior notice to watch out for issues in this regards. Since I need this for subsequent work I'm unable to wait for someone to test this within a timeframe of a week or longer.

kghbln:
So what I need to do is getting the pr in on sandbox, run the database update and see if there are worries?!

mwjames:
> So what I need to do is getting the pr in on sandbox, run the database update and see if there are worries?!

Since all tests passed, I don't expect any issues. It is just in case that something "funny" occurs during a table update in future then we might have to look at this PR.

mwjames:
I'm just raising awareness that 2.5 contains such change.

kghbln:
Just updated sandbox, ran "update.php" followed by "rebuildData.php" and I was not able to see irregularities nor did I get errors in the log. Seems cool.

PR #1756, opened 2016-08-04
Load Special:Browse content via the API back-end

mwjames:
This PR is made in reference to: N/A

- Display is served from an API request/response in order to increase responsiveness for pages with large fact sets
- `Special:Browse` synchronous display characteristics is still available using `output=legacy` or `$GLOBALS['smwgBrowseByApi'] = false`
- Design issues (html construction etc.) that are prominent in `Special:Browse` (and have moved to the `HtmlContentBuilder`) class has not been addressed as part of this PR and requires a follow-up

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -281,6 +281,12 @@ $GLOBALS['smwgBrowseShowAll'] = true;
 ##
 
 ###
+# Whether the browse display is to be generated using an API request or not.
+##
+$GLOBALS['smwgBrowseByApi'] = true;
+##
+
+###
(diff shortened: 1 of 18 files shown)


mwjames:
As in the outlined use case below (with 1.6 MB in retrievable data) doing a synchronous display (as it is now) will in most cases block the web browser for a length of time and create an unresponsive user interface. 

image

mwjames:
@SemanticMediaWiki/testers Anyone wants to test this before I merge this?

PR #1761, opened 2016-08-07
Add context to newPropertyValueByLabel

mwjames:

- If the property page defines a specific page context then localized annotations (e.g `[[具有类型::文本型]]`) can be correctly identified.
- A page content language defined by `SIL` can directly make use of this

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/dataitems/SMW_DI_Property.php
+++ b/includes/dataitems/SMW_DI_Property.php
@@ -388,7 +388,7 @@ class DIProperty extends SMWDataItem {
 	 *
 	 * @return DIProperty object
 	 */
-	public static function newFromUserLabel( $label, $inverse = false ) {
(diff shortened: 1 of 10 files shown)


mwjames:
@kghbln FYI, I saw some examples [0] where `Special:CreateProperty` generated content like `Cette propriété est de type [[A le type::URL]].` due to the `FR` content language. This PR should allow for those annotation to be recognised correctly while we retain our ability to switch the user language.

The canonical `Has type` will always work (== in any language) but now it is possible to rely on either the site content or page content language.

[0] http://sandbox.semantic-mediawiki.org/w/index.php?title=Attribut:G-Has_logo

kghbln:
@mwjames Great that you saw this. Did not get around reporting this yet. Now it is already tackled. Cool service.

Issue #1621, opened 2016-05-26
Localization and multilingual content


mwjames:
In light of the upcoming 2.4 release and with multilingual wikis becoming (or maybe already are) a hot-button topic, I gave it quick and dirty write-up on how SMW can be used to manage content in different languages. I'll invite interested parties to review and improve [0] the text and/or in case something is unclear to post a question here.

While alternative property labels (those with a language tag) are not yet supported, I think we come a long way from being mono restricted and provide a lean and mean conceptional framework for users to create multilingual content.

[0] https://www.semantic-mediawiki.org/wiki/Localization_and_multilingual_content

refs #594

nemobis:
Is there some part on which you most need/desire comments?

Do I understand correctly that in your opinion making SMW multilingual boils down to making links language-aware?

Issue #1452, opened 2016-03-12
Replace inContentLanguage


mwjames:
I just realized that most errors are displayed in or by using `inContentLanguage` which isn't really what we want, so either remove this or explicitly use `inLanguage( ... )`to ensure that messages are displayed in the user language selected.

http://sandbox.semantic-mediawiki.org/wiki/Issue/1243_%28Make_failed_queries_discoverable%29

Content language being FR, user language is EN but the error is displayed in FR.

mwjames:
@kghbln Maybe should make an analysis where such change makes most sense.

JeroenDeDauw:
What errors is this about? What you are saying makes sense to me for special pages and such things that are not cached. Not for things that are cached, such as the content of pages.

mwjames:
Mostly happened with #1768.

It may that we missed something but those should appear in the `Special:ProcessingErrorList` as not "translatable" when a user switches its language.

PR #1768, opened 2016-08-12
Make display of errors user language dep., refs #1293, #1452, #1621

mwjames:
This PR is made in reference to: #1293, #1452, #1621

-  Error messages are only being resolved when requested by an output
- `ProcessingErrorMsgHandler` and `ErrorMsgTextValue` will handle the specific encoding/decoding so it can be stored context independent 
- Error messages are stored with its identifier (instead of a lang dep. text) in order for its signature to remain stable and independent of a language representation 
- Switching a user language (via ULS) will now (for most DV's) display the message in a user specific context
- Removes `wfMessage` dependencies in `DataValue` objects
- Ties in with the effort of providing a better localized user experience [0]

This PR includes:
- [x] CI build passed

[0] https://www.semantic-mediawiki.org/wiki/Localization_and_multilingual_content


mwjames:
> ... will now (for most DV's) display the message in a user specific context

@kghbln For most of the errors after a newly rebuild run we should see messages that switch with the user language but some of the objects in [0] still contain things like `$this->addError( wfMessage( 'smw_wrong_namespace', $propertyNamespace )->inContentLanguage()->text() );` and need  to be converted to something like `$this->addErrorMsg( array( 'smw_wrong_namespace', $propertyNamespace ) );`

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/master/includes/datavalues

mwjames:
Changing languages will now have errors display in a user language context.

image

image

PS: As for the above [0] example, "FR" is the site content language hence "no" is not recognized as valid value. Once the [page content language](https://www.semantic-mediawiki.org/wiki/Page_content_language) is switched to "EN", the error should disappear.

[0] http://sandbox.semantic-mediawiki.org/wiki/Charles_de_gaulle

Issue #1763, opened 2016-08-09
Add `--quick` option to "rebuildFulltextSearchTable.php"


kghbln:
To be able to skip the five second countdown sequence it will be nice to have an additional ``--quick`` option to the "rebuildFulltextSearchTable.php" script.

mwjames:
Done with #1769.

kghbln:
> Done with #1769.

Awesome and [documented](https://www.semantic-mediawiki.org/wiki/Help:RebuildFulltextSearchTable.php) now!

PR #1769, opened 2016-08-12
FTS rebuilder to add --v/--quick option, refs #1481, #1763

mwjames:
This PR is made in reference to: #1481, #1763

- Adds `--v`/`--quick` to `rebuildFulltextSearchTable.php`

This PR includes:
- [x] CI build passed


mwjames:
`--v` will output things like:

```
Table(s) not used for indexing:

- smw_di_number                     Not a blob or URI table type.
- smw_di_bool                       Not a blob or URI table type.
- smw_di_time                       Not a blob or URI table type.
- smw_di_coords                     Not a blob or URI table type.
- smw_di_wikipage                   Not a blob or URI table type.
- smw_fpt_type                      Fixed property table that belongs to the _TYPE exempted property.
```

PR #1775, opened 2016-08-12
Fix time offset recognition

mwjames:

- Fix time offset recognition where something like `[[Has date::1 Jan 1971 13:45:23-3:30]]` should return
`1971-01-01T16:15:23`. This is/was at least broken since 1.9.2 (where it returns `1 January 1971 13:45:23`).
- Adds additional error msgs to clarify situations where values are unclear, incomprehensible,  or insufficient

This PR includes:
- [x] CI build passed


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1775

## Before

image

Issue #1708, opened 2016-07-07
External identifier type


mwjames:
Add a new type called `External identifier` (`[[Has type::External identifier]]`) to transform a string literal into an external resource uri.

The `External identifier` type will require a `External formatter uri` annotation in order for the property declaration to be valid, so that an input like `[[NDL ID::00564222]]` is referenced as `https://id.ndl.go.jp/auth/ndlna/00564222`. 

Other attributes such as `Allows pattern` or `Has uniqueness constraint` can be applied as well but are not required to form a valid property declaration.

![image](https://cloud.githubusercontent.com/assets/1245473/16670583/34acff48-449b-11e6-9584-5a634e083183.png)

mwjames:
https://www.semantic-mediawiki.org/wiki/Help:Type_External_identifier

kghbln:
@mwjames I just created special property docu for [external identifier](https://www.semantic-mediawiki.org/wiki/Help:Special_property_External_identifier). Could you please set up an usage example for it on sandbox?

mwjames:
>  I just created special property docu for external identifier. Could you please set up an usage example for it on sandbox?

You mean?

- https://sandbox.semantic-mediawiki.org/wiki/Attribut:EC_number
- https://sandbox.semantic-mediawiki.org/wiki/Attribut:P3520
- https://sandbox.semantic-mediawiki.org/wiki/Attribut:NDL_ID

kghbln:
> You mean?

I was wondering how the special property comes into play: https://sandbox.semantic-mediawiki.org/wiki/Attribut:External_identifier

mwjames:
> I was wondering how the special property comes into play: https://sandbox.semantic-mediawiki.org/wiki/Attribut:External_identifier

Actually there is no use case for this property alone since you always need a `External formatter URI`. I mostly added them so the there are reserved and cannot be occupied by a user defined property and hereby create confusion about the usage.

kghbln:
Thanks for the insight. Documented on wiki!

PR #1782, opened 2016-08-15
Update SMW_LanguageDe.php

kghbln:
This PR is made in reference to:
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1708
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1721

- Provide missing German translations for datatype names and special property names

This PR includes:
- [x] Tests (unit/integration) - already available
- [x] CI build passed

### DIFF
--- a/languages/SMW_LanguageDe.php
+++ b/languages/SMW_LanguageDe.php
@@ -49,16 +49,16 @@ class SMWLanguageDe extends SMWLanguage {
 		'_rec' => 'Verbund', // name of the record datatype
 		'_qty' => 'Maß', // name of the quantity datatype
 		'_mlt_rec' => 'Einsprachiger Text',
-		'_eid' => 'External identifier',
+		'_eid' => 'Externe Kennung'
 	);
 
 	protected $m_DatatypeAliases = array(
-		'URI'			=> '_uri',
-		'Ganze Zahl'		=> '_num',
-		'Dezimalzahl'		=> '_num',
-		'Aufzählung'		=> '_txt',
-		'Zeichenkette'		=> '_txt',
-		'Menge' 		=> '_qty',
+		'URI' => '_uri',
+		'Ganze Zahl' => '_num',
+		'Dezimalzahl' => '_num',
+		'Aufzählung' => '_txt',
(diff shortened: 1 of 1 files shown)


languages/SMW_LanguageDe.php
```diff
@@ -96,8 +96,8 @@ class SMWLanguageDe extends SMWLanguage {
 		'_PVAP' => 'Erlaubt Muster',
 		'_DTITLE' => 'Anzeigetitel',
 		'_PVUC' => 'Erlaubt Wert einmal',
-		'_PEID' => 'External identifier',
-		'_PEFU' => 'External formatter uri',
+		'_PEID' => 'Kennung zur externen URI',
```
kghbln: @mwjames I am assuming here that this special property will always be used for exteral identifiers in connection with constructing an URI, so no other external ...

Yes, I am deliberately translating the special property and datatype differently. So no worries.

mwjames: > special property will always be used for exteral identifiers in connection with constructing an URI

Yes

`_PEID` is just reserved, the most important part is the type `_eid` which controls the behaviour and allows the output [0, 1] to be exported as:

```
<property:NDL_ID rdf:datatype="http://www.w3.org/2001/XMLSchema#string">00432003</property:NDL_ID>
<skos:exactMatch rdf:resource="https://id.ndl.go.jp/auth/ndlna/00432003"/>
``` 
Property specification follows:

```
* Has type: [[Has type::External identifier]]
* External formatter uri: [[External formatter uri::https://id.ndl.go.jp/auth/ndlna/$1]]
```


If you add `[[External formatter uri::...]]` to another property specification that is not of type `External identifier` then it just gets ignored,


[0] http://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Parcourir/Q1339
[1] http://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial:Export_RDF/Q1339&syntax=rdf

kghbln: Great, thank you for your insight and confirming my understanding of this enhancement.

PR #1784, opened 2016-08-17
Don't call assert() with a string

jdforrester:
Moved upstream from https://gerrit.wikimedia.org/r/#/c/304772/


jdforrester:
cc @tstarling who wrote this.

JeroenDeDauw:
Thanks!

PR #1786, opened 2016-08-17
Extend testcases to include not like

kghbln:
This PR is made in reference to: #1129 

- test case for not like telephone number
- test case for not like e-mail address

This PR includes:
- [x] CI build passed

### DIFF
--- a/tests/phpunit/Integration/ByJsonScript/Fixtures/q-0702.json
+++ b/tests/phpunit/Integration/ByJsonScript/Fixtures/q-0702.json
@@ -30,7 +30,7 @@
 	],
 	"query-testcases": [
 		{
-			"about": "#0",
+			"about": "#0 search phone number +1-2012-555-5555",
 			"condition": "[[Has telephone number::+1-2012-555-5555]]",
 			"printouts" : [ "Has telephone number" ],
 			"parameters" : {
@@ -85,7 +85,7 @@
 			}
 		},
 		{
-			"about": "#3",
+			"about": "#3 search e-mail address Lorem@ipsum.org",
 			"condition": "[[Has email::Lorem@ipsum.org]]",
 			"printouts" : [],
 			"parameters" : {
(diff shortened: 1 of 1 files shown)


tests/phpunit/Integration/ByJsonScript/Fixtures/q-0702.json
```diff
@@ -117,6 +117,40 @@
...
+			"about": "#5 search phone number that does not contain 123",
+			"condition": "[[Has telephone number::!~*123*]]",
+			"printouts" : [],
+			"parameters" : {
+				"limit" : "10"
+			},
+			"queryresult": {
+				"count": "1",
+				"results": [
+					"Page/07/02/2#0##"
```
kghbln: was not sure about the `#0##` particle so I just added it too.

kghbln:
My first fiddling with tests. Let's see.

tests/phpunit/Integration/ByJsonScript/Fixtures/q-0702.json
mwjames: Very nice.

For some background, using the particular `query-testcases` is accessing query methods directly as provided by the system (== `Store::getQueryResult`) hence the results resemble the internal representation more closely. As for the `results` field those are fields you normally see as linked "main" subject in a table and is serialised as `Name#NamespaceId#Interwiki#SubobjetctName`.

On the opposite, the `parser-testcases` uses the output generated by MW after a page was saved with all its internal/external formatting which allows us to compare the `final` output (normally what we expect an end-user to see) and is the reason why for most of those cases we assert HTML strings against `expected-output:to-contain`/`expected-output:not-contain` expectations.

In short, `query-testcases` is more about the detail of a query and its expect result while  `parser-testcases` is about the formatting that is applied to it.

kghbln: Great, thank you for your insight. I have added the "wikidocu missing" label so I do not forget to do this on wiki for a wider audience to know this.

PR #1793, opened 2016-08-18
TimeValue to support #LOCL@... as output format, refs #1545

mwjames:
This PR is made in reference to: #1545

- Support language specific output allowing something like `#LOCL@es` to output a localized date for the annotated language instead of the user language.

This PR includes:
- [x] CI build passed


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1545_%28LOCL%29

PR #1796, opened 2016-08-18
Allow Special:Import to have early access to property specifications

mwjames:

- Allow `Special:Import` to have early access to property specifications

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/MediaWiki/Hooks/LinksUpdateConstructed.php
+++ b/src/MediaWiki/Hooks/LinksUpdateConstructed.php
@@ -71,6 +71,11 @@ class LinksUpdateConstructed {
 			$this->updateEmptySemanticData( $parserData, $title );
 		}
 
+		// Push updates on properties directly without delay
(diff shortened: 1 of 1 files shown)


mwjames:
SMW generally pushes changes to its StorageEngine in form of deferred updates [0] to avoid lock contention or race conditions. Unfortunately, when importing pages that comes at price that specifications of a property are only available after the `DeferredUpdates` has been processed which can be rather late for "normal" imported pages that rely on those specifications (mostly type information) since the `WikiImporter` posts pages in bulk without delaying transactions. 

The above observation is made in connection for when properties are imported first before any "subject" matter. For property specifications that are imported after a subject it should be clear that those specs can not be available by the time a subject makes a reference to it.

Since the `Import` class or its hooks doesn't identify itself as being "imported" during the process, we need to circumvent the situation in case a page is a property page (`SMW_NS_PROPERTY`).

[0] https://www.semantic-mediawiki.org/wiki/Deferred_updates

PR #1806, opened 2016-08-24
Use prototype instead of ES6 class construct, refs #1756

mwjames:
This PR is made in reference to: #1756

- Instead of ES6 `class` syntax use the prototype constructor to avoid older browsers being excluded
- Show an error message in case JS is disabled

This PR includes:
- [x] CI build passed


mwjames:
image

Issue #1808, opened 2016-08-24
Untested claim, factual claim, and the recoding of provenance data


mwjames:
## Synopsis

For most applications or users that use Semantic MediaWiki, making untested claims is the desired (or convenient) method to record data or data snippets, yet some users require factual claims to be recorded.

This issue is about adding metadata to a value declaration in an intuitive (meaning using the common and existing Semantic MediaWiki edit patterns) and extensible way. 

![reference](https://cloud.githubusercontent.com/assets/1245473/17970417/d87586fc-6ad6-11e6-934d-3fb5ad70c52d.gif)

## Preface

For clarification, the term untested claim used in the issue is understood as "incomplete claim for which evidence is yet unavailable" [0]. For example, when a user records a statement as in "John Doe's car is green", a reader of said statement has to trust (or believe) that this statement (with the claim John's car is green) is true without evidence or a possibility to verify it. Whether such statement corresponds with that of the real world or not is of secondary nature or in some cases is even undesirable.

Under the open world assumption [1] it is expected that a statement is true and represents a fact (instead of just being an opinion; what if the observer who made the claim has a deuteranomaly [2] condition) and thus is to be true without evidence or a possibility to test the claim against a reciprocal entity.

Statements denoting authority data such as "Berlin has a population of 3,520,061" [3] with a claim of "population of 3,520,061" being untested may represent challenges for some users (#985) and it is the objective of this issue to exhibit a nominal approach (opposed to what Wikidata provides but yet coherent with the existing Semantic MediaWiki data model) that would allow verification of individual claims.

If we understand a factual claim as "a statement that is supported by convergent evidence" then we require from a statement like "Berlin has a population of 3,520,061" additional information  [4] about the when, where, and by whom a particular claim (population being 3,520,061) was made. Recording provenance metadata and tracing the source [5, 6] of said claim will help transform an untested into a factual claim by allowing a verification of such claim.

## Technical concept and realization 

Technically there are different approaches (which I omit to describe and only selected the least invasive) that can be used for recording "provenance" data on an individual value level to support "empirical or analytical" statements in form of a reference to a claim.

Isolating a value (object) "3,520,061" from a subject (as we require it to be represented as single identifiable entity) is the first necessary step for a technical realization therefore we require a new `DataType` which is called `Reference`. `Reference` allows to define the required or expected provenance data fields that may be entered while annotating a property value.

[...]

PR #1809, opened 2016-08-24
Record type to support property name as index, refs #1808

mwjames:
This PR is made in reference to: #1808

- Besides supporting an index number as in `|?Foo=Bar|+index=1`, enable to use the property name `|?Foo=Bar|+index=Bar` as well.

This PR includes:
- [x] CI build passed


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1809
https://www.semantic-mediawiki.org/wiki/Help:Type_Record#Semantic_search

@kghbln FYI

PR #1812, opened 2016-08-24
Add ReferenceValue, refs #1808

mwjames:
This PR is made in reference to: #1808

- `ReferenceValue` and `ReferenceValueFormatter`
- `smwgDataTypePropertyExemptionList` to exclude listed `DataType`'s from being registered as property. In case of `Reference` the use as property to match the `DataType` semantics is of no help since the definition requires `Has fields` and it is expected that the `Reference` name to be a common property name hence we refrain from using it unless in a `DataType` context.

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -535,6 +535,20 @@ $GLOBALS['smwgPageSpecialProperties'] = array( '_MDAT' );
 $GLOBALS['smwgDeclarationProperties'] = array( '_PVAL', '_LIST', '_PVAP', '_PVUC' );
 ##
 
+###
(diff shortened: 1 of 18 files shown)


mwjames:
@kghbln Would you mind doing a quick sandbox update in order for me to create an example since I'm running out of support time for the coming week.

kghbln:
> Would you mind doing a quick sandbox update

Done!

kghbln:
Documented [datatype](https://www.semantic-mediawiki.org/wiki/Help:Type_Reference) and [configuration setting](https://www.semantic-mediawiki.org/wiki/Help:$smwgDataTypePropertyExemptionList)

PR #1821, opened 2016-09-06
add option to set action in api js

PeterTheOne:
This makes it possible to not only send request to `aks` api but also to the `compoundquery` api.

This PR includes:
- [ ] CI build passed

### DIFF
--- a/res/smw/api/ext.smw.api.js
+++ b/res/smw/api/ext.smw.api.js
@@ -98,9 +98,10 @@
 		 *
 		 * @return {jQuery.Promise}
 		 */
-		fetch: function( queryString, useCache ){
+		fetch: function( queryString, useCache, action ){
 			var self = this,
-				apiDeferred = $.Deferred();
+				apiDeferred = $.Deferred(),
+				action = action || 'ask';
 
 			if ( !queryString || typeof queryString !== 'string' ) {
 				throw new Error( 'Invalid query string: ' + queryString );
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
I don't think this is a good idea design wise, even if the compound query stuff was something part of SMW itself.

What are you trying to achieve with this change?

PeterTheOne:
In SemanticMaps I switch between `ask` and `compoundquery` depending on the query/situation and I want to use the caching functionality of `api.js`. I can only use it for `ask` now.

JeroenDeDauw:
This is client side caching (ie with localstorage) you are talking about right?

Is it important that you use the same cache as something else using this JS code that could end up running the same queries? I suspect not. And if not, then I'd not use (and bind to) the same mechanism/state. What I'd expect is to have some kind of caching decorator, or perhaps the code directly interacting with some caching class (which might or might not share its pool, depending on what you decide when constructing the object graph).

mwjames:
I don't really fancy a dilution of the interface. That being said `localstorage` is quite old and nowadays I would rely on indexedDB instead as it is done in [0] or [1] provided by [2].

[0] https://github.com/SemanticMediaWiki/SemanticCite/blob/master/res/scite.tooltip.js#L15-L19
[1] https://github.com/SemanticMediaWiki/WhatsNearby/blob/master/res/ext.whats.nearby.js#L494-L499
[2] https://github.com/onoi/shared-resources/blob/master/res/onoi.blobstore.js

kghbln:
This pull has not seen any updates in more than two years. Thus closing to make room for another effort. In the end the code changes are still here to pick them up.

PR #1822, opened 2016-09-06
add js geoitem geo

PeterTheOne:
Adds a js geoitem geo which I need for https://github.com/SemanticMediaWiki/SemanticMaps/ . This is still only a rough implementation (copy from dataitem.text.js).

This PR includes:
- [ ] CI build passed

### DIFF
--- a/res/smw/data/ext.smw.dataItem.geo.js
+++ b/res/smw/data/ext.smw.dataItem.geo.js
@@ -1,0 +1,88 @@
+/**
+ * SMW Text DataItem JavaScript representation
+ *
+ * @see  SMW\DIGeoCoord
+ *
+ * Implementation of dataitems that are geographic coordinates.
+ *
+ * @file
+ * @ingroup SMW
+ *
+ * @licence GNU GPL v2 or later
+ * @author Peter Grassberger < petertheone@gmail.com >
+ */
+( function( $, mw, smw ) {
+	'use strict';
+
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Thanks

Looks reasonable to me, though then I don't know this code. Leaving further review to @mwjames

res/smw/data/ext.smw.dataItem.geo.js
```diff
@@ -0,0 +1,88 @@
...
+		},
+
+		/**
+		 * Returns a plain geo representation
+		 *
+		 * @return {string}
+		 */
+		getGeo: function() {
+			return this.geo;
+		}
```
mwjames: What is "plain geo representation" mean? Are those or coordinates as string, does `this.geo` return an object of lat, long or is it more something like `52° 30' 59", 13° 24' 0"`?

The DI [0] returns  an array of `array( 'lat' => $this->latitude, 'lon' => $this->longitude )` so I'd expect something similar.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/dataitems/SMW_DI_GeoCoord.php#L96-L103

mwjames:
There is a lingering issue that got not addressed but I'm fine merging it since it covers the rudimentary use case of `geo`. Any issues with the formatting of the output should be addressed in a follow-up.

PR #1824, opened 2016-09-08
Add PropertyChainValue

mwjames:
This PR is made in reference to: # N/A (orthogonal refs #1812, #1819)

- Support for the property chain notation in printrequests to provide the means to request page nodes of the next chain member (if the returned data type is not a page type then no further iterations are executed as the match process only takes into account `_wpg` edges)
- Elevated by the introduction of #1812 to allow for a more convenient access to provenance values using the `SomeProperty.AnotherProperty` notation in printouts

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/datavalues/SMW_DV_Property.php
+++ b/includes/datavalues/SMW_DV_Property.php
@@ -235,6 +235,7 @@ class SMWPropertyValue extends SMWDataValue {
 			$this->m_wikipage = \SMW\DataValueFactory::getInstance()->newDataValueByItem( $diWikiPage, null, $this->m_caption );
 			$this->m_wikipage->setOutputFormat( $this->m_outformat );
 			$this->m_wikipage->setLinkAttributes( $this->linkAttributes );
+			$this->m_wikipage->setOptions( $this->getOptions() );
 			$this->addError( $this->m_wikipage->getErrors() );
 		} else { // should rarely happen ($value is only changed if the input $value really was a label for a predefined prop)
 			$this->m_wikipage = null;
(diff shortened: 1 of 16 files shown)


mwjames:
I would not recommend a query like the example below but the integration test contains it as "advanced" case where various properties are chained (including inverted) to return the expected printout.

Using the property chain in the condition and using it on the printout are two different distinct processes where matching a condition is about resolving and finding entities (either as row or edge) that adheres certain codified rules while the printout chain matching on the other hand is about creating a property path along the entity that was returned as part of a condition or rule match.   

```
{{#ask: [[Category:Country]] [[Category:P0434]]
 |?Located in.-Located in.-Located in.Capital of.Has subobject.Has population.Number
 |link=none
}}
```

## Features and limitations
- `PropertyChainValue` is added as internal `DV` to manage the chain, its last member, and the output manipulation
- The amount of chain members is currently not limited to a certain depth
- Properties identified as being part of a property chain are marked with `⠉` which shows all its members when being hovered over

image

image

mwjames:
@kghbln I fear that adding more information to [0] only increase the chance of users being lost instead of being a platform to create a sufficient understanding on how those options play together. 

[0] https://www.semantic-mediawiki.org/wiki/Help:Displaying_information

kghbln:
> @kghbln I fear that adding more information to [0] 

I realized this also during my summer documentation offensive where I advanced slowly into undocumented grounds. All the info there should be split similarly to what was done for the `#set` parser function.

mwjames:
Asking people to test this feature prior its merge won't create much response (I'm relying on past experience) therefore I'd say use the sandbox and give it a test drive.

kghbln:
[Now documented.](https://www.semantic-mediawiki.org/wiki/Property_chains_and_paths)

PR #1836, opened 2016-09-13
QueryPrinterFactory cleanup

JeroenDeDauw:
In preparation for making it support proper dependency injection in query printers

### DIFF
--- a/includes/QueryPrinterFactory.php
+++ b/includes/QueryPrinterFactory.php
@@ -7,21 +7,21 @@ use MWException;
 /**
  * Factory for "result formats", ie classes implementing QueryResultPrinter.
  *
- * @since 1.9
+ * @since 2.5 (since 1.9, renamed in 2.5)
  *
  * @ingroup SMW
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-final class FormatFactory {
+final class QueryPrinterFactory {
 
 	/**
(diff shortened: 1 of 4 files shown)


includes/QueryPrinterFactory.php
```diff
@@ -7,21 +7,21 @@
...
- * @since 1.9
+ * @since 2.5 (since 1.9, renamed in 2.5)
  *
  * @ingroup SMW
  *
  * @licence GNU GPL v2+
  * @author Jeroen De Dauw < jeroendedauw@gmail.com >
  */
-final class FormatFactory {
+final class QueryPrinterFactory {
```
mwjames: I think this should go into `SMW\Query\ResultPrinterFactory` (with printers being move to `SMW\Query\ResultPrinters`). I'd like to minmize arbitrary access to namespaced classes from the outside therefore this should be accessible either via `ApplicationFactory::getInstance()->getQueryPrinterFactory` or  `ApplicationFactory::getInstance()->getQueryFactory()->getPrinterFactory`.

JeroenDeDauw: Sounds good to me. Feel free to move it. To me the allowing of dependency injection is higher prio and I won't mix both changes in one PR.

I'm not up to speed with if this factory is being used outside of SMW. Perhaps that is not a good idea to begin with, as you can already register the query printers via the global vars. Or would there be a need for extensions to access instances of the printers?

PR #1733, opened 2016-07-12
Added PHP 7.1 to TravisCI

JeroenDeDauw:

PR #1837, opened 2016-09-13
Added PHP 7.1 to TravisCI

JeroenDeDauw:
Fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1733


mwjames:
As I pointed out in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1733#issuecomment-236476032 there is an issue with MW and 7.1 so running the coverage on 7.1 isn't going to be the best choice.

JeroenDeDauw:
I was hoping the issue to not be present in a more recent version of MW, but alas...

PR #1841, opened 2016-09-14
Revert "QueryPrinterFactory cleanup"

mwjames:
@JeroenDeDauw FYI

Reverts SemanticMediaWiki/SemanticMediaWiki#1836 as it causes a local "Warning</b>:  Class 'SMW\QueryPrinterFactory' not found in <b>...\SemanticMediaWiki\src\Aliases.php</b> on line <b>76</b>"

### DIFF
--- a/includes/FormatFactory.php
+++ b/includes/FormatFactory.php
@@ -7,21 +7,21 @@ use MWException;
 /**
  * Factory for "result formats", ie classes implementing QueryResultPrinter.
  *
- * @since 2.5 (since 1.9, renamed in 2.5)
+ * @since 1.9
  *
  * @ingroup SMW
  *
  * @licence GNU GPL v2+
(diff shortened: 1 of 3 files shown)


mwjames:
I suspect you didn't use a wiki to verify the change.

mwjames:
Probably caused by the `includes` folder being a classmap generated autoloader map instead of a PSR-4 and hereby causing outdated maps to be out of sync on.

JeroenDeDauw:
> I suspect you didn't use a wiki to verify the change.

Indeed not. If I needed to do that, I'd not be able to make any change.

> Probably caused by the includes folder being a classmap generated autoloader map instead of a PSR-4 and hereby causing outdated maps to be out of sync on.

Yeah I think so, the code looks good. How come your autoloader did not get rebuild?

PR #1868, opened 2016-09-24
Correct lowest supported MW version in 2.5.x

JeroenDeDauw:
I think this was a typo or confusion of the version number

### DIFF
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -109,7 +109,7 @@ Note that HHVM is only required if you do not use PHP.
 		<th>SMW 2.5.x</th>
 		<td><strong>5.5.0</strong><sup>[p.1]</sup> - latest</td>
 		<td>3.5.0 - latest</td>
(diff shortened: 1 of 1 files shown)


mwjames:
>I think this was a typo or confusion of the version number

I'm not sure about this but since you are changing the boundaries, you might want to adjust [0].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/.travis.yml

JeroenDeDauw:
It's essentially what we presented on last SMWCon: https://jeroendedauw.github.io/slides/smwcon/2015fall/yearly/#/14/1

1.23 is an LTS, 1.25 is not. I have no idea where 1.25 would be coming from apart from a typo

kghbln:
> 1.23 is an LTS, 1.25 is not. I have no idea where 1.25 would be coming from apart from a typo

I also believe that this branch should continue to support 1.23 since this is an LTS.

JeroenDeDauw:
Travis updated with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1967

Issue #1872, opened 2016-09-25
Retrieve URI longer than 255 chars


mwjames:
### Setup and configuration
- SMW version: all
- MW version: all
- DB (MySQL etc.):
### Issue

Currently the field `o_serialized` that stores the URI literal is defined as 255 CHAR field which means
that any URI longer than that will be truncated and when retrieved from the Store (as in case of a RDF export) it will contain an incomplete representation.
### Example

http://sandbox.semantic-mediawiki.org/wiki/Issue/1872

mwjames:
## Before

![image](https://cloud.githubusercontent.com/assets/1245473/20031144/25686414-a373-11e6-9411-d68d1ecef85f.png)

## After

![image](https://cloud.githubusercontent.com/assets/1245473/20031154/3b0e3226-a373-11e6-8b34-4c7b7c2d1fa3.png)

PR #1878, opened 2016-09-25
Store URI's longer than 255 chars in blob field, refs #1872

mwjames:
This PR is made in reference to: #1872

- Add a `o_blob` field to store URI's > 255 chars and allow retrieving them as intended by a user. The `o_serialized` is left unchanged to ensure that standard searches against this field is continuing to work. The full-text index (if enabled) will allow to search on the complete length.

This PR includes:
- [x] CI build passed


mwjames:
@kghbln The uri table requires a new field therefore after the merge `update.php` is required to run.

mwjames:
@kghbln FYI

kghbln:
Sandbox is now updated to include this pull.

Issue #1865, opened 2016-09-21
Support for preferred property labels


mwjames:
With the introduction of the `MonolingualTextValue` #1344, Semantic MediaWiki is able to provide a method to denote a text element of arbitrary length with a language tag and allowing it to distinguish such text by the same attribute.  

Adding preferred labels to a property specification is aimed at increasing the likelihood for a user to understand the concept of a particular property by defining a label in a language (as side from the canonical label) a user is most familiar with.
## Features and limitations
- Borrowing on the `skos:prefLabel` definition, the preferred label is understood as to only have one label per given language (while `skos:altLabel` can have many alternative labels but this is not part this issue)
- To separate a canonical label from a preferred label during the output, a small <sup>p</sup> hint will be shown along with the label
- The preferred label feature is aimed for the output and not for the annotation process, meaning that a reverse matching during an annotation process is not supported (`[[Has population::123]]` vs `[[人口@ja::123]]`, annotations are expected to use the canonical identifier or label)
- The user language is being used as selective attribute that will trigger a change of labels  (if available) 
## Example

```
Property: Has population

* [[Has preferred property label::population@fr]]
* [[Has preferred property label::población@es]]
* [[Has preferred property label::Bevölkerung@de]]
* [[Has preferred property label::人口@ja]]
```

![image](https://cloud.githubusercontent.com/assets/1245473/18725129/4c88773e-803f-11e6-91e8-84709c50a12c.png)

![image](https://cloud.githubusercontent.com/assets/1245473/18726074/204822b0-8043-11e6-9141-d79cab9031fb.png)

refs #594, #1856, 

[0] https://www.semantic-mediawiki.org/wiki/Localization_and_multilingual_content

mwjames:
https://www.semantic-mediawiki.org/wiki/Preferred_property_label (as quick introduction)

mwjames:
Can easily be combined with #1855 (`[[SomeProperty::@@@]]` or `[[SomeProperty::@@@en]]`) to create local-specific infoboxes.

mwjames:
https://www.semantic-mediawiki.org/wiki/Preferred_property_label

PR #1879, opened 2016-09-25
Preferred property label, refs #1865

mwjames:
This PR is made in reference to: #1865

- Adds support for localized preferred property labels
- `$GLOBALS['smwgDVFeatures']` adds `SMW_DV_PPLB` to support the use of preferred property labels and `SMW_DV_PROV_LHNT` to output a <sup>p</sup> hint marker on properties that use a preferred label

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -524,7 +524,7 @@ $GLOBALS['smwgPageSpecialProperties'] = array( '_MDAT' );
 # setting is not normally changed by users but by extensions that add new
 # types that have their own additional declaration properties.
 ##
-$GLOBALS['smwgDeclarationProperties'] = array( '_PVAL', '_LIST', '_PVAP', '_PVUC' );
+$GLOBALS['smwgDeclarationProperties'] = array( '_PVAL', '_LIST', '_PVAP', '_PVUC', '_PDESC', '_PPLB' );
 ##
 
 ###
@@ -1007,9 +1007,14 @@ $GLOBALS['smwgEnabledInTextAnnotationParserStrictMode'] = true;
 # - SMW_DV_NUMV_USPACE (Number/QuantityValue) to preserve spaces within
 # unit labels
 #
+# - SMW_DV_PPLB to support the use of preferred property labels
+#
+# - SMW_DV_PROV_LHNT (PropertyValue) to output a <sup>p</sup> hint marker on
+# properties that use a preferred label
+#
 # @since 2.4
 ##
-$GLOBALS['smwgDVFeatures'] = SMW_DV_PROV_REDI | SMW_DV_MLTV_LCODE | SMW_DV_PVAP | SMW_DV_WPV_DTITLE | SMW_DV_TIMEV_CM;
+$GLOBALS['smwgDVFeatures'] = SMW_DV_PROV_REDI | SMW_DV_MLTV_LCODE | SMW_DV_PVAP | SMW_DV_WPV_DTITLE | SMW_DV_TIMEV_CM | SMW_DV_PPLB | SMW_DV_PROV_LHNT;
 
 ##
 # Fulltext search support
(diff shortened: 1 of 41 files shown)


mwjames:
@kghbln After a turbo sprint weekend and as stipulated, the PR is more or less finished before the SMWCon. It is nearly feature complete and only requires manual testing and CI integration tests.

As for the usage:
- Select a property `Foo`
- Add `[[Has preferred property label::SomeLabel@SomeLanguage]]`
- Switching to `SomeLanguage` should display property `Foo` with `SomeLabel`<sup>p</sup> and pointing it to its canonical property `Foo`
- If you switch to a language that has not maintained a preferred label then the "normal" canonical property label is displayed (just as before)
- The property auto-complete function in `Special:Ask` and `Special:SearchByProperty` supports searching for localized labels which when matched will return the canonical form
- The canonical form is expected to be used for the annotation and query process (i.e. doing a reverse match as in `{{#ask: [[SomeLabel@SomeLanguage::SomeValue]] ... }}` is not expected to work)

mwjames:
@kghbln This PR requires to run `update.php`. The following content can be used as example [0, 1] after the merge:

- Property:P106 (https://gist.github.com/mwjames/06afb68c071c58abe841ee7b807f5160)
- Occupation (https://gist.github.com/mwjames/0ee589bee74a647b89664403513d34bf)

mwjames:
This PR took a lot of energy ...

kghbln:
> This PR took a lot of energy ...

I cannot even imagine all the effort! Another great one, indeed!!

> This PR requires to run update.php. The following content can be used as example [0, 1] after the merge:

I have updated the wiki and added the example pages:

- http://sandbox.semantic-mediawiki.org/wiki/Attribut:P106
- http://sandbox.semantic-mediawiki.org/wiki/Occupation

The only thing that seem so cause an issue is that one can only assign one value per language so `[Has preferred property label::occupation@en]]` and ` [[Has preferred property label::profession@en]]` will not work for the same property. However I think this is fine.

mwjames:
> The only thing that seem so cause an issue is that one can only assign one value per language so [Has preferred property label::occupation@en]] and [[Has preferred property label::profession@en]] will not work for the same property. However I think this is fine.

As per #1865 "Borrowing on the skos:prefLabel definition, the preferred label is understood as to only have one label per given language ..." and [0], so yes this was made on purpose.

Imaging you have `[[Has preferred property label::occupation@en]]` and `[[Has preferred property label::profession@en]]` on the same property what are you going to display on an instance that referrers to that property? You require a preference as decision parameter and this is what this label represents because you don't want to display one label and next time you display another label (by chance) without changing the context.

Property:P106 (which I pulled from Wikidata) demonstrates this issue where you have competing labels for the same language with the fist being registered on the property page being defined as preferred and any other label that is trying to occupy the same declarative space (with the same language) are marked as error.

image

There is a concept of alternative labels [1] but I'm not sure how helpful this is or how we would use it other than to clarify to a user that they are synonyms (we also have redirects that form equality to a property instance, expressed as `sameAs`).

But let's say, you would be able to do `[[Has preferred property label::occupation@en]]` and `[[Has alternative property label::profession@en]]` (using an alternative) how would we use this label? What would we do with this label?

image

image

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/dataitems/SMW_DI_Property.php#L212-L234

[1] https://www.w3.org/TR/skos-primer/#secalt

kghbln:
> so yes this was made on purpose.

I was just surprised to see all these errors. Did not guess that this was part of the demo.

> Property:P106 (which I pulled from Wikidata) demonstrates this issue where you have competing labels for the same language with the fist being registered on the property page being defined as preferred and any other label that is trying to occupy the same declarative space (with the same language) are marked as error.

Yeah, now I know the reason. I think the error ensures consistency so in the end it has to be decided which one should be used -> no competition and in the end no confusion. The clarification that labels are synonyms comes from the error note telling that the "canonical" label (= label already assigned) should be used.

> use it other than to clarify to a user that they are synonyms

Cannot think of anything else here.

mwjames:
> (using an alternative) how would we use this label? What would we do with this label?

On the matter of alternative labels, [0] contains an examples on how to use them more broadly in defining alternatives for an entity.

[0] http://sandbox.semantic-mediawiki.org/wiki/Triacylglycerol_lipase

Issue #1880, opened 2016-09-29
Invalid input syntax for bytea - SMW::writePropertyTableRowUpdates-insert-smw_di_blob


drhirn:
### Setup and configuration

- SMW version: 2.4.1
- MW version: 1.27.1
- DB (MySQL etc.): Postgres 9.4

### Issue

When running
```
sudo -u wwwrun php /srv/www/htdocs-wiki-doku/extensions/SemanticMediaWiki/maintenance/rebuildData.php -d 100
```
I get the following error:

```
Database error occured.
Query: INSERT INTO "smw_di_blob" (s_id,p_id,o_blob,o_hash) VALUES ('1764','933','C:\Program Files (x86)\AnotherFolderHere\AndAnotherFolderHereToo\AndTheNameOfTheFileHere.exe','C:\Program Files (x86)\AnotherFolderHerea30462340362b7e3c3073054bd572471')
Funktion: SMW::writePropertyTableRowUpdates-insert-smw_di_blob
Fehler: 22P02 ERROR:  invalid input syntax for typ bytea
LINE 1: ...b" (s_id,p_id,o_blob,o_hash) VALUES ('1764','933','C:\Progra...
```

### Steps to reproduce the observation:

I have a Semantic Forms form where I can save a file-path. Typ of the property is "text".
Resulting pages look like this for example:
```
{{Windows Service
|Name=JustATest
|Path=C:\Program Files (x86)\AnotherFolderHere\AndAnotherFolderHereToo\AndTheNameOfTheFileHere.exe
|StartupType=Automatic
|LogonUser=hwe,
|Hostname=ABCDEF01, ABCDEF02,
|Description=It does something
}}
```
It seems it has something to do with the length of the "path" combined with backslashes, but I'm not quite sure. Shorter "paths" do work. Or short paths with backslashes.

mwjames:
> Fehler: 22P02 ERROR:  invalid input syntax for typ bytea

This error relates to [0] caused by how postgres is handling escaped `\` character for table columns that use the `bytea` type. SMW uses `bytea` on `o_blob` fields that contain text strings longer than 255 chars.

[0] https://stackoverflow.com/questions/28631807/postgresql-error-invalid-input-syntax-for-type-bytea

mwjames:
Expected to be solved by #1885 (added an extra integration to test with > `MAX_HASH_LENGTH`)

PR #1885, opened 2016-10-05
Fix postgres bytea escape/unescape on blob fields, refs #1880

mwjames:
This PR is made in reference to: #1880

- "Error: 22P02 ERROR:  invalid input syntax for type bytea"

This PR includes:
- [x] CI build passed


kghbln:
@mwjames Since this was backported to 2.4.2: Shouldn't the lable be 2.4.2 instead of 2.5?

mwjames:
> Since this was backported to 2.4.2: Shouldn't the lable be 2.4.2 instead of 2.5?

Maybe in retrospect but the initial issue was solved in 2.5 and it just so happened that that it solved another (from the mailing-list) while the original issue appeared in a different context. Getting two for one so to speak.

kghbln:
Thanks for the info!

PR #1891, opened 2016-10-05
ParserParameterProcessor to support JSON typed annotation

mwjames:
This PR is made in reference to: #N/A

- This allows to add JSON typed annotation in the `#set` and `#subobject` parser function using the `@json` marker.
- Using the JSON format allows for creating structured annotation by means of electronic generators
- A depth of 3+ is not supported (for details see test cases)

```
{{#set:
    |@json={
    "foo": [
         "bar",
         "foobar"
    ]
}
}}
```

This PR includes:
- [x] CI build passed


mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/1891

mwjames:
@kghbln FYI

mwjames:
Using the `|` pipe symbol generally posses an issue within a parser function when used as part of the content representation and to avoid breaking the content `+pipe` can be used to signal that no breakage is expected for the preceding element (well, internally we clue the content together since the MW parser would split the content no matter what).

```
{{#set:
    |@json={
    "has text": "Some text with a | pipe symbol within ..."
}|+pipe }}

{{#subobject:
    |@json={
    "has text": "Some text with a | pipe symbol within ..."
}|+pipe }}
```

PS: I did forget to add them to the integration test, so it would be nice if someone could extend the test.

Issue #1825, opened 2016-09-08
Master (1.28-alpha) test failures


mwjames:
## Related 1.28 issues
- https://phabricator.wikimedia.org/T147550
- https://phabricator.wikimedia.org/T145947
- https://phabricator.wikimedia.org/T145840
- https://phabricator.wikimedia.org/T148332
- [ ] #1923

mwjames:
After a week the master test results are getting worse, now it reports with `There were 10 failures:` [0].

For the last two MW releases, I tried to be accommodating and made an effort to fix those and search for the cause but with upcoming MW release I can no longer effort investing time on fixing issues caused by WMF developers.

[0] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/159775308

```
4) SMW\Tests\Integration\MediaWiki\Hooks\FileUploadIntegrationTest::testFileUploadForDummyTextFile
Failed asserting that actual size 0 matches expected size 4.

/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/SemanticDataValidator.php:109
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/SemanticDataValidator.php:207
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/MediaWiki/Hooks/FileUploadIntegrationTest.php:122
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/MwDBaseUnitTestCase.php:125
/home/travis/build/SemanticMediaWiki/mw/maintenance/doMaintenance.php:110

5) SMW\Tests\Integration\MediaWiki\JobQueueDBIntegrationTest::testPropertyTypeChangeToCreateUpdateJob
Failed asserting that 0 is greater than 0.

/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/MediaWiki/JobQueueDBIntegrationTest.php:260
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/MwDBaseUnitTestCase.php:125
/home/travis/build/SemanticMediaWiki/mw/maintenance/doMaintenance.php:110

[...]

PR #1897, opened 2016-10-08
Work around deprecated/removed DatabaseBase::getSearchEngine

s7eph4n:
This PR is made in reference to: https://phabricator.wikimedia.org/T147683

DatabaseBase::getSearchEngine was removed in 1.28.

This PR includes:
- [x] Tests (unit/integration): Existing tests failed. No new tests necessary.
- [ ] CI build passed: No, but not because of this patch.

### DIFF
--- a/src/MediaWiki/Search/Search.php
+++ b/src/MediaWiki/Search/Search.php
@@ -4,6 +4,7 @@ namespace SMW\MediaWiki\Search;
 
 use Content;
 use DatabaseBase;
+use MediaWiki\MediaWikiServices;
 use RuntimeException;
 use SearchEngine;
 use SMW\ApplicationFactory;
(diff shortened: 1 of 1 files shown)


kghbln:
There is also a [unit test](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/c87e501aa908d55481d0538179d712a0695dbb44/tests/phpunit/Unit/MediaWiki/Search/SearchTest.php) involving this. This probably needs to be  adjusted too.

No backport to 2.4 required since this one is not meant to support MW 1.28+

s7eph4n:
No need to fix the unit test. The existing test fails without the patch, works with the patch.

kghbln:
Great! Everything is fluffy I guess. :)

JeroenDeDauw:
Yay MediaWiki, the new thing does not have dependency injection either. At least it's better named crap code...

JeroenDeDauw:
Thanks!

JeroenDeDauw:
@s7eph4n  Note that there is a follow up at https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1898 which highlights some things you could have done differently

mwjames:
That was a bit quick here because of:

> No need to fix the unit test. The existing test fails without the patch, works with the patch.

https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/165969263

Not really, now it returns with:

```
3) SMW\Tests\MediaWiki\Search\SearchTest::testGetDefaultFallbackSearchEngineForNullFallbackSearchType
Failed asserting that SearchEngineDummy Object (...) is an instance of class "SearchDatabase".
```

mwjames:
> There is also a unit test involving this. This probably needs to be adjusted too.

https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/165436085

The unit test is fine since we mock that part which isn't really helpful (no failure in above master test from two days ago) but if you are interested in a real integration test.

If I have time, I'll add an integration test so things like below will shine through.

````
Fatal error: Call to undefined method DatabaseMysqli::getSearchEngine() in ...\extensions\SemanticMediaWiki\src\MediaWiki\Search\Search.php on line 58
```

PS: There is another case https://phabricator.wikimedia.org/T147550 where WMF doesn't mind its own deprecation policy but luckily our integration test shouted out on this one.

PSS: 1.28/master is allowed to fail otherwise we go crazy on the extend exposition of WMF funnyism.

JeroenDeDauw:
> PSS: 1.28/master is allowed to fail

Muh... I want to be able to merge PRs if the CI is green

mwjames:
> I want to be able to merge PRs if the CI is green

I want to no have to FIX (or circumvent) issues not related to SMW-core but surely that's not going to happen.

mwjames:
#1825 is a prime example of what happens when you run master.

src/MediaWiki/Search/Search.php
```diff
@@ -4,6 +4,7 @@
 
 use Content;
 use DatabaseBase;
+use MediaWiki\MediaWikiServices;
```
mwjames: Access, reference, and usage to `MediaWiki\MediaWikiServices` should be confined to either `SMW\Tests\TestEnvrionment` [0] or `ApplicationFactory`.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/TestEnvironment.php#L95-L114

mwjames:
> If I have time, I'll add an integration test so things like below will shine through.

... And it shall be light #1904.

PR #1898, opened 2016-10-09
CS changes in SearchEngine

JeroenDeDauw:
Follow up to https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1897

### DIFF
--- a/src/MediaWiki/Search/Search.php
+++ b/src/MediaWiki/Search/Search.php
@@ -45,64 +45,6 @@ class Search extends SearchEngine {
 	}
 
 	/**
-	 * @param $type
-	 */
-	private function assertValidFallbackSearchEngineType( $type ) {
-
-		if ( $type === null ) {
-			return;
-		}
-
-		if ( !class_exists( $type ) ) {
-			throw new RuntimeException( "$type does not exist." );
-		}
-
-		if ( !is_subclass_of( $type, 'SearchEngine' ) || $type === 'SMWSearch' ) {
-			throw new RuntimeException( "$type is not a valid fallback search engine type." );
-		}
-	}
-
-	/**
-	 * @param $db
-	 * @return string
-	 */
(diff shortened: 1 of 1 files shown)


src/MediaWiki/Search/Search.php
```diff
@@ -163,6 +104,52 @@ private function searchFallbackSearchEngine( $term, $fulltext ) {
...
+
+	/**
+	 * @param string $className
+	 */
+	private function assertIsSearchEngineClassName( $className ) {
+		if ( !class_exists( $className ) ) {
+			throw new RuntimeException( "$className does not exist." );
+		}
+
+		if ( !is_subclass_of( $className, 'SearchEngine' ) || $className === 'SMWSearch' ) {
```
JeroenDeDauw: @s7eph4n This predicate is as you wrote it, and I'm not sure it's correct. We should not get an exception when `$className === 'SMWSearch'` right? So some brackets are missing... though even in that case, I'm not sure why you would have this, since `SMWSearch` is a subclass of `SearchEngine` no?

I should not have merged your PR before that is addressed though saw too late.

s7eph4n: It is intentional. If the fallback search is again an SWMSearch we run into a recursion.

JeroenDeDauw: Ah ok. In that case, does this PR look good to you?

kghbln: @s7eph4n @JeroenDeDauw I guess we now need some fluff.

kghbln: Ping

JeroenDeDauw:
Blah, why did this not get mawrged :(

kghbln:
> Blah, why did this not get mawrged :(

I guess too much noise the days I pinged.

PR #1916, opened 2016-10-09
SMWSearch: Fix unit test; Move access to MediaWikiServices to ApplicationFactory

s7eph4n:
This PR is made in reference to:
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1897#issuecomment-252457927
- https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1897#pullrequestreview-3417583

- fix SearchTest::testGetDefaultFallbackSearchEngineForNullFallbackSearchType
- remove direct accesses to MediaWikiServices from Search.php
- add ApplicationFactory::getMediaWikiServices
- add $callbackLoader->registerCallback( 'MediaWikiServices', ...) to SharedCallbackContainer
  (not sure if that is the correct thing to do)

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/MediaWiki/Search/Search.php
+++ b/src/MediaWiki/Search/Search.php
@@ -52,7 +52,11 @@ class Search extends SearchEngine {
 			throw new RuntimeException( "$type does not exist." );
 		}
 
-		if ( !is_subclass_of( $type, 'SearchEngine' ) || $type === 'SMWSearch' ) {
+		if ( $type === 'SMWSearch' ) {
+			throw new RuntimeException( 'SMWSearch is not a valid fallback search engine type.' );
+		}
+
+		if ( $type !== 'SearchEngine' && !is_subclass_of( $type, 'SearchEngine' ) ) {
 			throw new RuntimeException( "$type is not a valid fallback search engine type." );
 		}
 	}
--- a/src/MediaWiki/Search/Search.php
+++ b/src/MediaWiki/Search/Search.php
@@ -4,7 +4,6 @@ namespace SMW\MediaWiki\Search;
 
 use Content;
 use DatabaseBase;
-use MediaWiki\MediaWikiServices;
 use RuntimeException;
 use SearchEngine;
 use SMW\ApplicationFactory;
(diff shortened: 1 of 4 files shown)


src/MediaWiki/Search/Search.php
```diff
@@ -110,13 +118,30 @@ public function setDB( DatabaseBase $connection ) {
 		$this->fallbackSearch = null;
 	}
 
+
+	/**
+	 * @return \LoadBalancer
+	 */
+	protected function getLoadBalancer() {
+
```
mwjames: I would have thought you make this available to the `ApplicationFactory` as dedicated service hiding `MediaWikiServices` as much we can to avoid issues when things change again on the MW side.

The DI container would always return a `DBLoadBalancer` instance ( avoiding the `null` case) and internally makes the call to either `MediaWikiServices` or `LBFactory` but it is hidden from the consumer who calls `ApplicationFactory::getMediaWikiDbLoadBalancer`.

s7eph4n: Ok, no problem.

What about [`getDefaultSearchEngineTypeForDB`](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1916/files?diff=unified#diff-2d41b72f98fa4903663fb448b947b11dR76)?
Not sure if I should register a callback or just add the method to `ApplicationFactory` as is.

mwjames:
This doesn't seems to be an issue on 1.27+ [0] but all others return with a:

```
1) SMW\Tests\MediaWiki\Search\SearchTest::testGetDefaultFallbackSearchEngineForNullFallbackSearchType
RuntimeException: SearchEngine is not a valid fallback search engine type.
```

[0] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/builds/166264508

mwjames:
Expecting this to fix #1916.

s7eph4n:
This is becoming a nightmare. Could have sworn I've run the tests. Oh well...

mwjames:
Thanks.

PR #1917, opened 2016-10-10
Update de.json

kghbln:
* Fill authors array
* Provide missing translation

### DIFF
--- a/i18n/extra/de.json
+++ b/i18n/extra/de.json
@@ -1,6 +1,9 @@
 {
     "@metadata": {
-        "authors": []
+        "authors": [
+            "Karsten Hoffmeyer",
+            "Markus Krötzsch"
+        ]
     },
     "fallbackLanguage": "en",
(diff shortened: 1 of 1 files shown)


i18n/extra/de.json
```diff
@@ -198,4 +201,4 @@
             "Sun"
```
kghbln: @mwjames Must [these](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/d35763142ec4455e07c817e5f2a8cb25f4d8632c/i18n/extra/de.json#L174-L201) be 3 chars? In German we have only 2 chars shortcuts so ...

JeroenDeDauw: I doubt it needs to be 3, dunno tho

kghbln: I guess so too so  I made https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1919

PR #1919, opened 2016-10-11
Update de.json

kghbln:
Provide German translations and shortcuts for days.

### DIFF
--- a/i18n/extra/de.json
+++ b/i18n/extra/de.json
@@ -173,32 +173,32 @@
     ],
     "days": [
         [
-            "Monday",
-            "Mon"
+            "Montag",
+            "Mo"
         ],
         [
-            "Tuesday",
-            "Tue"
+            "Dienstag",
+            "Di"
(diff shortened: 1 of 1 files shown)


i18n/extra/de.json
```diff
@@ -173,32 +173,32 @@
...
         [
-            "Tuesday",
-            "Tue"
+            "Dienstag",
+            "Di"
         ],
         [
-            "Wednesday",
-            "Wed"
+            "Mittwoch",
```
JeroenDeDauw: Silly Germans, 3 is not the middle in [1, 7], that's 4 :)

kghbln: So I read that you have learned your German lessons. :P

JeroenDeDauw: nein

kghbln: Hä, so you are not into learning esoteric languages. :D

PR #1927, opened 2016-10-16
PropertyPage avoid unrestricted subproperty query

mwjames:

- Adds `$GLOBALS['smwgSubPropertyListLimit']` (default 25) to avoid running an unlimited SQL query. In case additional subproperties are available that are beyond the limit then a note will appear to inform about the restriction.

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

PR #1848, opened 2016-09-14
Add ExtraneousLanguage JSON format, refs #1691, #1561

mwjames:
This PR is made in reference to: #1691, #1561

This PR addresses or contains:

- Moves SMW specific language elements from PHP to JSON
- Support for partial fallback
- If a json file (e.g. `zh-hant.json`) contains `isLanguageRedirect` then only a `fallbackLanguage` is required
- The `CanonicalFallbackLanguage` is `en`
- Adds `LanguageFallbackFinder`, `LanguageFileContentsReader`

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

PR #1943, opened 2016-10-23
ExtraneousLanguage / Extend the CanonicalPropertyAliases members

mwjames:
This PR is made in reference to: #1848

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/ExtraneousLanguage/ExtraneousLanguage.php
+++ b/src/ExtraneousLanguage/ExtraneousLanguage.php
@@ -325,10 +325,14 @@ class ExtraneousLanguage {
 			'propertyAliases'
 		);
 
-		$canonicalPropertyAliases += $this->languageContents->getFromLanguageWithIndex(
-			$this->languageCode,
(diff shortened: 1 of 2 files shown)


mwjames:
> I wanted to draw attention to a PR that has the potential for a regression

@kghbln FYI; Speaking of regression, I found one while browsing [0].  Properties would appear as redlinks instead of the linking to the property page due to some of them missing the canonical form in the search `array` (lucky us the sandbox uses FR which allowed to spot this).

image

[0] http://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Parcourir/Athens-23_QUERYa9764e9cc227bd29aabe7aebfd41a997

kghbln:
>  (lucky us the sandbox uses FR which allowed to spot this)

It was indeed a pretty good idea to do the switch not only to another language than just English as the ultimate fallback language but also to a language like French which is somewhat familiar an providing some linguistic extras.

PR #1945, opened 2016-10-23
Add support for travis ci containter

paladox:

- ...
- ...
- ...

This PR includes:
- [ ] CI build passed

https://docs.travis-ci.com/user/migrating-from-legacy/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: false
+
 language: php
 
 matrix:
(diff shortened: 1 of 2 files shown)


mwjames:
Sorry, but you cannot  do this as other services like `SPARQL` triple stores rely on being installed using `sudo`.

paladox:
Oh but what does it need sudo for?

We can install packages by doing

addons:
  apt:
    packages:

paladox:
Or we can use the trusty beta with sudo?, but container is faster.

mwjames:
> Oh but what does it need sudo for?

You may look at [0].

> Or we can use the trusty beta with sudo?, but container is faster.

I know but stability of the integration platform is more important than speed. 

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/travis/install-services.sh

mwjames:
>  trusty beta with sudo?

One would have to verify that each job (and we have 10 of them) is doing an equally good execution compared to the current setup. Those services listed above have to run no matter what, whether they run with/without `sudo` doesn't really matter but the current setup required them to be setup with `sudo`.

paladox:
Ive added the package here https://github.com/travis-ci/apt-package-whitelist/pull/3453 for the container.

paladox:
the container runs on ubuntu precise which the tests were doing already does.

Issue #1751, opened 2016-08-02
2.4.2 release


mwjames:
Tracking issue for a possible (not yet defined) release.
- [x] #1823
- [x] #1883 
- [x] #1885 
- [x] #1887 
- [x] #1901

mwjames:
Personally, I will have no time for this so maybe asking on the mailing-list on who wants to help out (and I'm mean NOT Karsten) otherwise I will close this task on my next week's support slot.

JeroenDeDauw:
No need to close it next week. I'll try to get to it, but it might be afterwards.

kghbln:
This is confusing. Apparently the code for #1901 is already within the 2.4.x branch tough it was not backported. Probably I should back off here.

kghbln:
@JeroenDeDauw Something we could do about release rot. As far as I see there is already enough for a 2.4.3

JeroenDeDauw:
Things get a lot easier if the minor releases get done more frequently, ie 2.5.0 happening soon.

I don't understand your comment about 2.4.3, as there is no 2.4.2 yet

JeroenDeDauw:
I've merged the stuff, so these commits are now on 2.4.x. Someone double checking things did not get messed up in the rebase would be good, as @mwjames mentioned already.

JeroenDeDauw:
@SemanticMediaWiki/testers to try out the development version of 2.4.2, change the version of SMW in `composer.json` to `2.4.x@dev` and run `composer update`. We want to know if the functionality critical to you is in working order.

@mwjames could you provide some pointers on what exactly to test based on what might have gotten broken?

kghbln:
> I don't understand your comment about 2.4.3, as there is no 2.4.2 yet

I meant if we had release 2.4.2 a month ago we could already have done a 2.4.3 by now.

mwjames:
> could you provide some pointers on what exactly to test based on what might have gotten broken?

Some have been tested, others relate to postgres text/bytea storage, the most critical is #1887 that should fix once and for all the registration of Property namespaces independent of whether an extension with a `extension.json` is invoked or not.

kghbln:
"rebuildData.php" gives me this:
```
PHP Fatal error:  Class 'SMWDIHandlerWikiPage' not found in /.../extensions/SemanticMediaWiki/includes/storage/SQLStore/SMW_SQLStore3.php on line 212

[...]

Issue #1931, opened 2016-10-20
Exception from line 151 of MemcachedBagOStuff.php: Key contains invalid characters


s7eph4n:
### Setup and configuration

- SMW version: 2.4.1
- MW version: 1.27.0-rc.0 (f465524)
- DB: MySQL 5.1.73

### Issue
Exception from line 151 of /var/www/html/w/includes/libs/objectcache/MemcachedBagOStuff.php: Key contains invalid characters

Line 151 of MemcachedBagOStuff.php throws an error if the cache key contains anything but ASCII characters, so it is apparently supposed to be a feature.

### Stack trace
```
[maintenance]# php rebuildData.php -d 100 --ignore-exceptions --no-cache -v

Running for storage: SMWSQLStore3

Removing table entries (marked for deletion).

.............................................Exception encountered, of type "Exception"
[470fdde19df57e183827d2e7] [no req]   Exception from line 151 of /var/www/html/w/includes/libs/objectcache/MemcachedBagOStuff.php: Key contains invalid characters: WANCache:v:my_wiki:interwiki:création_de_la_fiche_smog_sol_ela3_mecaph_20/0_indice_a
Backtrace:
#0 /var/www/html/w/includes/libs/objectcache/MemcachedPhpBagOStuff.php(56): MemcachedBagOStuff->validateKeyEncoding(string)
#1 /var/www/html/w/includes/libs/objectcache/WANObjectCache.php(259): MemcachedPhpBagOStuff->getMulti(array)
#2 /var/www/html/w/includes/libs/objectcache/WANObjectCache.php(216): WANObjectCache->getMulti(array, array, array)
#3 /var/www/html/w/includes/libs/objectcache/WANObjectCache.php(807): WANObjectCache->get(string, NULL, array)
#4 /var/www/html/w/includes/libs/objectcache/WANObjectCache.php(779): WANObjectCache->doGetWithSetCallback(string, integer, Closure, array)
#5 /var/www/html/w/includes/interwiki/Interwiki.php(256): WANObjectCache->getWithSetCallback(string, integer, Closure)
#6 /var/www/html/w/includes/interwiki/Interwiki.php(101): Interwiki::load(string)
#7 /var/www/html/w/includes/interwiki/Interwiki.php(74): Interwiki::fetch(string)
#8 /var/www/html/w/includes/title/MediaWikiTitleCodec.php(290): Interwiki::isValidInterwiki(string)
[...]

PR #1955, opened 2016-10-25
Backport into 2.4.x

JeroenDeDauw:
Fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1751

@mwjames there where some merge conflicts in here... I'm not sure I resolved them correctly


mwjames:
I took a minute to correct the tests and added a small fix as suggested in #1931 to the 2.4.x backport list.

Before putting a release tag on it, someone should verify that 2.4.x generally works as expected as it would be a shame if we miss something.

kghbln:
> Before putting a release tag on it, someone should verify that 2.4.x generally works as expected as it would be a shame if we miss something. 

I could switch sandbox to this branch. Not sure if I will have to uninstall all other semantic extensions by doing so though. Branches and "@dev" usually do not mix.

Issue #1969, opened 2016-10-28
add native lua support for semantic mediawiki


oetterer:
### Setup and configuration
- SMW version: 2.5@dev
- MW version: 1.27.1
- PHP version: 5.6
- DB (MySQL etc.): mysql 5.5
### Proposal

I would like to suggest a native lua support for semantic mediawiki.
### Motivation

For projects using the [Scribunto](https://www.mediawiki.org/wiki/Extension:Scribunto) Extension the only way of accessing semantic data is by calling the parser function #ask via frame. Besides the fact that this causes a performance issue, to make the resulting semantic data available inside the lua module, one has to do some serious string parsing. Same goes when storing data via #set or #subobject, albeit the string parsing is not necessary.

To streamline this and make it possible for users to directly access smw's data inside lua modules, I would like to suggest creating a smw.lua module that returns the result of an #ask query directly as a table when called inside other lua modules.

Same goes for set and suboject: let users call a smw.lua function to directly store data, w/o having to rely on the slower frame:callParserFunction().
### Notes on existing solutions
- There is the way of using callParserFunction. This is a viable, if not slower way to store data. When it comes to retrieving data, one has to find a way to create a lua data table from a result string. This is tedious and prone to error
- [Extension: Semantic Scribunto](https://www.mediawiki.org/wiki/Extension:Semantic_Scribunto) allows to do an ask query and returns a json formatted string. With Scribunto >= 1.25 this can be parsed into a table. However, this extension relies on open read access to the api (see [Issue#8](https://github.com/SemanticMediaWiki/SemanticScribunto/issues/8) and provides no solution to write property or subobject data.

mwjames:
> I would like to suggest a native lua support for semantic mediawiki.

Lua (or for that matter Scribunto) are not essential for the working of SMW-core hence creating an interdependency on another MediaWiki extension will unlikely met with a favourable support as it increases maintenance cost, technical dept, and may cause unanticipated breakages [0]. Instead, Extension: Semantic Scribunto has been created with those users in mind that want a more tighter integration with Scribunto and SMW without making SMW-core directly depend on it.

In general, functionality that is not absolutely necessary to the overall mission of the software (which is to provide the means of storing and retrieving data in a structured way) is expected to be provided by add-on extensions that can be maintained independently and define necessary dependency boundaries.

> relies on open read access to the api (see Issue#8 and provides no solution to write property or subobject data.

[...]

PR #1970, opened 2016-10-28
add native lua capabilities to semantic mediawiki

oetterer:
This PR is made in reference to: #1969 

- provides a smw.ask function inside lua (mirrors functionality of parser function #ask)
- provides a smw.set function inside lua (mirrors functionality of parser function #set)
- provides a smw.subobject function inside lua (mirrors functionality of parser function #subobject)
- provides a smw.info function inside lua (mirrors functionality of parser function #info)
- provides a smw.getPropertyType to retrieve smw's internal property type for a given property

This PR includes:
A native LuaResultPrinter that is only registered, when inside Scribunto Lua module calls. This converts the result of an ask query into a multi dimensional array that is later parsed to the smw lua module

A LuaLibrary class that has one method for every adressed lua function. The functionality is provided by the respective existing parser function hook classes which are created by the existing ParserFunctionFactory. This maximizes reusability and minimizes maintenance on functionality updates.

The smw.lua module to use inside lua.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -449,5 +449,6 @@
 	"smw-datavalue-reference-invalid-fields-definition": "The [[Special:Types/Reference|Reference]] type expects a list of properties to be declared using the [https://www.semantic-mediawiki.org/wiki/Help:Special_property_Has_fields Has fields] property.",
 	"smw-parser-invalid-json-format": "The JSON parser returned with a \"$1\".",
 	"smw-property-preferred-title-format":"$1 ($2)",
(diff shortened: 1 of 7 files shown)


JeroenDeDauw:
Given the discussion on https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1969 about Semantic Scribunto, can this be closed?

oetterer:
Could you leave it for a few days? I would like to get @mwjames opinion on the solution because I would like to integrate it in his semantic scribunto extension.

kghbln:
For ref https://github.com/SemanticMediaWiki/SemanticScribunto/pull/19

Issue #1418, opened 2016-02-24
Add options for displaying images in query results (T32002)


gesinn-it:
continued from: https://phabricator.wikimedia.org/T32002

summary: allow to set the thumb size in query results like

```
{{#ask: ...
   ?my image property#130px
 }}
```

There has already been a patch for this which. See Phabricator.

Would be cool to get this feature into SMW.

kghbln:
The patch suggested:
```
--- SemanticMediaWiki/includes/datavalues/SMW_DV_WikiPage.php	Fri Aug 03 16:33:57 2012
+++ SemanticMediaWiki/includes/datavalues/SMW_DV_WikiPage.php	Fri Aug 03 16:33:30 2012
@@ -193 +193 @@
-				$defaultCaption = '|' . $this->getShortCaptionText() . '|frameless|border|text-top';
+				$defaultCaption = '|' . $this->getShortCaptionText() . '|frameless|border|' . $this->m_outformat . '|text-top';
```

mwjames:
> +               $defaultCaption = '|' . $this->getShortCaptionText() . '|frameless|border|' . $this->m_outformat . '|text-top';

Maybe the parsing should be a bit more sophisticated oppose to of just placing `$this->m_outformat` as plain string (sth. more expressive such as ` |?MyFile#-s[130px]`)

PS: I marked this issue as `easy` which means that I will __not__ engage and expect the community to pick them up but of course I will give assistance as needed.

PR #2009, opened 2016-11-12
Test file upload and #ask image annotation, refs 1418

mwjames:
This PR is made in reference to: #1418

- Support for things like `{{#ask: [[Has file::+]] |?Has file#120px;thumb;<b>{{#show: File:P0705.png |?Has caption |link=none}}</b>[[Extra]] |format=table |limit=1 }}`
- Using `|` as context marker is not possible therefore the declaration expects something like `120px;thumb;...` which is internally interpret as `120px|thumb| ...` for entities that belong to the `NS_FILE` namespace

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

PR #2017, opened 2016-11-18
Add localisation changes for ES

VikaCep:
- Text translations including week days for Spanish
- Changes the order of dateFormats preferences to match the language's (MDY to DMY).

### DIFF
--- a/i18n/extra/es.json
+++ b/i18n/extra/es.json
@@ -18,9 +18,9 @@
         "_tel": "Número de teléfono",
         "_rec": "Registro",
         "_qty": "Cantidad",
-        "_mlt_rec": "Monolingual text",
-        "_eid": "External identifier",
-        "_ref_rec": "Reference"
+        "_mlt_rec": "Texto monolingüe",
+        "_eid": "Identificador externo",
+        "_ref_rec": "Referencia"
     },
     "dataTypeAliases": {
         "URI": "_uri",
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Thanks! Merging in the hope you did not sneakily submit a bunch of insults in Spanish

i18n/extra/es.json
```diff
@@ -93,8 +93,8 @@
             "SMW_YM"
         ],
         [
-            "SMW_MDY",
             "SMW_DMY",
+            "SMW_MDY",
```
mwjames: This change influences the preference [0] in how the Time/Date is interpret from a string [1] therefore I'd like to ask users with ES language competence (@jaideraf) to comment on this to avoid introducing a sub-optimal change in response to [2].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/ExtraneousLanguage/ExtraneousLanguage.php#L427-L430
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/eff53c2ff1fd0365c823c4ae867aed57749edcf2/includes/datavalues/SMW_DV_Time.php#L461-L473
[2] https://www.mediawiki.org/w/index.php?title=Extension_talk:Semantic_Result_Formats/calendar_format&curid=40988&diff=2288909&oldid=2287003

VikaCep: I made that change due to a bug I noticed every time I made semantic queries, the dates were incorrect. In ES language the dates entered  have the DMY format but the Special:Browse page or the result of some semantic queries showed the day as month (when it's < 12) and the month as day. 

I'm from Argentina. Also checked in other Spanish speaking countries and the format should be DMY as shown in this map https://en.m.wikipedia.org/wiki/Date_format_by_country

jaideraf: @mwjames thanks for caring, but, I'm from Brazil :brazil:: I speak Portuguese, not Spanish. :wink: 
Just for record, brazilians use the date format DMY as well.

mwjames: >  I speak Portuguese, not Spanish

Dammit and for this faux pas I'd probably get arrested from the new Dumpf administration for cultural insensitivity.

PR #2022, opened 2016-11-19
ListResultPrinter / Fix usage of sep in format=template

mwjames:
This PR is made in reference to: # [0, 1]

- Most likely introduced with https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/66b776138b9c9607566cdc2f7ada8ffc77804bac

This PR includes:
- [x] CI build passed

[0] [Issues with the "sep" function in mode Query template.](https://www.semantic-mediawiki.org/wiki/Thread:Semantic-mediawiki.org:Community_portal/Issues_with_the_%22sep%22_function_in_mode_Query_template.)
[1] http://naruto.wikia.com/wiki/Thread:217742#29

### DIFF
--- a/includes/queryprinters/ListResultPrinter.php
+++ b/includes/queryprinters/ListResultPrinter.php
@@ -224,7 +224,7 @@ class ListResultPrinter extends ResultPrinter {
 		}
 
 		// Define separators for list items
-		if ( $this->params['format'] !== 'template' ){
+		if ( $this->params['sep'] !== '' ){
 			if ( $this->params['format'] === 'list' && $this->params['sep'] === ',' ){
 				// Make default list ", , , and "
 				$this->listsep = ', ';
--- a/tests/phpunit/Integration/ByJsonScript/Fixtures/f-0803.json
+++ b/tests/phpunit/Integration/ByJsonScript/Fixtures/f-0803.json
@@ -1,0 +1,47 @@
+{
+	"description": "Test `format=template` and `sep`/`named args` (...)",
+	"setup": [
+		{
+			"namespace": "SMW_NS_PROPERTY",
+			"page": "Has text",
+			"contents": "[[Has type::Text]]"
+		},
+		{
+			"namespace": "NS_TEMPLATE",
+			"page": "Example/F0803/Sep",
+			"contents": "<includeonly>{{{?Has text}}}</includeonly>"
+		},
+		{
+			"page": "Example/F0803/1",
+			"contents": "{{#subobject: |@category=F0803 |Has text=123 |Has text=456 }} {{#subobject: |@category=F0803 |Has text=abc }}"
+		},
+		{
+			"page": "Example/F0803/Q.1",
+			"contents": "{{#ask: [[Category:F0803]] |?Has text |format=template |template=Example/F0803/Sep |mainlabel=main |named args=yes |link=none |sep= &#32;&bull;&#32; |sort=Has text |order=asc }}"
+		}
+	],
+	"tests": [
+		{
+			"type": "parser",
+			"about": "#0",
+			"subject": "Example/F0803/Q.1",
+			"assert-output": {
+				"to-contain": [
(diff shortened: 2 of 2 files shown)


mwjames:
@kghbln FYI

kghbln:
Just fetched the latest code on sandbox. Works perfect! Thanks!

mwjames:
By coincidence, @kghbln found:

image

with no `sep=` parameter therefore falling back to `,` as default.

```
{{#ask:
 [[Category:SMW releases]]
 [[Language code::en]]
 [[Is duplicate page::false]]
 [[Version::!Semantic MediaWiki 1.6.2]]
 | ?Version
 | ?Has release date#ISO
 | ?Has highlights
 | ?Is compatible with
 | format=template
 | sort=Version, Has release date
 | order=descending
 | limit=100
 | introtemplate=Version history intro
 | template=Version history
 | outrotemplate=Version history outro
 | intro=Here is the complete <b>version history</b>:
}}
```

kghbln:
> By coincidence, @kghbln found:

Great, that we found this prior to the release.

kghbln:
@mwjames Should I open a separate issue allowing to track this?

mwjames:
> Should I open a separate issue allowing to track this?

Maybe for the best. Also what I need is some kind of baseline that defines the behavior to avoid any future regression. We caught this just by mere chance!

```
{{#ask: [[List::2022]]
 |?Has page
 |?Has page
 |format=ul
}}
```
Expects to output `Some, Some`
 
Issue: The default `sep` is set to `,` which is a bit tricky because it cannot be decided whether a user made a deliberate decision or not. I will change the default to be empty and in case it is a list/ul/ol then the output will use `,` as default if no other `sep` was given. `template` by default will have no `sep` to avoid the case we just found.

s7eph4n:
Just for my understanding: This means that when I upgrade to 2.5 I have to modify all `{{#ask}}` calls with `format=template` and explicitly add a `sep` parameter? Or should it be `listsep`?

mwjames:
If I remember correctly (off the top of my head), #2331 introduced
`$smwgResultFormatsFeatures` as means to allow to stick with pre-2.5
behaviour, so setting it to `SMW_RF_NONE` should avoid having to
update any of your existing queries. Of course, this also makes it
impossible to use the changes addressed in #2022.

I hope this helps otherwise please see the examples attached on each issue.

PS: Currently, my memory is a bit fuzzy on the issue, so the answer is
the best I can give you right now.

Cheers

s7eph4n:
Ah, ok. Guess you had to break something to make it work properly. Just wish it wouldn't bite me now. :P
I'll change the `{{#ask}}` calls. That parameter is nice, but I will have forgotten all about it by tomorrow.

kghbln:
[Needs more doculove](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ATemplate_format&type=revision&diff=51666&oldid=49433)

stefahn:
I just came across this issue, since commas were missing on my wiki(s) since I updated SMW from 1.x to 2.5.5.

I use many ask queries and I didn't want to change any single one of them individually. Also it seems I didn't get the point of the valuesep thing. So I just added this line to my LocalSettings.php and all commas reappeared:
`$smwgResultFormatsFeatures = SMW_RF_NONE;`

However I find it strange that default output behaviour changes, when I update SWM.
Why is the default value of `sep` empty and not a comma?

kghbln:
> Also it seems I didn't get the point of the valuesep thing.

The example should make it clear: https://sandbox.semantic-mediawiki.org/wiki/Template_sep_valuesep

> However I find it strange that default output behaviour changes, when I update SMW.

Sometimes changes have to be made to bring more goodness. You always get occasional default setting changes, not just with SMW but e.g also with MW, but there is in most cases a way to retain the previous bahaviour to allow for migrating and you indeed found it: https://www.semantic-mediawiki.org/wiki/Help:$smwgResultFormatsFeatures

Perhaps in this case we should probably have added this to a breaking changes section in the RELEASE NOTES.

Issue #972, opened 2015-04-09
Template named args to require '?' as prefix


mwjames:
Having read [0] with "This behavior seems odd to me because it prevents SMW from using templates that are also meant for normal use by users without modifying those templates to accept alternative parameters with question marks ...", I was wondering about the reason as to why `?` is required.

The line that makes it a requirement can be seen here  (`$fieldName = '?' . $field->getPrintRequest()->getLabel();`)  [1] but no obvious technical reason can be cited to make this a hard requirement. @kghbln Any idea why `?` was required when this feature was added?

Using `Template:ListOutput` manually
```
{{ListOutput
 |TemplateListOutput-1=Foo
 |TemplateListOutput-2=bar
}}
```
should be no different as in using it for/by
```
{{#ask: [[TemplateListOutput-1::+]]
 |?TemplateListOutput-1
 |?TemplateListOutput-2
 |template=ListOutput
 |format=template
 |named args=yes
 |sep=;
}}
```

[0] 	https://semantic-mediawiki.org/wiki/Thread:Help_talk:Template_format/Clarifying_something_with_named_args

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/queryprinters/ListResultPrinter.php#L408

kghbln:
I cannot tell but the requirement for this"?" appears pretty noticeable. So I think there was at least at that time a good reason for doing so. I guess that either @mkroetzsch or @JeroenDeDauw added this to SMW so one of them may have an info on this.

mwjames:
> I cannot tell but the requirement for this"?" appears pretty noticeable.

Well, I don't know why therefore the usage of named args for the category printer #974 is done without `?` which may seem counterintuitive to the user (in comparison to named args in template output) but I see no reason to continue something that has no merit or a clear technical specification.

kghbln:
> I see no reason to continue something that has no merit or a clear technical specification.

Neither do I. I just figured to have some information about the rationale back then may be interesting.

mkroetzsch:
The ? is to distinguish print requests from ask parameters. Without it, it would be impossible to parse ask parser function calls unambiguously. The choice of ? (as opposed to some other symbol or syntactic change) was motivated by its intuitive appeal, since printouts are part of what you are asking for.

mkroetzsch:
[...]

PR #2024, opened 2016-11-19
ListResultPrinter / Add `template arguments` parameter, refs 972

mwjames:
This PR is made in reference to: #972, [0]

- Adds the `template arguments` parameter with `'numbered', 'named', 'legacy'` where numbered being `{{{1}}}`, named is interpret as `{{{Has test}}}`, and legacy to require the `?` `{{{?Has text}}}` which is equal to `named args`
- `named args` is used with priority (to avoid any incompatibilities with existing output) even in cases where `|template arguments=named` is denoted

This PR includes:
- [x] CI build passed

[0] https://semantic-mediawiki.org/wiki/Thread:Help_talk:Template_format/Clarifying_something_with_named_args

### DIFF
--- a/includes/queryprinters/ListResultPrinter.php
+++ b/includes/queryprinters/ListResultPrinter.php
@@ -403,11 +403,18 @@ class ListResultPrinter extends ResultPrinter {
 			$value = '';
 			$fieldName = '';
 
-			if ( $this->mNamedArgs ) {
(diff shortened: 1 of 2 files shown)


mwjames:
@kghbln FYI

mwjames:
http://sandbox.semantic-mediawiki.org/wiki/Issue/2024

PR #2029, opened 2016-11-19
Reorganize how DefaultSettings are loaded

mwjames:
This PR is made in reference to: # [0]

- Only `"SemanticMediaWiki.php"` is autoloaded in `composer.json`
- `DefaultSettings.php` returns an array and is assigned in `SemanticMediaWiki::load` to $GLOBALS and can as in case of [0] be modified to adhere custom specific processing 

This PR includes:
- [x] CI build passed

[0] https://github.com/Wikia/app/pull/11922

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -16,1184 +16,1210 @@ if ( !defined( 'MEDIAWIKI' ) ) {
   die( "This file is part of the Semantic MediaWiki extension. It is not a valid entry point.\n" );
 }
 
-###
-# This is the path to your installation of Semantic MediaWiki as seen on your
-# local filesystem. Used against some PHP file path issues.
-# If needed, you can also change this path in LocalSettings.php after including
-# this file.
-##
-$GLOBALS['smwgIP'] = dirname( __FILE__ ) . '/';
-$GLOBALS['smwgExtraneousLanguageFileDir'] = __DIR__ . '/i18n/extra';
-##
+return array(
 
-###
-# Semantic MediaWiki's operational state
-#
-# It is expected that enableSemantics() is used to enable SMW otherwise it is
-# disabled by default. disableSemantics() will also set the state to disabled.
-#
-# @since 2.4
-##
-$GLOBALS['smwgSemanticsEnabled'] = false;
-##
+	###
+	# This is the path to your installation of Semantic MediaWiki as seen on your
+	# local filesystem. Used against some PHP file path issues.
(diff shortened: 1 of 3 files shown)


DefaultSettings.php
```diff
@@ -16,1184 +16,1210 @@
...
-###
-# Settings for inline queries ({{#ask:...}}) and for semantic queries in
-# general. This can especially be used to prevent overly high server-load due
-# to complex queries. The following settings affect all queries, wherever they
-# occur.
-##
-$GLOBALS['smwgQEnabled'] = true;   // (De)activates all query related features and interfaces
-$GLOBALS['smwgQMaxLimit'] = 10000; // Max number of results *ever* retrieved, even when using special query pages.
-$GLOBALS['smwgIgnoreQueryErrors'] = true; // Should queries be executed even if some errors were detected?
+return array(
```
JeroenDeDauw: y u no `[]`

load.php
```diff
@@ -8,26 +8,60 @@
...
+		include_once __DIR__ . '/src/Defines.php';
+		include_once __DIR__ . '/src/GlobalFunctions.php';
+
+		$settings = include __DIR__ . '/DefaultSettings.php';
+
+		foreach ( $settings as $key => $value ) {
+			$GLOBALS[$key] = $value;
+		}
+
+		unset( $settings );
```
JeroenDeDauw: huh... if you don't want it in scope, then simply don't put it in it in the first place. You can inline the include in the foreach predicate, or do that and factor the include into a getDefaultSettings function, though I think the later is overkill here

DefaultSettings.php
mwjames: For now I rely on `array` to avoid another 2.4.2 back-port issue.

load.php
mwjames: > the include into a getDefaultSettings function, though I think the later is overkill here

Wikia/app#11922 changed the settings file to get access to individual settings hence the anticipation of this by providing that in a form without someone having to edit DefaultSettings.

They have to make some changes eventually but I'd rather see them to do those changes in this file with less potential of conflict.

mwjames: >  You can inline the include in the foreach predicate,

I experimented with that but somehow the readability suffered.

JeroenDeDauw: Not sure you got what I suggested. Change here: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2035

PR #2036, opened 2016-11-20
StringValueFormatter / Allow to reduce the text output length

mwjames:

- A simple numeric value added to the output modifier `|?Has text#20` is interpret as requested length and hereby reduces the output text (if the overall string length is larger than the requested length then an additional `…` is displayed)
- Pending the last word to the requested length, the length may vary given that a whole word is returned as to the requested length (this does not apply for languages that use a different/no word separation)

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2036

PR #2037, opened 2016-11-20
ResultFieldMatchFinder / Support |+lang= filter in PrintRequests

mwjames:
This PR is made in reference to: #1344

- `|+lang=` is to support filtering of results that match a specific language and is applied during query post-processing after the condition has retrieved matchable entities. (The `PrintRequest` and hereby the DV may match text in different languages and unless the query condition was designed in a way that would made them reducible, it would display them all)
- Using the post-filtering mechanism allows that a query condition can be indifferent towards possible language matches, yet enables to only select the one that match the denoted `lang` value
- `|+lang=` filtering is only applied to instances of type `Monolingual text`

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/dataitems/SMW_DI_Container.php
+++ b/includes/dataitems/SMW_DI_Container.php
@@ -4,6 +4,8 @@
  */
 
 use SMW\DataItemException;
+use SMW\DIProperty;
+use SMWDIBlob as DIBlob;
 
 /**
  * Subclass of SMWSemanticData that is used to store the data in SMWDIContainer
@@ -178,6 +180,18 @@ class SMWDIContainer extends SMWDataItem {
 
 	public function getSortKey() {
 		return '';
+	}
+
+	/**
+	 * @since 2.5
+	 *
+	 * @param string $sortKey
+	 */
+	public function addCompositeSortKey( $sortKey ) {
+		$this->m_semanticData->addPropertyObjectValue(
+			new DIProperty( '_SKEY' ),
+			new DIBlob( $this->m_semanticData->getSubject()->getSortKey() . '#' . $sortKey )
+		);
 	}
 
 	public function getSerialization() {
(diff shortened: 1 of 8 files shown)


mwjames:
@kghbln FYI

http://sandbox.semantic-mediawiki.org/wiki/Issue/2037

kghbln:
> |+lang= filtering is only applied to instances of type Monolingual text

Does this also make sense, though in a reduced degree to "Has property description"? like e.g. [here on sandbox](http://sandbox.semantic-mediawiki.org/wiki/Main_Page#Sandbox%20provided%20imported%20vocabulary)

Going further I think it will be nice to use this feature it in conjunction with a variable which is providing the language code from the users preference setting. I will have a look if there is an extension for this around.

kghbln:
> I will have a look if there is an extension for this around.

Great, found it: [MyVariables](https://www.mediawiki.org/wiki/Extension:MyVariables)

mwjames:
> Does this also make sense, though in a reduced degree to "Has property description"?

Lucky us, `Has property description` and `Has preferred property label` are defined as Monolingual text type hence `|+lang=` can be applied here as well.

kghbln:
> Lucky us, Has property description and Has preferred property label are defined as Monolingual text type hence |+lang= can be applied here as well.

I have the feeling that this does not just work by chance. ;)

> Great, found it: MyVariables

Now also available on sandbox.

kghbln:
@mwjames 

Hmm, I did a [field test with "Has property description"](https://sandbox.semantic-mediawiki.org/wiki/Issue/2037/Has_property_description) While it works for "fr" i.e. showing nothing is only works for the first property "Foaf:homepage" and not for "Foaf:knows" or "Foaf:name" when using "en" and "de". The latter two show both languages.

mwjames:
> ... not for "Foaf:knows" or "Foaf:name" when using "en" and "de". The latter two show both languages.

This happens when you only test half a meal and declare it as delicious. Should be fixed with #2048 containing a similar test case.

kghbln:
> Should be fixed with #2048 containing a similar test case.

It works! Perfect! :+1:

PR #2044, opened 2016-11-26
Modernize SpecialAdmin

mwjames:

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI, I needed space for an upcoming PR therefore some links have been reshuffled with all functions now being embedded in the skin output.

image

image

image

mwjames:
To a certain degree this should also address https://github.com/Wikia/app/pull/11916. If `smwgAdminRefreshStore` is disabled then no active connection is established to the `JobQueueLookup`.

mwjames:
As for testing:

- Follow all links and press all buttons to confirm that all previous functions do work as before
- Test with `smwgAdminRefreshStore` enabled/disabled

kghbln:
Overall comment :+1:  :)

> Follow all links and press all buttons to confirm that all previous functions do work as before

:+1: 

> Test with smwgAdminRefreshStore enabled/disabled

it works but: 

- case "enabled"  
When I click the button I get the French version of "smw_smwadmin_datarefreshprogress" though a translated German version ships with the code. Still I have the feeling that this may not be a regression introduced with this pull.

- case "disabled"  
It will be nice if the button not just vanishes but also to have a message showing up saying :"You do not have the permission on this wiki to start a data refresh from this special page". This is not a regression but an enhancement.

Further comments:
- I have the feeling that the breadcrumb allowing me to return to this special page is too small in font size. Also it could say "Return to Administrative Functions". Still it is good to have the return to link on top of the page.
- The Supplementary Functions could probably be listed only in one wide column instead of two.

res/smw/ext.smw.css
```diff
@@ -498,6 +498,15 @@ label.smw-form-checkbox {
           column-gap: 1em;
 }
 
+.smw-column-twofold-responsive {
+  -webkit-columns: 2 20em;
```
mwjames: > The Supplementary Functions could probably be listed only in one wide column instead of two.

@kghbln FYI

src/MediaWiki/Specials/Admin/LinkSection.php
```diff
@@ -0,0 +1,124 @@
...
+
+	/**
+	 * @since 2.5
+	 *
+	 * @return string
+	 */
+	public function getForm() {
+		$html =	Html::rawElement( 'h2', array(), Message::get( array( 'smw-smwadmin-information-section-title' ), Message::TEXT, Message::USER_LANGUAGE ) );
+		$html .= Html::rawElement( 'p', array(), Message::get( array( 'smw-smwadmin-information-section-intro' ), Message::TEXT, Message::USER_LANGUAGE ) );
+		$html .= Html::rawElement( 'div', array( 'class' => 'smw-column-twofold-responsive' ),
```
mwjames: > The Supplementary Functions could probably be listed only in one wide column instead of two.

@kghbln

src/MediaWiki/Specials/Admin/OutputFormatter.php
```diff
@@ -0,0 +1,136 @@
...
+
+		if ( defined( 'JSON_PRETTY_PRINT' ) ) {
+			return json_encode( $input, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE );
+		}
+
+		return FormatJson::encode( $input, true );
+	}
+
+	private function createParentLink() {
+		return Html::rawElement(
```
mwjames: > breadcrumb allowing me to return to this special page is too small in font size. Also it could say "Return to Administrative Functions"

Function that creates the breadcrumb on all pages.

src/MediaWiki/Specials/Admin/DataRepairSection.php
```diff
@@ -0,0 +1,194 @@
...
+			$prog = $refreshjob->getProgress();
+
+			$progressBar = Html::rawElement(
+				'div',
+				array( 'style' => 'float: left; background: #DDDDDD; border: 1px solid grey; width: 300px;' ),
+				Html::rawElement( 'div', array( 'style' => 'background: #AAF; width: ' . round( $prog * 300 ) . 'px; height: 20px; ' ), '' )
+			);
+
+			$this->htmlFormRenderer
+				->addParagraph( Message::get( 'smw_smwadmin_datarefreshprogress' ) )
```
mwjames: >  I get the French version of "smw_smwadmin_datarefreshprogress"

@kghbln Should be `Message::get( 'smw_smwadmin_datarefreshprogress', Message::TEXT, Message::USER_LANGUAGE )`

kghbln:
Then looking a https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2046 we should also have a legend for the operational statistics page. Actually I thought about this before but forget to mention it.

res/smw/ext.smw.css
kghbln: Great, so there is a class to manipulate this locally. :+1:

mwjames: #2056 uses `smw-admin-supplementary-linksection`

src/MediaWiki/Specials/Admin/LinkSection.php
mwjames: #2056 uses `smw-admin-supplementary-linksection`

src/MediaWiki/Specials/Admin/DataRepairSection.php
mwjames: Fixed with #2056

mwjames:
To clarify `Special:SMWAdmin` can still be used and was made an alias of `Special:SemanticMediaWiki` since using the ambiguous `Special:Admin` name would most likely cause an issue in case another extension uses that name for a special page.

PR #2052, opened 2016-11-26
Escape some messages shown in Special:Browse

Nikerabbit:
This PR is made in reference to: https://phabricator.wikimedia.org/T85864

- Proper escaping of some interface messages in Special:Browse

This PR includes:
- [ ] CI build passed

This does not fix escaping of `smw-datavalue-wikipage-empty` or `smw-browse-invalid-subject` because SMW\Message does not seem to  support sensible escaping of parameters such as MediaWiki's Message class' rawParam.

### DIFF
--- a/src/MediaWiki/Specials/Browse/HtmlContentBuilder.php
+++ b/src/MediaWiki/Specials/Browse/HtmlContentBuilder.php
@@ -336,7 +336,7 @@ class HtmlContentBuilder {
 
 		if ( $noresult ) {
 			$html .= "<tr class=\"smwb-propvalue\"><th> &#160; </th><td><em>" .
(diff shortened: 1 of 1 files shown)


kghbln:
Thanks for the pr and notifying about the issue. Recorded here: https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2053

mwjames:
Thanks.

> This does not fix escaping of smw-datavalue-wikipage-empty or smw-browse-invalid-subject because SMW\Message does not seem to support sensible escaping

I'll give it some thinking. Reason for our `Message` is to keep the reference on the key and arguments from a process that triggers such a message (not the literal message itself, it gets only translated when an output makes a request).

`return $this->addErrorMsg( array( 'smw-datavalue-wikipage-empty' ), Message::ESCAPED );`

mwjames:
- `'smw-datavalue-wikipage-empty'` done with 6602295815d77

Issue #1855, opened 2016-09-21
@@@ syntax to generate a property link


mwjames:
Creating a property link (as to be used in infoboxes or by templates) within a wiki-text requires some additional work and even though it can be done it will not allow to take advantage of auto-property descriptions (displayed as tooltip etc.) or in future to change to a preferred label (based on the user language).

To generate a property link within a text or template, the special syntax `[[Foo::@@@]]` is introduced, where `@@@` will signal to the `InTextAnnotationParser` to handle it as non-value annotation and instead output a property link as seen in the examples.

Going with the special syntax seems more appropriate (and is less work) than introducing yet another parser function.

## Examples

![image](https://cloud.githubusercontent.com/assets/1245473/18569275/e576eb12-7ba1-11e6-96c2-caef7a89ad18.png)

![image](https://cloud.githubusercontent.com/assets/1245473/18569314/1d94f552-7ba2-11e6-85af-dd3ab3186427.png)

![image](https://cloud.githubusercontent.com/assets/1245473/18569445/e6aa40b4-7ba2-11e6-9176-ea35448f84e6.png)

mwjames:
@kghbln FYI

kghbln:
* https://www.semantic-mediawiki.org/wiki/Help:Property_links

krabina:
> Placeholder: https://www.semantic-mediawiki.org/wiki/Help:Property_links

I added some more wikidoku for this....

krabina:
This is a great feature, but there are several use cases where in you wiki you don't want user to visit property pages. 

So my question is if it would be possible to use the feature of having the preferred property label and property description showing up in a tooltip, but without the possibility to actualle click on it an got to the property page?
Could be achieved by something like [[Foo::@@@|@@@]].

The workaround is to use something like
`{{#show:Property:Foo|?has preferred property label|+lang={{USERLANGUAGECODE}}}}` 
to display the correct name of the poperty and then also do the tooltip in a similar way which is a lot of coding and querying in templates...

mwjames:
> So my question is if it would be possible to use the feature of having the
> preferred property label and property description showing up in a tooltip,
> but without the possibility to actualle click on it an got to the property
> page?
> Could be achieved by something like [[Foo::@@@|@@@]].

If you create a feature request for `[[Foo::@@@|@@@]]` (where `|@@@]]`
indicates a no link) then I may have a look at it.

krabina:
[...]

PR #2069, opened 2016-12-03
Add PropertyValue::OPT_HIGHLIGHT_LINKER

mwjames:
This PR is made in reference to: #1855 

This PR includes:
- [x] CI build passed


mwjames:
image

Ensures that a context can be set where the linker is to generate an appropriate output.

Issue #2016, opened 2016-11-18
Avoid pulling in too new PHP code in the build script


JeroenDeDauw:
For https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/build/release/build_tarballs.sh, it will pull in SMW dependencies needing [insert new PHP version] if the build machine supports it, thus breaking all consumers that don't have this new version.

legoktm:
I believe you can use <https://getcomposer.org/doc/06-config.md#platform> to set your lowest supported PHP version, and then it'll pull in dependencies that match that platform requirement.

JeroenDeDauw:
This script needs to be modified to insert the platform information into the composer.json file and then do a composer update https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/build/release/build_tarballs.sh

kghbln:
I knew the file but did on think of

> This script needs to be modified to insert the platform information into the composer.json file

So something like
```sh
PHPVERSION="$3"

if [ "$PHPVERSION" == "" ]; then
	sed -e '"support": {\[.*\]},/a "platform": { "php": "${PHPVERSION}" },/n' composer.json' 
fi
```

legoktm:
Might be easier to just manipulate the JSON file with PHP instead of trying to do it in sed? :)

kghbln:
> Might be easier to just manipulate the JSON file with PHP instead of trying to do it in sed? :)

So my suggestion does not work which also does not come as a surprise. Just tried to get painkillers moving. ;) I guess I should rather leave this to Jeroen.

JeroenDeDauw:
I actually tried the sed approach but this bash magic really is no my thing

> Might be easier to just manipulate the JSON file with PHP instead of trying to do it in sed? :)

This requires changing the bash script into a PHP script, so perhaps not

legoktm:
Is using PHP a deal breaker? You could also use `php -r ...` or I suspect you could probably use jq if you really want to use bash.

kghbln:
Thought it will have been nice I do not see this as a hold up for SMW 2.4.4 since I can add the platform magic prior to creating the release tarballs.

Why is bash so bad? Is is just quaint or so? I think it might just be a matter of figuring out the regex fu for sed.

JeroenDeDauw:
I'm fine with the script being PHP. PR welcome.

[...]

PR #2081, opened 2016-12-09
Update composer.json

kghbln:
This PR is made in reference to: #2016 

- Fakes platform packages, i.e. the minimum requirement for php for building our tarballs.

This PR includes:
- [ ] CI build passed

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,9 @@
 		"wiki": "https://www.semantic-mediawiki.org/wiki/",
 		"source": "https://github.com/SemanticMediaWiki/SemanticMediaWiki"
 	},
+	"platform": {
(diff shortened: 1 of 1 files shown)


kghbln:
Should probably also be backported to 2.4.x

JeroenDeDauw:
Eh... Don't think we should include this, as it will prevent pulling in packages that require PHP > 5.3 and will stick with obsolete versions. We somehow want this to be included for the tarball build script though.

kghbln:
> Eh... Don't think we should include this, as it will prevent pulling in packages that require PHP > 5.3 and will stick with obsolete versions.

Ah, I get it. No worries.

Issue #2078, opened 2016-12-09
MySQL 5.7 causes "SELECT list; this is incompatible with DISTINCT"


kghbln:
### Setup and configuration

- MediaWiki 	1.27.1 (8c593e7) 23:49, 26 October 2016
- PHP 	7.0.8-0ubuntu0.16.04.3 (apache2handler)
- MySQL 	5.7.16-0ubuntu0.16.04.1
- Semantic MediaWiki 2.4.3 (https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/6661acdd8c8420a7e07c178aad47590d086c83d6) as well as master (https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/fd8ea67a7c66aa58513b3b67ee24b4182504dc09)

### Issue
During the server upgrade [from Ubuntu 15.10 to Ubuntu 16.04.1](https://github.com/SemanticMediaWiki/semantic-mediawiki.org/issues/1) the database server was also upgraded from version 5.6 to 5.7 which now results in:
```
Exception encountered, of type "RuntimeException"
[273d99a4e9af388f22675441] /wiki/Semantic_MediaWiki RuntimeException from line 186 of /.../w/extensions/SemanticMediaWiki/src/MediaWiki/Database.php: A database error has occurred. Did you forget to run maintenance/update.php after upgrading? See: https://www.mediawiki.org/wiki/Manual:Upgrading#Run_the_update_script
```
### Query error
```
Query: SELECT DISTINCT t6.smw_id AS id,t6.smw_title AS t,t6.smw_namespace AS ns,t6.smw_iw AS iw,t6.smw_subobject AS so,t6.smw_sortkey AS sortkey FROM `smw_object_ids` AS t6 INNER JOIN `smw_di_time` AS t1 ON t6.smw_id=t1.s_id INNER JOIN `smw_fpt_lcode` AS t3 ON t1.s_id=t3.s_id INNER JOIN `smw_di_blob` AS t4 ON t1.s_id=t4.s_id WHERE (t1.p_id='80' AND ((t3.o_hash='en')) AND ((t4.o_hash='false') AND t4.p_id='48542')) AND t6.smw_iw!=':smw' AND t6.smw_iw!=':smw-delete' AND t6.smw_iw!=':smw-redi' ORDER BY t1.o_sortkey DESC LIMIT 9
[...]

PR #2082, opened 2016-12-10
Use distinct sortfield list (MySQL 5.7), refs 2078

mwjames:
This PR is made in reference to: #2078

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/SQLStore/QueryEngine/QueryEngine.php
+++ b/src/SQLStore/QueryEngine/QueryEngine.php
@@ -353,13 +353,17 @@ class QueryEngine implements QueryEngineInterface {
 		$connection = $this->store->getConnection( 'mw.db.queryengine' );
 		list( $startOpts, $useIndex, $tailOpts ) = $connection->makeSelectOptions( $sqlOptions );
 
+		$sortfields = implode( $qobj->sortfields, ',' );
+		$sortfields = $sortfields ? ', ' . $sortfields : '';
+
 		$sql = "SELECT DISTINCT ".
 			"$qobj->alias.smw_id AS id," .
 			"$qobj->alias.smw_title AS t," .
 			"$qobj->alias.smw_namespace AS ns," .
 			"$qobj->alias.smw_iw AS iw," .
 			"$qobj->alias.smw_subobject AS so," .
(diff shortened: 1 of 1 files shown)


mwjames:
@kghbln This one needs manual confirmation since we cannot test 5.7 (only 5.6) on Travis and MariaDB 10.1 doesn't choke on this very issue.

kghbln:
@mwjames The field test tells that we are back in business. I can however only test SMW and not related extensions since as always Composer does not allow to mix dev with stable. So we will need a backport and release 2.4.4 if we want to continue with the whole lot on smw.o.

Only one thing is realise are there nasty commas on [Version history](https://www.semantic-mediawiki.org/wiki/Help:SMW_History) we are now getting. This may however be unrelated to this issue here.

kghbln:
I will be off for about 2 hours.

mwjames:
> Only one thing is realise are there nasty commas on Version history we are now getting. This may however be unrelated to this issue here.

Relates to #2022 where when no `sep=` is defined it falls back to `,` (let's discussion this on #2022)

mwjames:
> So we will need a backport and release 2.4.4 if we want to continue with the whole lot on smw.o.

Okay, I'll make a backport but leave the admin stuff (tags etc.) to you guys.

kghbln:
@mwjames Thanks a lot for your immediate fix!

mwjames:
> @mwjames Thanks a lot for your immediate fix!

The backport fails on Travis so I need to spend time on it (d....)!

kghbln:
> The backport fails on Travis so I need to spend time on it (d....)!

That's indeed truly desagreable. However, perhaps there are good changes that this will also show a way for master as soon as we test against MySQL 5.7 when using Ubuntu 16.04.

mwjames:
> That's indeed truly desagreable. 

I found the error, this happens when you need to copy line by line because master has already changed quite a bit.

kghbln:
> I found the error

I am glad that is was not too time consuming.

mwjames:
> I am glad that is was not too time consuming.

Travis went green https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/cb75adb6fa9cd3d96926ba975806c4ce278fdaf1.

Issue #2090, opened 2016-12-10
template format / `sep` parameter falling back to default when not specified.


kghbln:
### Setup and configuration

- MediaWiki 	1.27.1 (8c593e7) 23:49, 26 October 2016
- PHP 	7.0.8-0ubuntu0.16.04.3 (apache2handler)
- MySQL 	5.7.16-0ubuntu0.16.04.1
- Semantic MediaWiki master (https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/999fcb76b8697f1c168bc9951715eb425edd4e5f)

### Issue
Originally posted here: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2022#issuecomment-266194451

By coincidence, @kghbln found:

![image](https://cloud.githubusercontent.com/assets/1245473/21072527/7df1849e-bec6-11e6-86e6-b17aff1ab078.png)

with no `sep=` parameter therefore falling back to `,` as default.

```
{{#ask:
 [[Category:SMW releases]]
 [[Language code::en]]
 [[Is duplicate page::false]]
 [[Version::!Semantic MediaWiki 1.6.2]]
 |?Version
 |?Has release date#ISO
 |?Has highlights
 |?Is compatible with
 |format=template
 |sort=Version, Has release date
 |order=descending
 |limit=100
 |introtemplate=Version history intro
 |template=Version history
 |outrotemplate=Version history outro
 |intro=Here is the complete <b>version history</b>:
}}
```

kghbln:
Quoting @mwjames https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2022#issuecomment-266207046

Also what I need is some kind of baseline that defines the behavior to avoid any future regression. We caught this just by mere chance!

```
{{#ask: [[List::2022]]
 |?Has page
 |?Has page
 |format=ul
}}
```
Expects to output `Some, Some`
 
Issue: The default `sep` is set to `,` which is a bit tricky because it cannot be decided whether a user made a deliberate decision or not. I will change the default to be empty and in case it is a list/ul/ol then the output will use `,` as default if no other `sep` was given. `template` by default will have no `sep` to avoid the case we just found.

kghbln:
> Issue: The default `sep` is set to 

Sounds reasonable to me.

kghbln:
Works perfect! Thanks a lot!

PR #2091, opened 2016-12-10
Template format, use empty sep when not specified, refs 2022, 2090

mwjames:
This PR is made in reference to: #2022, #2090

This PR includes:
- [x] CI build passed


mwjames:
@kghbln `f-0104.json` contains some baseline tests for `format=list, ul, ol, template` which is similar to what can be seen at [0] (additional test definitions are welcome, just follow the example).

[0] https://sandbox.semantic-mediawiki.org/wiki/Issue/2090

Issue #250, opened 2014-03-28
smw_querytoolarge needs PLURAL support


JeroenDeDauw:
Import from https://phabricator.wikimedia.org/T20515

mediawiki:smw_querytoolarge should be using PLURAL,
depending on the number of items in the list in $1.

--  Purodha Blissenbach 2009-04-19 09:34:03 UTC

mwjames:
@kghbln ?

kghbln:
@JeroenDeDauw Still have no idea how SMW compiles this message.

mwjames:
@kghbln Can we kick this one?

kghbln:
We need an additional parameter here that allows to count the number of query conditions displayed in $1 in order to be able to change

"smw_querytoolarge": "The following query conditions <code>$1</code> could not be considered due to this wiki's restrictions on query size or depth."

to

"smw_querytoolarge": "The following query {{PLURAL:$2|condition|conditions}} <code>$1</code> could not be considered due to this wiki's restrictions on query size or depth."

While changing the system message as such to support plural is no problem for me I do not know how to make SMW count the number of different conditions.

Since the error reporting was improved in SMW 2.4 it might not indeed be the right time to tackle this issue.

kghbln:
Back when I figured that the same applies to the messages "smw_noqueryfeature", "smw_noconjunctions" and "smw_nodisjunctions". Nowadays I believe this is not the case since `$1` in these three messages just present a uncountable chuck of "ignored" ask query code.

PR #2097, opened 2016-12-10
Count log (PLURAL support), refs #250

mwjames:
This PR is made in reference to: #250

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/query/SMW_Query.php
+++ b/includes/query/SMW_Query.php
@@ -365,7 +365,8 @@ class SMWQuery implements QueryContext {
 			if ( count( $log ) > 0 ) {
 				$this->errors[] = Message::encode( array(
 					'smw_querytoolarge',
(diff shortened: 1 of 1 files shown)


includes/query/SMW_Query.php
```diff
@@ -365,7 +365,8 @@ public function applyRestrictions() {
 			if ( count( $log ) > 0 ) {
 				$this->errors[] = Message::encode( array(
 					'smw_querytoolarge',
-					str_replace( '[', '&#x005B;', implode( ', ', $log ) )
+					str_replace( '[', '&#x005B;', implode( ', ', $log ) ),
+					count( $log )
 				) );
```
mwjames: @kghbln This one should give you a count value.

kghbln: @mwjames Well if one knows how to do it when it is actually not very hard at all.

PR #1251, opened 2015-11-06
Support for QueryResult caching

mwjames:
Most of the job is already done by the `QueryDependencyLinksStore` to track and update query dependencies.

To eliminate unnecessary SQL/SPARQL connections `QueryResultCache` ought to cache a subject list returned from the `QueryEngine`.

We don't cache the string result of a printer == it means we don't interfere with how the printer manipulates the data displayed, `QueryResultCache` as the name suggests caches the return object (aka `QueryResult`) from the QueryEngine before it is forwarded to an individual printer.

If for some reason `QueryDependencyLinksStore` is not enabled then auto-invalidation of the `QueryResultCache` items can not occur and instead (same as of now, meaning manual intervention using the purge button) setting `$GLOBALS['smwgQueryResultCacheRefreshOnPurge'] = true;` can
be used so that during an purge action event the cache is invalidated for all queries stored with a corresponding article (aka subject).

image
## Features and limitations
- `limit=0`, `format=debug`, and `Special:Ask` queries are not cached
- `redis` is the preferred external StoreEngine in order to avoid issues during serialization as well as in terms of the amount of data to be stored/requested
- `$GLOBALS['smwgQueryResultCacheType']` is set to `CACHE_NONE` which means the feature is disabled, choosing an appropriate type is left to a administrator
- `$GLOBALS['smwgQueryResultCacheLifetime'] = 60 * 60 * 24; // a day` declares the lifetime of a cached item

refs  #1035, #1117

PR #2099, opened 2016-12-11
Improve computation of query signature (Description::getHash)

mwjames:
This PR is made in reference to:  #1251

- Using the plain query string `[[Foo::123]][[Bar::abc]]` as part of the signature (among limit, offset, sort/order) is not good enough to filter queries with the same conditional expression because `[[Foo::123]][[Bar::abc]]` would yield a different hash compared to `[[Bar::abc]][[Foo::123]]`
- `[[Foo::123]][[Bar::abc]]` and `[[Bar::abc]][[Foo::123]]` are equal expressions in terms of the conditional match process therefore a signature should account for that as well to ensure that queries can sufficiently be cached or compared
- `Description::getHash` will take into account the canonization and uses a sorted compound to represent a signature

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/query/SMW_Query.php
+++ b/includes/query/SMW_Query.php
@@ -445,21 +445,25 @@ class SMWQuery implements QueryContext {
 		// elements that directly influence the result list
 		$serialized = array();
 
-		$serialized['conditions'] = $this->getQueryString();
+		// Don't use the QueryString, use the canonized hash to ensure that
+		// [[Foo::123]][[Bar::abc]] returns the same ID as [[Bar::abc]][[Foo::123]]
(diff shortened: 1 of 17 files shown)


mwjames:
@kghbln For users who don't use the query cache nothing will change, for those who enable it queries will get a new ID in order to match those with the same signature. This will cause a temporary increase in outdated entities (_QUERY* subjects being marked as outdated due to a new assigned ID hash).

mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2099

image

kghbln:
Affirmative. Works cool on sandbox!

PR #2031, opened 2016-11-19
SearchByProperty value escape, follow-up on 2005

mwjames:
This PR is made in reference to: #2005

This PR addresses or contains:

This PR includes:
- [ ] Tests (unit/integration)
- [x] CI build passed

PR #2109, opened 2016-12-17
Special:Browse replace decode with unescape, refs #2031

mwjames:
This PR is made in reference to: #2031

- In #2031 auto-generate links were prefixed with `:` (`:Foo` doesn't represent an actual title therefore it is used as indicator) to apply internal escaping/unescaping
- `:` is used for `Special:Browse` as well to distinguish between machine/human generate links

This PR includes:
- [x] CI build passed


mwjames:
@kghbln If the sidebar contains a cached link to `Special:Browse` then it is possible that because of the missing `:` prefix encoded pages may lead to a false negative therefore simply purging the page should ensure a correct link.

[0] contains some manual linked titles that would currently create issues as some of the characters (-25, -20, -2D, %) would be tried for decoding even though there are part of the title itself.

[0] https://sandbox.semantic-mediawiki.org/wiki/Issue/2109

kghbln:
> If the sidebar contains a cached link

I disabled sidebar caching for now since this allows for more and immediate flexibility.

The pull works great but I discovered the [% being omitted](https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Parcourir/Issue/2109/%25Some@%25) form the property name.

mwjames:
> The pull works great but I discovered the % being omitted form the property name.

`@` is a reserved character which should be avoided as part of a property name (apparently I don't follow the principle when creating examples) because of `[[%Some@%::%Some@%]]`, [0] tries to match `@`.

Doing `[[%Some%::%Some@%]]` should work as expected.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/dataitems/SMW_DI_Property.php#L427-L430

kghbln:
> @ is a reserved character which should be avoided as part of a property name

I should have thought about this myself.

> Doing [[%Some%::%Some@%]] should work as expected.

:+1:

PR #2111, opened 2016-12-17
ValueDescription::getQueryString to set DV output option

mwjames:

This PR includes:
- [x] CI build passed


mwjames:
This should help to avoid some decode/encode visibility issues as in case of [0]:

image

image


[0] https://sandbox.semantic-mediawiki.org/wiki/Issue/2111

PR #2135, opened 2016-12-31
Add PropertyTypeFinder (smwgFixedProperties)

mwjames:

- Simplify `$smwgFixedProperties` customizing where the type is now read from the property instead of having it declared as customizing which would create issues when the type on the property page and the `$smwgFixedProperties` setting are different from each other
- Table name is no longer the complete md5 hash instead `substr( base_convert( md5( $tableName ), 16, 32 ), 0, 12 )` is used to get a shorter ID 
- Property page now displays a message that any change to its type requires some admin activity

```
$GLOBALS['smwgFixedProperties'] = array(
	'Has fixed test'
);
```

image

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -731,39 +731,26 @@ return array(
 	##
 
 	##
-	# These are fixed properties, i.e. user defined properties having a
-	# dedicated table for them. Entries in this array have the following format:
+	# List of user-defined fixed properties
 	#
-	# 		property_key => property_type.
+	# Listed properties are managed by its own fixed table (instad of a
+	# shared one) to allow for sharding large datasets with value assignments.
 	#
-	# The 'property_key' is the title of the property (with underscores instead
-	# of _ and capital first letter).
-	# The 'property_type' denotes the type of the property and has to be one of the following:
-	#		SMWDataItem::TYPE_BLOB
-	#		SMWDataItem::TYPE_URI
-	#		SMWDataItem::TYPE_WIKIPAGE
-	#		SMWDataItem::TYPE_NUMBER
-	#		SMWDataItem::TYPE_TIME
-	#		SMWDataItem::TYPE_BOOLEAN
-	#		SMWDataItem::TYPE_CONTAINER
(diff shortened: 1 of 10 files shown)


mwjames:
@kghbln FYI

kghbln:
How awesome is this???!!! Great and many thanks!!! Just [documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AFixed_properties&type=revision&diff=48112&oldid=47004). 

What about RDF stores? Will this work for them too? I suppose so.

kghbln:
This is a highlight thus https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/7f3fe529a7e2f6bbd9d1aec45feabe8c8884afd5

mwjames:
> This is a highlight thus 7f3fe52

At first, I just wanted to scratch an itch where a `DataItem` object reference would appear in the `LocalSettings` and I ended up fixing the whole "Fixed Properties" issue.

> Property names must use underscores and not spaces or you will see MWException errors on some page actions.

Properties without an underscore do no longer cause a `MWException`, meaning that `Has_fixed_text` and `Has fixed text` will be handled as equivalent [0].

> What about RDF stores? Will this work for them too? I suppose so.

This issue relates to the creation of extra DB tables with a RDF store not being linked to internal changes of the `SQLStore` therefore declaring properties as fixed table properties will have no effect on the `SPARQLStore` or for that matter to the RDF export.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2135/files#diff-53920baeff0ad9c3bad870946bb97034R221

kghbln:
> Properties without an underscore do no longer cause a MWException, meaning that Has_fixed_text and Has fixed text will be handled as equivalent [0].

Good to know that there are less worries. :)

> This issue relates to the creation of extra DB tables with a RDF store not being linked to internal changes of the SQLStore therefore declaring properties as fixed table properties will have no effect on the SPARQLStore or for that matter to the RDF export.

Thanks a lot for this info. I guess me having a better understanding of RDF store will be preferable in future.

PR #2137, opened 2016-12-31
UsageStatisticsListLookup to add TOTALPROPS

mwjames:

- `TotalPropertiesCount` is not to count property subobjects 
- `UsedPropertiesCount` to use an INNER JOIN to match only valid properties

image

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

Issue #1963, opened 2016-10-26
`$smwgLinksInValue` leading to PHP crashes when set to "true"


kghbln:
### Setup and configuration

- Semantic MediaWiki: 2.5.0-alpha (9a82356)
- MediaWiki: 	1.27.1 (a52d35d)
- PHP: 	5.6.27
- MariaDB: 	10.0.27

### Issue

The help page on [`$smwgLinksInValues`](https://www.semantic-mediawiki.org/wiki/Help:$smwgLinksInValues) still warns about possible PHP crashes if set the configuration parameter is set to "true" due to an eight year old [warning](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/1.3/includes/SMW_Settings.php#L101-L110).

Quote @mwjames: 

> So it is time to reevaluate the truthfulness of this statement or at least given an indication of how many links would break the wiki. For example a text with xyz characters and x amount of links in values would break the wiki for the PCRE y version.

It is however now possible to avoid changing this setting by using the `#set'` parser function as these [examples show](https://www.semantic-mediawiki.org/wiki/Demo:Links_in_values) but what about in-text annotations?

To be able to test I have changed the [configruation of sandbox](http://sandbox.semantic-mediawiki.org/w/index.php?title=SMW_Sandbox%3ASetup_information&type=revision&diff=4400&oldid=4393) to `$smwgLinksInValues = true;`.

A test page for this issue can be found [here](http://sandbox.semantic-mediawiki.org/wiki/Issue/1963).

kghbln:
On the spot which caused https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1961#issuecomment-256417048 to exit I get this in the promt:

```
(3371/3427)     Finished processing ID 3376 (Utilisateur:Lalquier)
(3373/3427)     Finished processing ID 3378 (This_page_supports_semantic_in-text_annotations_(e.g._"Is_specified_asWorld_Heritage_Site")_to_build_structured_and_queryable_content_provided_by_Semantic_MediaWiki._For_a_comprehensive_description_on_how_to_use_annotations_or_the#0#ask_parser_function,_please_have_a_look_at_the_getting_started,_in-text_annotation,_or_inline_queries_help_page.This_page_supports_semantic_in-text_annotations_(e.g._"Is_specified_asWorld_Heritage_Site")_to_build_structured_and_queryable_content_provided_)
[...]

PR #2153, opened 2017-01-08
$smwgLinksInValues to support Obfuscator approach

mwjames:
This PR is made in reference to: #1963

- Add `SMW_LINV_OBFU` as approach to resolve LinkInValues, `SMW_LINV_PCRE` is kept to avoid BC breakages for users who have `$smwgLinksInValues` enabled

This PR includes:
- [x] CI build passed


mwjames:
[0] is provided to enable parsing of so called "LinksInValues" (`[[SomeProperty::Foo [[link]] in value]]`). The problem with the current PCRE approach is that PHP can easily crash #1963 and is therefore a feature normally disabled despite its usefulness.

To avoid mentioned issues, a different approach is sought to be introduced by this PR including a different interpretation of the `$smwgLinksInValues` setting.

`$smwgLinksInValues` will except `false`, `true` (deprecated with 2.5 to keep BC, use `SMW_LINV_PCRE` instead), `SMW_LINV_PCRE`, `SMW_LINV_OBFU`.

The following text is used to test the `SMW_LINV_OBFU` approach:

`
[[Has text::[[Lorem ipsum]] dolor sit amet, [[Has page::consectetur adipiscing elit]]. In [http://example.org/eleifend eleifend] cursus dignissim. Suspendisse potenti. [[Has page::Curabitur aliquet|Bar]] malesuada ex sed efficitur. [[Nullam id ante|Foo]] eget lacus rutrum tincidunt eu et lacus. Aliquam sed consequat orci. Nulla sodales neque magna, nec sollicitudin erat aliquam ut. Sed sed vulputate tortor, in [[vestibulum nibh]] [[Has number::42]] [[Has date::01-01-2020]]]].
`

[0] https://www.semantic-mediawiki.org/wiki/Help:$smwgLinksInValues

mwjames:
[0] is provided as test instrument to verify the performance impact on a single page when `SMW_LINV_OBFU`  is enabled. [0] is expected to be copied and handled by a single page.

PS: [0] would most certainly kill the Parser (and hereby the system) if the legacy `SMW_LINV_PCRE` approach is used.

[0] https://gist.github.com/mwjames/78354d4b5d36878145dbcdee2a9daa75

mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2153

mwjames:
@kghbln FYI

kghbln:
Now [documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgLinksInValues&type=revision&diff=49651&oldid=49638) @mwjames Thanks for the poke.

kghbln:
Now [serving](https://sandbox.semantic-mediawiki.org/w/index.php?title=SMW_Sandbox%3ASetup_information&type=revision&diff=5731&oldid=5705) on sandbox.

PR #2159, opened 2017-01-14
RepositoryConnection to expose RepositoryClient, refs 1620

mwjames:
This PR is made in reference to: #1620 

- Allows the `smwgSparqlQueryEndpoint` to be `false` and in those cases will the query request redirected to the `SQLStore` query engine

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/SPARQLStore/RepositoryClient.php
+++ b/src/SPARQLStore/RepositoryClient.php
@@ -72,7 +72,7 @@ class RepositoryClient {
 	/**
 	 * @since 2.2
 	 *
-	 * @return string
+	 * @return string|false
 	 */
 	public function getQueryEndpoint() {
 		return $this->queryEndpoint;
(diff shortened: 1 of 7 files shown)


kghbln:
> Allows the smwgSparqlQueryEndpoint to be false and in those cases the query request will be redirected to the base store (== SQLStore) query engine

@mwjames Just to clarify: So this means in case the wiki is using sql instead of sparql the request gets redirected instead of ending up in void.

mwjames:
> So this means in case the wiki is using sql instead of sparql the request gets redirected instead of ending up in void.

Slightly different. A user who uses the `SPARQLStore` would normally send all query requests to the `QueryEngine` defined by the endpoint.

In case of `smwgSparqlQueryEndpoint = false` the query answering is redirected to the `SQLStore` instead so that a surge of `SPARQL` queries can be avoided. Whether this makes really sense or not is a different question.

kghbln:
Thanks a lot for clarifying!  [Documented in the wiki.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgSparqlQueryEndpoint&type=revision&diff=55914&oldid=55912)

PR #2176, opened 2017-01-15
Experimental feature to filter duplicate query segments, refs 2099

mwjames:
This PR is made in reference to: #2099

- Add `Description::getFingerprint` to compute a signature that allows to compare description instances
- Add `Description::setMembership` to track members of a distinct segment within
the description instance
- Add `smwgQFilterDuplicates` as experimental setting to optimize and filter duplicate conditions (Foo OR Foo === Foo, Foo.Bar OR Foo remains) from a query segment

This PR includes:
- [x] CI build passed


mwjames:
@kghbln In #2099 with the condition stated the query signature would change and in order for `smwgQFilterDuplicates` to work sufficiently, I had to re-adjust how the signature (aka fingerprint) is computed therefore we should see on the sandbox a small spike on new query ID's given the new model (`smwgQueryResultCacheType` or `smwgQFilterDuplicates` would force the usage of the new model, yet if both are disabled then everything remains the same for the user and the system).

The idea behind `smwgQFilterDuplicates` is to filter some query segments (part that represents individual SQL statements of a query) that can be duplicate and would create an additional temp table which in return would lengthen the time the SQL engine needs to answer the query.

`smwgQFilterDuplicates` is an experimental feature and if enabled tries to match the fingerprint of a description and replaces the reference if matched before the `QueryEngine` becomes aware of the segments.

The impact of described optimization is unknown.

kghbln:
Now [documented.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgQFilterDuplicates&type=revision&diff=51014&oldid=50003)

kghbln:
Now [live on sandbox](https://sandbox.semantic-mediawiki.org/w/index.php?title=SMW_Sandbox%3ASetup_information&type=revision&diff=5903&oldid=5731)

kghbln:
Renaming of [`$smwgQStrictComparators`](https://www.semantic-mediawiki.org/w/index.php?title=Help:$smwgQStrictComparators&curid=9369&diff=56057&oldid=56053) now documented, too.

PR #2177, opened 2017-01-15
Move SubSemanticData to its own class

mwjames:

- Moves `SubSemanticData` operations to its own `SMW\DataMode\SubSemanticData` component
- Moves the `ContainerSemanticData` class from `DIContainer` to `SMW\DataMode\ContainerSemanticData`

This PR includes:
- [x] CI build passed


mwjames:
@kghbln This can cause a "Fatal error: Call to a member function getSubSemanticData() on array in /var/www/htdocs/mw/02100/w/extensions/SemanticMediaWiki/includes/SemanticData.php on line 282" because the `ParserCache` (owned by MW) contains a serialized version of the  `SemanticData` when it was a simple `array`.

mwjames:
I try to come up with a contingency for cases where the object contains an "old" array object instead of the `SubSemanticData`.

PR #2227, opened 2017-02-04
Check codified property specification requirements, refs 2221

mwjames:
This PR is made in reference to: #2221 

This PR addresses or contains:

- Show users clear messages on the property page in case some codified requirements (in `PropertySpecificationReqExaminer`) are missing from the property specification
- Move code for the HTML generation to `PropertyPageMessageHtmlBuilder`

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

PR #2232, opened 2017-02-04
Support for "Is edit protected", refs 2227

mwjames:
This PR is made in reference to: #2227

- Adds `Is edit protected` as an administrative property to restrict edit access to an annotated entity (page) to avoid accidental data modification

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1371,4 +1371,27 @@ return array(
 	'smwgQueryResultCacheRefreshOnPurge' => true,
 	##
 
+	###
+	# Protect page edits
+	#
+	# To prevent accidental changes of content especially to those of property
+	# definitions, this setting allows with the help of the `Is edit protected`
+	# property to prevent editing on pages that have annotate the property with
+	# `true`.
+	#
+	# Once the property is set, only users with the listed user right are able
+	# to edit and/or revoke the restriction on the selected page.
+	#
+	# `smw-pageedit` has been deployed as extra right to be distinct from existing
+	# edit protections
+	#
+	# To enable this functionality either assign `smw-pageedit` or any other
+	# right to the variable to activate an edit protection.
+	#
+	# @since 2.5
+	# @default false
+	##
+	'smwgEditProtectionRight' => false,
+	##
+
 );
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -338,6 +338,7 @@
 	"smw-qp-empty-data": "Requested data could not be displayed due to some insufficient selection criteria.",
 	"right-smw-admin": "Access administration tasks (Semantic MediaWiki)",
 	"right-smw-patternedit": "Edit access to maintain allowed regular expressions and patterns (Semantic MediaWiki)",
+	"right-smw-pageedit": "Edit access for <code>Is edit protected</code> annotated pages (Semantic MediaWiki)",
 	"action-smw-patternedit": "edit regular expressions used by Semantic MediaWiki",
 	"group-smwadministrator": "Administrators (Semantic MediaWiki)",
 	"group-smwadministrator-member": "{{GENDER:$1|administrator (Semantic MediaWiki)}}",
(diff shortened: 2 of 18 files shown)


mwjames:
For sometime now, I was locking for a smart (== none intrusive, yet powerful) way to prevent users from accidental changing data or property specifications, especially when you have authority data some kind of check and balance should hinder a change to a specification to avoid altered types or worse have those changes flood the job queue and invalidate depending data sets.

With #2227 (`PropertySpecificationReqExaminer`, `PropertyPageMessageHtmlBuilder`), I was able to add a feature to 2.5 that is small in terms of LOC but does what is required to hinder accidental data modifications.

## Workflow 

The workflow as very simple, "... any user is qualified to add the `Is edit protected` property to a subject, yet only a user with a dedicated permission can edit or revoke the protection to an entity after it has been added ..."

A user without an appropriate right (which is maintained in  `smwgEditProtectionRight`) will be given a notice and is unable to edit, delete, or move that page.

image

In case `smwgEditProtectionRight` is disabled (== `false`), the `Is edit protected` property page will notify a user about the circumstances.

image

## Technical notes

- `Is edit protected` is a predefined property
- `smwgEditProtectionRight` is added as setting to disable, enable the functionality and provide the name of the right to check against
- `smw-pageedit` is deployed as extra right to distinguish it from any other right and is assigned to the `smwcurator` role by default
- Lookup and validation of the `Is edit protected` property is handled by `EditProtectionValidator` and is optimized to cache a request any minimize any Store/DB connections
- Only the `TitleQuickPermissions` hook  is used to control and influence the protection behaviour  because "... checked first in the `Title::checkQuickPermissions` ... are the most basic of permissions needed to perform ..."

mwjames:
@kghbln FYI, Once merged it would be marvelous if you could add `$GLOBALS['smwgEditProtectionRight'] = 'smw-pageedit';` to the configuration.

PS: I haven't restricted the functionality to only work on the Property NS which means that any page can be protected using the `Is edit protected` annotation.

kghbln:
I have finally done this. Utmost cool! Especially fixed properties and properties of imported vocabularies are born to be used together with this  feature.

kghbln:
Documentation:

- [ ] User groups and right - [done](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AUser_rights_and_user_groups&type=revision&diff=49130&oldid=48352)
- [ ] Special property
- [ ] Configuration parameter

mwjames:
> I have finally done this. Utmost cool! 

I'll try to squeeze in a change to have it combined using `$wgRestrictionLevels[] = 'smw-pageedit';` [0]:

image

Settings and work done in this PR are a prerequisites but the update on the protect action event to set the `Is edit protected` property is a bit shaky right now therefore I can't promise that it works from the protect action menu.

[0] https://www.mediawiki.org/wiki/Manual:$wgRestrictionLevels

kghbln:
I can very well imagine that this is not quite as trivial. A lot of users will also used this via a semantic form so both approaches should be somehow compatible if possible.

mwjames:
> Especially fixed properties and properties of imported vocabularies are born to be used together with this feature.

Yes, when I saw changes on the sandbox to same properties, I thought that this sort of thing probably  happens fairly frequent and is most likely caused unintentionally, yet creates unfortunate disruptions therefore a solution should be part of SMW-core to hinder such actions.

> For this feature to work the "smw-pageedit" user right must be added to the configuration parameter $smwgEditProtectionRight in "LocalSettings.php".

`$smwgEditProtectionRight` can contain any right, it is just that "smw-pageedit" is provided for convenience to make the intention more distinct. `Is edit protected` bounds itself to check against what is maintained in `$smwgEditProtectionRight` which happens in the `EditProtectionValidator`.

mwjames:
> A lot of users will also used this via a semantic form so both approaches should be somehow compatible if possible.

I suspect any `edit` is protected which should include SF as well (I don't have it installed nor have I tested it explicitly).

mwjames:
> I'll try to squeeze in a change to have it combined using $wgRestrictionLevels[] = 'smw-pageedit'; [0]:

This ended up to be a bit more challenging than anticipated but #2249 contains the necessary changes.

> Documentation:

@kghbln To provide you with some support on the documentation front, I added:

- https://www.semantic-mediawiki.org/wiki/Edit_protection
- https://youtu.be/AT4j78i4xW8

kghbln:
> $smwgEditProtectionRight can contain any right, it is just that "smw-pageedit" is provided for convenience to make the intention more distinct

I wanted to elaborate more on this on the configuration parameter level. Still a mention on this page is not bad either.

> This ended up to be a bit more challenging than anticipated but #2249 contains the necessary changes.

Thanks you for being brave and having the required stamina!

> To provide you with some support on the documentation front, I added: https://www.semantic-mediawiki.org/wiki/Edit_protection

Great, since this also answers some questions I would have posted here. Just fluffed and tweaked a bit.

kghbln:
Is it possible to add an array of rights to `$smwgEditProtectionRight`?

mwjames:
> Is it possible to add an array of rights to $smwgEditProtectionRight?

That would require some changes. Any particular use case in mind that would require multiple allowed edit rights?

For example, if `[[Is edit protected::true]]` is set then we force a protection level and in case it is only one the case is simple but when you have multiple which one do you want to enforce? The first one in the array as convention or do you want to add a `smwgEditProtectionEnforcedRight` setting?

mwjames:
> Is it possible to add an array of rights to $smwgEditProtectionRight?

I pushed a c88ee7c114 change that would allow `$GLOBALS['smwgEditProtectionRights'] = array( 'smw-pageedit', 'editsemiprotected' );` (PS: mind the name change to the plural form). I haven't really tested it and the question about  the enforced right remains open (currently it uses the first right from the array).

kghbln:
> That would require some changes. Any particular use case in mind that would require multiple allowed edit rights?

Cool. Thank you for going at it. I was primarily asking for the purpose of documentation but I can imagine cases where I would not like a user to be part of the curator group due to the regex editing connected with it. However this person may very well be capable of doing all other edits on protected pages. Multiple rights would allow for more flexibility. However it would alternatively be possible to create an extra group or use an existing one and assign `smw-pageedit` to it. So allowing multiple rights to be set is not required to allow for flexibility in this case if required. For the sake of simplicity it may probably be better to keep as is ...

Issue #2233, opened 2017-02-04
RDF output of BC dates seems not complies with XML schema


toniher:
### Setup and configuration

Tested in sandbox: https://sandbox.semantic-mediawiki.org/wiki/Zero_year

RDF output: https://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial:Export_RDF/Zero_year&syntax=rdf

### Issue

Output of XML dates for years in BC era seem not to comply with XML Schema definition:

1.1 Version: https://www.w3.org/TR/xmlschema11-2/#dateTime (previous definition version was different, though: https://www.w3.org/TR/xmlschema-2/#dateTime)

toniher:
I have a problem here that might be related to tests definition: https://travis-ci.org/toniher/SemanticMediaWiki/jobs/198528079 Despite I defined on '#3' skip-on virtuoso, it goes through RDF output test anyway: https://github.com/toniher/SemanticMediaWiki/blob/master/tests/phpunit/Integration/JSONScript/TestCases/p-0451.json
Is there any way to fully skip it for Virtuoso?

mwjames:
> I have a problem here that might be related to tests definition: https://travis-ci.org/toniher/SemanticMediaWiki/jobs/198528079 Despite I defined on '#3' skip-on virtuoso, it goes through RDF output test anyway: https://github.com/toniher/SemanticMediaWiki/blob/master/tests/phpunit/Integration/JSONScript/TestCases/p-0451.json
Is there any way to fully skip it for Virtuoso?

Yes there is [0], use the `meta` field to define:

```
	"meta": {
		"skip-on": {
			"virtuoso": "Virtuoso 6.1 rdfs / subproperty/subcategory hierarchies are not supported"
		},
		"version": "2",
		"is-incomplete": false,
		"debug": false
	}
```

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/JSONScript/TestCases/q-0802.json

mwjames:
> Yes there is

... and is now documented [0].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/JSONScript/README.md#skipping-a-test-or-mark-as-incomplete

toniher:
[...]

Issue #2218, opened 2017-01-28
Accepting negative ISO dates


toniher:
### Setup and configuration

Tested in sandbox instance at: https://sandbox.semantic-mediawiki.org/wiki/Negative_date

### Issue

I was wondering whether could be possible to accept negative dates in a format such as:
-0341-01-01T00:00:00Z

I think this format is ISO 8601, isn't it? 

I actually retrieved this date from Wikidata: https://www.wikidata.org/wiki/Special:EntityData/Q43216.json So, it might be convenient to accept this format without further parsing.

toniher:
I included a couple of additional examples, from the same source, which actually shows it's failing also replacing 'starting -' with a BC suffix. I imagine because of 00 month and day.

mwjames:
> I included a couple of additional examples

:+1: 

> I think this format is ISO 8601, isn't it?

A quick look on the net [0, 1] shows that dates like `-0166-11-01T17:00:00+00:00` are accepted and represent `5th of November 167 BC at 5pm`, so yes negative dates are supported in ISO.

> I was wondering whether could be possible to accept negative dates in a format such as:
-0341-01-01T00:00:00Z

Just to clarify this isn't a regression (tested it on an old 1.9.2 which equally fails for the examples).

## Solution 

I have spent 10 min on issue and therefore did not extensively test this (but works for the cited examples). It requires extra integration tests [2] and to pass of our test suite.

```
* [[Date::-0641-01-01T00:00:00Z]]
* [[Date::-0642-00-00T00:00:00Z]]
* [[Date::0342-00-00T00:00:00Z BC]]
* [[Date::600 BC]]
* [[Date::0341-01-01T00:00:00Z BC]]
* [[Date::-0741-01-01T00:00:00Z BC]]
```
```
@@ -327,10 +327,27 @@ class SMWDITime extends SMWDataItem implements CalendarModel {
 
 		return new DateTime( $time );
 	}
 
 	/**
+	 * @since 2.5
+	 *
+	 * @param string $iso
+	 *
+	 * @return SMWDITime|false
+	 */
+	public static function newFromISO( $iso ) {
+
+		// 0341-01-01T00:00:00Z BC / -0341-01-01T00:00:00Z BC
+		if ( strpos( $iso, 'BC' ) !== false && substr( $iso, 0, 1 ) !== '-' ) {
+			$iso = '-' . $iso;
+		}
+
[...]

PR #2234, opened 2017-02-05
Fix for storing and formatting dates in a compatible way with ISO8601 & XML Schema. Allowing ISO 8601 input

toniher:
This PR is made in reference to: #2218 and #2233 

This PR includes:
* Tests added: p-0450.json and p-0451.json.
* Tests modified: p-0413.json, p-0414.json, MediaWiki/Import/TimeDataTypeTest.php 

Now ISO dates are following the convention that 1 BC is 0000 and 2 BC is -0001, and so on...

Special care should be taken when wikis containing BC dates would be updated.


kghbln:
A squash and merge classic. :) Thanks a lot for your pull!

toniher:
Let me know if I could do anything else or more tests needed in your opinion!

mwjames:
Before I continue, let me thank you for the effort and dedication (especially in dealing with the tests).

> Let me know if I could do anything else or more tests needed in your opinion!

At a quick glance it looks ok but I want to review the "Handle offset" in `newFromISO` to understand what is required here (and if there isn't a better way).

 > Special care should be taken when wikis containing BC dates would be updated.

Since we would break the RDF output (which doesn't say this is wrong and apparently no one has complained about the difference in BCE/CE handling before), I'd rather make the break with 3.0 otherwise we would require a BC (backwards compatibility) switch in order for people to use the existing RDF format.

It may seem nitpicky but a break is a break and we either provide a migration path for the change or break on a version (oppose to what normally happens in the MediaWiki vicinity) that allows for that to happen which according to semver [0] would be 3.x.

Of course, if you say you need the feature now and by 2.5 then we need to do something to avoid a format break.

[0] http://semver.org/ "... MAJOR version when you make incompatible API changes ..." (API changes include exchange to formats like JSON, XML, or RDF including where a format or its data no longer result in the same interpretation)

toniher:
I agree that more work may be needed for time offsets (timezones). What I'm doing in the code above is practically ignoring that information when retrieved (actually merging the value into its UTC equivalent) from a ISO 8601 input. This means that, let's say, when the output is rendered in RDF, the detail about the timezone is lost. That's bad and I think I would rather change it (more testing would be needed).

I see your point regarding compatibility changes. However, especially if 3.0 may become too far away, I think it's worth to see if present situation can be improved. 
As a compromise, do you think it's worth a parameter that could let administrator choose whether 1 BC is 0000 or -0001 would be enough (keeping current -0001 as default option)? Being strict another parameter may also be needed for '-' vs '--' (or '−' proper) and even for zero left padding...

mwjames:
> As a compromise, do you think it's worth a parameter that could let administrator choose whether 1 BC is 0000 or -0001 

If want this in 2.5 then we require something that leaves the output as is and have a setting to change to the new output format.

PS: I had to do a similar thing to support the IRI export [0] and avoid BC for existing outputs.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/DefaultSettings.php#L1069-L1083.

kghbln:
@toniher In case you have some spare time this summer it will be great if you could work on this. :)

kghbln:
This pull has not seen any updates in basically two years. Thus closing to make room for another effort.

Issue #1757, opened 2016-08-04
MySQL "enforce_gtid_consistency=true" will cause a "Error: 1787 When @@GLOBAL.ENFORCE_GTID_CONSISTENCY = 1, the statements CREATE TEMPORARY TABLE and DROP TEMPORARY TABLE ..."


mwjames:
### Setup and configuration

- SMW version: all
- MW version: 1.28.0-alpha
- PHP: 5.6.8
- DB: MySQL 5.6.24-log
- ICU: 54.1

### Issue

Adding the following setting to `my.ini`

```
gtid_mode=ON
enforce_gtid_consistency=true
log_bin=/xampp/mysql/log/bin_log
log_slave_updates=true
```

```
Exception encountered, of type "RuntimeException"
[c8533ce4a56b1eab94b3b9a4] /mw-28-00/index.php/Main_Page RuntimeException from line 249 of ...\extensions\SemanticMediaWiki\src\MediaWiki\Database.php: A database error has occurred. Did you forget to run maintenance/update.php after upgrading? See: https://www.mediawiki.org/wiki/Manual:Upgrading#Run_the_update_script
Query: CREATE TEMPORARY TABLE `t3`( id INT UNSIGNED KEY ) ENGINE=MEMORY
Function: SMW\SQLStore\QueryEngine\QuerySegmentListProcessor::resolveHierarchyForSegment
Error: 1787 When @@GLOBAL.ENFORCE_GTID_CONSISTENCY = 1, the statements CREATE TEMPORARY TABLE and DROP TEMPORARY TABLE can be executed in a non-transactional context only, and require that AUTOCOMMIT = 1. (localhost)

#0 ...\includes\db\Database.php(901): DatabaseBase->reportQueryError('When @@GLOBAL.E...', 1787, 'CREATE TEMPORAR...', 'SMW\\SQLStore\\Qu...', false)
#1 ...\extensions\SemanticMediaWiki\src\MediaWiki\Database.php(247): DatabaseBase->query('CREATE TEMPORAR...', 'SMW\\SQLStore\\Qu...', false)
#2 ...\extensions\SemanticMediaWiki\src\SQLStore\QueryEngine\QuerySegmentListProcessor.php(303): SMW\MediaWiki\Database->query('CREATE TEMPORAR...', 'SMW\\SQLStore\\Qu...')
[...]

PR #2235, opened 2017-02-05
Backport PR #1758 to 2.4.x branch.

crlf0710:
Since #1757 is a show stopper for certain configuration, and 2.5's release schedule is not decided yet,  i want to try back-porting #1758  back to 2.4.x branch.

This PR includes:
- [x] CI build passed


crlf0710:
@mwjames Hi, what do you think? Any chance to include this in a future release of 2.4.x branch?

Notice that some inline documentation still says "since 2.5" though.

mwjames:
> Since #1757 is a show stopper for certain configuration, and 

Well, I only know of Wikia to use the "Global Transaction Identifier" actively.

> 2.5's release schedule is not decided yet

We expect that to happen in March [0].

> @mwjames Hi, what do you think? Any chance to include this in a future release of 2.4.x branch?

If it is really urgent then of course we can merge this to the 2.4.x and since this isn't your normal "bug/regression" backport, I'd would count on @kghbln to release this as separate 2.4.6 to isolate it in case something unanticipated happens (I'm not running any 2.4.x, so I cannot check against an actual instance). The tests pass so there shouldn't be any issue but as always the devil is in the details.

> Notice that some inline documentation still says "since 2.5" though.

Nothing to worry about.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/milestones

crlf0710:
Hi, thanks for the reply!

My hosting service's database provider has their "Global Transaction Identifier" turned on, which caused this problem. It occurs only with certain category queries, leading us have to workaround those queries.

Good to know that 2.5 will come in March, however i bet it will take some time after it is released before my mediawiki hosting service is able to finish evaluating the new version, upgrading their stuff, adapt to the feature changes, do testing etc.

Actually it is not in a very urgent status (current queries have been workaround-ed), but I will be very grateful if there be a separate 2.4.6 version that ship this fix before the 2.5 release, then i can go to persuade them do the upgradation soon.

Thanks a lot!

kghbln:
Since there are no objections for @JeroenDeDauw and @mwjames I could indeed merge and release as 2.4.6 during next week.

mwjames:
> Since there are no objections for @JeroenDeDauw and @mwjames I could indeed merge and release as 2.4.6 during next week.

No objections from my side, so I leave this to you guys!

PR #2240, opened 2017-02-10
properly set headers in SpecialSearchByProperty.php

FO-nTTaX:
This makes the SpecialPage adhere to stuff like `$wgDefaultRobotPolicy`


JeroenDeDauw:
+0.5

PR #2243, opened 2017-02-11
Add Property/Concept NS to $wgContentNamespaces/$wgNamespacesToBeSearchedDefault

mwjames:

- `$wgContentNamespaces` to allow custom namespaces to be acknowledged as containing useful content
- `$wgNamespacesToBeSearchedDefault` to indicate which namespaces are enabled for searching by default

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/NamespaceManager.php
+++ b/src/NamespaceManager.php
@@ -213,6 +213,24 @@ class NamespaceManager {
 			$smwNamespacesSettings,
 			$this->globalVars['smwgNamespacesWithSemanticLinks']
 		);
+
+		/**
+		 * Allow custom namespaces to be acknowledged as containing useful content
+		 *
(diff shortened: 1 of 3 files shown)


mwjames:
@kghbln FYI

kghbln:
[Documented on smw.o](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ANamespaces&type=revision&diff=49245&oldid=48010).

@mwjames At least the setting about changing the content namespaces is debatable. I personally am indifferent about "Property" but I see "Concept" as a dynamic category which I do not think is content in every situation. I guess it will be possible to override the setting as I documented it. Also we should mention in the release notes that "updateArticleCount.php" should be run after an upgrade from an earlier version by referring to the docu I added. Thus people are also deliberately invited to think about this.

PR #2244, opened 2017-02-11
Report similarity on property labels

mwjames:

- Adds `Special:PropertyLabelSimilarity` to list properties with a calculate syntactic similarity

This PR includes:
- [x] CI build passed


mwjames:
It can be difficult to figure by hand which properties are most likely cover the same concepts (based on the property label) [0] therefore `Special:PropertyLabelSimilarity`  will help with that effort by calculating the syntactic similarity and provide a list such as:

image

The page only reports those similarities and does not provide a function for merging or deleting reported properties, instead the user should use the information to create redirects [1] or delete duplicate properties. 

Of course, sometimes the syntactic similar of a property may in fact be a false positive on the assumption that both represent the same concept and to eliminate those cases `smwgSimilarityLookupExemptionProperty` setting contains the name (which by default is `owl:differentFrom`) of a property that is used to mark and exempt properties from each by declaring them to be different from each other.

Furthermore, to give the report a bit more flexibility, the threshold of the similarity distance can be freely adjusted.

@kghbln FYI

[0] https://www.semantic-mediawiki.org/wiki/Property_similarity
[1] https://www.semantic-mediawiki.org/wiki/Redirects

mwjames:
The Type ID checkbox is to help making a quick assumption about the nature of a property and as it can be seen on [0] because sometimes results reveal similar names but different types.

```
    {
        "property": [
            {
                "label": "Defaultvalue1",
                "type": "_txt"
            },
            {
                "label": "Defaultvalue4",
                "type": "_dat"
            }
        ],
        "similarity": 92.31
    },
```

What's more, the user has to select a limit to minimize the computational effort (the comparison is exponential with size of the list members).

[0] https://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial%3APropertyLabelSimilarity&limit=500&offset=0&threshold=90&type=yes

kghbln:
Also at https://www.semantic-mediawiki.org/wiki/Help:Special:PropertyLabelSimilarity

PR #2256, opened 2017-02-12
Updating ca translation

toniher:
Updating ca translation of extra messages


kghbln:
@toniher Thanks a lot for providing the translations! In the meantime we have another special property you may want to add: [en.json#L91](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/i18n/extra/en.json#L91) Also do not forget the magic words or aliases which are also located in `/i18n/extra/`

PR #2264, opened 2017-02-16
Fixed a number of special pages to properly set headers

FO-nTTaX:
I fixed one of these recently, but realized that there are more of these in the codebase. This makes the SpecialPages adhere to stuff like `$wgDefaultRobotPolicy`


mwjames:
Thanks.

PR #2267, opened 2017-02-18
Obfuscator improve proc. time, refs 2153

mwjames:
This PR is made in reference to: #2153

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/InTextAnnotationParser.php
+++ b/src/InTextAnnotationParser.php
@@ -262,7 +262,7 @@ class InTextAnnotationParser {
 	 *
 	 * @return string
 	 */
-	public static function getRegexpPattern( $linksInValues = false ) {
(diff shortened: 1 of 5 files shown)


mwjames:
@kghbln FYI

Parsing a page with a large pool of annotations like [0] in terms of enabled/disabled "links in value" option (aka `$GLOBALS['smwgLinksInValues'] = SMW_LINV_OBFU;`) should no longer show any significant difference in parsing time.

[0] https://gist.github.com/mwjames/a5d844c58d8806dcaa19ec13002442f2

mwjames:
In case someone gets a message like "Warning: preg_match_all(): Compilation failed: recursive call could loop indefinitely at offset 20 in /in/uEUQA on line 6" then s(he) use an outdated PHP version as shown in https://3v4l.org/uEUQA.

PR #2275, opened 2017-02-19
Use onoi/callback-container 2.x

mwjames:

- Requires `onoi/callback-container:~2.0`

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

If people complain that something is missing (class etc.) then it is not because the software has an issue but rather users where using `git` or something else to pull SMW and missed the step in receiving the latest compatible library. Anyway, people who use `git` need to ensure that they run `composer update` in order to fetch the required library.

kghbln:
Setting now [documented](https://www.semantic-mediawiki.org/wiki/Help:$smwgServicesFileDir).

PR #2282, opened 2017-02-25
Add PropertyFormatter/Parser service, refs 2275

mwjames:
This PR is made in reference to: #2275

- Adds `DataValueServiceFactory` and `DataValueServices.php`
- Label/name parsing has been moved from `PropertyValue` to a dedicated `PropertyValueParser` class
- Services are constructed via `DataValueServiceFactory` and injected into an instance
- Adds `smwgPropertyInvalidCharacterList` to define a list of invalid characters (default is `'[', ']' , '|' , '<' , '>', '{', '}', '+', '%'`) that are not expected in a label and would result in an error

This PR includes:
- [x] CI build passed


mwjames:
@kghbln The default in `smwgPropertyInvalidCharacterList` (`'[', ']' , '|' , '<' , '>', '{', '}', '+', '%'`) may seem a bit strict but I don't think that  `'+', '%'`  belong in a property label. On the contrary if users think they do, they can adjust the setting accordingly.

mwjames:
@kghbln If you have 5 min, could you verify that properties like `File:Image for Semantic-mediawiki.org.png|Semantic` [0] can no longer be created.

[0] https://wikiapiary.com/w/index.php?title=Special:Properties

kghbln:
> If you have 5 min, could you verify that properties like File:Image for Semantic-mediawiki.org.png|Semantic [0] can no longer be created.

Affirmative. When using PageForms "This appears to be a cross-site request forgery; canceling save." is shown, when trying to create manually the error message provided by SMW ist shown. No property gets created.

kghbln:
Now [documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgPropertyInvalidCharacterList&type=revision&diff=49716&oldid=49640) on wiki.

PR #2295, opened 2017-03-04
Add AllowsListValue

mwjames:

- Maintain a list of allowed list via `MediaWiki:Smw allows list ...` using `[[Allows value list::...]]` instead of individual `[[Allows value:: ...]]` annotations
- Expects the form of a hierarchy list which in its raw form (without the levels) isn't used directly by SMW but other processes may want to to create multilevel dropdown input forms using the hierarchy given in the  `MediaWiki:Smw allows list ...` page
- `[[Allows value list::Foo]]` is linked to `MediaWiki:Smw allows list Foo`
- Each valid list item in `MediaWiki:Smw allows list ...` is expected to start with `*` (e.g `* Bar` or `** Foobar`)
- Mixing of  `[[Allows value:: ...]]` and `[[Allows value list::...]]` within a property specification is allowed

This PR includes:
- [x] CI build passed


mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2295

mwjames:
A different perspective on using `Allows value list` is that changes to a reference list is not coupled by updating depending property values. For a property that has a large value pool that requires validation, an updated list will no automatically force a surge in update jobs due to a changed content list.

Whether this a positive or negative outcome is (was) not part of this PR and if such dispatch job is required, a new discussion should be opened for elaborating on related use cases.

PS: Of course, any new value assignment to a property will use the updated list.

Issue #1965, opened 2016-10-26
Rework "INSTALL.md"


kghbln:
### Issue

Since the upcoming 2.5 branch of SMW breaks compatibility with versions of MediaWiki 1.23.0 and lower the "INSTALL.md" should be reworked as it makes a lot of now unnecessary references to earlier versions and does not at all mention the worries introduced by MediaWiki 1.25+ ("composer.local.json"), etc.

mwjames:
"... Thanks, and in the mean time I can see in a readme.md that MS SQL is not supported by Semantic Mediawiki - would be nice if that was clearly stated in the installation guide..." [0]

[0] https://www.mediawiki.org/wiki/Topic:Tep32zhstc7x2gro

kghbln:
> would be nice if that was clearly stated in the installation guide...

Well, what about reading [COMPAT](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/COMPATIBILITY.md). Anyway why not adding a note and also include Oracle which is another one not supported.

kghbln:
> "... Thanks, and in the mean time I can see in a readme.md that MS SQL is not supported by Semantic Mediawiki - would be nice if that was clearly stated in the installation guide..." [0]

We are already clearly referring to the compat document in the installation document. However an explicit info about these two backends was missing there. Thus just added with a7e5abb.

Issue #1752, opened 2016-08-02
2.5.0 release


mwjames:
Tracking issue for the 2.5 release.

- [Release notes](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/RELEASE-NOTES.md)
- [Open issues for 2.5](https://github.com/SemanticMediaWiki/SemanticMediaWiki/milestones/SMW%202.5.0)
- [2.5 issues already closed](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues?q=milestone%3A%22SMW+2.5.0%22+is%3Aclosed)
- [Changes since 2.4](https://github.com/SemanticMediaWiki/SemanticMediaWiki/compare/2.4.0...master)

## rc.1 target: 2017-03-07

- [x] Shape rel notes
  - [x] Add missing commints tagged with [releasenotes](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pulls?q=is%3Apr+is%3Aopen+label%3Areleasenotes)
- [x] Update version number
  - [x] load.php
  - [x] COMPATIBILITY.md
- [x] Create tag
- [x] Create tarballs
- [x] Update `semantic-mediawiki.org`
  - [x] Run `composer update`
  - [x] Touch `LocalSettings.php`
  - [x] Update the SMW configuration 
  - [x] Check [Special:Version](https://www.semantic-mediawiki.org/wiki/Special:Version)
- [x] Ping testers group
- [x] Announce via mail
- [x] Do a release tweet on Twitter

## Release target: 2017-03-13

- [x] Finalize rel notes (include contributors (`git shortlog -sn 2.4.0...2.5.0`))
  - [x] Add missing commits tagged with [releasenotes](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pulls?q=is%3Apr+is%3Aopen+label%3Areleasenotes)
  - [x] Check [open issues](https://github.com/SemanticMediaWiki/SemanticMediaWiki/milestones/SMW%202.5.0)
- [ ] At least document the version highlights on wiki
  - [x] Full-text search
  - [x] Provenance data recording
  - [ ] Property chain and language filter support in print request
  - [x] Edit protection
  - [x] Preferred property label support
  - [x] Query result cache
  - [x] Links in values
  - [x] Fixed properties
  - [ ] Special page "SemanticMediaWiki"
- [x] Update version number (2.5.0)
  - [x] load.php
[...]

PR #2299, opened 2017-03-04
Update INSTALL.md

kghbln:
This PR is made in reference to: #1965 

- Update to reflect usage with versions of MediaWiki compatible with this version of Semantic MediaWiki

This PR includes:
- [ ] CI build passed

### DIFF
--- a/docs/INSTALL.md
+++ b/docs/INSTALL.md
@@ -55,7 +55,7 @@ Remember to add a comma to the end of the preceding line in this section.
 
 ##### Step 4
 
-Run the following command your shell:
+Run the following command in your shell:
 
     php composer.phar update --no-dev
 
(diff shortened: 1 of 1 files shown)


kghbln:
I guess this should be it. Perfection is always possible but basically it should be fluffy now.

kghbln:
@mwjames Oh, this was quick and dirty. :)

mwjames:
> Oh, this was quick and dirty. :)

I canceled the TravisCI to avoid blocking other tests on changes that only affect documentation. Generally, when changing documentation `[skip ci]` can be used to avoid running the test suite.

https://docs.travis-ci.com/user/customizing-the-build#Skipping-a-build

kghbln:
> I canceled the TravisCI to avoid blocking other tests on changes that only affect documentation. Generally, when changing documentation [skip ci] can be used to avoid running the test suite.

Thanks for the note! I have see that one can do this but thought that it may involve more voodoo.

kghbln:
Refs https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1752

PR #2324, opened 2017-03-11
IdTaskHandler to allow string match for ID

mwjames:
This PR is made in reference to: #2268

- Allows for the ID to be a string to filter on the sortkey and return pairs of matches

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI I wanted to investigate something on `semantic-mediawiki.org` but realized that `Object ID lookup and disposal` can only work with numeric IDs. I tweak the input a bit with this PR so that now one can also use a string to find matching IDs and work its way through.

For example, `sandbox.semantic-mediawiki.org` with input `P106` returns:

image


If the results are not unique (as in case for a single dedicated ID) then the form fields will remove any input hint.

kghbln:
Very nice improvement!

PR #2325, opened 2017-03-11
Add `#-ia` output option filter for text results

mwjames:
This PR is made in reference to: #1314, https://github.com/SemanticMediaWiki/SemanticCite/issues/40

- If for some use cases a text that contains an in-text annotation wishes to retain that content in its raw form then `#-ia` can be used to direct that action otherwise it is filter as described in #1314
- Adds some extra `format=list` integration tests

This PR includes:
- [x] CI build passed


kghbln:
@mwjames Thanks for [cranking up the documenation](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Annotation_markup_output_formatter)

Issue #2329, opened 2017-03-11
Use of escape sequences in template output poduces different results (sep= ...)


Contrafibularity:
### Setup and configuration

- SMW version: 2.5rc
- MW version: 1.28
- PHP version: 5.6.30-0+deb8u1 (apache2handler)

### Issue

See https://sandbox.semantic-mediawiki.org/wiki/Issue_with_double_sep_in_template_format

When used with the template format, <code>sep</code> is intended to delimit the property values that are passed to the template. However, in the first example provided on the sandbox page, it also ends up delimiting the results (hence the semi-colon), which I assume is a (recent) bug. 

I'm not sure what the cause is though I suspect that it may have something to do with the number of prinouts parameters used by the template.

mwjames:
There was a fix for sep (#2022, #2090) which produces the expected results but the MediaWiki `Parser` is working its magic which can be seen when we not use `#ask` and instead simply add the templates manually (this involves no SMW).

```
{{Show example for normal sep|1=CityA|?Page=CityA|?Capital of=CapitalA|#=0|userparam=Test test}};{{Show example for normal sep|1=CityB|?Page=CityB|?Capital of=CapitalB|#=1|userparam=Test test}};{{Show example for normal sep|1=CityC|?Page=CityC|?Capital of=CapitalC|#=2|userparam=Test test}}
```
```
{{Show example for double sep|1=CityA|?Page=CityA|?Capital of=CapitalA|#=0|userparam=Test test}};{{Show example for double sep|1=CityB|?Page=CityB|?Capital of=CapitalB|#=1|userparam=Test test}};{{Show example for double sep|1=CityC|?Page=CityC|?Capital of=CapitalC|#=2|userparam=Test test}}
```

The usage is the same except for the template name!

https://sandbox.semantic-mediawiki.org/wiki/Issue/2329

mwjames:
Before 2.5 the `sep` would not be published but #2022 fixes that so that the content returned to MW would now correctly contain `{{Template| ...}} sep {{Template|...}} sep ...`. As demonstrated above, MW is using a different schema to generate the final output depending of how escape sequences `\n` (newline) are used within those templates.

SMW is only involved in generating the template content (not the parsing) with the output and parsing being done by MediaWiki alone.

Contrafibularity:
Thanks for clearing that up - I forgot about that particular MW-specific usage of semi-colons, so that explains. 

[...]

PR #2331, opened 2017-03-12
Add smwgResultFormatsFeatures, refs 2022, 2329

mwjames:
This PR is made in reference to: #2022, #2329

- Adds `$smwgResultFormatsFeatures` setting with enabled `SMW_RF_TEMPLATE_OUTSEP` flag (#2022)
- Adds `valuesep` for the `format=template` when `SMW_RF_TEMPLATE_OUTSEP` is enabled to avoid ambiguity of the `sep` parameter semantics

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -553,6 +553,17 @@ return array(
 	'smwgResultAliases' => array( 'feed' => array( 'rss' ) ),
 	##
 
+	##
+	# Result printer features
+	#
+	# - SMW_RF_NONE
+	# - SMW_RF_TEMPLATE_OUTSEP, #2022 (use the sep parameter as outer separator)
+	#
+	# @since 2.3
+	##
+	'smwgResultFormatsFeatures' => SMW_RF_TEMPLATE_OUTSEP,
+	##
+
 	### Predefined sources for queries
 	# Array of available sources for answering queries. Can be redefined in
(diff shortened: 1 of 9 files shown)


mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2331

kghbln:
[Needs more doculove](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ATemplate_format&type=revision&diff=51666&oldid=49433)

cicalese:
I just discovered this change when preparing to upgrade from SMW 2.4.2 to SMW 2.5.2. My wikis have many places where an #ask query uses a template result format and one of the fields returned is multi-valued. I typically use #arraymap (from PageForms) to separate the values, splitting them at the comma separator, for further processing. However, without valuesep specified, the separator is now defaulting to a space rather than a comma even when I try to turn off the feature with $smwgResultFormatsFeatures. I am faced with the proposition of finding all of the #ask queries in all of my wikis that use a template result format and then examining all of those templates to see if I use any #arraymap calls and then returning to the query to add valuesep=,. Is the fact that valuesep defaults to a space rather than a comma a bug or a conscious choice?

kghbln:
[Documented](https://www.semantic-mediawiki.org/wiki/Help:$smwgResultFormatsFeatures)

Issue #2342, opened 2017-03-18
Invalid data value annotations not directly shown for datatype "Text"


kghbln:
### Setup and configuration

- SMW version: 2.5.0
- MW version: 1.28.0
- PHP version: 5.6
- DB (MySQL etc.): 10

### Issue
In contrast to earlier versions of Semantic MediaWiki <= 2.4.x erroneous values are no longer shown when annotated. Just the warning icon is show at the spot which allows to hover for information about the error.

I believe that it is better to be able to directly see the incorrectly annotated value instead of having to hover for it.

### Example
[Jane Do](https://sandbox.semantic-mediawiki.org/wiki/Do,_Jane)

mwjames:
Hi,

> In contrast to earlier versions of Semantic MediaWiki <= 2.4.x erroneous

I believe this isn't correct for the `Text` type because the source
from 1.9.2 [0] (and it is equal to what 2.5.0 does) says something
like "When the value is invalid return an empty string to the caller".

If you want to change the behaviour then you would need to adjust the
source (I haven't test the change nor did I run the test suite itself
on this).

```
@@ -41,10 +41,11 @@ class SMWStringValue extends DataValue {

 		if ( $value === '' ) {
 			$this->addErrorMsg( 'smw_emptystring' );
 		}

+		$this->m_caption = $value;
 		$this->m_dataitem = new DIBlob( $value );
 	}
```

```
@@ -47,11 +47,11 @@ class StringValueFormatter extends DataValueFormatter {
 		if ( $this->dataValue->getCaption() !== false && $type ===
self::HTML_SHORT ) {
 			return smwfXMLContentEncode( $this->dataValue->getCaption() );
 		}

 		if ( !$this->dataValue->isValid() ) {
-			return '';
+			return $this->dataValue->getCaption();
 		}
```

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/1.9.2/includes/datavalues/SMW_DV_String.php#L119-L120

Cheers

kghbln:
> I believe this isn't correct for the `Text` type

Just verified this. Indeed I was wrong. 

> If you want to change the behaviour

I still believe that this should be consistent for all datatypes so ...

kghbln:
I can imagine that the reason for datatype "Text" not having this feature is that in former times allows value restrictions were probably only applied to datatype "String". Since these were merged.

PR #2348, opened 2017-03-24
Allow showing annotations even if they are improper

kghbln:
This PR is made in reference to: #2342 

- Datatype "Text" - Allow showing annotations even if they are improper

Coding by @mwjames via https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2342#issuecomment-287533327

This PR includes:
- [x] CI build passed

### DIFF
--- a/tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
+++ b/tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
@@ -34,7 +34,7 @@
 			"subject": "Example/P0452/2",
 			"assert-output": {
 				"to-contain": [
-					"<span class=\"smw-highlighter smwttinline\" data-type=\"4\" data-state=\"inline\" data-title=\"Error\" title=\"&quot;Do&quot; is not in the list (Doe, Doe) of allowed values for the &quot;Has text&quot; property.\">"
+					"Do<span class=\"smw-highlighter\" data-type=\"4\" data-state=\"inline\" data-title=\"Error\" title=\"&quot;Do&quot; is not in the list (Doe) of allowed values for the &quot;Has text&quot; property.\">"
 				]
 			}
 		}
--- a/tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
+++ b/tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
@@ -34,10 +34,7 @@
 			"subject": "Example/P0452/2",
 			"assert-output": {
 				"to-contain": [
-					"<span class=\"smw-highlighter smwttinline\" data-type=\"4\" data-state=\"inline\" data-title=\"Error\" data-hasqtip=\"0\" aria-describedby=\"qtip-0\">"
-				],
-				"not-contain": [
-					"Do"
+					"<span class=\"smw-highlighter smwttinline\" data-type=\"4\" data-state=\"inline\" data-title=\"Error\" title=\"&quot;Do&quot; is not in the list (Doe, Doe) of allowed values for the &quot;Has text&quot; property.\">"
 				]
 			}
 		}
--- a/tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
+++ b/tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
@@ -1,0 +1,58 @@
+{
+	"description": "Test in-text `_txt` datatype in combination with an \"Allows value\" output (#2342, `wgContLang=en`, `wgLang=en`)",
+	"setup": [
+		{
+			"page": "Has text",
+			"namespace":"SMW_NS_PROPERTY",
+			"contents": "[[Has type::Text]], [[Allows value::Doe]]"
+		},
+		{
+			"page": "Example/P0452/1",
+			"namespace":"NS_MAIN",
+			"contents": "[[Has text::Doe]]"
+		},
+		{
+			"page": "Example/P0452/2",
+			"namespace":"NS_MAIN",
+			"contents": "[[Has text::Do]]"
+		}
+	],
+	"tests": [
+		{
+			"type": "parser",
+			"about": "#0 test output of the valid [[Has text::Doe]] annotation",
+			"subject": "Example/P0452/1",
+			"assert-output": {
+				"to-contain": [
+					"Doe"
+				]
+			}
+		},
+		{
+			"type": "parser",
+			"about": "#1 test output of the invalid [[Has text::Do]] annotation",
+			"subject": "Example/P0452/2",
+			"assert-output": {
+				"to-contain": [
+					"<span class=\"smw-highlighter smwttinline\" data-type=\"4\" data-state=\"inline\" data-title=\"Error\" data-hasqtip=\"0\" aria-describedby=\"qtip-0\">"
+				],
+				"not-contain": [
+					"Do"
+				]
+			}
+		}
+	],
+	"settings": {
+		"wgContLang": "en",
+		"wgLang": "en",
+		"smwgNamespacesWithSemanticLinks": {
+			"NS_MAIN": true,
+			"SMW_NS_PROPERTY": true
+		}
+	},
+	"meta": {
+		"version": "2",
+		"is-incomplete": false,
+		"debug": false
+	}
+}
--- a/includes/dataitems/SMW_DI_Error.php
+++ b/includes/dataitems/SMW_DI_Error.php
@@ -21,8 +21,14 @@ class SMWDIError extends SMWDataItem {
 	 */
 	protected $m_errors;
 
-	public function __construct( $errors ) {
+	/**
+	 * @var string
+	 */
+	private $userValue;
+
+	public function __construct( $errors, $userValue = '' ) {
 		$this->m_errors = $errors;
+		$this->userValue = $userValue;
 	}
 
 	public function getDIType() {
@@ -31,6 +37,15 @@ class SMWDIError extends SMWDataItem {
 
 	public function getErrors() {
 		return $this->m_errors;
+	}
+
+	/**
+	 * @since 3.0
+	 *
+	 * @return string
+	 */
+	public function getUserValue() {
+		return $this->userValue;
 	}
 
 	public function getSortKey() {
--- a/includes/datavalues/SMW_DataValue.php
+++ b/includes/datavalues/SMW_DataValue.php
@@ -149,6 +149,11 @@ abstract class SMWDataValue {
 	private $infoLinksProvider = null;
 
 	/**
+	 * @var string
+	 */
+	private $userValue = '';
+
+	/**
 	 * @var DataValueServiceFactory
 	 */
 	protected $dataValueServiceFactory;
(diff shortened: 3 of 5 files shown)


kghbln:
The [field test](https://sandbox.semantic-mediawiki.org/wiki/Do,_Jane) shows that it works. Also the error.log is empty for this.

kghbln:
Obviously test need to be adapted, too. Cool, hhvm passes. :)

JeroenDeDauw:
What is happening! :)

bunny_on_my_keyboard

mwjames:
> Obviously test need to be adapted, too. Cool, hhvm passes. :)

Only the "benchmark" (which runs on HHVM) tests pass hence this needs a bit more work.

```
There were 6 failures:

1) SMW\Tests\DataValues\ValueFormatters\CodeStringValueFormatterTest::testFormat with data set #1 ('foo', 1, NULL, '<div class="smwpre">foo</div>')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<div class="smwpre">foo</div>'
+'foo'

/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Unit/DataValues/ValueFormatters/CodeStringValueFormatterTest.php:40

2) SMW\Tests\DataValues\ValueFormatters\CodeStringValueFormatterTest::testFormat with data set #2 ('foo', 2, NULL, '<div class="smwpre">foo</div>')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<div class="smwpre">foo</div>'
+'foo'

/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Unit/DataValues/ValueFormatters/CodeStringValueFormatterTest.php:40

3) SMW\Tests\DataValues\ValueFormatters\CodeStringValueFormatterTest::testFormat with data set #8 ('<foo>', 2, NULL, '<div class="smwpre">&lt;foo&gt;</div>')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'<div class="smwpre">&lt;foo&gt;</div>'
+'&lt;foo&gt;'

/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Unit/DataValues/ValueFormatters/CodeStringValueFormatterTest.php:40

4) SMW\Tests\DataValues\ValueFormatters\StringValueFormatterTest::testFormatWithReducedLength
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Lorem ipsum dolor sit amet consectetuer â€¦'
+'Lorem ipsum dolor sit amet consectetuer justo Nam quis lobortis vel. Sapien nulla enim Lorem enim pede lorem nulla justo diam wisi. Libero Nam turpis neque leo scelerisque nec habitasse a lacus mattis. Accumsan tincidunt Sed adipiscing nec facilisis tortor Nunc Sed ipsum tellus'

/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Unit/DataValues/ValueFormatters/StringValueFormatterTest.php:97

5) SMW\Tests\InTextAnnotationParserTest::testTextParse with data set #14 (0, array(array(true), 4, true, true), '[[Text::Bar [http://example.org/Foo Foo]]] [[Code::Foo[1] Foobar]]', array('Bar [http://example.org/Foo Foo] <div class="smwpre">Foo&#91;1]&#160;Foobar</div>', 2, array('Text', 'Code'), array('Bar [http://example.org/Foo Foo]', 'Foo[1] Foobar')))
Failed asserting that 'Bar [http://example.org/Foo Foo] Foo[1] Foobar' contains "Bar [http://example.org/Foo Foo] <div class="smwpre">Foo&#91;1]&#160;Foobar</div>".

/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Unit/InTextAnnotationParserTest.php:140

6) SMW\Tests\Integration\JSONScript\JsonTestCaseScriptRunnerTest::executeTestCases with data set "p-0402.json" ('/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/JSONScript/TestCases/p-0402.json')
Failed on `#1 (: as leading text element to be converted to an indent)` for <p>//::Page/::Foo:Bar is not allowed as title;&#160;::Text/::Foo:Bar <a rel="nofollow" class="external free" href="http://example.org/:::Foo:Bar">http://example.org/:::Foo:Bar</a>
</p>
<!-- 
NewPP limit report
Cached time: 20170324155511
Cache expiry: 86400
Dynamic content: false
[SMW] Inâ€text annotation parser time: 0.004 seconds
CPU time usage: 0.007 seconds
Real time usage: 0.007 seconds
Preprocessor visited node count: 1/1000000
Preprocessor generated node count: 4/1000000
Postâ€expand include size: 0/2097152 bytes
Template argument size: 0/2097152 bytes
Highest expansion depth: 1/40
Expensive parser function count: 0/100
-->

<!-- 
Transclusion expansion time report (%,ms,calls,template)
100.00%    0.000      1 - -total
-->
 with (StringContains) [ <dd>Text/::Foo:Bar ]
Failed asserting that 1 matches expected 2.
```

kghbln:
@mwjames You may hijack this branch. To me it seems that the test are also checking datatype "Code" which was not touched.

mwjames:
> You may hijack this branch. To me it seems that the test are also checking datatype "Code" which was not touched.

As to evidence by the failing tests some implicit behaviour was expecting to work with the caption hence I took a slightly different approach to solve the issue.

I left out an integration test but I would be nice you could try to write one as a follow-up.

kghbln:
> I left out an integration test but it would be nice if you could try to write one as a follow-up.
Merge state

I guess a test will be similar to the one we have for date and should go into this [directory](https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/master/tests/phpunit/Integration/JSONScript/TestCases). Could you point to it?

mwjames:
> the one we have for date and should go into this directory.

Yes that's the one. I thought that this is mentioned somewhere in [0] and if you think [0] lacks expressiveness on how to design and conduct an integration test then we should see to improve this based on your experience.

To avoid having others to "suffer" the same fate, I created a quick introduction video [1] in hopes to provide some more practical inspiration.

Happy testing!

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/master/tests/phpunit/Integration/JSONScript#designing-an-integration-test
[1] https://youtu.be/7fDKjPFaTaY

PS: Of course, I could write the test myself but that would have no educational value whatsoever.

kghbln:
> To avoid having others to "suffer" the same fate

Found the directory but did not look for README's. Senior moments. :(

> Of course, I could write the test myself but that would have no educational value whatsoever.

That's what I dreaded. ;)

tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
```diff
@@ -0,0 +1,58 @@
+{
```
kghbln: Let's see what happens. I guess we could also use a "Allows value list" test in here.

tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
```diff
@@ -0,0 +1,58 @@
...
+		},
+		{
+			"type": "parser",
+			"about": "#1 test output of the invalid [[Has text::Do]] annotation",
+			"subject": "Example/P0452/2",
+			"assert-output": {
+				"to-contain": [
+					"<span class=\"smw-highlighter smwttinline\" data-type=\"4\" data-state=\"inline\" data-title=\"Error\" data-hasqtip=\"0\" aria-describedby=\"qtip-0\">"
+				],
+				"not-contain": [
```
kghbln: Hmm, this should go since the page still shows the value even if it is improper.

tests/phpunit/Integration/JSONScript/TestCases/p-0452.json
```diff
@@ -34,10 +34,7 @@
 			"subject": "Example/P0452/2",
 			"assert-output": {
 				"to-contain": [
-					"<span class=\"smw-highlighter smwttinline\" data-type=\"4\" data-state=\"inline\" data-title=\"Error\" data-hasqtip=\"0\" aria-describedby=\"qtip-0\">"
-				],
-				"not-contain": [
-					"Do"
+					"<span class=\"smw-highlighter smwttinline\" data-type=\"4\" data-state=\"inline\" data-title=\"Error\" title=\"&quot;Do&quot; is not in the list (Doe, Doe) of allowed values for the &quot;Has text&quot; property.\">"
```
kghbln: This should not fail but it does. Hmm ...

mwjames:
> That's what I dreaded. ;)

## Methodology

The general approach (or methodology) for testing is:

- [ ] Rebase the PR to ensure that it is tested against the latest master
- [ ] Create the examples "Example/P0452/1", "Example/P0452/2" locally on the wiki and observe the results, Do they contain the expected results?
- [ ] Run individual tests locally with the setup and compare the results
- [ ] Once the test passes, run the whole suite (`composer phpunit`) to ensure that nothing broke during the process or change 

## Questions

Let me ask some questions.

While creating the test case did you run the test locally as shown in the video (e.g. `composer phpunit /d/xampp/htdocs/mw-master/extensions/SemanticMediaWiki/tests/phpunit/Integration/JSONScript/JsonTestCaseScriptRunnerTest -- --filter 'p-0452.json'`)?

When looking at the failed test, the output shows:

> `Failed on `#1 test output of the invalid [[Has text::Do]] annotation` for <div class="mw-parser-output">Do<span class="smw-highlighter" data-type="4" data-state="inline" data-title="Error" title="&quot;Do&quot; is not in the list (Doe) of allowed values for the &quot;Has text&quot; property."><span class="smwtticon warning"></span><div class="smwttcontent">"Do" is not in the list (Doe) of <a href="/mw-master/index.php/Property:Allows_value" title="Property:Allows value">allowed values</a> for the "Has text" property.</div></span>`

and the failed assertion is:

> `with (StringContains) [ <span class="smw-highlighter smwttinline" data-type="4" data-state="inline" data-title="Error" title="&quot;Do&quot; is not in the list (Doe, Doe) of allowed values for the &quot;Has text&quot; property."> ]`

> Failed asserting that 0 matches expected 1.

It shows a difference in the expected `Do&quot; is not in the list (Doe, Doe) of ...` vs. the actual `"Do" is not in the list (Doe) of ...` output (happens because of an outdated master this was tested against).

Did the second  "Example/P0452/2" case actually test the "Allow showing annotations even if they are improper"? For example, the "most" important part isn't tested which is "Do<span class="smw-highlighter" data-type="4" data-state=..." because `Do` isn't allowed and we expect `Do` to be visible as simple value, it would not be visible without this PR.

kghbln:
> Let me ask some questions.

Actually I did not run the test locally prior to adding it here. :( I just typed it in my gedit window

> Did the second "Example/P0452/2" case actually test

This was kinda stupid not to add. Indeed.

> If you think it does what it should and you are satisfied with the results then feel free to squash and merge.

Well it works for now. Will add a second test with allows value list.

kghbln:
[Documented on wiki.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AType_Text&type=revision&diff=57136&oldid=57132)

PR #2357, opened 2017-03-25
Add DeprecationNoticeTaskHandler, refs 2343

mwjames:
This PR is made in reference to: #2343

This PR addresses or contains:

- `Special:SemanticMediaWiki` adds a section and lists deprecated settings that were detected to be used on the wiki

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

Issue #2343, opened 2017-03-18
Remove deprecated configuration parameters


kghbln:
Starting with SMW 2.5.0 this configuration parameter was depreciated in favour of the `$smwgAdminFeatures` configuration parameter. I believe an overly long migration period is not required for this configuration parameter. Thus it may be removed.

mwjames:
Done so in #2362 with a little twist that if it is still used, it will appear in `Special:SemanticMediaWiki`. If you want to remove it then you have to:

- Remove the legacy mapping in `Settings`
- Move the entry in `DeprecationNoticeTaskHandler` from `deprecatedConfigList` to `removedConfigList`

![image](https://cloud.githubusercontent.com/assets/1245473/24321304/7bf57c4a-118c-11e7-8e7f-675fe86a0a67.png)

kghbln:
>  Done so in #2362 with a little twist that if it is still used, it will appear in Special:SemanticMediaWiki. If you want to remove it then you have to:

Great idea! Thanks a lot.

kghbln:
This was originally only about `$smwgAdminFeatures` but now adding in `smwgQueryDependencyPropertyExemptionlist` and `smwgQueryDependencyAffiliatePropertyDetectionlist` as stated #2384

krabina:
Guess this can this be closed? the above mentioned PRs are merged

PR #2362, opened 2017-03-26
Deprecate $smwgAdminRefreshStore, refs 2343

mwjames:
This PR is made in reference to: #2343, #2357

- In case it is still used, it will show up on #2357

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -126,7 +126,6 @@ class Settings extends Options {
 			'smwgDeclarationProperties' => $GLOBALS['smwgDeclarationProperties'],
 			'smwgDataTypePropertyExemptionList' => $GLOBALS['smwgDataTypePropertyExemptionList'],
 			'smwgTranslate' => $GLOBALS['smwgTranslate'],
-			'smwgAdminRefreshStore' => $GLOBALS['smwgAdminRefreshStore'],
 			'smwgAutocompleteInSpecialAsk' => $GLOBALS['smwgAutocompleteInSpecialAsk'],
 			'smwgAutoRefreshSubject' => $GLOBALS['smwgAutoRefreshSubject'],
 			'smwgAdminFeatures' => $GLOBALS['smwgAdminFeatures'],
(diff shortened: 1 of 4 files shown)


includes/Settings.php
```diff
@@ -287,4 +288,15 @@ private function doIterate( $key, $options ) {
 		throw new SettingNotFoundException( "'{$key}' is not a valid settings key" );
 	}
 
+	private static function initLegacyMapping( &$configuration ) {
+
+		// When adding a deprecated mapping, extend the DeprecationNoticeTaskHandler
+		// with the corresponding entry
+
+		// Deprecated in 2.5, to be removed with 3.1
```
JeroenDeDauw: Removal in a feature release? I suppose this is not truly a user-space thing... but still...

Perhaps we should specify somewhere what exactly we guarantee to not break in non-breaking releases

mwjames: > Removal in a feature release?

Well, there are two releases a part. Furthermore, this consolidates the literal description (found in release notes, setting notes etc.) to an actional representation by making #2357 to detect those and inform the user more actively about pending future removal.

See also https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2343#issuecomment-289252430.

mwjames: > Perhaps we should specify somewhere what exactly we guarantee to not break in non-breaking releases

The `Settings` legacy mapping and `DeprecationNoticeTaskHandler` allows for a more controlled process instead of just dropping the ball, so for now the process would is envisioned as:

- Remove the setting from `DefaultSettings.php` (done so to allow for a detection on whether `$GLOBALS` still contains a reference to it or not infused by the `LocalSettings.php`)
- Remap the setting in case it has been replaced by something else
- Add an entry in `DeprecationNoticeTaskHandler` to be either categorized as deprecated or removed

mwjames:
Back-ported to 2.5.x with 4c004e4.

includes/Settings.php
kghbln: > Removal in a feature release? I suppose this is not truly a user-space thing... but still...

Doing an extra release just to remove something ... 3.0 is big so watch out for changes.

kghbln:
> Back-ported to 2.5.x with 4c004e4.

https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/560abb1736923ebc629217ff1fa77ab253053da2 probably needs to be back-ported too. I just set um twn for this change.

PR #2365, opened 2017-03-26
Track persistent update marker (MW 1.29+)

mwjames:

- MW 1.29 made the `LinksUpdate` a `EnqueueableDataUpdate` which creates updates as `JobSpecification` (refreshLinksPrioritized) and depending on the scheduler creates a possibility of running an update more than once on the same RevID
- Track the RevID and skip any reoccurring updates for the same RevID (persistent update marker is stored for an hour)

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/ParserData.php
+++ b/includes/ParserData.php
@@ -5,6 +5,7 @@ namespace SMW;
 use ParserOutput;
 use SMWDataValue as DataValue;
 use Title;
+use Onoi\Cache\Cache;
 
 /**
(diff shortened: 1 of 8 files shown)


mwjames:
@kghbln This needs to be back-ported as well. I'll merge it later in order for it to be available on the sandbox but the back-port should only happen next week or thereafter to allow for some inspection time.

mwjames:
Back-ported to 2.5.x with f5a30dd.

PR #2369, opened 2017-03-30
Remove "successfully" from smw-admin-setupsuccess

amire80:
According to
https://www.mediawiki.org/wiki/Localisation#Avoid_jargon_and_slang


kghbln:
Thanks for the patch. In this context here I do not fully agree however if that's what MediaWiki and co. should do... I am always happy when being told that I did something successfully. It's like a cheer up. Then it comes to failures the situation is indeed different.

PR #2376, opened 2017-04-01
3.0.0 footer logo

mwjames:
This PR is made in reference to: https://phabricator.wikimedia.org/T44128

- Current logo contains unbalanced letter positions and uses a blurred font (with uneven edges)
- Current sunflower takes unprecedented focus
- Current logo design is in stark contrast to existing footer logos

This PR includes:
- [ ] CI build passed


mwjames:
@kghbln FYI

* New logo is created from ground up (using Photoshop)
* Using `Semantic MediaWiki` as name inclusion is a bit too long to provide a balanced footer logo and was replaced by `Powered with semantic` as indication for semantic functionality provided by `Semantic MediaWiki` without repeating `MediaWiki` as the platform is provided by `MediaWiki`
* The sunflower was replaced by a simpler logo (renting some part of it) and may serve a more professional image of the software product that went through various changes

PS: The logo (and hereby I as creator) don't claim it to be the best possible solution but it provides an artwork for direct replacement. Users or interested parties who want to see some changes to the existing logo or to the proposal are encouraged to provide artworks of their own.

JeroenDeDauw:
The text "powered with semantic" makes little sense to me. Also, I find the next in the new image harder to read.

JeroenDeDauw:
I actually don't know who created the current one. Would be good to get their input. Might be @mkroetzsch

kghbln:
Just this week I thought about the footer logo too. No kidding.

> Current logo contains unbalanced letter positions and uses a blurred font (with uneven edges)

Probably people will not recognize this since it is so small however this should be tackled too

> Current sunflower takes unprecedented focus

Indeed, as much as I like sunflowers (one of my favourite flowers) I am all for it to drop this focus

> Current logo design is in stark contrast to existing footer logos

I'd rather stick to Semantic MediaWiki and drop the "served by", "powered by", "brought by", "provided by", "made with" or whatever may come to mind here. People will have an isolated look at the footer logo so most of them will not manage to autocomplete "semantic" with MediaWiki.

mwjames:
I'll leave the merge window a bit more stretched and consider any counter proposal, but as it stands the current logo will vanish with 3.0.

> I'd rather stick to Semantic MediaWiki

If you want to stick with "Semantic MediaWiki" then you have to use the upper/lower case combination (as for the "brand" name) and that makes it very difficult to find a font that doesn't suck on the image size without blurring or creating and unbalanced view.

>  Also, I find the next in the new image harder to read.

I explicitly noted that "... proposal are encouraged to provide artworks of their own."

kghbln:
> but as it stands the current logo will vanish with 3.0.

what's why I added the 3.0 milestone

> If you want to stick with "Semantic MediaWiki" 

or just "semantics" still without this "powered by" etc. stuff

kghbln:
> Current new logo does not overall improve over the old one IMO

I think we have gotten that far. ;)

mwjames:
> Current new logo does not overall improve over the old one IMO

Please provide a counter proposal otherwise I will override this block.

mkroetzsch:
I am not a designer and always open to suggestions regarding visual layout and fonts. However, please keep in mind why we have the button in the first place. It is part of SMW's branding activities, and it does not make sense to replace it with a button that is not consistent with SMW's branding. In particular, removing the project's name and logo would make the button rather pointless. Of course one can change the branding, but this should be done consistently.

As I read the original request, the main issue is that the button is too prominent and should attract less attention from the main page. This should be fixable by changing the colour scheme.

mwjames:
> However, please keep in mind why we have the button in the first place. It is part of SMW's branding activities, and it does not make sense to replace it with a button that is not consistent with SMW's branding.

I'm very well aware of that fact but the colour schema seems un-proportional aggressive (especially the blue) and the sunflower as single identifying factor does no longer reflect the software as to striving towards a more professional endeavor. I think the flower as design has outlived its representing factor and seems rather immature in retrospect.

Artwork is a matter of opinion and I value opinion to the point that there are constructive to the process or show effort in improving a situation. Arguments for the sake of "something is wrong" isn't a way forward on a volunteer project.

> The logo (and hereby I as creator) don't claim it to be the best possible solution but it provides an artwork for direct replacement. 

Again, any effort from interested parties are welcome and are highly encouraged therefore I'm looking forward to proposals that improve the current status quo.

kghbln:
I will try to get my sis to create a logo - main and footer.

kghbln:
My idea is to provide two or three alternatives to choose from. An open vote if desired and no open discussion. There are currently major "issues" which prevent my sister from creating these right away so it will be July hopefullly.

s7eph4n:
Found the original discussion:
https://sourceforge.net/p/semediawiki/mailman/semediawiki-user/thread/200911181544.09280.markus%40semantic-mediawiki.org/#msg24001383

kghbln:
> Found the original discussion:
https://sourceforge.net/p/semediawiki/mailman/semediawiki-user/thread/200911181544.09280.markus%40semantic-mediawiki.org/#msg24001383

That' s a good wrap up for the briefing. Thanks a lot.

kghbln:
We should have progress here be the end of the month. Sit tight.

kghbln:
> We should have progress here by the end of the month. Sit tight.

For known reasons I cannot expand on in publiv it was not possible to provide here. While it would have been really nice to have a new logo with 3.0.0 I guess having it for 3.1.0 will be fine too.

PR #2384, opened 2017-04-02
Tidy (DeprecationNoticeTaskHandler), refs 2357

mwjames:
This PR is made in reference to: #2357

- Using `smwgDeprecationNotices` (hidden from the normal user and is not to be used outside of SMW-core) instead of changing the `DeprecationNoticeTaskHandler` code directly

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with d39f1e6.

PR #2388, opened 2017-04-02
TextByChangeUpdater change update approach, refs 1481

mwjames:
This PR is made in reference to: #1481

A certain update sequence could allow for a text element to remain in the
index due to the missing information in the diff record about a removed
element.

Instead of using the insert-diff to replace text elements, use the delete-diff
to remove all items from the index and after that insert related data from
the primary data container. This approach ensures the content is in sync with
the `SemanticData` primary data update.

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with 74afabe.

PR #2389, opened 2017-04-02
ParserCachePurgeJob count links from the source

mwjames:
This PR is made in reference to: #1421

- Use the count of the original source list instead of the composite list to ensure that all chunks can be fetched

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with b1800eb.

PR #2397, opened 2017-04-08
SemanticData::getPropertyValues to return always an indexed array

mwjames:

- Ensures `SemanticData::getPropertyValues` always returns an index array independent of https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/2.5.x/includes/SemanticData.php#L405-L409

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with 2af43cb.

PR #2398, opened 2017-04-08
AskParserFunction to support @deferred mode

mwjames:
Some `#ask` result formats are able to defer the result display with the help of JavaScript but those formats still require the initial `#ask` query to be executed while the page content is build making it possible that the page display is delayed due to some complex or long running queries being executed synchronously with the build process.

The overcome the issue of a `#ask` to require an in-process execution (means during the page content build process), this PR introduces some minor modification to decouple query execution and display from a page build process. 

## Synopsis

The enemy of any up-to-date query is the `ParserCache` as it prevents most embedded queries to contain updated results. `@deferred` as option is yet another method (besides `QueryDependencyLinksStore` and the `ParserCachePurgeJob`) that can circumvent the  `ParserCache` (without disabling it) by providing an inception point for queries (those marked with `@deferred`) to be updated via the MediaWiki API after MediaWiki has processed the page and generated the static HTML content.

## Query examples

Non-deferred query, requires the result to be available during the page build process.

```
{{#ask: [[Category:Lorem ipsum]]
 |?Has property description=Description 
 |limit=3
}}
```

Deferred query
```
{{#ask: [[Category:Lorem ipsum]]
 |?Has property description=Description 
 |limit=3
 |@deferred
}}
```

deferred-query-2

## Technical notes

image

As the `@deferred` name suggests, the query request is postponed until a page is build and send to the browser and only then a request (initiated by the `smw.deferred.query`) is made to the MediaWiki API to generate the result and place it within the marked area.

There is no modification of the result format itself with `smw.deferred.query` being only responsible to request a `api.post` for `action=parse`.

- Added `@deferred` as fixed identifier to a `#ask` query that will signal the parser function to defer any activities for those printers that support `ResultPrinter::isDeferrable`
- `#ask` and `#show` are both supported given that a result format supports `isDeferrable`
- The `smw.deferred.query` script will execute the query request on behalf of the embedded `#ask` query
- The `smw.deferred.query` script will trigger a `smw.deferred.query` event to inform result context related objects that the result has been generated and is loaded
- Out-of-the box support is added for `CategoryResultPrinter`, `EmbeddedResultPrinter`, `ListResultPrinter`, and `TableResultPrinter`
- `ResultFormat::isExportFormat` and `limit=0` are excluded from the deferred request as those queries don't create any DB/Store connection while building a reference link 
- `@control=slider` was added to allow generating a slider which can be used to dynamically alter a result display without the need to reload a page
- If the `QueryDependencyLinksStore` is enabled then the initial query (the one that contains the embedded definition) will be tracked but any modified query via the `@control` will not.
- If the `QueryCache` is enabled then the initial query will be stored with the embedded settings, yet any modified query  via the `@control` will only be cached for the time of the non-embedded cache duration.
- image indicates a queued request

### JS event hook

The tooltip uses the event to ensure that after a deferred parse content that requires a tooltip display is correctly initialized for the context.

```
// Listen to the smw.deferred.query event
mw.hook( 'smw.deferred.query' ).add( function( context ) {
	 self.initFromContext( context );
} );
```

### How to make a result format deferrable?

Example for the SRF gallery format:

- Add `SRF\Gallery::isDeferrable` and return `true`
- Modify `ext.srf.gallary.slideshow.js` with 

```
 	$( document ).ready( function() {
 
+		var gallery = new srf.formats.gallery();
+
 		$( '.srf-gallery-slideshow' ).each(function() {
-			var gallery = new srf.formats.gallery();
 			gallery.slideshow( $( this ) );
 		} );
+
+		// Listen to the smw.deferred.query event
+		mw.hook( 'smw.deferred.query' ).add( function( context ) {
+			gallery.slideshow( context );
+		} );
+
 	} );
```

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/query/SMW_Query.php
+++ b/includes/query/SMW_Query.php
@@ -121,7 +121,7 @@ class SMWQuery implements QueryContext {
 
 		// stating whether this query runs in an inline context; used to
 		// determine proper default parameters (e.g. the default limit)
-		if ( $context === self::INLINE_QUERY ) {
+		if ( $context === self::INLINE_QUERY || $context === self::DEFERRED_QUERY ) {
 			$inline = true;
 		}
 
--- a/includes/query/SMW_QueryProcessor.php
+++ b/includes/query/SMW_QueryProcessor.php
@@ -10,6 +10,7 @@ use SMW\ApplicationFactory;
 use SMW\Message;
 use SMW\Query\QueryContext;
 use SMW\Query\ResultFormatNotFoundException;
+use SMW\Query\DeferredQuery;
 
 /**
  * This file contains a static class for accessing functions to generate and execute
@@ -43,8 +44,8 @@ class SMWQueryProcessor implements QueryContext {
 	 *
 	 * @return Param[]
 	 */
-	public static function getProcessedParams( array $params, array $printRequests = array(), $unknownInvalid = true ) {
-		$validator = self::getValidatorForParams( $params, $printRequests, $unknownInvalid );
+	public static function getProcessedParams( array $params, array $printRequests = array(), $unknownInvalid = true, $context = null ) {
+		$validator = self::getValidatorForParams( $params, $printRequests, $unknownInvalid, $context );
 		$validator->processParameters();
 		return $validator->getParameters();
 	}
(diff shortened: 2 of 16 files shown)


mwjames:
https://www.semantic-mediawiki.org/wiki/Result_printer_(deferred_mode)

https://sandbox.semantic-mediawiki.org/wiki/Issue/2398

kghbln:
Also:
* [`@deferred` query marker](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Deferred_query_marker)
* [`@control` query marker](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Control_query_marker)

jamesmontalvo3:
I just learned about the `@deferred` option. Very cool. Faster initial page load is awesome. Thanks @mwjames!

krabina:
None of the examples in sandbox work. Is this a sandbox problem or a bug in the deferred feature?
https://sandbox.semantic-mediawiki.org/wiki/Issue/2398

kghbln:
The browser complains about #3626 Not sure if this is the cause.

mwjames:
> The browser complains about #3626 Not sure if this is the cause.

Yes, this is the cause.

PR #2387, opened 2017-04-02
Special:Browse improve mobile display (replace <table> with <div>)

mwjames:
This PR is made in reference to: #

This PR addresses or contains:

- Replaces the use of `<table>` with `<div>` (together with `display: table;`)
- Provides better responsiveness towards smaller screens 
- Avoids `MobileFrontend` display  issues (uneven table columns) as seen in 

image

This PR includes:
- [ ] Tests (unit/integration)
- [ ] CI build passed

PR #2409, opened 2017-04-15
Add HtmlTable, refs 2387

mwjames:
This PR is made in reference to: #2387

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/Utils/HtmlTable.php
+++ b/src/Utils/HtmlTable.php
@@ -1,0 +1,145 @@
+<?php
+
+namespace SMW\Utils;
+
+use Html;
+
+/**
+ * @license GNU GPL v2+
+ * @since 3.0
+ *
+ * @author mwjames
+ */
+class HtmlTable {
+
+	/**
+	 * @since 3.0
+	 *
+	 * @param string $html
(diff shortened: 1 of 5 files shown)


src/Utils/HtmlTable.php
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * @since 3.0
+	 *
+	 * @param string $html
+	 * @param array $attributes
+	 *
+	 * @return string
+	 */
+	public static function table( $html = '', array $attributes = array() ) {
```
JeroenDeDauw: such static

mwjames: Well, its better than having `span` and `div`'s being sprinkled all over the code. Nothing wrong on using statics where an instance isn't required.

JeroenDeDauw: static

Like this you cannot provide an alternative implementation though

JeroenDeDauw: Clearly one of those cases where using static is not a total disaster, but still. Esp since you delegate to MediaWiki's Html... I'd be using an instance

PR #2413, opened 2017-04-15
UniquenessConstraintValueValidator to enforce NO_DEPENDENCY_TRACE

mwjames:

- Avoids unnecessary ID creation for a temporary query when a uniqueness constraint isn't cached and requires an active DB connection

This PR includes:
- [x] CI build passed


mwjames:
Failed caused by Scrutinizer — Errored (Disabled Scrutinizer as it started to throw errors)

PR #2420, opened 2017-04-15
TableResultPrinter to support a datatable output

mwjames:

- Adds native support for a datatable output that provides in-table search and navigational display support for larger datasets by relying on the https://github.com/DataTables/DataTables/releases/tag/1.10.13 plug-in
- It addressed accents searches [0] (PS: This is fixed only for the core datatable output not for the SRF:DataTables format)
- `class=datatable` is used to switch the output display, internal code generation remains untouched and  therefore is part of the `TableResultPrinter`
- Requires `onoi/shared-resources:~0.4` (https://github.com/onoi/shared-resources/commit/5f41bc737de659bdbc5dffd8fea7d91ddd93ad77)
- `ext.smw.tableprinter.css` contains some CSS rules to make it more align with the overall SMW design instead of the default DataTables `display` CSS
- `ext.smw.tableprinter.js` modifies the HTML table to be transformable by the DataTables plug-in and adds required translations

image

[0] https://www.semantic-mediawiki.org/wiki/Help_talk:Datatables_format#In-table_search_issue_2349

This PR includes:
- [x] CI build passed


mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2420

kghbln:
So this is basically a new result printer added to SMW core. Very cool.

kghbln:
It will be nice to allow the printer's name be translatable for use on special page "Ask" with "smw_printername_datatable"

i18n/en.json
```diff
@@ -531,5 +531,21 @@
 	"smw-patternedit-protection": "This page is protected and can only be edited by users with the appropriate <code>smw-patternedit</code> [https://www.semantic-mediawiki.org/wiki/Help:Permissions permission].",
 	"smw-pa-property-predefined_edip": "\"$1\" is a predefined property provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki] to indicate whether editing is protected or not.",
 	"smw-pa-property-predefined-long_edip": "While any user is qualified to add this property to a subject, only a user with a dedicated permission can edit or revoke the protection to an entity after it has been added.",
-	"smw-query-reference-link-label" : "Query reference"
+	"smw-query-reference-link-label" : "Query reference",
+	"smw-format-datatable-emptytable": "No data available in table",
+	"smw-format-datatable-info": "Showing _START_ to _END_ of _TOTAL_ entries",
+	"smw-format-datatable-infoempty": "Showing 0 to 0 of 0 entries",
```
kghbln: @mwjames Shouldn't this be something like "No results are shown"?

mwjames:
Hi,

>
> @mwjames Shouldn't this be something like "No results are shown"?

For details see, [0, 1].

[0] https://datatables.net/manual/i18n
[1] https://datatables.net/plug-ins/i18n/

Cheers

mwjames:
> So this is basically a new result printer added to SMW core. Very cool.

No.

Technically speaking it is a `format=table` or `format=broadtable`
being displayed as datatable (hence the `class` parameter) instead of
a wikitable.

kghbln:
> For details see, [0, 1].

So that may be changed. :)

> Technically speaking it is a `format=table` or `format=broadtable` being displayed as datatable (hence the `class` parameter) instead of a wikitable.

Ah, I get it. All other formats use the "class" parameter for CSS adaption, so I think it will be better to call this "tabletype" like it is e.g. similarly done with "charttype" for the "Jqplotseries" printer.

mwjames:
Hi,

> Ah, I get it. All other formats use the "class" parameter for CSS adaption,
> so I think it will be better to call this "tabletype" like it is e.g.
> similarly done with "charttype" for the "Jqplotseries" printer.

Semantically, it is the same table only with a different set of CSS
rules (and so it happens with some JS) but in the end it is ruled by
CSS hence the `class` parameter.

For example, the SRF:DataTables is different (while it may use the
same plug-in in a different version) in this regard because it doesn't
use a normal `<table>` and is created and updated via the
SMW/MediaWiki API.

kghbln:
> Semantically, it is the same table only with a different set of CSS rules (and so it happens with some JS) but in the end it is ruled by CSS hence the `class` parameter.

So this is a fixed class naming schema which still can be adapted. 

Another plus for "datatable". It works with the "MobileFrontend" extension while "DataTables" does not.

mwjames:
Hi,

The main objective was to allow a table to be searchable (from within
the cells and not via the browser ctrl+f key) and the options were
either to write our own title JS and plug it into the html output or
to use a well tested lib that does that without much change to the
existing format.

The SRF:DataTables does a lot more and is therefore distinguished from
the `class=datatable` output.

Another aspiration was that a "data" table can be displayed (with 100+
rows) without shattering (in terms of length) the visibility of wiki
page. After all the table and template format are probably the most
used formats to display queried data.

Cheers

mwjames:
Hi,

> So this is a fixed class naming schema which still can be adapted.

The plug-in itself comes with some predefined classes. For example,
you can use `class=datatable compact` and the output will be squeezed
a bit more.

`ext.smw.tableprinter.css` contains the special CSS rules for SMW and
if you compare it with the default datatable output then you'll see a
slightly different theme.

> Another plus for "datatable". It works with the "MobileFrontend" extension
> while "DataTables" does not.

The problem with the "MobileFrontend" is that it manipulates the
`<table>` CSS which is really hard to come by and is the reason a
table cannot be displayed in 100% width.

Cheers

mwjames:
Hi,

> The problem with the "MobileFrontend" is that it manipulates the
> `<table>` CSS which is really hard to come by and is the reason a
> table cannot be displayed in 100% width.

This is the rule that makes the width go bonkers because of the `
!important` attribute and invalidates any other CSS width table rule.

```
.content table {
    display: table;
    width: auto !important;
```

kghbln:
Thanks for your infos on this. This will help for further reference!

kghbln:
[Now documented](https://www.semantic-mediawiki.org/w/index.php?title=Help:Table_format&curid=317&diff=51315&oldid=47750]

mwjames:
> This is the rule that makes the width go bonkers because of the ` !important` attribute and invalidates any other CSS width table rule.

Fixed with https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/70a3e5f2a5a231e1ac20e1f8cdb900b7623707b9 (based on http://stackoverflow.com/questions/11178673/how-to-override-important).

kghbln:
Note: The feature request for the "datatables" format was added [here](https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/211).

pravoporyadok:
Greetings!
I use the template format and use it to display a custom table, but I would like to initialize the search on it, is it possible to somehow pass the initialization of the search module on the table in the request?

Issue #2065, opened 2016-12-03
Local-specific (ICU) sorting and collation


mwjames:
MySQL docs describe a collation as a "set of rules for comparing characters in a character set" [0].

Currently it is not possible to support local-specific sorting/collation (e.g. using ICU) because the sortfield [1] defined by the `smw_object_ids` table is at the same used as string matching field making it impossible to store sort modified representation of a value (which in case of ICU would not be a plain string but a computed UTF-8 character set).

To support local-specific sorting (preferable using the same setting as in $wgCategoryCollation) some changes are required:

- Add an extra sort field that can store  the representation returned by the ICU collator
- `DataItemHandler` is to return the expected sortfield instead of the current fixed `smw_sortkey`
- Modify `DescriptionInterpreter` and `EntityLookup` to use the new sortfield instead of  `smw_sortkey`

refs #1383, #2040

[0] http://dev.mysql.com/doc/refman/5.7/en/charset-general.html
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/SQLStore/TableSchemaManager.php#L108

## See also

- http://php.net/manual/en/intro.intl.php
- https://www.mediawiki.org/wiki/Manual:$wgCategoryCollation

DengroWikibase:
Thanks a lot for this, James! It's good to see things moving forward on this front.


________________________________

mwjames:
#2429 contains a prototype, yet we are missing specific examples that would allow a design of some integration tests for at least the following settings:

* `$wgCategoryCollation = 'uca-default';`
* `$wgCategoryCollation = 'uppercase';`

I only found [0] as authority source but this is a bit narrow to create something from it.

@kghbln Would be nice if we could somehow consolidate all issues relating to collation in one place.

[0] https://dev.mysql.com/doc/refman/5.7/en/charset-collation-effect.html
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/CONTRIBUTING.md#feature-requests

PS: As I reminder, please read [1].

mwjames:
Addressed with #2429 (3.0.x)

PR #2429, opened 2017-04-22
Add entity specific collation support (smw_sort), refs 2065

mwjames:
This PR is made in reference to: #2065

- Prototype on how to extend SMW and support a flexible (based on `$smwgEntityCollation`) collation mapping for entities.
- `$smwgEntityCollation` should correspond to the `$wgCategoryCollation` setting (also in regards to selected argument values), yet it is kept separate to have a better control over changes in regards to the collation, sorting, and display of values.
- As mentioned in #2065, a new field `smw_sort` is required that can hold a collated sort value and no longer conflicts with the search and match `smw_sortkey` field. 
- The `smw_sortkey` becomes a field that stores a literal value for a search hence it will be renamed in a follow-up to reflect that intention. 
- This PR alters the DB schema and can therefore not be back-ported.
- When changing the `$smwgEntityCollation` option it is expected that a user applies the same caution as for the standard MediaWiki system with either `rebuildData.php` or `updateEntityCollation.php` being run immediately post the change to ensure that data are updated and correspond to the setting.
- `$smwgEntityCollation` uses `identity` as default setting and will produce the same sorting results as without this change

## Example

Output example of the ID_TABLE with a `$wgCategoryCollation = 'uppercase';` setting.

image

## Update

In case the table is updated and data exists but without the `smw_sort` field being present, the `update.php` or `setupStore.php` will initiate a post processing where the content of `smw_sortley` is copied once to the `smw_sort` field during the setup. The message will be similar to:

```
Checking post creation activities ...
   Table smw_object_ids ...
   ... copying smw_sortkey to smw_sort ... done.
   ... done.
```
Depending on the size of the table, this process may take a moment. Any update on the collation/sort field hereafter has to be done executing the `updateEntityCollation.php` script.

The `updateEntityCollation.php` maintenance script will output something like:

```
$ php maintenance/updateEntityCollation.php

[ Notes ]

- $smwgEntityCollation              uca-default-u-kn
- $wgCategoryCollation              numeric

The setting of $smwgEntityCollation and $wgCategoryCollation are different
and may result in an inconsitent sorting display for entities.

[ Update ]

- Selecting rows ...                1 to 100659
- Updating the smw_sort field ...     0% (313/100659)

```

## Technical notes

- `SMW_SPARQL_QF_COLLATION` (see `DefaultSettings.php`) which allows to replicate collation specific information using `swivt:sort`
- `Collator` is the interface to MediaWiki's `Collation` class
- `TableFieldUpdater` is the sole responsible class to update the `smw_sortkey` and `smw_sort` field
- `TableIntegrityExaminer` will copy the content of `smw_sortkey` once at the time `smw_sort` is added to the schema
- `TableBuilder::getProcessLog` returns a list of activities occurred during the table update and is used in `TableIntegrityExaminer`
- `SortPropertyValueResourceBuilder` now handles the export of `_SKEY` resources
- `updateEntityCollation.php` script provides a tool for mass updates on the occasion that the `$smwgEntityCollation` setting is changed

This PR includes:
- [x] CI build passed


mwjames:
@kghbln Some testing guidelines:

- Run `update.php` or `setupStore.php` and you expect to see `... copying smw_sortkey to smw_sort ... done.`  only the first time the script is run
- `smw_sort` field should be visible in the `smw_object_ids` table with the same content as `smw_sortkey`
- Change `$smwgEntityCollation` to something like `numeric` or `uca-default-u-kn` [0]
- After the setting has changed, you are required to run `updateEntityCollation.php` with the result that the content in `smw_sort` has changed accordingly 
- Sorting in [1] should be different from [2] (using the default `identity` setting)

[0] https://www.mediawiki.org/wiki/Manual:$wgCategoryCollation
[1] https://sandbox.semantic-mediawiki.org/wiki/Issue/2429/NumericSorting
[2] https://cloud.githubusercontent.com/assets/1245473/25782835/abf402c8-338c-11e7-9e0b-940bae59503e.png

mwjames:
> Run update.php or setupStore.php and

I had to run `Special:SemanticMediaWiki` `updatetables` (as replacement for `update.php` or `setupStore.php`) otherwise the sandbox would have been blocked on:

```
Original exception: [7e298e0712bc0dbc0186f5b2] /wiki/Issue/2429/NumericSorting DBQueryError from line 1054 of /var/www/htdocs/mw/02100/w/includes/libs/rdbms/database/Database.php: A database query error has occurred. Did you forget to run your application's database schema updater after upgrading?
Query: SELECT smw_id,smw_sortkey,smw_sort FROM `02100_smw_object_ids` WHERE smw_title = 'Issue/2429/NumericSorting' AND smw_namespace = '0' AND smw_iw = '' AND smw_subobject = '' LIMIT 1
Function: SMWSql3SmwIds::getDatabaseIdAndSort
Error: 1054 Unknown column 'smw_sort' in 'field list' (localhost)
```

> expect to see ... copying smw_sortkey to smw_sort ... done. only the first time the script is run

The output contained:

```
Checking post creation activities ...
   Table smw_object_ids ...
   ... copying smw_sortkey to smw_sort ... done.
   ... done.
```

kghbln:
Documented:
- [`$smwgSparqlQFeatures`](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgSparqlQFeatures&type=revision&diff=57790&oldid=52081)
- [`$smwgEntityCollation`](https://www.semantic-mediawiki.org/wiki/Help:$smwgEntityCollation)
- [updateEntityCollation.php](https://www.semantic-mediawiki.org/wiki/Help:Maintenance_script_"updateEntityCollation.php")

kghbln:
@mwjames [Code docu](https://doc.semantic-mediawiki.org/namespaceSMW_1_1Maintenance.html) does not seem to have notes about the maintenance script

mwjames:
> Code docu does not seem to have notes about the maintenance script

Yes, because `updateEntityCollation.php` has no class in `SMW\Maintenance` as it uses `TableFieldUpdater` directly to push related updates when running the script.

kghbln:
Thanks for the information. Just wanted to make sure! :)

kghbln:
@mwjames Another question: The default for `$smwgEntityCollation` is "identity" is different to MW's `$wgCategoryCollation` which is "uppercase". So running "updateEntityCollation.php" should through a not about this incongruent setting even though the sysadmin did not change a thing. My theory is that SMW was always different in this respect however I one chooses to make a change in this respect the sysadmin should probably do congruent settings for both configuration parameters?!?

mwjames:
> The default for $smwgEntityCollation is "identity" is different to MW's $wgCategoryCollation which is "uppercase". 

Well, the default SMW sort mode has been that equally of "identity" therefore it should remain so since we don't want to break existing queries that would otherwise carry a different sort match. If users want to change the setting then of course it is suggested to match both.

kghbln:
> Well, the default SMW sort mode has been that equally of "identity"  ...

That's what I have guessed. Thanks for confirming!

PR #2435, opened 2017-04-29
Filter and armor against invisible chars

mwjames:
This PR is made in reference to: http://wikimedia.7.x6.nabble.com/I-really-can-t-understand-what-s-happening-with-my-form-template-tp5073264p5074223.html

- Armor against so called non-printing characters [0, 1] to avoid facts and properties appearing as equal when in fact there are not due to some hidden characters
- Remove visible and invisible shyness character  

[0] https://en.wikipedia.org/wiki/Right-to-left_mark
[1] https://en.wikipedia.org/wiki/Left-to-right_mark

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/datavalues/SMW_DataValue.php
+++ b/includes/datavalues/SMW_DataValue.php
@@ -14,6 +14,7 @@ use SMW\Localizer;
 use SMW\Message;
 use SMW\Options;
 use SMW\Query\QueryComparator;
+use SMW\Utils\CharArmor;
 
 /**
  * Objects of this type represent all that is known about a certain user-provided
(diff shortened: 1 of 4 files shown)


mwjames:
Those non-printing characters or control chars are mostly relevant to left-to-right scripts (such as English and Cyrillic) and right-to-left scripts (such as Persian, Arabic, Syriac and Hebrew) according to [0].

@kghbln Some outreach to the mentioned groups would be nice in order to evaluate the impact towards those users.

If you look at the `p-0450.json` test then you can not directly distinguish the one with and without control chars and without the `CharArmor` they would result in different facts and properties.

One way to the see the differences is by using a git client where it shows `â€Ž` or similar and is marking the position of invisible characters.

image


[0] https://en.wikipedia.org/wiki/Right-to-left_mark

Issue #2439, opened 2017-05-03
Removing the property and concept namespace from `$wgNamespacesToBeSearchedDefault` not possible


kghbln:
### Setup and configuration

- MediaWiki 	1.27.2 (1c409c5) 01:42, 7 April 2017
- PHP 	5.6.30 (cgi-fcgi)
- MySQL 	5.6.28-76.1-log
- Semantic MediaWiki	2.5.1 (568d058) 01:52, 22 April 2017

### Issue
https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2243 added the "Propery" and "Concept" namepaces to the `$wgNamespacesToBeSearchedDefault` array. Removing the property and concept namespace from `$wgNamespacesToBeSearchedDefault` is not possible:

I tried:
```php
$wgNamespacesToBeSearchedDefault = [
        NS_MAIN => true,
        NS_FILE => true,
	NS_GLOSSARY => true
        ];
```
and
```php
$wgNamespacesToBeSearchedDefault = [
        NS_MAIN => true,
        NS_FILE => true,
	SMW_NS_PROPERTY => false,
	SMW_NS_CONCEPT => false,
	NS_GLOSSARY => true
        ];
```
before and after the inclusion of Semantic MediaWiki with no change in behavior.

PR #2444, opened 2017-05-06
NamespaceManager avoid reset of user settings, refs 2439

mwjames:
This PR is made in reference to: #2439

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/NamespaceManager.php
+++ b/src/NamespaceManager.php
@@ -53,6 +53,7 @@ class NamespaceManager {
 		}
 
 		$this->addNamespaceSettings();
+		$this->addExtraNamespaceSettings();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


mwjames:
@kghbln FYI

mwjames:
Back-ported to 2.5.x with 8c9c4c3.

kghbln:
@mwjames Hmm ... there is still something in the water. I have added

```php
$wgNamespacesToBeSearchedDefault = [
        NS_MAIN => true,
        NS_FILE => true,
	SMW_NS_PROPERTY => true
        ];
```
to "LocalSettings.php" for sandbox but now neither Property nor Concept are searched by default.

mwjames:
>  there is still something in the water. I have added

Are you sure because my local test showed the opposite. The change was made so that any setting the exists are not overridden and in case they don't exists are added with a true value.

You can check the setting by adding `var_dump( $wgNamespacesToBeSearchedDefault )` to [0].

[0] https://github.com/wikimedia/mediawiki/blob/c64e2b3fede0d7793cfefa0ff67f2ddad67febd9/includes/user/User.php#L1563-L1595

kghbln:
Ah, oops indeed. Now it is also working in sandbox. Probably the change still somehow had to propagate through caching etc. Sorry for the noise!

kghbln:
I may note that `$wgNamespacesToBeSearchedDefault` must be done after the inclusion of SMW as far as semantic namespaces are concerned. Otherwise one will get as to be expected the following PHP notices:

```
Notice: Use of undefined constant SMW_NS_PROPERTY - assumed 'SMW_NS_PROPERTY' in /w/LocalSettings.php on line xxx
Notice: Use of undefined constant SMW_NS_CONCEPT - assumed 'SMW_NS_CONCEPT' in /w/LocalSettings.php on line xxx
```

Issue #2440, opened 2017-05-05
JS Date object in ext.smw.dataItem.time.js can cause date conversion issues


jamesmontalvo3:
The title may indicate that this belongs in the SRF list of issues, but I believe it should be here. See "Issue" section below for details.

### Setup and configuration

Issue present on the following:
- semantic-mediawiki.org
- sandbox.semantic-mediawiki.org
- My staging server
  - MW 1.27.2
  - SMW 2.5.1
  - SRF 2.4.2
- My staging server with SMW at dev-master
  - MW 1.27.2
  - SMW 3.0.0-alpha (022591a)
  - SRF 2.4.2
- My staging server with SMW and SRF at dev-master
  - MW 1.27.2
  - SMW 3.0.0-alpha (022591a)
  - SRF 2.5.0-alpha (1641870)

Issue **not** present on the following:
- My production server:
  - MW 1.25.5
  - SMW 2.4.0-rc1
  - SRF 2.3
- My staging server with SMW downgraded form 2.5.1 to 2.4.6
  - MW 1.27.2
  - SMW 2.4.6
  - SRF 2.4.2
- My staging server with SMW and SRF downgraded
  - MW 1.27.2 (1c409c5)
  - SMW 2.4.6
  - SRF 2.3

### Issue

Initially this may appear to be a Semantic Result Formats issue, but as you can see from the versions tested above the problem arises in SRF when upgrading SMW from 2.4.6 to 2.5.1. Since minor versions should be backwards compatible I think this is an SMW issue.

When using the `eventcalendar` format with SMW 2.5.1, events are display one year in the past. Additionally, they are often offset by a month into the past (e.g. a total of 13 months in the past). 

There are also JS errors, though they differ by version. See example below.

### Examples

In the following examples, the page includes the event calendar followed by a table display of the events. On the event calendar you will need to click back several times to find the events.

* [smw.o](https://www.semantic-mediawiki.org/wiki/User:Jamesmontalvo3/Test) events show on calendar in fall 2011 (should be in fall 2012). 
* [sandbox](https://sandbox.semantic-mediawiki.org/wiki/Utilisateur:Jamesmontalvo3) events show on calendar in spring 2016 (should be spring 2017).

#### Screenshot of smw.o JS errors
![image](https://cloud.githubusercontent.com/assets/716482/25765903/0d143ba6-31b5-11e7-8f40-0b6eab7b1720.png)

#### Screenshot of sandbox JS errors
[...]

PR #2451, opened 2017-05-08
Construct this.date after determining components

jamesmontalvo3:
Prior to this commit, this.date was constructed with just the year and then used setXYZ() methods (e.g. setMonth(), setDate()) to update the date as required based upon other values. This was causing bad dates as noticed in SemanticResultFormats `eventcalendar` format, where events showed up approximately a year in the past. This commit changes it so this.date is constructed with the full date information rather than being updated one component at a time.

This PR is made in reference to: #2440 (Pull requests used for troubleshooting: #2441, #2443)

This PR includes:
- [x] CI build passed

### DIFF
--- a/res/smw/data/ext.smw.dataItem.time.js
+++ b/res/smw/data/ext.smw.dataItem.time.js
@@ -52,39 +52,47 @@
 			}
 		}
 
-		// SMW seralization format with:
+		// SMW serialization format with:
 		if ( this.raw !== null ) {
 			var date = this.raw.split( '/' );
 
 			// [0] contains the calendar model
 			this.calendarModel = date[0];
-
-			this.date = new Date( date[1] );
 			this.precision = FLAG_YEAR;
 
-			// Note: January is 0, February is 1, and so on
 			if ( typeof date[2] !== 'undefined' ) {
-				this.date.setMonth( ( date[2] - 1 ) );
+				// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date
+				// Note: January is 0, February is 1, and so on
+				date[2] = date[2] - 1;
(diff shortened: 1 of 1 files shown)


mwjames:
> @kghbln requested your review on: SemanticMediaWiki/SemanticMediaWiki#2451

I'd appreciated if you could replace the date getter methods with
their UTC equivalent (as noted on the other thread) now that
`Date.UTC` is explicitly used.

PS: I'm only doing weekend sprints (incl. merge windows) but anyone
else is free to do a merge if it is really urgent.

Cheers

jamesmontalvo3:
So you're saying basically anywhere in this file where `getMonth` is used it should be replaced with `getUTCMonth`? And similar replacements for other time components?

jamesmontalvo3:
I made the changes I believe were required, and datatables and eventcalendar are working:

Eventcalendar working:
image

Datatables working:
image

mwjames:
> I made the changes I believe were required, and datatables and eventcalendar are working:

Thanks for the confirmation.

I'm glad that we found a solution with the help of the various parties involved especially given that issue presented or better shrouded itself obscure under certain conditions.

jamesmontalvo3:
Will this be pulled into the 2.5.x branch and given a tag? It seems like it should since it's a fix to a bug in 2.5.x. Thanks for everything!

mwjames:
> Will this be pulled into the 2.5.x branch and given a tag?

Later in the day I'll prepare a back-port in order for it to be deployable with 2.5.2.

jamesmontalvo3:
Thanks!

kghbln:
> Later in the day I'll prepare a back-port in order for it to be deployable with 2.5.2.

Cool, I could do the release early next week.

mwjames:
Back-ported to 2.5.x with 8a9bef2.

PR #2459, opened 2017-05-13
Travis switch to trusty environment

mwjames:
This PR is made in reference to: https://blog.travis-ci.com/2017-05-04-precise-image-updates

- Travis support wrote "Thanks for writing in. HHVM was unfortunately excluded from our last round of Ubuntu Precise 12.04 updates. Could you try using `dist: trusty` in your travis.yml and running the build again? "

This PR includes:
- [ ] CI build passed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
+# https://docs.travis-ci.com/user/languages/php
 language: php
+sudo: required
+
+# https://docs.travis-ci.com/user/trusty-ci-environment/
+dist: trusty
 
 matrix:
   fast_finish: true
   include:
-    - env: DB=mysql; MW=REL1_27; TYPE=coverage; PHPUNIT=4.8.*
+    - env: DB=mysql; MW=REL1_27; PHPUNIT=4.8.*
       php: 7.0
     - env: DB=mysql; MW=REL1_28; FUSEKI=2.4.0
       php: 5.5
@@ -17,10 +22,10 @@ matrix:
       php: 5.5
     - env: DB=mysql; MW=REL1_26; BLAZEGRAPH=1.5.2; PHPUNIT=4.8.*
       php: 5.6
-    - env: DB=mysql; MW=REL1_25; TYPE=benchmark; PHPUNIT=4.8.*
-      php: hhvm
-    - env: DB=sqlite; MW=master; PHPUNIT=4.8.*
-      php: hhvm
+    - env: DB=mysql; MW=REL1_25; TYPE=benchmark; PHPUNIT=5.7.*
+      php: hhvm-3.12
+    - env: DB=sqlite; MW=master; PHPUNIT=5.7.*
+      php: hhvm-3.12
     - env: DB=sqlite; MW=REL1_27; TYPE=composer; PHPUNIT=4.8.*
       php: 5.6
     - env: DB=mysql; MW=REL1_25; TYPE=relbuild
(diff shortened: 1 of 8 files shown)


mwjames:
The problem is that tests are run but the HHVM environment returns with a segmentation fault. Trying to output the stacktrace log doesn't reveal anything.

https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/231785737

```
Time: 4.09 minutes, Memory: 185.69MB

OK (2 tests, 2 assertions)
Core dumped: Segmentation fault
Stack trace in /tmp/stacktrace.6208.log
Segmentation fault (core dumped)
[37;41mScript php ../../tests/phpunit/phpunit.php -c phpunit.xml.dist handling the phpunit event returned with error code 139[39;49m

travis_time:end:014cd60c:start=1494650049551329151,finish=1494650355444313206,duration=305892984055
[0K
[31;1mThe command "bash ./tests/travis/run-tests.sh" exited with 139.[0m
travis_time:start:0dc591e0
[0K$ if ls /tmp/stacktrace* 1> /dev/null 2>&1; then cat /tmp/stacktrace*.log; fi

PHP Stacktrace:


travis_time:end:0dc591e0:start=1494650355452192275,finish=1494650355462260814,duration=10068539
[0K
[32;1mThe command "if ls /tmp/stacktrace* 1> /dev/null 2>&1; then cat /tmp/stacktrace*.log; fi" exited with 0.[0m
```

https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/231785738

```
FAILURES!
Tests: 4753, Assertions: 14344, Failures: 4, Skipped: 21.
Core dumped: Segmentation fault
Stack trace in /tmp/stacktrace.6824.log
Segmentation fault (core dumped)
Script php ../../tests/phpunit/phpunit.php -c phpunit.xml.dist handling the phpunit event returned with error code 139


The command "bash ./tests/travis/run-tests.sh" exited with 139.
$ if ls /tmp/stacktrace* 1> /dev/null 2>&1; then cat /tmp/stacktrace*.log; fi

PHP Stacktrace:



The command "if ls /tmp/stacktrace* 1> /dev/null 2>&1; then cat /tmp/stacktrace*.log; fi" exited with 0.
```

JeroenDeDauw:
+1

mwjames:
>  HHVM environment returns with a segmentation fault

After some back and forth and switching versions it became clear the both hhvm-15 and hhvm-18 would create a segmentation fault, selecting hhvm-12 on the other hand would allow to get it pass without failing on the OS.

It's been a while but all test jobs pass again.

image

kghbln:
> both hhvm-15 and hhvm-18 would create a segmentation fault,

Does this mean that we have a support ceiling when it comes to hhvm and should this be noted in the compat docu?

mwjames:
> Does this mean that we have a support ceiling when it comes to hhvm and should this be noted in the compat docu?

No. The tests would all pass on either version with the only problem that __after__ the tests have finished the OS would output a seg fault and have the travis build fail. I have no idea what's causing it but other people seem to have similar issues [0].

Before the change we run on HHVM 3.6, now it is HHVM 3.12, and maybe in future (without a seg fault) we are able to run an more dated version.

[0] https://github.com/travis-ci/travis-ci/issues/5634

kghbln:
That's indeed an interesting variant. Thanks for the insight on this and cool to see everything pop up in green again. Cool!

mwjames:
Of course, it goes without saying that we are cheating on the hhvm#6899 issue but since we are not at fault here, we just skip those three tests in regards to `.888500 vs. .888499`.

mwjames:
> both hhvm-15 and hhvm-18 would create a segmentation fault,

... and it seems the reason for those segfaults can be found in https://github.com/facebook/hhvm/issues/7779#issuecomment-300917617 given the same symptoms as outlined above.

https://phabricator.wikimedia.org/T165074

mofarrell:
The latest release `3.19.3` should include the patch for that issue.  We suggest testing on travis using docker.  Our newest binaries don't support trusty, so I believe it is the only way to test on travis.

Facebook's XHP project includes a travis configuration that runs hhvm in docker.
https://github.com/facebook/xhp-lib

Issue #2453, opened 2017-05-10
Inquiry: SMW updates on article delete


TK-999:
### Setup and configuration

- SMW version: 2.4.0
- MW version: 1.19
- PHP version: 7.0.7
- DB (MySQL etc.): MySQL 5.7

### Issue
I'd like to inquire about the update jobs which are triggered by SMW on article deletion.

As I understand, the `ArticleDelete` hook handler creates and dispatches an instance of `UpdateDispatcherJob`, which then handles creating and dispatching individual `UpdateJob`s as needed. My question is as follows:

 https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/MediaWiki/Jobs/UpdateDispatcherJob.php#L99 gets all properties for the page that was deleted, and schedules an `UpdateJob` for all pages which share any of these properties. I'd like to ask about this update - how does the page delete affect the semantic data of other pages which share this property? Are there situations where it's possible to avoid having to update all pages which share any property with the page that was deleted without corrupting data?

mwjames:
Hi,

> how does the page delete affect the semantic data of
> other pages which share this property?

A deleted page maybe used as reference on other pages (displayed as
blue link) and after a delete, a user would expect to see a "red" link
on no longer existing references. The `UpdateJob` is mostly dispatched
to correct (invalidate parser cache, null edit) any possible
references for annotations or queries on those connected pages.

After an entity is deleted (== page), Semantic MediaWiki has no longer
any knowledge or reference that would allow it to query references
therefore the `UpdateDispatcherJob` is invoked before the actual data
is removed from the Store to fetch any properties that connects them.

Why use all properties and not query only possible matches? As I said
before, as soon as the data are removed from the Store, we are no
longer able to query them therefore we choose the more broader
approach and avoid having to run queries during the `ArticleDelete`
transaction and possibly block the transaction and instead post the
workload into the job queue for deferred processing. Of course, by the
time the job queue runs `ArticleDelete` was processed and the Store no
longer has any reference to the deleted entity with
`UpdateDispatcherJob` acting on what was transferred in the job
parameters and execute `Store::getAllPropertySubjects`.

Why not use the incoming properties to limit the pool of properties
with a reference? I made a comment in [0] but never follow-up to see
which is more efficient and requires less resources pre/post the
`ArticleDelete` task.

On second thought, maybe it would be more effective (and efficient) to
just use the incoming properties (as those are indicate a known
reference).

```
[...]

PR #2460, opened 2017-05-13
ArticleDelete to restrict pool of properties in update dispatcher, refs 2453

mwjames:
This PR is made in reference to: #2453

- Only use the incoming properties to restrict the list of potential referenced entities that require an update after a subject has been deleted.

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/MediaWiki/Hooks/ArticleDelete.php
+++ b/src/MediaWiki/Hooks/ArticleDelete.php
@@ -4,7 +4,11 @@ namespace SMW\MediaWiki\Hooks;
 
 use SMW\ApplicationFactory;
 use SMW\DIWikiPage;
+use SMW\SemanticData;
 use SMW\EventHandler;
+use Wikipage;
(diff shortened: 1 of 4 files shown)


mwjames:
@kghbln I cannot really quantify the impact in light of the #2453 comments but the list of follow-up `UpdateJob` should be less.

kghbln:
> @kghbln I cannot really quantify the impact in light of the #2453 comments but the list of follow-up UpdateJob should be less.

I will have a look once in a while. I usually only see this en passant since my cron does the jobs round the clock. However in comparison to Wikia I only deal with homeopathic numbers. I think having a smaller number is good.

TK-999:
@mwjames , thank you very much! I will communicate this and will get back in touch once we've reapplied it. :)

mwjames:
Back-ported to 2.5.x with 3b6e30d.

Issue #1234, opened 2015-10-27
SMWSQLStore3Readers::getInProperties performance


benhutchins:
I'm working with a wiki that has a large data set, and seeing that currently I'm unable to edit a large number of pages due to a performance issue resulting from `SMWSQLStore3Readers::getInProperties`.

`includes/storage/SQLStore/SMW_SQLStore3_Readers.php`:`getInProperties` results in a generated SQL statement that uses a very inefficient `INNER JOIN`.

The generated query is `SELECT DISTINCT smw_title,smw_sortkey FROM "smw_object_ids" INNER JOIN "smw_di_wikipage" AS t1 ON t1.p_id=smw_id WHERE t1.o_id='19683';`

``` sql
EXPLAIN ANALYZE VERBOSE SELECT DISTINCT smw_title,smw_sortkey FROM "smw_object_ids" INNER JOIN "smw_di_wikipage" AS t1 ON t1.p_id=smw_id WHERE t1.o_id='19683';

                                                                             QUERY PLAN                                                                             
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Unique  (cost=27277.13..28047.51 rows=66035 width=54) (actual time=49404.123..78673.174 rows=647 loops=1)
   Output: smw_object_ids.smw_title, smw_object_ids.smw_sortkey
   ->  Sort  (cost=27277.13..27533.92 rows=102717 width=54) (actual time=49404.119..76402.701 rows=101614 loops=1)
         Output: smw_object_ids.smw_title, smw_object_ids.smw_sortkey
         Sort Key: smw_object_ids.smw_title, smw_object_ids.smw_sortkey
         Sort Method: external merge  Disk: 5040kB
         ->  Hash Join  (cost=8849.69..15212.81 rows=102717 width=54) (actual time=971.847..4724.746 rows=101614 loops=1)
               Output: smw_object_ids.smw_title, smw_object_ids.smw_sortkey
               Hash Cond: (t1.p_id = smw_object_ids.smw_id)
               ->  Bitmap Heap Scan on mediawiki.smw_di_wikipage t1  (cost=1940.48..4644.44 rows=102717 width=4) (actual time=24.473..1161.580 rows=101614 loops=1)
                     Output: t1.p_id
                     Recheck Cond: (t1.o_id = 19683)
[...]

PR #2461, opened 2017-05-13
Improve Store::getInProperty performance, refs 1234

mwjames:
This PR is made in reference to: #1234

- The following [content](https://gist.github.com/mwjames/103afa6ecd53b666de075e69cc1970ca) was used to simulate some high data load
- Changes on pre/post to the query plan can be found [here](https://gist.github.com/mwjames/0bbe876b11ec4e134cddd0c46345b7d5)
- The PR in particular addresses the `Store::getInProperties` performance with a __significant improvement__ in query execution time by extending the `INNER JOIN` with a subquery to pre-select and filter possible matches before being materialized. One example entity with > 100 incoming properties and an overall table size that covers ~1M triples demonstrated the following changes pre/post this PR.
  - MySQL ( `841.3639ms` ↘ `39.1431ms`)
  - Postgres (`1282.6319ms` ↘ `59.8922ms`) 
- The suggestion from #1234 on using something like  ` ... WHERE smw_id IN (SELECT p_id FROM mediawiki."smw_di_wikipage" WHERE o_id='19683’);` did not create the expected improvement on MySQL hence a different query was selected that showed favorable query plans on both MySQL and Postgres.

To eliminate filesort and other expensive operations we needed to:

* Create additional composite indexes to ensure the query planner choose the right one with the objective of using an index
* Eliminate `ORDER BY` as it is expected to returns all matches and sort the results during a post processing
*  Filter possible matches before being materialized

```sql
SELECT DISTINCT smw_title,smw_sortkey,smw_iw FROM "smw_object_ids"
INNER JOIN (
  SELECT p_id FROM "smw_di_wikipage" WHERE o_id='266' GROUP BY p_id
 ) AS t1 ON t1.p_id=smw_id
WHERE smw_iw!=':smw' AND smw_iw!=':smw-delete'
```

Tested on:

```
MediaWiki	1.30.0-alpha (3d26a96)
PHP	7.1.1 (apache2handler)
MariaDB	10.1.21-MariaDB
ICU	57.1
```
```
MediaWiki	1.26.2
PHP	5.6.8 (apache2handler)
PostgreSQL	9.3.12
ICU	54.1
```

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1495,4 +1495,18 @@ return array(
 	'smwgEntityCollation' => 'identity',
 	##
 
+	##
+	# Entity lookup specific features
+	#
+	# - SMW_EL_NONE applies no query or schema changes
+	#
+	# - SMW_EL_INPROP enables a new query form for selecting incoming properties
+	#   (#1234)
+	#
+	# @since 3.0
+	# @default false
+	##
+	'smwgEntityLookupFeatures' => SMW_EL_INPROP,
+	##
+
 );
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -174,6 +174,7 @@ class Settings extends Options {
 			'smwgSimilarityLookupExemptionProperty' => $GLOBALS['smwgSimilarityLookupExemptionProperty'],
 			'smwgPropertyInvalidCharacterList' => $GLOBALS['smwgPropertyInvalidCharacterList'],
 			'smwgEntityCollation' => $GLOBALS['smwgEntityCollation'],
+			'smwgEntityLookupFeatures' => $GLOBALS['smwgEntityLookupFeatures'],
 		);
 
 		self::initLegacyMapping( $configuration );
--- a/includes/storage/SQLStore/SMW_SQLStore3_Readers.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3_Readers.php
@@ -33,6 +33,11 @@ class SMWSQLStore3Readers {
 	private $factory;
 
 	/**
+	 * @var SqlEntityLookupResultFetcher
+	 */
+	private $sqlEntityLookupResultFetcher;
+
+	/**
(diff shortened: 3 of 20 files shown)


mwjames:
This requires either `update.php` or `setupStore.php` to be run.

@kghbln One thing to be noted is due to the added indices, the `update.php` might take a bit longer the first time as those are created and depend largely on the size of the DB.

kghbln:
@mwjames This surely will come out with a bang! The performance improvements seen are quite impressive! WOW!

> One thing to be noted is due to the added indices, the update.php might take a bit longer the first time as those are created and depend largely on the size of the DB.

We should note this is the RELEASE NOTES in any case.

DefaultSettings.php
```diff
@@ -1495,4 +1495,18 @@
...
+	##
+	# Entity lookup specific features
+	#
+	# - SMW_EL_NONE applies no query or schema changes
+	#
+	# - SMW_EL_INPROP enables a new query form for selecting incoming properties
+	#   (#1234)
+	#
+	# @since 3.0
+	# @default false
```
kghbln: @mwjames Shouldn't this say `SMW_EL_INPROP`

mwjames:
Well, the setting is internal in case something seems fishy in order
to easily switch back and compare results without reworking the whole
code from the inside.

I hope that by the time of SMW 3.0 the setting can vanish and yes, the
default value is different.

kghbln:
> I hope that by the time of SMW 3.0 the setting can vanish and yes, the default value is different.

You mean the whole `smwgEntityLookupFeatures` parameter or just "false" as the internal setting?

DefaultSettings.php
kghbln: See https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2758

kghbln:
[Documented on wiki](https://www.semantic-mediawiki.org/wiki/Help:$smwgEntityLookupFeatures)

PR #2471, opened 2017-05-20
Add $smwgUseCategoryRedirect to handle redirects/errors on categories

mwjames:
This PR is made in reference to: http://wikimedia.7.x6.nabble.com/Query-doesn-t-show-all-rows-tp5074992p5075018.html

- Introduces`$smwgUseCategoryRedirect` (enabled by default) to allow finding redirects on categories, use them as annotation value, and if necessary track errors on invalid redirects to make them discoverable beyond `Special:Browse`

image

vs.

image

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -294,6 +294,15 @@ return array(
 	##
 
 	###
+	# Resolves redirects and errors in connection with categories
+	#
+	# @since 3.0
+	# @default true
+	##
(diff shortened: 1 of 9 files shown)


kghbln:
[Documented on wiki.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgUseCategoryRedirect&type=revision&diff=57212&oldid=51705)

PR #2475, opened 2017-05-21
Capitalize predefined property only, refs 673, 2282

mwjames:
This PR is made in reference to: #673, #2282, 
http://wikimedia.7.x6.nabble.com/various-semantic-properties-related-isssues-tp5074872p5074873.html

- In case of `$wgCapitalLinks=false`, restrict property name uppercase conversion on special properties only 

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with a3499b6.

Issue #2469, opened 2017-05-20
Limit the number of queries allowed per page


Reasno:
### Setup and configuration

- SMW version:2.4.6
- MW version:1.28
- PHP version:HHVM 3.19
- DB (MySQL etc.):MySQL 5.6

### Issue

After an investigation in the wiki I maintain, I found out the number of queries user made through smw per page is staggering. The most usual abuse is that they want to chain queries in an exponential way. For example: #ask all countries in a page, then #ask all cities in each country and again #ask all members of city councils in each city. 

Having browsed through Manual: Performance Tuning, I noticed that those tips were more focused on limit complex individual query. In vanilla MediaWiki there are expensive parser function counter which help us to limit db query numbers. I wonder if smw offers something similar? Or if SMW supports expensive parser function counter per se?

mwjames:
> I wonder if smw offers something similar?

No.

> if SMW supports expensive parser function counter per se

Not to my knowledge and I would have to inform myself before I can answer this question in a more detailed fashion on whether this would allow to reduce/limit the the amount of #ask per page and to what extend.

> want to chain queries 

On the topic of chaining results, [0] may reduce the amount and the need of chained queries.

>The most usual abuse is that they want to chain queries in an exponential way.

In cases where queries have the same signature, [1] may help reduce and reuse results from a previous computing cycle. The feature is experimental but we can see from the smw.org statistics [2] that reuse is likely, yet we are still conducting research on how much it can actually lower the requirements for computing resources.

More insights and use cases to the mentioned issue are welcome and any pre-work on the "expensive parser function counter" would help lower the time in gathering information about the topic and possible limitations.

As for "Limit the number of queries allowed per page", you could count the number of #ask iterations per page and cut off (or bail-out) if a specific threshold is reached. Of course, those are discriminatory restrictions without any analysis of the queries or context itself. It would be similar to when a shop owner provides a flat of 10 apples per customer without the customer being able to choose between good or bad apples, all s(he) knows is that 10 apples will be provided by the time they leave the shop.

[0] https://www.semantic-mediawiki.org/wiki/Help:Property_chains_and_paths
[1] https://www.semantic-mediawiki.org/wiki/Help:Query_result_cache
[2] https://gist.github.com/mwjames/94bfcd8ca3a3d322b1905199c3828178

Reasno:
Thanks for the fast reply!

[...]

PR #2476, opened 2017-05-21
Count and limit expensive #ask/#show functions, refs 2469

mwjames:
This PR is made in reference to: #2469 

- `$smwgQExpensiveThreshold` (Threshold in seconds that define an expensive #ask/#show function) and
- `$smwgQExpensiveExecutionLimit` (Limit of expensive #ask/#show functions)
-  It counts all expensive functions (as per definition) and cut-off any #ask/#show that follows the reached limit in a semi sequential process with the exception that a #ask/#show (limit=0) can still be executed even though the counter has reached the limit. In case of `$smwgQExpensiveThreshold=0` any #ask/#show will be counted towards the limit.
- The limit is disabled by default.

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -472,6 +472,31 @@ return array(
 	##
 
 	###
+	# Expensive threshold
+	#
+	# The threshold defined in seconds denotes the ceiling as to when a #ask or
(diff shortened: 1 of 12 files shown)


mwjames:
image

Issue #2479, opened 2017-05-26
Invalid non-numeric limit passed to limitResult()


Nikerabbit:
### Setup and configuration

- SMW version: 2.5.2
- MW version: 1.30.0
- PHP version: 3.9.12 (hhvm)
- DB (MySQL etc.): 10.0.29

### Issue

Produces a [stack trace](https://www.semantic-mediawiki.org/wiki/Help:Identifying_bugs) and outputs:

```
RuntimeException from line 229 of ////extensions/SemanticMediaWiki/src/MediaWiki/Database.php: Invalid non-numeric limit passed to limitResult()

#0 ////includes/libs/rdbms/database/Database.php(1381): Wikimedia\Rdbms\Database->limitResult()
#1 ////includes/libs/rdbms/database/Database.php(1327): Wikimedia\Rdbms\Database->selectSQLText()
#2 ////extensions/SemanticMediaWiki/src/MediaWiki/Database.php(224): Wikimedia\Rdbms\Database->select()
#3 ////extensions/SemanticMediaWiki/includes/storage/SQLStore/SMW_SQLStore3_Readers.php(504): SMW\MediaWiki\Database->select()
#4 ////extensions/SemanticMediaWiki/includes/storage/SQLStore/SMW_SQLStore3_Readers.php(592): SMWSQLStore3Readers->getPropertySubjects()
#5 ////extensions/SemanticMediaWiki/src/SQLStore/EntityStore/DirectEntityLookup.php(86): SMWSQLStore3Readers->getAllPropertySubjects()
#6 ////extensions/SemanticMediaWiki/includes/storage/SQLStore/SMW_SQLStore3.php(225): SMW\SQLStore\EntityStore\DirectEntityLookup->getAllPropertySubjects()
#7 ////extensions/SemanticMediaWiki/includes/articlepages/SMW_PropertyPage.php(246): SMWSQLStore3->getAllPropertySubjects()
#8 ////extensions/SemanticMediaWiki/includes/articlepages/SMW_PropertyPage.php(80): SMWPropertyPage->getPropertyValueList()
#9 ////extensions/SemanticMediaWiki/includes/articlepages/SMW_OrderedListPage.php(186): SMWPropertyPage->getHtml()
#10 ////extensions/SemanticMediaWiki/includes/articlepages/SMW_OrderedListPage.php(94): SMWOrderedListPage->showList()
[...]

PR #2481, opened 2017-05-27
RequestOptions to cast int on limit/offset, refs 2479

mwjames:
This PR is made in reference to: #2479

- Cast int on limit and offset
- Adds an integration test (p-1001.json) to verify that an input like `1foo` on a property page doesn't cause an exception

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with 7c3900f.

PR #2487, opened 2017-05-27
Fix for finallistsep not showing up

s7eph4n:
When using the list format `$this->numRows` contains the current row number starting from 0, so when comparing it to the number of rows in the result you need to add 1.
Then you need to make sure that the normal listsep is returned only for rows strictly less than the row count. For row number == row count the finallistsep needs to be returned.

(Generally `$this->numrows` should probably start from 1, but I don't  want to dig in right now and fix all the usages.)

This PR includes:
- [ ] CI build passed

### DIFF
--- a/includes/queryprinters/ListResultPrinter.php
+++ b/includes/queryprinters/ListResultPrinter.php
@@ -325,7 +325,7 @@ class ListResultPrinter extends ResultPrinter {
 
 		if ( $this->numRows > 0 && $this->isPlainlist() )  {
 			// Use comma between "rows" other than the last one:
-			return ( $this->numRows <= $res->getCount() ) ? $this->listsep : $this->finallistsep;
+			return ( $this->numRows + 1 < $res->getCount() ) ? $this->listsep : $this->finallistsep;
 		}
 
 		if ( $this->rowSortkey !== '' ) {


mwjames:
Do you have simple example where this becomes obvious?

s7eph4n:
See https://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial:Requ%C3%AAter&q=%5B%5BCategory%3AExample+sepvaluesep%5D%5D&p=format%3Dlist%2Flink%3Dall%2Fheaders%3Dshow%2Fsearchlabel%3D...-20further-20results%2Fsep%3D%2C&po=%3FHas+text%0A&eq=no#embed_show

Expected: Template sep valuesep (Has text 123, 456), and Template sep valuesep (Has text abc, efg)

Observed: Template sep valuesep (Has text 123, 456), Template sep valuesep (Has text abc, efg)

s7eph4n:
Before merging, maybe let's discuss  #2488.

kghbln:
> Before merging, maybe let's discuss #2488.

So we basically all agree that the "list" format could do a major re-factoring. I understand that this pull fixes an issue with the current version. So merging this will not hurt, will it?

s7eph4n:
Well, merging this would re-instate a behavior that apparently nobody missed and that I'd like to remove in the re-factoring. Weißt schon, schlafenden Hunde und so. 😛

kghbln:
> Well, merging this would re-instate a behavior that apparently nobody missed and that I'd like to remove in the re-factoring. Weißt schon, schlafende Hunde und so. :stuck_out_tongue:

Fair enough, though we have a very high latency when it comes to upgrading and missing beloved features. However it this is not going to be part of the re-factoring ...

We can leave this open and close as soon as the re-factoring was done. Thanks for the info!

s7eph4n:
Abandoned.

PR #2491, opened 2017-06-03
Provide a ChunkedIterator, refs 1938

mwjames:
This PR is made in reference to: #1938

- Adds a `ChunkedIterator` to avoid a possible OOM in cases where outdated entities reach a level of insanity 
- `EntityIdDisposerJob` to post a `Databae::commitAndWaitForReplication` where available (1.28+) after each chunk to release the transaction queue

http://wiki.teamliquid.net/dota2/Special:Statistics

image

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/IteratorFactory.php
+++ b/src/IteratorFactory.php
@@ -4,6 +4,7 @@ namespace SMW;
 
 use SMW\Iterators\ResultIterator;
 use SMW\Iterators\MappingIterator;
+use SMW\Iterators\ChunkedIterator;
 
 /**
  * @license GNU GPL v2+
@@ -34,6 +35,18 @@ class IteratorFactory {
 	 */
 	public function newMappingIterator( $iterable, callable $callback ) {
 		return new MappingIterator( $iterable, $callback );
+	}
+
+	/**
+	 * @since 3.0
+	 *
+	 * @param Iterator/array $$iterable
+	 * @param integer $chunkSize
+	 *
+	 * @return ChunkedIterator
+	 */
+	public function newChunkedIterator( $iterable, $chunkSize = 500 ) {
+		return new ChunkedIterator( $iterable, $chunkSize );
 	}
 
 }
--- a/src/Iterators/ChunkedIterator.php
+++ b/src/Iterators/ChunkedIterator.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace SMW\Iterators;
+
+use IteratorIterator;
+use Traversable;
+use ArrayIterator;
+use InvalidArgumentException;
+use RuntimeException;
+
+/**
+ * @see Guzzle::ChunkedIterator
+ *
+ * @license GNU GPL v2+
+ * @since 3.0
+ */
+class ChunkedIterator extends IteratorIterator {
+
+	/**
+	 * @var integer
+	 */
+	private $chunkSize = 0;
+
+	/**
+	 * @var array
+	 */
+	private $chunk;
+
+	/**
+	 * @since 3.0
+	 *
+	 * @param Traversable|array $iterator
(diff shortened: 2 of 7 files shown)


JeroenDeDauw:
So this is for performance reasons? What exactly is the issue that is avoided here? I had some guesses though it looks like none of those are the case after checking the code

mwjames:
> So this is for performance reasons? What exactly is the issue that is avoided here?

"EntityIdDisposerJob to post a Databae::commitAndWaitForReplication where available (1.28+) after each chunk to release the transaction queue"

mwjames:
Only during the foreach chunking is the result iterator being resolved on its members and exposes the row elements hence avoiding any exhausting array operations before and after the iteration.

JeroenDeDauw:
Ah yes, somehow I missed that line. That makes sense.

Somehow a counterpart to https://github.com/JeroenDeDauw/BatchingIterator/ which helps on the reading side, where this one helps on the writing one

kghbln:
This is what I had in April since I forgot to add cron:

20170430_39992_spezialstatistik cropped

It took the script 24 hours to complete and the size of the DB dropped from 5.7 GB to 3.7 GB. So yes, a good idea. Besides, that particular wiki produces 200,000 outdated entities a month.

mwjames:
> It took the script 24 hours to complete and the size of the DB dropped from 5.7 GB to 3.7 GB. So yes, a good idea. Besides, that particular wiki produces 200,000 outdated entities a month.

One thing to note, I did not add `Database::commitAndWaitForReplication` on the `rebuildData` process since running it from the command line or cron-job should handle the transaction queue (more precisely the MediaWiki Maintenance class) itself. Of course, if it necessary we can change that if that makes the process a bit more quicker.

In general, the iterator will not speed-up anything but is to avoid any expensive array operations which may occur given the amount of entries one could reach.

> Besides, that particular wiki produces 200,000 outdated entities a month.

What are user doing to reach that level or better what do they change?

kghbln:
Thanks for the info.

> Of course, if it necessary we can change that if that makes the process a bit more quicker.

I think it will be nice if the process may be accelerated somehow. Interestingly clearing just 200,000 outdated entities takes just an hour plus another hour shallow update. So the consecutive runs were faster than the very first one. 

> What are user doing to reach that level or better what do they change?

This is a private wiki so I did not have a look in detail. However, they are using heaps of properties (about 100 a page) and queries on a page which are heavily interconnected within the wiki. Just add high editing activity to it. Just last week they moved  a couple of pages and the SMWUpdate job took about 650,000 ms each to complete. This wiki is an exception since all other are pretty smooth.

mwjames:
> I think it will be nice if the process may be accelerated somehow. Interestingly clearing just 200,000 outdated entities takes just an hour

That's going to be a bit difficult without some qualitative analysis because the process here is fairly simple:
- Select all entities with `smw_iw=:smw:delete` (returns an iterator)
- `EntityIdDisposerJob` selects a single ID (chunking is addressed above)
- `PropertyTableIdReferenceDisposer::cleanUpTableEntriesById` is called and is encapsulated as `beginAtomicTransaction` to go through each registered table and remove any reference to the ID. At the end some events to `cached.prefetcher.reset` and `factbox.cache.delete` are triggered to remove any remaining cache entries for a matched subject.

Each single delete is done on a ID based field which should have an appropriate index..

mwjames:
Back-ported to 2.5.x with ca36069.

kghbln:
> That's going to be a bit difficult without some qualitative analysis because the process here is fairly simple:

Let's rather keep it as is and gain some more experience. There is always time for further improvements if there is the need to.

Issue #2494, opened 2017-06-03
Improved property change propagation


mwjames:
## Issue

As noted in [0] the process is to ensure that any relevant change to a property specification is being propagated and related entities are updated to reflect those changes (type, unit etc.). In recent MW releases relying on a sequential processing of updates and job queue is no longer guaranteed for processes that do not make that order explicit due to the deferring update logic. 

Aside from the update sequence, an elevating factor is the amount of possible updates required and was already noted in [1] with "FIXME: this will kill large wikis! Use incremental updates!".

The issue is that once the type (or any other declarative property attribute) is changed, using `Store::getAllPropertySubjects` can no longer create a list of subjects that match the old specification (a new type means new table and storage etc.), so fetching all members has to happen before a new specification is "activated".

## Approach

The solve the outlined insufficiencies, a new process is proposed where in case of a change, the property gets locked with the update being suspended until a `ChangePropagationDispatchJob` is scheduled which would first execute `Store::getAllPropertySubjects` (to find all related entities before the update) before updating the property with the new property specification.

- `PropertyChangePropagationNotifier` will detect any change to the listed declarative properties and suspend the update by setting property `_CHGPRO` and stores the new specification as serialized JSON blob
- A detected change will trigger a `ChangePropagationDispatchJob`
- The property page is being locked with the help of `PermissionPthValidator` and `EditProtectionValidator` and a notice will be shown as long as property `_CHPRO` remains part of the `SemanticData` annotation
- When `ChangePropagationDispatchJob` is run:
  - Fetch  `Store::getAllPropertySubjects` 
  - Schedule updates via `ChangePropagationUpdateJob` for all selected subjects and once finished 
  - Schedule a new `ChangePropagationUpdateJob` to add the new `SemanticData` from the serialized `_CHGPRO` blob

The serialized JSON blob is stored with property `_CHGPRO` because attempts failed when using it as simple job parameter with something like [2] due to limited space job parameters can occupy and failed on a unserialized error during job execution.

![image](https://user-images.githubusercontent.com/1245473/27986652-9b19416c-643c-11e7-878f-29178bf8d43a.png)

![image](https://user-images.githubusercontent.com/1245473/27986665-d3306de6-643c-11e7-8df2-77cd1fef6387.png)

[0] https://www.semantic-mediawiki.org/wiki/Property_type_change_propagation
[...]

PR #2495, opened 2017-06-03
Improve property change propagation

mwjames:
This PR is made in reference to: #2494, https://phabricator.wikimedia.org/T154425#3323418

- `Store::getPropertySubjects` and `Store::getAllPropertySubjects` will now return an iterator instead of an array to avoid a possible OOM (see migration-guide-3.0.md)
- Added `AppendIterator` to deal with different iterators in a coherent way
- `SQLStore::fetchSemanticData` (private) avoids displaying any property that has been marked deleted or outdated
- `PropertyChangePropagationNotifier` is to detect changes according to `$smwgDeclarationProperties`
- Added `ChangePropagationDispatchJob` to schedule `ChangePropagationUpdateJob` and find all related entities using `ChangePropagationEntityFinder`
- Added `ChangePropagationUpdateJob` as wrapper around `UpdateJob` to easily count remaining jobs for the specific task using `JobQueueGroup::getSize` to circumvent an implementation of a tracking mechanism for estimating pending jobs
- Estimating `ChangePropagationUpdateJob` (`JobQueueGroup::getSize` caches the the result and only counts on jobs not on titles) comes with some caveats:
  - Since we can not distinguish jobs per title (== property page), each property page will show a message about pending jobs to inform users that change propagation is in progress and
  - Due to lag in `JobQueueGroup::getSize`, the amount of jobs displayed is only an estimation not an exact number since deduplication may alter the numbers at any time
  - Due to random execution in the job scheduler numbers can vary when `ChangePropagationDispatchJob` hasn't been scheduled yet
- Added `PropertySpecificationReqExaminer::reqLock` to identify whether a property page requires a lock state
- Added `$smwgChangePropagationProtection` (default setting is `true`) as emergency setting in case some `ChangePropagationDispatchJob` went missing and the property page needs to be unlocked (should not normally not be necessary), a `purge` should be able the fetch a missing update by scheduling a new job
- `ChangePropagationEntityFinder` may return a very long list of matches as part of an iterator and to avoid having to use a DB connection or cache provider to temporary store the list of entities before it can be scheduled as `ChangePropagationUpdateJob`, `TempFile` is being used to created chunked temp files that contain a list of 1000 entities per batch and are dispatched using `ChangePropagationDispatchJob` in an iterative process to avoid an exhaustion of memory
- On `ChangePropagationEntityFinder::isTypePropagation`, a type change my create duplicate jobs at first and is expected to be handled by the deduplication in the job scheduler due to an extra match process on `PropertyTableInfoFetcher::getDefaultDataItemTables` to find rogue references caused by a change in table and ID assignments

This PR includes:
- [x] CI build passed


mwjames:
- Tests are missing
- Depends on the outcome of https://phabricator.wikimedia.org/T154425#3312440

mwjames:
> Depends on the outcome of https://phabricator.wikimedia.org/T154425#3312440

and apparently "... regression is due to a pre-existing bug being exposed with transaction rounds ...", if it weren't for this PR, I wouldn't have tested this and ... (I stop ranting here, especially on the fact that MW-core has no integration tests whatsoever).

kghbln:
Documented:
- [$smwgChangePropagationProtection](https://www.semantic-mediawiki.org/wiki/Help:$smwgChangePropagationProtection)
- TODO

DefaultSettings.php
```diff
@@ -697,15 +697,23 @@
...
-	# setting is not normally changed by users but by extensions that add new
-	# types that have their own additional declaration properties.
+	# pages that use P must be processed again.
+	#
+	# For example, if _PVAL (allowed values) for a property change, then pages
+	# must be processed again. This setting is not normally changed by users but
+	# by extensions that add new types that have their own additional declaration
+	# properties.
+	#
+	# @since 1.6
```
kghbln: @mwjames This setting was actually introduced with 1.5.0

mwjames: ... one this one will be renamed to reflect that it is actually a `smwgChangePropagationWatchlist`

mwjames:
> Setting this to false will disable edit protection of a property page while a property change propagation is being processed or is in a "frozen" state allowing to trigger the process manually again. 

Let me clarify some subtleties (important so it doesn't get _lost in translation_).

* When you change specification constraints and output characteristics (type, allows value etc.) related entities may need reevaluation (aka need to be parsed with the new spec.)
* To make sure we get all the subjects related to a property we need to run a query before (not after) the change but since such list can be huge (see #948), we need to defer the process which means:
  * Halt the update of the new spec and store it away temporarily (`_CHPRO`)
  * Schedule a `ChangePropagationDispatchJob ` and lock the page since we have halted the update with the property containing the old spec. until the primary update (executed as part of the job) has been completed
  * Once `ChangePropagationDispatchJob` is run it should:
    * Find all related subjects to the property before the update to the new spec.
    * After the query, get the new spec and update the property with that spec.
    * Now, that the property contains the new spec. schedule `ChangePropagationUpdateJob` for all subjects found (pages parsed will use the new spec.)

Locking the page is not because we want to overturn the "will of the user" but to keep the spec. changing while the `ChangePropagationDispatchJob` hasn't been executed yet and hereby avoiding competing updates on the primary specification of a property.

`$smwgChangePropagationProtection` was added to give an administrator some control to suspend the lock. For example, if someone cleans the job table and `ChangePropagationDispatchJob` wasn't finished at this point then a property will continue to be annotated with the `_CHPRO` property (internal property that stores the new spec.) therefore the spec. annotation remains and so does the locked.

As outlined above, we want to make some estimation about the work required and we rely on  MediaWiki to give us some figures about how many of the `ChangePropagationUpdateJob` remain. MediaWiki's `JobQueueGroup::getSize` does some internal caching so it should be no surprise that we might display some false positive (MW reports number X even though the table is empty). Why? because we don't want to count the table on our own (#2596 ,#2591) especially when it contains millions of rows.

The only reason for the existence of `ChangePropagationUpdateJob` is it to be able to get a number about the work required, the class itself calls `UpdateJob` to do the parsing and update process.

One thing to note, we get only one number (MediaWiki cannot distinguish on other criteria, and we cannot track start and end of a specific queue because jobs are executed randomly) about `ChangePropagationUpdateJob` work being required which means that if you change two properties simultaneously then the number shown is accumulated. Since we want to inform about the status, we show on all properties the same message "that work is ongoing and users should be wary about other changes" to avoid flooding the job queue.

I know this a bit technical but I think the process should be understood as to what, why, how and the possible consequences.

PS: After I answered my own inquiry [0], the process will most likely be expanded to category pages when changing `_SUBC` (Subcategory of) to ensure that hierarchies are updated when the
structure changes itself.

[0] https://phabricator.wikimedia.org/T176540

kghbln:
> Let me clarify some subtleties (important so it doesn't get lost in translation).

Thanks a lot! That's a bit to chew on, especially when it comes to integrate this into the wiki docu. ;)

PR #2496, opened 2017-06-03
Normalize message value arguments

mwjames:

- Avoid to break JSON encoding for arguments that contain things like `"Expression error: Unexpected < operator." is not recognized as a Boolean (true/false) value.`

image

http://wiki.teamliquid.net/dota2/index.php?title=Special:Ask&q=[[Has+processing+error+text%3A%3A%2B]]&po=%3FHas+improper+value+for|%3FHas+processing+error+text&p=class%3Dsortable-20wikitable-20smwtable-2Dstriped&eq=no&limit=100&bTitle=processingerrorlist&bMsg=smw-processingerrorlist-intro

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/Message.php
+++ b/src/Message.php
@@ -111,7 +111,16 @@ class Message {
 			$type = self::TEXT;
 		}
 
-		return json_encode( array_merge( (array)$type, (array)$message ) );
+		$message = (array)$message;
+		$encode = array();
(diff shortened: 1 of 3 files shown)


mwjames:
Back-ported to 2.5.x with fb3d604.

Issue #1912, opened 2016-10-09
Case insensitive LIKE matching for _wpg type entities (MySQL)


mwjames:
Some notes to a question [0] that came up on the mailing-list:

> As far as I know it was not possible in SMW 2.3, did it change or maybe
> there is some kind of workaround exists?

To support case insensitive matching for `_wpg` type entities one could make the following changes to the SMW 2.5 table settings (this is only tested in connection with MySQL, I'm not sure this works with  SQLite or Postgres).

`!~/~` queries will use the `smw_sortkey` field of the ID_TABLE (`smw_object_ids`) to match expressions, yet exact queries like `[[Has title::CAseInSensitiveSearch]]` will be matched against the `smw_title` field of the same table which remains case sensitive and therefore only returns exact matches to the `CAseInSensitiveSearch` value.

I'm not sure about the performance implications by changing from `VARBINARY(255)` to `VARCHAR(255) CHARSET utf8 COLLATE utf8_general_ci` and therefore hasn't been considered to be the default setting. Also, SQLite and Postgres would require similar table settings to ensure feature parity.
## Example

```
Has title -> Page type

[[Has title::CAseInSensitiveSearch]]

{{#ask: [[Has title::~CASEIN*]]
 |?Has title
}}

{{#ask: [[Has title::~casein*]]
 |?Has title
}}
```
## SQLStore

`SMWSQLStore3SetupHandlers`

```
-                   'smw_sortkey' => $dbtypes['t']  . ' NOT NULL',
+                   'smw_sortkey' => $dbtypes['s'] . ' NOT NULL',
```

`MySQLTableBuilder`

```
            case 'sort':
-           return 'VARCHAR(255)';
+           return 'VARCHAR(255) CHARSET utf8 COLLATE utf8_general_ci';
```
## SPARQLStore

`ValueDescriptionInterpreter`

```
-           return new FilterCondition( "$comparator( ?$joinVariable, \"$pattern\", \"s\")", array() );
+           return new FilterCondition( "$comparator( ?$joinVariable, \"$pattern\", \"i\")", array() );
        }

        if ( $dataItem instanceof DIUri ) {
            return new FilterCondition( "$comparator( str( ?$joinVariable ), \"$pattern\", \"i\")", array() );
        }
@@ -248,11 +248,11 @@ class ValueDescriptionInterpreter implements DescriptionInterpreter {
        $filterVariable = $this->compoundConditionBuilder->getNextVariable();

[...]

PR #2499, opened 2017-06-10
Add `smwgFieldTypeFeatures` (SMW_FIELDT_CHAR_NOCASE), refs 1912

mwjames:
This PR is made in reference to: #1912, http://wikimedia.7.x6.nabble.com/Non-case-sensitive-search-possible-tp5075393p5075402.html

- Adds a new `smwgFieldTypeFeatures` setting with flag `SMW_FIELDT_CHAR_NOCASE` that can alter the field type to make a field search case insensitive after the #1912 model
- The setting and flag `SMW_FIELDT_CHAR_NOCASE` are disabled by default as " ... no extensive analysis has been performed on how the switch from VARBINARY to a collated VARCHAR field type affects the search performance"
- Extends `smwgSparqlQFeatures` to support `SMW_SPARQL_QF_NOCASE` (`SPARQLStore`)
- Changing the setting requires to run `setupStore.php` or `update.php`

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -152,6 +152,8 @@ return array(
 	#   as in case of for example "Ennis, Ennis Hill, Ennis Jones, Ennis-Hill,
 	#   Ennis-London"
 	#
+	# - SMW_SPARQL_QF_NOCASE to support case insensitive pattern matches
+	#
 	# Please check with your repository provider whether SPARQL 1.1 is fully
 	# supported or not, and if not SMW_SPARQL_QF_NONE should be set.
 	#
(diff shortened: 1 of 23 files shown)


mwjames:
Okay, this took more effort than previously anticipated but this should give users the power to switch field types more easily and gain case insensitivity searches for LIKE/NLIKE when one wants to avoid the full-text index.

Limitations and restrictions are outlined in the `DefaulSettings.php` to the `smwgFieldTypeFeatures` setting and as a goody, I'll throw in the `SMW_SPARQL_QF_NOCASE` flag for `SPARQLStore` so that LIKE/NLIKE will be equally case insensitive.

mwjames:
The examples are part of an integration test:

- `[[Has page::noCASE]] [[Has text::noCASE]]`
- `[[Has page::NOcase]] [[Has text::NOcase]]`
- `[[Has page::NOoNcase]] [[Has text::NOoNcase]]`

and support (or not) the following queries:

- `[[Has page::~*noCa*]]`
- `[[Has text::~*NOcA*]]`
- `[[Has page::!~*noCa*]]`
- `[[Has text::!~*NOcA*]]`
- `[[Has text::nocase]]` it uses no ~ approximation, yet extends the match depth for an exact insensitive match (use case `[[Has keyword::SEARCH]]`, `[[Has keyword::search]]`, `[[Has keyword::sEaRch]]` )
- `[[Has text::!nocase]]`
- `[[Has page::!nocase]]` not supported on the NOCASE setting for _wpg types since a different field is used to match the DBKEY which is not made case insensitive
- `[[Has page::nocase]]` same as above


PS: If enabled then full-text index (#1481)only comes into effect for `~/!~` queries.

mwjames:
* https://sandbox.semantic-mediawiki.org/wiki/Issue/2499/Keyword
* https://sandbox.semantic-mediawiki.org/wiki/Issue/2499

Issue #1585, opened 2016-05-15
1.28.0-alpha / Change T135170 (PurgeAction) causes viewing disruption 


mwjames:
### Setup and configuration

- SMW version: all
- MW version: 1.28.0-alpha
- DB (MySQL etc.):

### Issue

During testing of \[0] (as it was small enough and relates to the whole TransactionProfiler topic, GET/POST updates) it emerged that it is visually impacting the viewing experience.

Now, every time one does a refresh, a  logged-in user is being redirected to a confirmation screen which causes a disruption in the reading flow. Prior to [0], a normal refresh (or purge) would have not caused a redirection for logged-in users.

![image](https://cloud.githubusercontent.com/assets/1245473/15240687/97c890a0-18eb-11e6-80bb-1b39258bd62a.png)

While the intention of [0] is understood, it is not clear whether the change is in favour of users that do not edit Wikipedia.

refs #1499

[0] https://phabricator.wikimedia.org/T135170, https://gerrit.wikimedia.org/r/#/c/288462/

mwjames:
After several of hours trying working with the patch, I had to remove it as turned out to be utterly inconvenient.

mwjames:
@kghbln @cicalese @hexmode I would appreciate some input on this from third-party wiki users.

If this is a none issue then I will close this one.

PS: [0] is considered not a real option as it would require develop time of volunteers to match the change.

PSS: [0] has mentioned a security aspect which is considered as well to be a really weak argument (we are talking about logged-in users, anonymous users are required to confirm even before [0]).

[0] https://phabricator.wikimedia.org/T135170#2310146

kghbln:
> Now, every time one does a refresh, a logged-in user is being redirected to a confirmation screen which causes a disruption in the reading flow.

That's what anons have to do. So if I understand the issue correctly in future also logged in user have to do this? If this is the case than it is not just an third-party issue but also a wmf issue since it is imperative that this does not happen. I believe this is not even worth discussing, it is just ridiculous to even consider this.

mwjames:
> That's what anons have to do. So if I understand the issue correctly in future also logged in user have to do this? 

Yes.

with GWicke clarifying:
> "The only difference is that all users will be shown a form with a submit button, while previously this was only done for anons / those without the specific rights.".

Feel free to comment on the phab issue.

>  I believe this is not even worth discussing, it is just ridiculous to even consider this.

Apparently it is.

[...]

PR #2502, opened 2017-06-10
Provide POST purge link, to avoid confirmation

kghbln:
This PR is made in reference to: #1585 

- Resource module to make a POST purge action possible avoiding confirmation

This PR includes:
- [x] CI build passed

### DIFF
--- a/res/Resources.php
+++ b/res/Resources.php
@@ -192,6 +192,11 @@ return array(
 		'scripts' => 'smw/util/ext.smw.util.purge.js',
 		'messages' => array(
 			'smw-purge-failed'
+		),
+		'position' => 'top',
+		'targets' => array(
+			'mobile',
+			'desktop'
 		)
 	),
 	
--- a/res/Resources.php
+++ b/res/Resources.php
@@ -186,6 +186,15 @@ return array(
 		'dependencies' => 'jquery.ui.autocomplete',
 		'targets' => array( 'mobile', 'desktop' )
 	),
+	
+	// Purge resources
+	'ext.smw.purge' => $moduleTemplate + array(
+		'scripts' => 'smw/util/ext.smw.util.purge.js',
+		'messages' => array(
+			'smw-purge-failed'
+		)
+	),
+	
 	// Special:Ask
 	'ext.smw.ask' => $moduleTemplate + array(
 		'scripts' => 'smw/special/ext.smw.special.ask.js',
--- a/res/smw/util/ext.smw.util.purge.js
+++ b/res/smw/util/ext.smw.util.purge.js
@@ -9,20 +9,28 @@
  * @ingroup SMW
  *
  * @licence GNU GPL v2+
- * @author samwilson
+ * @author samwilson, mwjames
  */
 
+/*global jQuery, mediaWiki, smw */
+/*jslint white: true */
 
-mw.loader.using( [ 'mediawiki.api', 'mediawiki.notify' ] ).then( function () {
+( function( $, mw ) {
 
-	$( "#ca-purge a" ).on( 'click', function ( e ) {
-		var postArgs = { action: 'purge', titles: mw.config.get( 'wgPageName' ) };
-		new mw.Api().post( postArgs ).then( function () {
-			location.reload();
-		}, function () {
-			mw.notify( mw.msg( 'smw-purge-failed' ), { type: 'error' } );
+	'use strict';
+
+	mw.loader.using( [ 'mediawiki.api', 'mediawiki.notify' ] ).then( function () {
+
+		$( "#ca-purge a" ).on( 'click', function ( e ) {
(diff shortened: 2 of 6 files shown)


src/MediaWiki/Hooks/SkinTemplateNavigation.php
```diff
@@ -45,6 +45,7 @@ public function __construct( SkinTemplate &$skinTemplate, array &$links ) {
 	public function process() {
 
 		if ( $this->skinTemplate->getUser()->isAllowed( 'purge' ) ) {
+			$skin->getOutput()->addModules( 'ext.smw.purge' );
```
kghbln: I suspect that this may be very wrong. I guess this is as far as imitation can get me. :(

kghbln:
@mwjames I hope that this is not to big of an disaster. I tried to make an educated guess regarding the placement of the script.

res/Resources.php
```diff
@@ -186,6 +186,15 @@
...
 		'targets' => array( 'mobile', 'desktop' )
 	),
+	
+	// Purge resources
+	'ext.smw.purge' => $moduleTemplate + array(
+		'scripts' => 'smw/util/ext.smw.util.purge.js',
+		'messages' => array(
+			'smw-purge-failed'
+		)
+	),
```
kghbln: I had the feeling that "targets" should be added too and was a bit puzzled why the Purge extension does not have it.

mwjames: > I had the feeling that "targets"

... can be added, more importantly is `'position' => 'top',` so that the module is loaded as soon as possible.

kghbln: > 'position' => 'top',

Yeah, forgot to mention this. the targets made me go off target.

src/MediaWiki/Hooks/SkinTemplateNavigation.php
mwjames: Here we come to a more delicate issue on the reference of `$skin->getOutput()`.

Lets take the `$skin` variable, PHP is expected to provide a reference to the variable with the name `$skin` but when you look in the code then you cannot find any reference or declaration with that name. We should select `$this->skinTemplate` because this is a variable available to us and doing `$this->skinTemplate->getOutput()...` should just do what we want.

kghbln: If I had a thorough look at the rest of the function `$this->skinTemplate->getOutput()...` was not far away from being figured out. :|

res/smw/util/ext.smw.util.purge.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @file
+ * @ingroup SMW
+ *
+ * @licence GNU GPL v2+
+ * @author samwilson
+ */
+
+
+mw.loader.using( [ 'mediawiki.api', 'mediawiki.notify' ] ).then( function () {
```
mwjames: A small form issue, wrapping the function avoids some nasty GLOBAL pollution.

```
/*global jQuery, mediaWiki, smw */
/*jslint white: true */

( function( $, mw ) {

	'use strict';

	mw.loader.using( [ 'mediawiki.api', 'mediawiki.notify' ] ).then( function () {

		$( "#ca-purge a" ).on( 'click', function ( e ) {
			var postArgs = { action: 'purge', titles: mw.config.get( 'wgPageName' ) };
			new mw.Api().post( postArgs ).then( function () {
				location.reload();
			}, function () {
				mw.notify( mw.msg( 'smw-purge-failed' ), { type: 'error' } );
			} );
			e.preventDefault();
		} );

	} );

}( jQuery, mediaWiki ) );
```

kghbln: > A small form issue, wrapping the function avoids some nasty GLOBAL pollution.

I trust you on this one! :)

mwjames:
... and because we have a `SkinTemplateNavigationTest` test, we need to adjust it as well:

```
@@ -45,15 +45,23 @@ class SkinTemplateNavigationTest extends \PHPUnit_Framework_TestCase {
 
 		$user->expects( $this->atLeastOnce() )
 			->method( 'isAllowed' )
 			->will( $this->returnValue( true ) );
 
+		$output = $this->getMockBuilder( '\OutputPage' )
+			->getMock();
+
 		$skinTemplate = $this->getMockBuilder( '\SkinTemplate' )
 			->getMock();
 
 		$skinTemplate->expects( $this->atLeastOnce() )
+			->method( 'getOutput' )
+			->will( $this->returnValue( $output ) );
+
+		$skinTemplate->expects( $this->atLeastOnce() )
```

kghbln:
> ... and because we have a SkinTemplateNavigationTest test, we need to adjust it as well:

Yay! :)

kghbln:
@mwjames  very cool. Backported to 2.5.x with a527bbeb719dda62ca5d0d92a53e4948b91bd2b6

Issue #2509, opened 2017-06-14
Support for user's timezone preference?


jamesmontalvo3:
### Setup and configuration

- SMW version: 2.5.2
- MW version: 1.27.2
- PHP version: 5.6.30
- DB (MySQL etc.): 5.5.52-MariaDB

### Issue

Uploading an image (using Upload Wizard, but I don't think that's relevant) results in this history:

![image](https://user-images.githubusercontent.com/716482/27142420-93a3cfc4-50f0-11e7-903a-b5fee20b4edf.png)

But going to Special:Browse shows this:

![image](https://user-images.githubusercontent.com/716482/27142452-b07e6172-50f0-11e7-8cc6-cea9ce79c6c7.png)

The file page "history" section shows the time in my preferences-defined timezone (US Central). SMW property shows it in UTC. This is confusing for people using the "Modification date" property.

kghbln:
Thanks for reporting. Just to clarify the situation concerning the setup: Is the server set to UTC or the wiki or both?

mwjames:
>  my preferences-defined timezone (US Central). SMW property shows it in UTC. This is confusing for people using the "Modification date" property.

On the matter of storage and retrieval, the information is retrieved
from [0,1] and has been so since the very beginning therefore it is
not a bug with Page::getTimestamp returning the "MW timestamp of last
article revision".

On the matter of display and output characteristics, local (user)
preferences are currently not supported (I think I noted that on
another ticket recently) and it would require some rework of how dates
are to be displayed if a user chooses a preference.

I have a PR (is a bit dated though) that addresses the display issue
to a larger extend but it is missing unit and integration tests, and
I'm not currently not invested to improve the PR but if someone wants
to write the missing part as required, I will gladly share it. (PS:
...and yes, tests are required otherwise it won't get merged!)

- Unit test on methods that have been altered
- Unit test on new classes
- Integration test to verify that LOCL#TO works

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/PropertyAnnotators/PredefinedPropertyAnnotator.php#L88-L89

[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/MediaWiki/PageInfoProvider.php#L56-L58

kghbln:
> I think I noted that on another ticket recently

I guess it is this issue: https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/205

[...]

PR #2515, opened 2017-06-17
Add support for date display with local preference, refs 1545, 2509

mwjames:
This PR is made in reference to: #1545, #2509

- User can enable the preference (`smw-prefs-general-options-time-correction`) to display local corrected dates in `Special:Browse` and the property page
- Corrected dates will be hinted with a `ᴸ` as in `05:41:20, 31 May 2017 ᴸ`
- Query output formatter can use `LOCL#TO` as option to adapt a local preference
- Use `ParserOptionsRegister` (MW 1.30+) to register an extra options key (`localTime`) with the `ParserCache` to ensure that users with different settings can switch the view without retaining a cached (== old settings) page view

This PR includes:
- [x] CI build passed


src/DataValues/Time/LocalTime.php
```diff
@@ -0,0 +1,87 @@
...
+	 * @param DateTime $dateTime
+	 * @param User|null $user
+	 *
+	 * @return DateTime
+	 */
+	public static function adjustTimeByPreference( DateTime $dateTime, User $user = null ) {
+
+		$tz = false;
+
+		if ( $user === null  ) {
```
gesinn-it: shouldn't this be `if ( $user )`? We did a test with the fix, and now it works instead of using the defaults.

mwjames:
> +		if ( $user === null  ) {
>
> shouldn't this be `if ( $user )`? We did a test with the fix, and now it
> works instead of using the defaults.

Probably what I wanted here was something a long the line of:

```
if ( $user === null  ) {
 $user = $GLOBALS['wgUser'];
}

$tz = $user instanceof User ? $user->getOption( 'timecorrection' ) : false;
```

It is used as fallback to ensure that the method can be tested with an
injected `User` object, if it is tested, which obviously is not.

Also instead of calling `LocalTime` directly, call
`Localizer::getInstance()->getLocalTime( $dateTime )` to avoid
`TimeValueFormatter` to depend on a MW object state (e.g wgUser) with:

 - Localizer::getLocalTime calling out to
`LocalTime::getLocalizedTime( $dateTime, $GLOBALS['wgUser'] );` and
- Move `LocalTime` to the \SMW\MediaWiki namespace as it is a pure
MediaWiki thing that relies entirely on MW settings and object state.

gesinn-it:
@mwjames , okay, we'll try to help with adding tests.

Is `https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/local-time/tests/phpunit/Unit/DataValues/Time` the right place to add the UTs for _LocalTime_?

mwjames:
> `https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/local-time/tests/phpunit/Unit/DataValues/Time`
> the right place to add the UTs for _LocalTime_?

If we neglect my last comment, then yes.

mwjames:
https://www.semantic-mediawiki.org/wiki/Local_time

mwjames:
In order to show local times on special pages (SMW), the option has to be enabled using the user preference.

image

Dates with a modified local time are marked with a `ᴸ` where `ᴸ` will reveal the ISO date on a mouseover event. 

image

PR #2518, opened 2017-06-17
Avoid "PHP Notice: A non well formed numeric value encountered" on Title::touched

mwjames:
This PR is made in reference to: https://phabricator.wikimedia.org/T167943

- Mostly relates to `postgres` where `Title::touched` would return `2017-06-15 08:36:55+00` instead of a timestamp

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with a851f8d.

PR #2482, opened 2017-05-27
Add TransactionalDeferredCallableUpdate

mwjames:
This PR is made in reference to: #2198, https://phabricator.wikimedia.org/T152583#3292654, https://phabricator.wikimedia.org/T165714 

This PR addresses or contains:

- Moves transaction specific handling from `DeferredCallableUpdate` to `TransactionalDeferredCallableUpdate`
- Add `TransactionalDeferredCallableUpdate::commitWithTransactionTicket` to reduce a possible transaction lag where a `TransactionTicket` is available
- `PageUpdater` is to make use of `TransactionalDeferredCallableUpdate` instead of trying to manage `DeferredUpdates` on its own

#2198 tried to cover most use case where a `SMW\DeferredCallableUpdate::doUpdate: Cannot flush snapshot because writes are pending` could appear but when a transaction is started as job but not through the commandLine (cronJob) then under certain circumstance the following can be observed where `MediaWiki::doPostOutputShutdown` triggers a pending job.

```
[smw] SMW\Utils\BufferedStatsdCollector::recordStats (as DeferredCallableUpdate a4663009fe01e8c985f3dda0fe6a173b)
[...]

PR #2524, opened 2017-06-24
CommitWithTransactionTicket only on DeferrableUpdate, refs 1796, 2482, 2523

mwjames:
This PR is made in reference to: #1796, #2482, #2523

- Due to https://phabricator.wikimedia.org/T168347 somewhere an exception is thrown when a `transactionTicket` is used while not being a separate transaction and causes the property update (which is not made deferred due to #1796) to be silently fail without a DB update or log entry
- Ensure that only marked `isDeferrableUpdate` can use a `transactionTicket`

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with 5e86d9c.

Issue #2529, opened 2017-06-24
InternalParseBeforeLinks.php: Argument 1 passed to ... must be of the type array, boolean given


zampaleu:
### Setup and configuration

- SMW version: 2.4.6
- MW version: 1.28
- PHP version: 7.0
- DB (MySQL etc.): MariaDB 10.029

### Issue

Produces a [stack trace](https://www.semantic-mediawiki.org/wiki/Help:Identifying_bugs) and outputs:

```
Checking predefined properties ...
   ... space for internal properties already allocated.
   ... writing properties ...
   ... done.

Database initialized completed.

Import of vocabulary.json ...
   ... creating MediaWiki:Smw import skos contents ...
[8b061c90dc5f9a32dbc09662] [no req]   TypeError from line 54 of /var/www/llg/solab/extensions/SemanticMediaWiki/src/MediaWiki/Hooks/InternalParseBeforeLinks.php: Argument 1 passed to SMW\MediaWiki\Hooks\InternalParseBeforeLinks::setEnabledSpecialPage() must be of the type array, boolean given, called in /var/www/llg/solab/extensions/SemanticMediaWiki/src/MediaWiki/Hooks/HookRegistry.php on line 243
```

### Backtrace

```
#0 /var/www/llg/solab/extensions/SemanticMediaWiki/src/MediaWiki/Hooks/HookRegistry.php(243): SMW\MediaWiki\Hooks\InternalParseBeforeLinks->setEnabledSpecialPage(boolean)
#1 /var/www/llg/solab/includes/Hooks.php(186): SMW\MediaWiki\Hooks\HookRegistry->SMW\MediaWiki\Hooks\{closure}(Parser, string, StripState)
#2 /var/www/llg/solab/includes/parser/Parser.php(1309): Hooks::run(string, array)
#3 /var/www/llg/solab/includes/parser/Parser.php(451): Parser->internalParse(string)
#4 /var/www/llg/solab/includes/content/WikitextContent.php(330): Parser->parse(string, Title, ParserOptions, boolean, boolean, NULL)
#5 /var/www/llg/solab/includes/content/AbstractContent.php(497): WikitextContent->fillParserOutput(Title, NULL, ParserOptions, boolean, ParserOutput)
#6 /var/www/llg/solab/includes/page/WikiPage.php(2071): AbstractContent->getParserOutput(Title, NULL, ParserOptions)
[...]

PR #2530, opened 2017-06-24
InternalParseBeforeLinks cast smwgEnabledSpecialPage setting late, refs 2529

mwjames:
This PR is made in reference to: #2529

- Cause for #2529 is a not anticipated value in `smwgEnabledSpecialPage` when set via `InternalParseBeforeLinks::setEnabledSpecialPage`
- Validate the value later instead of relying on an early type check which avoids punishment of a user during normal operations or as seen in #2529 during the setup

This PR includes:
- [x] CI build passed


mwjames:
Back-ported to 2.5.x with ad32a26.

PR #2533, opened 2017-06-30
Add newQueryEngineDatabaseConnectionProvider for temporary table queries on DB_REPLICA

jamesmontalvo3:
This PR removes the runtime exception found in #2532 when using a DB_MASTER + DB_SLAVE setup. At present I'm seeing more lag then I like, so I'll do some profiling before being ready for this to merge.

This PR includes:
- [ ] Tests (unit/integration)
- [ ] CI build passed

Issue #2532, opened 2017-06-29
Multi-database environment, RuntimeException on Error: 1146 Table 'wiki_eva.t3' doesn't exist


jamesmontalvo3:
### Setup and configuration

- SMW version: 2.5.2
- MW version: 1.27.2
- PHP version: 5.6.30
- DB (MySQL etc.): 5.5.52-MariaDB DB_MASTER + (1) DB_REPLICA

### Issue

The following error occurs when attempting to view a page with SMW content on it with a DB_MASTER and a single DB_SLAVE configured. The error does **NOT** occur in the following scenarios:

* If the DB_SLAVE is removed from `$wgDBservers` in LocalSettings.php (i.e. a normal single-DB configuration)
* If SMW+SRF are removed via composer and all other SMW-based extensions are removed from LocalSettings.php (i.e. remove SMW and things dependent upon it from the configuration)
* If SMW and all SMW-based extensions are **enabled**, and the second database server is in `$wgDBservers`, but all instances of `DB_SLAVE` in SMW's codebase are converted to `DB_MASTER` via `cd /path/to/smw && sudo find . -type f -exec sed -i 's/DB_SLAVE/DB_MASTER/g' {} +` (i.e. don't ever use DB_SLAVE)

Due to the tests above, I believe that there is at least one case of `DB_SLAVE` that should be `DB_MASTER`. On SMW v2.5.2 I count nine instances of `DB_SLAVE`, so it shouldn't be too difficult to narrow down which are the offenders. I will figure that out tomorrow, but wanted to get this issue in before I go to sleep.

Error message:
```
Exception encountered, of type "RuntimeException"
[WVPYjvrJNCT5cyqAw0wKuQAAAAk] /eva/index.php/STS-72 RuntimeException from line 226 of /opt/htdocs/mediawiki/extensions/SemanticMediaWiki/src/MediaWiki/Database.php: A database error has occurred. Did you forget to run maintenance/update.php after upgrading? See: https://www.mediawiki.org/wiki/Manual:Upgrading#Run_the_update_script
[...]

PR #2534, opened 2017-07-01
Add $smwgLocalConnectionConf, refs 2532, 2533

mwjames:
This PR is made in reference to: #2532, #2533

- `$smwgLocalConnectionConf` to allow for modifications on connection providers
- As noted "Changes to these settings should ONLY be made by trained professionals to avoid unexpected or unanticipated results ..."
- In certain environments a setting like `$smwgLocalConnectionConf['mw.db.queryengine'] = [ 'read' => DB_SLAVE, 'write' => DB_SLAVE ]`  can be more favourable 
- Setting `[ 'read' => DB_SLAVE, 'write' => DB_MASTER ]` will be logged as `[smw] [mw.db.queryengine] connection provider with {"read":-1,"write":-2}`

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -75,6 +75,34 @@ return array(
 	##
 
 	###
+	# Local connection configurations
+	#
+	# Allows to modify connection characteristics for providers that are used by
+	# Semantic MediaWiki.
+	#
+	# Changes to these settings should ONLY be made by trained professionals to
+	# avoid unexpected or unanticipated results when using connection handlers.
+	#
+	# Available DB index as provided by MediaWiki:
+	#
+	# - DB_SLAVE or DB_REPLICA (1.28+)
+	# - DB_MASTER
+	#
+	# @since 3.0
+	##
+	'smwgLocalConnectionConf' => array(
+		'mw.db' => array(
+			'read'  => DB_SLAVE,
+			'write' => DB_MASTER
+		),
+		'mw.db.queryengine' => array(
+			'read'  => DB_SLAVE,
+			'write' => DB_MASTER
+		)
+	),
+	##
+
+	###
 	# Configure SPARQL database connection for Semantic MediaWiki. This is used
 	# when SPARQL-based features are enabled, e.g. when using SMWSparqlStore as
 	# the $smwgDefaultStore.
--- a/includes/Settings.php
+++ b/includes/Settings.php
@@ -53,6 +53,7 @@ class Settings extends Options {
 			'smwgSemanticsEnabled' => $GLOBALS['smwgSemanticsEnabled'],
 			'smwgEnabledCompatibilityMode' => $GLOBALS['smwgEnabledCompatibilityMode'],
 			'smwgDefaultStore' => $GLOBALS['smwgDefaultStore'],
+			'smwgLocalConnectionConf' => $GLOBALS['smwgLocalConnectionConf'],
 			'smwgSparqlDatabaseConnector' => $GLOBALS['smwgSparqlDatabaseConnector'],
 			'smwgSparqlDatabase' => $GLOBALS['smwgSparqlDatabase'],
 			'smwgSparqlQueryEndpoint' => $GLOBALS['smwgSparqlQueryEndpoint'],
--- a/includes/Setup.php
+++ b/includes/Setup.php
@@ -105,11 +105,11 @@ final class Setup {
 
 		$connectionManager->registerConnectionProvider(
 			'mw.db',
-			$mwCollaboratorFactory->newMediaWikiDatabaseConnectionProvider()
+			$mwCollaboratorFactory->newMediaWikiDatabaseConnectionProvider( 'mw.db' )
 		);
 
 		// Connection can be used to redirect queries to another DB cluster
-		$queryengineConnectionProvider = $mwCollaboratorFactory->newMediaWikiDatabaseConnectionProvider();
+		$queryengineConnectionProvider = $mwCollaboratorFactory->newMediaWikiDatabaseConnectionProvider( 'mw.db.queryengine' );
 		$queryengineConnectionProvider->resetTransactionProfiler();
 
 		$connectionManager->registerConnectionProvider(
--- a/src/MediaWiki/DatabaseConnectionProvider.php
+++ b/src/MediaWiki/DatabaseConnectionProvider.php
@@ -3,6 +3,9 @@
 namespace SMW\MediaWiki;
 
 use SMW\DBConnectionProvider;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerAwareInterface;
+use RuntimeException;
 
 /**
  * @license GNU GPL v2+
@@ -10,7 +13,12 @@ use SMW\DBConnectionProvider;
  *
  * @author mwjames
  */
-class DatabaseConnectionProvider implements DBConnectionProvider {
+class DatabaseConnectionProvider implements DBConnectionProvider, LoggerAwareInterface {
+
+	/**
+	 * @var string
+	 */
+	private $provider;
 
 	/**
 	 * @var Database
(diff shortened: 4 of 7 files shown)


mwjames:
@jamesmontalvo3 FYI

This PR achieves the same thing as #2533 but instead of creating an extra class we use a configuration to modify a connection provider.

jamesmontalvo3:
I updated `composer.local.json` to pick up this change to SMW and allow an acceptable SRF version:

```
"mediawiki/semantic-media-wiki": "dev-master#005783f1f4cad711b035d6d83052590500c77641",
"mediawiki/semantic-result-formats": "*",
```

Then after running `update.php` to pick up SMW 3.0 changes I added the following to `LocalSetting.php`:

```
$smwgLocalConnectionConf['mw.db.queryengine'] = [ 'read' => DB_SLAVE, 'write' => DB_SLAVE ];
```

Trying to load pages I'm getting what seems to be the same error:

```
Exception encountered, of type "RuntimeException"
[WVgt1p3iuBJVqTtTlDZkLwAAAAI] /eva/index.php/STS-72 RuntimeException from line 228 of /opt/htdocs/mediawiki/extensions/SemanticMediaWiki/src/MediaWiki/Database.php: A database error has occurred. Did you forget to run maintenance/update.php after upgrading? See: https://www.mediawiki.org/wiki/Manual:Upgrading#Run_the_update_script
Query: SELECT DISTINCT t0.smw_id AS id,t0.smw_title AS t,t0.smw_namespace AS ns,t0.smw_iw AS iw,t0.smw_subobject AS so,t0.smw_sortkey AS sortkey,t8.o_sortkey FROM `smw_object_ids` AS t0 INNER JOIN `smw_fpt_inst` AS t2 ON t0.smw_id=t2.s_id INNER JOIN `t3` AS t3 ON t2.o_id=t3.id INNER JOIN `smw_di_wikipage` AS t4 ON t2.s_id=t4.s_id INNER JOIN `smw_di_time` AS t8 ON t0.smw_id=t8.s_id WHERE ((t4.p_id='1908')) AND (t8.p_id='2116') AND t0.smw_iw!=':smw' AND t0.smw_iw!=':smw-delete' AND t0.smw_iw!=':smw-redi' ORDER BY t8.o_sortkey DESC LIMIT 6 
Function: SMW\SQLStore\QueryEngine\QueryEngine::getInstanceQueryResult
Error: 1146 Table 'wiki_eva.t3' doesn't exist (<redacted db_slave ip address>)
```

Special:Version says the following, though, so maybe Composer isn't doing what I think it should be doing.

`3.0.0-alpha (f81f11c) 16:56, 30 June 2017`

mwjames:
>"mediawiki/semantic-media-wiki": "dev-master#005783f1f4cad711b035d6d83052590500c77641",
> "mediawiki/semantic-result-formats": "*",

:+1: 

> $smwgLocalConnectionConf['mw.db.queryengine'] = [ 'read' => DB_SLAVE, 'write' => DB_SLAVE ];

Your standard log should contain something like:

```
[smw] [mw.db] connection provider with {"read":-1,"write":-2}

... 

[smw] [mw.db.queryengine] connection provider with {"read":-1,"write":-1}`
```

jamesmontalvo3:
Was getting errors due to a bad Composer version specified. This loads the correct SMW:

```diff
-    dev-master#005783f1f4cad711b035d6d83052590500c77641
+    dev-conn#005783f1f4cad711b035d6d83052590500c77641
```

With this I am no longer getting the database errors in a multi-database-server setup.

Also, I mentioned in #2533 that I was experiencing lag. This was due to the extreme debug measures I had enabled. With them disabled I am getting better-than-single-DB performance.

@mwjames Do you need me to test anything in particular to be able to merge this? Also, would it be possible to cherry-pick this onto the 2.5.x branch?

jamesmontalvo3:
Also, I'm getting the following javascript alert on some pages:

> blockUI requires jQuery v1.3 or later!  You are using v1.11.3

According to [1], this appears to be because when I specified `*` as the Composer version of Semantic Result Formats it installed v1.9. I initially had `2.4.x` specified, but that threw Composer errors when I had to upgraded to SMW v3.0 for this change. I thought by specifying `*` Composer would give me that latest version of SRF that worked with the rest of my requirements (e.g. SMW), but evidently not. This isn't related to this issue exactly, but since I mentioned the `*` above I thought I'd mention here.

[1] https://www.semantic-mediawiki.org/wiki/Thread:Talk:Semantic_Result_Formats/blockUI_requires_jQuery_v1.3_or_later!_You_are_using_v1.11.1

mwjames:
> Also, would it be possible to cherry-pick this onto the 2.5.x branch?

Back-ported to 2.5.x with d7077b8.

kghbln:
[Now documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgLocalConnectionConf&type=revision&diff=52202&oldid=52161)

jamesmontalvo3:
@kghbln @mwjames thanks!

TK-999:
@jamesmontalvo3 Note that using `DB_REPLICA` / `DB_SLAVE` for temporary table operations in MediaWiki world may have one caveat - if the global `wgReadOnly` flag is set for some reason (maintenance/promoting new master etc.), then MediaWiki will [prevent](https://github.com/wikimedia/mediawiki/blob/master/includes/libs/rdbms/database/Database.php#L901) temporary table operations even if the RDBMS actually supports them in read-only mode. This behavior cannot be circumvented without patching core.

hexmode:
> To be honest, I'm a bit confused about how temporary tables interact with replication. My assumption was that TEMP tables are not replicated at all.

They are not replicated. They are per-connection.

You can see this by opening up two connections from the command line to the same DB:

* Window 1
```
$ mysql -u$DB_USER -p$DB_PASS -h10.141.0.55 $DB_NAME
mysql> CREATE TEMPORARY TABLE foobar (id int) ENGINE=MEMORY;
Query OK, 0 rows affected (0.00 sec)
mysql> show create table foobar
    -> ;
+--------+------------------------------------------------------------------------------------------------------+
| Table  | Create Table                                                                                         |
+--------+------------------------------------------------------------------------------------------------------+
| foobar | CREATE TEMPORARY TABLE `foobar` (
  `id` int(11) DEFAULT NULL
) ENGINE=MEMORY DEFAULT CHARSET=latin1 |
+--------+------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql> 
```

* Window 2 on the same client machine
```
$ mysql -u$DB_USER -p$DB_PASS -h10.141.0.55 $DB_NAME
mysql> show create table foobar;
ERROR 1146 (42S02): Table '$DB_NAME.foobar' doesn't exist
mysql> 
```

I ran into the same problem that @darenwelsh experienced this week and that led to this patch.

The current solution relies on MediaWiki giving you *the same* connection to the DB that you used earlier in the session.  This will undoubtedly lead to problems later when MW decides to spread queries out over multiple replicas during the same session.

@brightbyte's idea of providing a `temp` key to `$smwgLocalConnectionConf` makes sense, since SMW could ensure that any connection request made to the `temp` stack used the same connection as was used earlier in the session.

mwjames:
>    To be honest, I'm a bit confused about how temporary tables interact with replication. My assumption was that TEMP tables are not replicated at all.

> They are not replicated. They are per-connection.

Those statements are not correct as it depends on the ` binlog_format` as noted in [0]. I introduced `$smwgLocalConnectionConf` so users can accommodate to their environment as long as users understand the implications.

> @brightbyte's idea of providing a temp key to $smwgLocalConnectionConf makes sense, since SMW could ensure that any connection request made to the temp stack used the same connection as was used earlier in the session.

I'm not sure what you mean by "temp key" but SMW is getting its connection details from `LoadBalancer::getConnectionRef` and isn't involve in a specific logic as to how those connections are created, closed, or reused. If creating temp tables isn't allowed for a slave or the `SELECT` (for the same request) cannot (or is unable to) access previously created temp tables (those holding onto the same connection instance) then this is an issue with how the `LoadBalancer` treats those requests involving temp tables and there is nothing we can do, or be concerned about on the SMW side. 

> is rather confusing and scary, since it implies that data is written to the replica - which is a very bad idea. But if I understand correctly, the intention is to only use that for temporary tables. But why, then, is this 

The document states "... ONLY be made by trained professionals to avoid unexpected or unanticipated results ...". Again, `$smwgLocalConnectionConf` allows to reroute those query requests to either the salve/master for read and write to prevent issues in case temp tables aren't replicated but that is only provided so users can cope with the issue of MW being unable to handle temp tables correctly in a replicated environment.

PS: To lower expectations, I'm unable to investigate issues related to multi-database setups (due to time constraints) and I would only consider providing support if someone creates a docker instance that has a multi-db setup running (including MW and description) so that any installation effort is reduced to downloading the docker container and executing `docker run`.

[0] https://dev.mysql.com/doc/refman/5.7/en/replication-features-temptables.html

PR #2536, opened 2017-07-08
Add SMW_FIELDT_CHAR_LONG, refs 2499

mwjames:
This PR is made in reference to: #2499

- `$smwgFieldTypeFeatures` setting add flag `SMW_FIELDT_CHAR_LONG` to allow for a 300 char long field (blob and uri datatype) to store indexable (== querable) text content without relying on a full-text index
- 300 char (-32 char) which is longer than the default 72 char (-32 char) and short enough to be indexable without further technical modifications
- `SMW_FIELDT_CHAR_NOCASE | SMW_FIELDT_CHAR_LONG` can be combined to make the long field type also case insensitive (limitation as outlined in #2499 applies)
- No analysis has been performed on how performance is impacted by the length extension in regards to index and update performance
- Requires to run `rebuildData.php` to adjust the field content length

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1599,6 +1599,22 @@ return array(
 	#
 	# [0] https://techblog.dorogin.com/case-insensitive-like-in-sqlite-504f594dcdc3
 	#
+	# SMW_FIELDT_CHAR_LONG - Extends the size to 300 chars for text pattern
+	# match (DIBlob and DIUri) fields.
+	#
+	# By default, those fields are limited to 72 chars that limits search depth
+	# in exchange for index size and performance. Extending fields to 300 allows
+	# to run LIKE/NLIKE matching on a larger text body without relying on a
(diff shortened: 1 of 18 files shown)


mwjames:
@kghbln FYI

>  ... to allow for a 300 char long field (blob and uri datatype) to store indexable 

For record as to why 300, this due to a restriction on indexable field length where any higher value would most likely result in a `#1071 - Specified key was too long; max key length is 1000 bytes` error [0].

[0] https://stackoverflow.com/questions/3489041/mysqlerror-specified-key-was-too-long-max-key-length-is-1000-bytes

kghbln:
Now serving on [sandbox.](https://sandbox.semantic-mediawiki.org/w/index.php?title=SMW_Sandbox%3ASetup_information&type=revision&diff=6896&oldid=6682)

PR #2551, opened 2017-07-20
Allow adding globals in TestEnvironment; allow extending permitted settings in JsonScripts

s7eph4n:
In sum this patch will allow to define additional allowed globals for JsonScript tests in sub-classes of JsonTestCaseScriptRunnerTest.
It will also allow adding globals that are not already defined in the TestEnvironment.

Use case: SRF's "filtered" format will support $srfgMapProvider, that is unset by default. See https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/272

* add getPermittedSettings() to JsonTestCaseScriptRunnerTest
* allow undefined global settings in JsonTestCaseScriptRunnerTest::prepareTest
* track new globals in a dedicated array in TestEnvironment and unset them on tearDown
* add delete($key) method to Options

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -35,6 +35,15 @@ class Options {
 	}
 
 	/**
+	 * @since 3.0
+	 *
+	 * @param string $key
+	 */
+	public function delete( $key ) {
+		unset( $this->options[ $key ] );
+	}
+
+	/**
 	 * @since 2.3
 	 *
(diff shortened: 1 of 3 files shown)


mwjames:
I'll most likely provide a a follow-up, yet merging as-is so that you can continue on the SRF part.

Issue #2489, opened 2017-05-31
Search of part of string is not working with FulltextSearch enabled


sommer-gei:
### Setup and configuration

Setup at https://sandbox.semantic-mediawiki.org/wiki/Special:Version

### Issue

Hi!

In my local environment I enabled `$smwgEnabledFulltextSearch = true;` and my not case-sensitive search worked like a charme! :-)

But then I realised that the **search for a part of property string isn’t working** anymore. − Is that a known problem?

Thanks in advance!

----
**Example** (for my local MW installation and the SMW [sandbox](https://sandbox.semantic-mediawiki.org)):

Search for `[[Foaf:name::~*john*]]` is working well: https://sandbox.semantic-mediawiki.org/w/index.php?title=Special:Ask&x=-5B-5BFoaf%3Aname%3A%3A~%2Ajohn%2A-5D-5D

Search for `[[Foaf:name::~*ohn*]]` is **not** working well: https://sandbox.semantic-mediawiki.org/w/index.php?title=Special:Ask&x=-5B-5BFoaf%3Aname%3A%3A~%2Aohn%2A-5D-5D

(ref. for this issue: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1481#issuecomment-305208840)

mwjames:
Hi,

> Search for `[[Foaf:name::~*john*]]` is working well:
> https://sandbox.semantic-mediawiki.org/w/index.php?title=Special:Ask&x=-5B-5BFoaf%3Aname%3A%3A~%2Ajohn%2A-5D-5D
>
> Search for `[[Foaf:name::~*ohn*]]` is **not** working well:
> https://sandbox.semantic-mediawiki.org/w/index.php?title=Special:Ask&x=-5B-5BFoaf%3Aname%3A%3A~%2Aohn%2A-5D-5D

I'm pretty sure this relates to the min token setting which we check
in [0, 1] and somewhere on the original issue I left some comments
about it.

`>=` (in [0, 1]) means that min is used which by default is 3 as far
as I remember but I cannot recall what the MySQL/MariaDB docu has to
say about it. There is also a difference between InnoDB and Myisam.

If you switch to lower token size (~*oh*) as min then the system
switches on using the "old" LIKE comparison which of course is not a
full-text index.

[...]

PR #2558, opened 2017-07-23
Add like:/nlike: comparator, refs 1481, 2489, 2536

mwjames:
This PR is made in reference to: #1481, #2489, #2536

- `~/!~` uses a preferred matching approach depending on the environment (available and for an enabled full-text index with  a "MATCH ..." operation)
- `like:/nlike:` is introduced to always use the standard `LIKE` / `NOT LIKE` SQL syntax for a pattern match

This PR includes:
- [x] CI build passed


mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2499

PR #2568, opened 2017-07-26
Fix #2567: Start template parameters on new line

### LINKED ISSUE #2567: Template format: Template call not parsed
### Setup and configuration

- SMW version: 3.0.0-alpha (c99ff74)
- MW version: 1.29
- PHP version: 5.6.30
- DB (MySQL etc.): MariaDB 10.0.31

### Issue

Steps to reproduce the observation (recommendation is to use the [sandbox](https://sandbox.semantic-mediawiki.org)):
(See https://sandbox.semantic-mediawiki.org/wiki/Issue/2567)
1. Have a property of type Text, e.g. `Has text`
2. Have a page assigning a multi-line text to that property, where the last line of the text starts with a `=`
3. Have a query where
   * the query result contains the page
   * the Text type property is one of the printouts, e.g. `|?Has text`
   * `|format=template`
   * `|named args=yes`
   * the template is any existing template, e.g. `|template=Test`

**Expected:** Some text according to the template
**Observed:** The template call produced by the `#ask` is put on the page verbatim without parsing.
**Probable cause:** The MW parser chokes on the `=` appearing as a first character in a line and again later on the same line where it separates the template parameter name from its value.

s7eph4n:
This PR is made in reference to: #2567 

- Start each parameter of the template calls created by the `template` format  on a new line

This PR includes:
- [x] CI build passed

Fixes #2567

### DIFF
--- a/src/MediaWiki/Renderer/WikitextTemplateRenderer.php
+++ b/src/MediaWiki/Renderer/WikitextTemplateRenderer.php
@@ -40,7 +40,7 @@ class WikitextTemplateRenderer {
 		$this->template .= '{{'. $templateName;
 
 		foreach ( $this->fields as $key => $value ) {
-			$this->template .= "|$key=$value";
+			$this->template .= "\n|$key=$value";
 		}
 
 		$this->template .= '}}';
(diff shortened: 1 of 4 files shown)


mwjames:
Nice.

This should also solve `{{#set: SomeProperty=SomeValue |template=ValueContainsEqualSign }}`.

mwjames:
Back-ported to 2.5.x with c8d6718.

Issue #2549, opened 2017-07-13
2.5.1 update logic and the use of self-referenced queries as input for value annotations


gesinn-it:
### Setup and configuration

- SMW version: 2.5.1
- MW version: 1.27.x
- PHP version: 7
- DB (MySQL etc.): MySQL

### Issue

I create pages with Page Forms writing a template. As part of the template, I set a property depending on the result of an `#ask` query:

```
{{#vardefine:partOfSite|{{#ask:[[-Part Of Building::{{FULLPAGENAME}}]]|?Part Of Site=|mainlabel=-|link=none|format=list}} }} 
{{#set:Part Of Site={{#var:partOfSite}}}}
```

I'm aware of [0] which is circumnavigated using [1]. But this should not be relevant here.

Starting with 2.5.1, property `Part Of Site` is not created anymore. I tried with downgrading to `2.5.0` and `2.4.5`. Both versions work as expected.

I tried `$smwgEnabledQueryDependencyLinksStore = true;` because I thought that this is related. No change.

Any idea, what change could have caused this issue?

[0] https://www.semantic-mediawiki.org/wiki/Help:Self-Referencing
[1] https://www.mediawiki.org/wiki/Extension:SemanticDependencyUpdater

kghbln:
Just to make sure: This also does not work in 2.5.3?

gesinn-it:
@kghbln not working with 2.5.3

mwjames:
> @kghbln not working with 2.5.3

Standard unit and integration tests pass for #set, the involvement of
#var .. or any other extension is not tested and also not part of SMW
core, and unless there is a stack trace or bisect there is not much
assistance we can provide.

PS: I have on several occasions here (and on wiki) explained that
relying on self-references values might not result in the expected
result because it depends on when and how values are available by the
time #ask is executed. Since we rely on deferred updates that process
can even more ambiguous due to different execution paths run by
MediaWiki.

PSS: There are some technical "tricks" you might be able to use but I
have no plans to look at them more closely in near future. For
example, you could track a #ask that contains a self-reference and
schedule an update job (a simple null edit isn't enough).

gesinn-it:
... before using `#var` I was directly using the result of `#ask` in `#set`. Same effect, not working since 2.5.1. For me, this pattern has been working very stable until 2.5.1 and in many setups. I'm wondering what has caused this breaking now...

mwjames:
[...]

Issue #2375, opened 2017-04-01
Data process, storage, and query update


mwjames:
Some notes about data processing and storage in connection with MediaWiki in general and Semantic MediaWiki specifically.

Simple text like `Foo ... Bar` is parsed, processed, and stored as blob in a regular MediaWiki database table with parsing being responsible for identifying wikitext specific syntax and expressions to transform `== Title ==` into a valid HTML representation.

Besides the "normal" text, MediaWiki provides so called parser functions and tags (`{{#someFunction: ...}}`, `<someTag/>` etc.) that extend the parsing processing with an extension mechanism for additional processing of raw text.

Due to the nature of this extension mechanism, parser functions are processed first and encapsulated in order to be separated from the embedding "normal" text which means that functions like `#ask`, `#set`, and `#subobject` are processed before any other text and implicitly explain why queries can __only__ return what is already stored.

Query results (via `#ask`) are queried before any other parsing operation and value changes (includes new or deleted values) can only be represented as part of a query after it has been stored which is after the parsing of a page has been finalized. The outcome of that cascading parsing process is that in case of self-referencing queries (with value annotations or subjects being part of the query condition or its results set) are required to be refreshed before they can display a change (edit → store → refresh).

```
. Create (or edit) page
	|- add annotation ([[Foo::Bar]])
	|- add annotation ({{#set: ...}})
	|- add query ({{#ask: [[Foo::Bar]] ... }})
	|
.. Processing page (parsing and text manipulation)
	|	|- parse #ask (== answer the query)
	|	|- parse #set (== Text -> DataValue)
	|	|- parse text
	|		|- InTextAnnotationParser transform Foo::Bar == Text -> DataValue
	|
... Update (storage)
		|- MediaWiki wikitext storage
		|- Semantic MediaWiki data storage (SQLStore, SPARQLStore)

```

@kghbln For reference and in case something is unclear then this issue should be used to address those questions.

PR #2572, opened 2017-07-29
Add PostProcHandler and `queryref` update task, refs 1696

### LINKED ISSUE #1696: Self-reference queries and updates
This issue contains a simple reference implementation for users interested in having self-reference queries updated upon a save request.

As described in [0, 1], setting values based on self-reference queries will have its limits because one cannot query something that hasn't been stored yet. In MediaWiki, any parser function that is executed before the storage requires a null-edit to fetch results from the DB when a condition is pointing to itself after the save event has been finished. 

Test case:

```
== Values set manually ==
{{#set:
 |Some other property=Foo
}}

[[Some property::Bar]]

== Values set by query ==

* [[Another property set by a query:: {{#show: {{FULLPAGENAME}} |?Some property |link=none|mainlabel=-}}]]
* [[Another property set by a query:: {{#show: {{FULLPAGENAME}} |?Some other property |link=none|mainlabel=-}}]]

== Validation ==
{{#ask: [[{{FULLPAGENAME}}]]
 |?Some other property
 |?Some property
 |?Another property set by a query
}}
```

To initiate a null-edit in case of a self-reference, some additional steps are necessary to cater for an auto update. Nested self-references will not be updated as those would need an "update on an update" due to the same principles.

`$GLOBALS['smwgEnabledQueryDependencyLinksStore'] = true;` plus add a hook to the `AskParserFunction`.

``` php
@@ -184,10 +184,12 @@ class AskParserFunction {
        // self-reference
        if ( $this->circularReferenceGuard->isCircularByRecursionFor( $queryHash ) ) {
            return '';
        }

+       \Hooks::run( 'SMW::Ask::BeforeResultFromQueryComplete', array( $contextPage, $this->query, $this->parserData->getOutput() ) );
```

Together with the `SMW::Ask::BeforeResultFromQueryComplete` and `PageContentSaveComplete` [2] hook.

``` php
/*
 * Occurs after the save page request has been processed
 *
 * @see https://www.mediawiki.org/wiki/Manual:Hooks/PageContentSaveComplete
 */
\Hooks::register( 'PageContentSaveComplete', function ( $article, $user, $content, $summary, $isMinor, $isWatch, $section, $flags, $revision, $status, $baseRevId ) {

    if ( $revision === null ) {
        return true;
    }

    $editInfoProvider = new \SMW\MediaWiki\EditInfoProvider( $article, $revision, $user );

    // If a self-reference was detected then post a null-edit into the deferred update
    // queue to allow for a processing after others have finished
    $deferredCallableUpdate = ApplicationFactory::getInstance()->newDeferredCallableUpdate( function() use( $editInfoProvider, $article ) {

        wfDebugLog( 'smw', 'before fetchEditInfo' );
        $output = $editInfoProvider->fetchEditInfo()->getOutput();

        if ( $output->getProperty( \SMW\Query\SelfRefMarker::ID ) ) {
            wfDebugLog( 'smw', 'doPurge' );
            $article->doPurge();
            $article->getTitle()->invalidateCache();
        }

    } );

    $deferredCallableUpdate->pushToDeferredUpdateList();

    return true;
};

\Hooks::register( 'SMW::Ask::BeforeResultFromQueryComplete', function ( $contextPage, $query, $parserOutput ) {

    $selfRefMarker = new \SMW\Query\SelfRefMarker( $parserOutput );
    $selfRefMarker->findAndMark( $query );

    return true;
};
```

``` php
<?php

namespace SMW\Query;

use SMWQuery as Query;
use ParserOutput;
use SMW\Query\Language\ValueDescription;
use SMW\Query\Language\Conjunction;
use SMW\Query\Language\Disjunction;

/**
 * @license GNU GPL v2+
 *
 * @author mwjames
 */
class SelfRefMarker {

    const ID = 'smw-query-selfref';

    /**
     * @var ParserOutput
     */
    private $parserOutput;

    /**
     * @param ParserOutput $parserOutput
     */
    public function __construct( ParserOutput $parserOutput ) {
        $this->parserOutput = $parserOutput;
    }

    /**
     * Mark the ParserOutput to signal post-processing that this entity contains
     * a query with a self-reference and need post-save attention.
     *
     * @param  Query $query
     */
    public function findAndMark( Query $query ) {
        if ( $this->hasSelfReference( $query->getContextPage(), $query->getDescription() ) ) {
            $this->parserOutput->setProperty( self::ID, true );
        }
    }

    private function hasSelfReference( $contextPage, $description ) {

        if ( $description instanceOf ValueDescription ) {
            if ( $contextPage->equals( $description->getDataItem() ) ) {
                return true;
            }
        }

        if ( $description instanceof Conjunction || $description instanceof Disjunction ) {
            foreach ( $description as $desc ) {
                if ( $this->hasSelfReference( $contextPage, $desc ) ) {
                    return true;
                };
            }
        }

        return false;
    }

}
```

[0] https://www.semantic-mediawiki.org/wiki/Help:Self-Referencing
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1478#issuecomment-202170690
[3] https://github.com/wikimedia/mediawiki/blob/8dc191edce050117223e4e00b6a9b6cbc7a6c870/includes/page/WikiPage.php#L1912-L1962, NewRevisionFromEditComplete vs PageContentSaveComplete with PageContentSaveComplete being executed `onTransactionIdle` mode

mwjames:
This PR is made in reference to: #1696

This implements a different approach from what has been outlined in #1696 due to some limitations on the part of:

One cannot reliably detect (while parsing a #ask from within the `Parser` instance) whether a #ask query is part of an embedded parsing process (i.e. part of a [[ ... ]] annotation or #set annotation) and trying to only use an entity name `{{FULLPAGENAME}}` as decision parameter is insufficient. It could cause a query like `{{#show: {{FULLPAGENAME}} |?Some property |link=none|mainlabel=-}}` to always being identified as to contain a self-reference even though it is not embedded in an annotation, yet it would be marked as to require a post processing update.

To clearly identify which `#ask`/`#show` is used as input for an annotation process, it is proposed to mark those queries with a `@annotation` parameter as in `[[Another property:: {{#show: {{FULLPAGENAME}} |?Some property |link=none|mainlabel=-|@annotation}}]]` to distinguish them from a simple `{{#show: {{FULLPAGENAME}} |?Some property |link=none|mainlabel=-}}` query.

### Technical notes 

* `AskParserFunction` will use `PostProcHandler::addQueryRef` on queries marked with a `@annotation` parameter and store references with `ParserOutput::setExtensionData` in order for succeeding processes to access that information without relying on a `Cache` or `DB` connection
* `OutputPageParserOutput` hook will inspect the `ParserOutput` and in case of a matching result, adds a `HTML` element at the top of a page together with a resource module (`ext.smw.postproc`)
* Once the page has been loaded client-side, the earlier invoked resource module will initiate an API request using the `smwtask` module to run the required `UpdateJob` and when finished forces a reload of the page to display the newly computed and annotated values
* This approach requires that the API is available and writable, see also:
  * https://www.mediawiki.org/wiki/Manual:$wgEnableAPI 
  * https://www.mediawiki.org/wiki/Manual:$wgEnableWriteAPI
  * https://www.mediawiki.org/wiki/API:Restricting_API_usage

This PR includes:
- [x] CI build passed

Fixes #1696

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -558,5 +558,6 @@
 	"smw-format-datatable-sortascending": ": activate to sort column ascending",
 	"smw-format-datatable-sortdescending": ": activate to sort column descending",
 	"smw-category-invalid-redirect-target": "Category \"$1\" contains an invalid redirect target to a non-category namespace.",
-	"smw-parser-function-expensive-execution-limit": "The parser function has reached the limit for expensive executions (see [https://www.semantic-mediawiki.org/wiki/Help:$smwgQExpensiveExecutionLimit $smwgQExpensiveExecutionLimit])."
+	"smw-parser-function-expensive-execution-limit": "The parser function has reached the limit for expensive executions (see [https://www.semantic-mediawiki.org/wiki/Help:$smwgQExpensiveExecutionLimit $smwgQExpensiveExecutionLimit]).",
+	"smw-postproc-queryref": "The page was marked as to be refreshed due to some required post processing."
 }
--- a/includes/Setup.php
+++ b/includes/Setup.php
@@ -190,6 +190,7 @@ final class Setup {
 		}
 
 		$this->globalVars['wgAPIModules']['smwinfo'] = '\SMW\MediaWiki\Api\Info';
+		$this->globalVars['wgAPIModules']['smwtask'] = '\SMW\MediaWiki\Api\Task';
 		$this->globalVars['wgAPIModules']['ask']     = '\SMW\MediaWiki\Api\Ask';
 		$this->globalVars['wgAPIModules']['askargs'] = '\SMW\MediaWiki\Api\AskArgs';
 		$this->globalVars['wgAPIModules']['browsebysubject'] = '\SMW\MediaWiki\Api\BrowseBySubject';
--- a/res/Resources.php
+++ b/res/Resources.php
@@ -186,7 +186,7 @@ return array(
 		'dependencies' => 'jquery.ui.autocomplete',
 		'targets' => array( 'mobile', 'desktop' )
 	),
-	
+
 	// Purge resources
 	'ext.smw.purge' => $moduleTemplate + array(
 		'scripts' => 'smw/util/ext.smw.util.purge.js',
@@ -199,7 +199,20 @@ return array(
 			'desktop'
 		)
 	),
-	
+
+	// Postproc resources
+	'ext.smw.postproc' => $moduleTemplate + array(
+		'scripts' => 'smw/util/ext.smw.util.postproc.js',
+		'position' => 'top',
+		'messages' => array(
+			'smw-postproc-queryref'
+		),
+		'targets' => array(
+			'mobile',
+			'desktop'
+		)
+	),
+
 	// Special:Ask
 	'ext.smw.ask' => $moduleTemplate + array(
 		'scripts' => 'smw/special/ext.smw.special.ask.js',
--- a/res/smw/util/ext.smw.util.postproc.js
+++ b/res/smw/util/ext.smw.util.postproc.js
@@ -1,0 +1,51 @@
+/*!
+ * This file is part of the Semantic MediaWiki Reload module
+ * @see https://www.semantic-mediawiki.org/wiki/Help:Purge
+ *
+ * @since 3.0
+ *
+ * @file
+ * @ingroup SMW
+ *
+ * @licence GNU GPL v2+
+ * @author mwjames
+ */
+
+/*global jQuery, mediaWiki, smw */
+/*jslint white: true */
+
+( function( $, mw ) {
+
+	'use strict';
+
+	/**
+	 * @since 3.0
+	 */
+	mw.loader.using( [ 'mediawiki.api', 'mediawiki.notify' ] ).then( function () {
+
+		$( '.smw-postproc' ).each( function() {
+
+			var queryRef = $( this ).data( 'queryref' );
+
+			if ( queryRef !== '' ) {
+				mw.notify( mw.msg( 'smw-postproc-queryref' ), { type: 'info', autoHide: false } );
+
+				var postArgs = {
+					'action': 'smwtask',
+					'subject': $( this ).data( 'subject' ),
+					'taskType': 'queryref',
+					'taskParams': JSON.stringify( queryRef )
+				};
+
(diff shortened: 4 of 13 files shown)


mwjames:
@kghbln FYI

I created an example [0] which demonstrates the approach including a cascading value annotation that requires multiple computations (hence updates) where input is taken from an alternation.

As I outlined in [1], the reliance on some non-guaranteed updates triggered by `refreshLinksPrioritized` would eventually fail on [0] as it would not allow a cascading update to be triggered in order to ensure that all annotation dependencies are re-computed and set.

As for the update, we rely on JS and the API to trigger an update from the client-side (browser-side) to create an event-based mechanism that doesn't rely on the job scheduler for those tasks. As long as the process is pending, a small notification is shown. Each `@annotation` on a query will cause a complete re-parse of the entire page using `UpdateJob` in the background as we cannot isolate a single annotation that may or may not share a dependency.

We mark a page for post processing (PostProc) before annotations are stored and can therefore not efficiently judge whether changes to a page related to a specific annotation that uses a query. Any change to a page with an embedded `@annotation` will trigger a `PostProc` therefore the use of `@annotation` should be done with care to avoid unnecessary updates or a parsing stampede.

Limitations and requirements are as listed above.

image

[0] https://sandbox.semantic-mediawiki.org/wiki/Issue/2572
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2549#issuecomment-315545252

kghbln:
@mwjames Again thanks for [cranking up documentation](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Annotation_query_marker)

mwjames:
@jamesmontalvo3, refs https://www.semantic-mediawiki.org/wiki/Thread:Help_talk:Ask_syntax/Annotation_query_marker/Is_this_advocating_self-referencing%3F

>Is this advocating self-referencing? 
> [[SomeProperty::{{#show: {{FULLPAGENAME}} |?AnotherProperty |link=none |@annotation }}]]

We have advertised against it for years, yet users will use that form no matter what we say and will be surprised by the results due to how parsing, updating, and querying works given the sequential nature (see above and #2375) of the process.

`@annotation` has been introduced to make those in-query/by-query annotations distinguishable hereby allowing to apply a different update process but as noted above "... change to a page with an embedded `@annotation` will trigger a PostProc therefore the use of `@annotation` should be done with care to avoid unnecessary updates ..."

PR #2585, opened 2017-07-30
Stop IRC notifications for every PR

JeroenDeDauw:


kghbln:
In guess this is allright in case nobody of you uses this as a "notification hub" system about new pull requests. This has little value for community users I believe.

JeroenDeDauw:
This is indeed for the devs in the IRC. I'm not using it, and doubt anyone else is. And given there is little user chatter in the IRC, these messages make it hard to find the actual user chatter.

PR #2600, opened 2017-08-06
Add create restriction on properties, refs 2232

mwjames:
This PR is made in reference to: #2232

- Adds `$smwgCreateProtectionRight` setting
- If enabled, restrict creation of new properties to user with the assigned right only
- https://www.semantic-mediawiki.org/wiki/Authority_mode

image

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI, I'll leave this open for a week in case you want to comment.

kghbln:
Indeed a nice enhancement which supports data curation and gardening! One question: How does this integrate with special page "WantedProperties" which currently has a slightly different scope? I can imagine that such properties which were not yet created could be listed there too - somehow marked in a special way.

mwjames:
> integrate with special page "WantedProperties" which currently has a slightly different scope? I can imagine that such properties which were not yet created could be listed there too 

I'll try to squeeze in a follow-up on this.

mwjames:
> I'll try to squeeze in a follow-up on this.

Done in #2615.

kghbln:
https://www.semantic-mediawiki.org/wiki/Help:Create_protection - to be fluffed

mwjames:
>https://www.semantic-mediawiki.org/wiki/Help:Create_protection - to be fluffed

I gave it a quick note.

kghbln:
> I gave it a quick note.

Cool, my earlier comment was more of a not to self. Nice. [`$smwgCreateProtectionRight`](https://www.semantic-mediawiki.org/wiki/Help:$smwgCreateProtectionRight) is documented too. :)

mwjames:
@kghbln Quick question, Do you think we make the collection display of the same size as the infobox (280px) as the current box feels a bit overwhelming when reading (focusing on) the text.

kghbln:
Yeah, sure. Since I do not use mobile devices I have no idea how the wiki appears on respective clients. For me personally on a desktop things are already pretty tiny.

PR #2632, opened 2017-08-19
Property page check for property label uniqueness, refs 2227

mwjames:
This PR is made in reference to: #2227, #2630 

- Checks the uniqueness of a property label and if necessary shows a warning 
- https://www.semantic-mediawiki.org/wiki/Property_uniqueness

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

mwjames:
@kghbln To clarify:

Why do I still see the message "The "20171107 Issue title" label was matched to at least one other property ..." even though `20171107 Issue Title` has been deleted?

Because instead of using a redirect, annotations may still be referenced (late job execution etc.) to the `20171107 Issue Title` as property therefore the `ID_TABLE` contains an ID with `20171107 Issue Title` and the uniqueness check is matching `20171107 Issue title` and `20171107 Issue Title`.

image

image

image

In the specific example above, checking `20509` should reveal that it no longer holds any references and can therefore be disposed.

PR #2659, opened 2017-09-01
 Removed long deprecated constant

Nasty:
This made issues with PageForms extension.


JeroenDeDauw:
Thanks! I've updated the release notes in 9c040adb0bbf509133b474c990f8daa648b9eb54

mwjames:
> This made issues with PageForms extension.

What kind of issue? Both extensions (SMW and PageForms) are not interdependent hence any declared CONST should not cause any issue on either side.

A bit more context would highly appreciated.

PR #2663, opened 2017-09-09
Specify hierarchy depth restriction in query condition, refs 2662

### LINKED ISSUE #2662: Hierarchies, queries, and depth restriction
http://bots.wmflabs.org/~wm-bot/logs/%23semantic-mediawiki/20170907.txt with 
> [13:30:52] <traumschule>	 ... but I am confused to select pages from a category via #ask not showing images and category contenty.
[13:34:53] <traumschule>	 {{#ask: |[[Category:fix]] |[[:+]] }} is a workaround but it shows pages in selected categories as well

The above is orthogonal to this issue but elevated the question of "How to restrict depth on a per query basis".

Generally, users do not recognize that hierarchies are resolved by default with queries yielding results that contain a larger match pool (as it includes members of the subcategories). The semantics of subcategories and subproperties are so that members are included in a set of a higher order class/property therefore resolving them is the expected behaviour.

## Issue

Semantic MediaWiki by default [0, 1] resolves subcategories and subproperties as part of a hierarchy with the levels as defined in [0, 1] and is applied to all queries where a hierarchy is detectable. It is currently not possible to restrict the depth on an individual query, forcing a user to choose between either or without providing a method to intercept or model specific use cases where hierarchy inferences are expected to be suspended temporarily.

```
Category: Foo
	|- Category: Bar
	|- Category: Foobar
		|- Category: Foobar/1
			|- Category: Foobar/1/1
		|- Category: Foobar/2
```
Given the levels in [0, 1], the query `{{#ask: [[Category: Foo]] }}` would resolve all members of the hierarchy tree without a possibility to restrict its level for this particular query.

## Proposal

To eliminate the outlined deficiency, the following syntax `[[...|+depth=...]]` is proposed to allow to define an individual level on a per query basis where [0] and [1] also function as maximum level for a possible definition within a query condition.

```
{{#ask: [[Category: Foo|+depth=0]]
 |?Category
}}
```

The query `[[Category: Foo|+depth=0]]` would only display members of the `Category: Foo` without resolving any other subcategory.

For the `SPARQLStore` we would only be able to support `|+depth=0` and `|+depth=1` achieved by modifying the path operator [2, 3] as in `*` or `?` (e.g. `rdfs:subClassOf*`, `rdfs:subClassOf?`) while any other level would require additional computation [4] and seems unnecessary for the SMW project.

[0] https://www.semantic-mediawiki.org/wiki/Help:$smwgQSubcategoryDepth
[1] https://www.semantic-mediawiki.org/wiki/Help:$smwgQSubpropertyDepth

[2] https://www.w3.org/TR/sparql11-query/#pp-language
[3] https://www.w3.org/TR/sparql11-query/#propertypath-arbitrary-length
[4] https://stackoverflow.com/questions/18126949/limit-the-sparql-query-result-to-first-level-in-hierarchy

refs: https://www.semantic-mediawiki.org/wiki/Help:Inferencing#Subcategories, https://www.semantic-mediawiki.org/wiki/Help:Inferencing#Subproperties

mwjames:
This PR is made in reference to: #2662 (#1003, #1012)

- Introduces the syntax of `+depth=`  in connection with resolving hierarchies for conditions like `[[Category:Foo|+depth=0]]` and `[[SomeProperty:SomeValue|+depth=0]]`
- Limitations for the `SPARQLStore` as described in #2662 by supporting only `0, 1, or any`

This PR includes:
- [x] CI build passed

Fixes #2662

### DIFF
--- a/includes/query/SMW_QueryParser.php
+++ b/includes/query/SMW_QueryParser.php
@@ -8,6 +8,7 @@ use SMW\Query\Language\ConceptDescription;
 use SMW\Query\Language\Description;
 use SMW\Query\Language\NamespaceDescription;
 use SMW\Query\Language\SomeProperty;
+use SMW\Query\Language\Disjunction;
 use SMW\Query\Language\ThingDescription;
 use SMW\Query\Parser\DescriptionProcessor;
 use SMW\Query\QueryToken;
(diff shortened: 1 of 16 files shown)


kghbln:
Thanks for [cranking on documenation](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Hierarchy_depth_restriction)

PR #2270, opened 2017-02-18
Add query parameters to profile

mwjames:
This PR is made in reference to: #

This PR addresses or contains:

- Add ability to record query parameters, this would subsequently enable SMW to recreate query results using a background job

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

PR #2146, opened 2017-01-01
Track query source in profile

mwjames:
This PR is made in reference to: # N/A

This PR addresses or contains:

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

PR #2673, opened 2017-09-16
Add query scode to profile, refs 2270, 2572

mwjames:
This PR is made in reference to: #2270, #2572

- Adds the `Query scode` property to be able to track an internal state (using a number as status code) on when a `#ask` uses `@annotation` or `@deferred` as special execution mode

This PR includes:
- [x] CI build passed

Fixes #


kghbln:
@mwjames This one is enabled by default?

kghbln:
I believe it is not: [Documented on wiki.](https://www.semantic-mediawiki.org/wiki/Help:Special_property_Query_scode)

kghbln:
> @mwjames This one is enabled by default?

When looking at the logic of this [page](https://www.semantic-mediawiki.org/wiki/Help:$smwgQueryProfiler) I am not so sure any longer.

kghbln:
> When looking at the logic of this page I am not so sure any longer.

Just had a peep at sandbox. Indeed it is enabled by default.

mwjames:
> Just had a peep at sandbox. Indeed it is enabled by default.

First I wanted to introduce a general `Status code` property but then I thought this
may collide with many other user defined properties of the same name so I settled for `Query scode` but now after a week, I'm not so happy with the name, I might need to rename it (will keep the number coding system).

We also have #2146, which appears when a federated query (remote) is used.

The idea behind it is very simple, if you wanted to find queries with a special execution mode you'd be out of luck hence the status code (or execution mode, OR a better name??).

Using a simple number code should be enough to fill that gap and allows to extend the list in future and hopefully avoids the need to introduce other properties.

100 - `@annotation`
200 - `@deferred`
300 - halted due to expensiveness (#2476) (isn't used yet!!)

kghbln:
@mwjames Thanks a lot for your insight on this.

>  I settled for Query scode but now after a week, I'm not so happy with the name,

Probably "Query mode code" as in execution mode is a better name for it.

kghbln:
I have added the "requires follow-up" label to prevent this from being forgotten.

jaideraf:
> I'm not so happy with the name, I might need to rename it (will keep the number coding system).

Maybe it's too late (sorry), but what about simply "`Query status code`"? 
Or, "`Query mode code`" as suggested by @kghbln?
We already have a name convention problem [here](https://www.semantic-mediawiki.org/wiki/Help:Special_properties): some property names start with "Has..." other don't. We have to be carefull to not introduce a third name convention here.

kghbln:
@mwjames I think is makes sense to rename. "Query status code" makes sense to me. Most of the profiling related special properties do not have a Has ... so this could only be an alias.

mwjames:
> I think is makes sense to rename. "Query status code" makes sense to me. Most of the profiling related special properties do not have a Has ... so this could only be an alias.

I was reviewing the issue and while the "code" part in the name isn't incorrect, I somehow feel the "status code" expresses a false semantic and therefore I'd rather go with `Query state` where "state" connotes a mode of existence which is closer to what the content of the property is to represent.

kghbln:
> I was reviewing the issue and while the "code" part in the name isn't incorrect, I somehow feel the "status code" expresses a false semantic and therefore I'd rather go with Query state where "state" connotes a mode of existence which is closer to what the content of the property is to represent.

Fair enough. This makes sense to me.

kghbln:
@mwjames Do you think you could still give this a shot with "Query state" if it does not consume too much time? This would really be great.

PR #2677, opened 2017-09-16
TableResultPrinter to support +width parameter

mwjames:

- Allows to define ` |+width=...` as in `|?Has number |+width=50%` to format the width of a table column and hereby easily enables different tables on the same page to be formatted with the same width without relying on a template format.

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/queryprinters/TableResultPrinter.php
+++ b/includes/queryprinters/TableResultPrinter.php
@@ -184,6 +184,7 @@ class TableResultPrinter extends ResultPrinter {
 		if ( count( $dataValues ) > 0 ) {
 			$sortKey = $dataValues[0]->getDataItem()->getSortKey();
 			$dataValueType = $dataValues[0]->getTypeID();
+			$printRequest = $resultArray->getPrintRequest();
 
 			if ( is_numeric( $sortKey ) ) {
 				$attributes['data-sort-value'] = $sortKey;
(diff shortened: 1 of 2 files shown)


mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2677

kghbln:
Thanks for the [docu](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Printout_column_width)

kghbln:
@mwjames 

This [does not apply to the mainlabel column](https://sandbox.semantic-mediawiki.org/wiki/Issue/2677/mainlabel) (same for `|+align=...`) I am not sure if `mainlabel` should listen to this since we have a good alternative way to achieve this without much extra syntax.

mwjames:
> This does not apply to the mainlabel column (same for |+align=...) I am not sure if mainlabel should listen

The mainlabel is not suppose to, the internal parsing code [0] isn't meant to not take more than a label argument for the `mainlabel` field.

Even if you would extend the [0] part, your `Special:Ask` and [1] for the mainlabel input field aren't made to be used as "everything goes" field like the textarea for the printout therefore your outlined alternative is the way to go.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/query/SMW_QueryProcessor.php#L252-L305
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/query/SMW_QueryProcessor.php#L189-L208

kghbln:
> The mainlabel is not suppose to, the internal parsing code [0] isn't meant to not take more than a label argument for the mainlabel field.

Thanks for confirming. Documented [here](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AAsk_syntax%2FPrintout_column_width&type=revision&diff=54886&oldid=54409) and [there](https://www.semantic-mediawiki.org/w/index.php?title=Help:Ask_syntax/Printout_column_content_alignment&curid=13302&diff=54887&oldid=54408) on the wiki.

PR #2693, opened 2017-09-23
Do not allow PHP 7.1 failures

JeroenDeDauw:
This is only taking 8 minutes like the other ones, not 35


mwjames:
> This is only taking 8 minutes like the other ones, not 35

Yeah, because I took off the coverage run otherwise it would require more than 50 min and Travis would halt the job by any means.

JeroenDeDauw:
Right ok. So good to go?

PR #2696, opened 2017-09-23
Add API smwbrowse module

mwjames:
This PR is made in reference to: #

This PR addresses or contains:

- Adds `smwbrowse` as API module (required for an upcoming feature) and currently supports the search for
  - properties
  - categories
  - concepts
- Does a similar thing like `browsebyproperty`  but that module was not easily extendable therefore rewrote it from scratch, eventually it will be deprecated and removed
- It is expected to incorporated `browsebysubject` at a later state so that this module can be scrapped as well
- Classes and responsibilities are split into:
  - `Browse` (API, MediaWiki)
  - `ListLookup` (primary selector),
  - `ListAugmentor` (adding additional field components) with
    - `property` to support:
      - `description` to pull the `Has property description` in a selected language
      - `prefLabal` to fetch the `Has preferred property label` in a selected language
      - `usageCount` retrieve the usage count of the property
  - `LookupCache` caches each request for 30 min and by this avoids to introduce some complex invalidation strategy, yet keeping requests close to up-to-date information

## Example

```
'api.php?action=smwbrowse&browse=property&params={ "limit": 10, "offset": 0, "search": "Date" }',
'api.php?action=smwbrowse&browse=property&params={ "limit": 10, "offset": 0, "search": "Date", "description": true }',
'api.php?action=smwbrowse&browse=property&params={ "limit": 10, "offset": 0, "search": "Date", "description": true, "prefLabel": true }',
'api.php?action=smwbrowse&browse=property&params={ "limit": 10, "offset": 0, "search": "Date", "description": true, "prefLabel": true, "usageCount": true }',
'api.php?action=smwbrowse&browse=category&params={ "limit": 10, "offset": 0 }',
'api.php?action=smwbrowse&browse=category&params={ "limit": 10, "offset": 0, "search": "Date" }',
'api.php?action=smwbrowse&browse=concept&params={ "limit": 10, "offset": 0 }'
'api.php?action=smwbrowse&browse=concept&params={ "limit": 10, "offset": 0, "search": "Date" }'

```

## Output
```json
{
    "query": {
        "Has_description": {
            "label": "Has description",
            "key": "Has_description",
            "description": {
                "en": "Simple descriptive explanatory text property."
            },
            "prefLabel": {
                "en": ""
            }
        },
        "_PDESC": {
            "label": "Has property description",
            "key": "_PDESC",
            "description": {
                "en": "&quot;Has property description&quot; is a predefined prop..."
            },
            "prefLabel": {
                "en": ""
            }
        }
[...]

PR #2699, opened 2017-09-23
Special:Ask add input assistance to condition textarea

mwjames:
This PR is made in reference to: #2696 

- Adds the https://github.com/ichord/At.js plug-in (License: MIT)
- The plug-in allows us to provide an input assistance for the condition textarea with search support for:
  - `p:` properties
  - `c:` categories, and
  - `con:` concepts

ask-input-assistance

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

DefaultSettings.php
```diff
@@ -884,6 +889,7 @@
 		'smwgPropertiesCacheExpiry' => 3600,
 		'smwgStatisticsCache' => true,
 		'smwgStatisticsCacheExpiry' => 3600,
+		'api.browse' => 3600
 	),
```
kghbln: @mwjames Cannot this be named somewhat more consistent like e.g. `smwgApiBrowse`?

mwjames: I'm considering to rename the others because there is some cruft, maybe something like:

* `special.statistics` which can be either a number or `false` and would kill one of the settings
* `special.properties`
* `special.unusedproperties`
* `special.wantedproperties`

kghbln: Makes sense to me. Also this would indeed make configuration a bit easier. Go for it. :)

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgCacheUsage&type=revision&diff=54814&oldid=54578)

PR #2703, opened 2017-09-25
Update en.json

kghbln:
This PR is made in reference to: #2232 

- Add missing system message used on "Special:ProtectedPages".

Note that the part within the brackets is usually added automatically by MediaWiki. The SMW implementation works a bit different thus adding this part here too since I do not think that it is worth automatizing the display of the allowed user groups. `{{PLURAL}}` was added for the sake of avoiding complaints though always the plural version is used.

This PR includes:
- [x] CI build passed

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -395,7 +395,7 @@
 	"right-smw-admin": "Access administration tasks (Semantic MediaWiki)",
 	"right-smw-patternedit": "Edit access to maintain allowed regular expressions and patterns (Semantic MediaWiki)",
 	"right-smw-pageedit": "Edit access for <code>Is edit protected</code> annotated pages (Semantic MediaWiki)",
-	"restriction-level-smw-pageedit": "protected (only {{PLURAL:$1|eligible user|eligible users}})",
+	"restriction-level-smw-pageedit": "protected (only eligible users)",
 	"action-smw-patternedit": "edit regular expressions used by Semantic MediaWiki",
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
+1

Nikerabbit:
Unless the number is hardcoded, the complaints are [legitimate](https://www.mediawiki.org/wiki/Localisation#Be_aware_of_PLURAL_use_on_all_numbers). Also message documentation in qqq.json should be added.

kghbln:
> Unless the number is hardcoded, the complaints are legitimate.

That's true but there is actually no number at all. In the past there have been complaints by translators that plural should be added since plural is theoretically necessary possible.

> Also message documentation in qqq.json should be added.

[Message docu](https://translatewiki.net/wiki/MediaWiki:Restriction-level-smw-pageedit/qqq) was already added yesterday directly to twn.

JeroenDeDauw:
> Mission: avoiding complaints

image

(I'm just goofing around, no need to take this serious)

kghbln:
Actually there is indeed an issue since I am trying to add dead code to solve an issue with the users and not the software. This is only possible since MediaWiki seems to instinctively choose plural in case of uncertainty. This again could be a source of regression in the future if MediaWiki changes here. Hmm, better remove plural...

Nikerabbit:
No number at all? Okay, then don't use plural and state in the qqq that it's not possible to know the exact number.

kghbln:
> Okay, then don't use plural and state in the qqq that it's not possible to know the exact number.

Done!

PR #2712, opened 2017-09-29
Always define cell datatype class for query printouts, refs 2586

### LINKED ISSUE #2586: Empty css-class information in result table
## Setup and configuration

- SMW version: 2.5.1
- MW version: 1.27.1
- PHP version: 5.6.30
- DB (MySQL etc.): mysql

## Issue

Inline queries, that contain empty printouts, generate html, that misses a class-name.

### Example:
Given the following inline query
```
{{#ask:
 [[Category:Ex]]
 |?Titel
 |?Dokumentart
 |?vom
 |format=table
}}
```
produces the following HTML:
```
[..]
<tr data-row-number="1" class="row-odd">
  <td class="Titel smwtype_wpg">xyz</td>
  <td></td>
  <td data-sort-value="2457959.3189815" class="vom smwtype_dat">24.07.2017</td>
</tr>
[..]
```
Notice, that the 2nd "TD" element provides no css-class-information. It would have been expected that it contains this class information: 
```
<td class="Dokumentart smwtype_wpg">xyz</td>
```
Apparently, columns that contain no value miss the css-class information.

This is unfortunate, if you want to purge columns in css-media queries, i.e. for small-size devices. In this case, a class-selector would match the table's column header, but it would miss the (empty) cells.

TK-999:
This PR is made in reference to: #2586 

For certain inline queries which contain empty printouts, the generated HTML table cell does not contain proper `class` attribute. We must ensure that the `class` attribute is always defined, even if the cell is empty.

This PR includes:
- [x] CI build passed

Fixes #2586

### DIFF
--- a/includes/queryprinters/TableResultPrinter.php
+++ b/includes/queryprinters/TableResultPrinter.php
@@ -178,13 +178,20 @@ class TableResultPrinter extends ResultPrinter {
 			$dataValues[] = $dv;
 		}
 
-		$attributes = array();
+		$printRequest = $resultArray->getPrintRequest();
+		$dataValueType = $printRequest->getTypeID();
+
+		$cellTypeClass = $dataValueType !== '' ? ' smwtype' . $dataValueType : '';
+
+		// We would like the cell class to always be defined, even if the cell itself is empty
+		$attributes = [
+			'class' => $columnClass . $cellTypeClass
+		];
+
 		$content = null;
 
 		if ( count( $dataValues ) > 0 ) {
 			$sortKey = $dataValues[0]->getDataItem()->getSortKey();
-			$dataValueType = $dataValues[0]->getTypeID();
-			$printRequest = $resultArray->getPrintRequest();
 
 			if ( is_numeric( $sortKey ) ) {
 				$attributes['data-sort-value'] = $sortKey;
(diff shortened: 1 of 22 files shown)


TK-999:
I'll take a look at the test failures

mwjames:
As general note:

While the change in itself isn't causing an issue, the modification of
the output is and is causing our integration tests (those that actual
test the integration with MW) to fail.

You may not be familiar with [0] but those tests are our defense line
against regressions especially those caused by MW releases and of
course those from ourselves if we change the behaviour.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/JSONScript/README.md

PS: The JSON may seem counter intuitive to a hard-core developer but
it frees us from writing integration tests (and hereby I mean raw
PHPUnit) one by one.

Cheers

TK-999:
@mwjames Yeah, I've updated them and they should pass after I correct my trivial copy-and-paste fails :) I've added explicit sorting to attributes so that changing their order of insertion won't break tests in the future.

mwjames:
> I've added explicit sorting to attributes so that changing their order of insertion won't break tests in the future.

:+1: 

> Eliminate final test failure

Thanks for the perseverance to see this through, I know it can be daunting but its the only way we can keep some form of quality assurance.

TK-999:
>Thanks for the perseverance to see this through, I know it can be daunting but its the only way we can keep some form of quality assurance.

Hah, no problem, it's actually a really nice declarative approach to integration tests. And congrats on the 87% coverage 😉

PR #2718, opened 2017-09-30
Datatable to support ad-hoc (JSON,CSV ...) export, refs

mwjames:
This PR is made in reference to: #2420

- Adds JSON, CSV, RSS, and RDF as export option directly to the output

image

This PR includes:
- [x] CI build passed

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -603,6 +603,7 @@
 	"smw-format-datatable-previous": "Previous",
 	"smw-format-datatable-sortascending": ": activate to sort column ascending",
 	"smw-format-datatable-sortdescending": ": activate to sort column descending",
+	"smw-format-datatable-toolbar-export": "Export",
 	"smw-category-invalid-redirect-target": "Category \"$1\" contains an invalid redirect target to a non-category namespace.",
 	"smw-parser-function-expensive-execution-limit": "The parser function has reached the limit for expensive executions (see [https://www.semantic-mediawiki.org/wiki/Help:$smwgQExpensiveExecutionLimit $smwgQExpensiveExecutionLimit]).",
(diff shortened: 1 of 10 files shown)


mwjames:
@kghbln FYI, Needs a bit of testing once merged!

kghbln:
https://sandbox.semantic-mediawiki.org/w/index.php?title=Issue/2718

Very cool! So far additional field tests showed that  the results were as expected. I think however that it needs a bit of tweaking for narrow screens / mobile devices.

grafik

kghbln:
PS I believe to know [from where](https://sandbox.semantic-mediawiki.org/wiki/Books) you go this idea. :)

mwjames:
> I think however that it needs a bit of tweaking for narrow screens / mobile devices.

Yes, but those are CSS only and specific to a screen maybe using `flex...` or so could be a solution. It is a bit of work to find the right soluton for small screens (e.g `@media screen and (max-width: 420px)` etc.) but related files are:

*  https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/res/smw/ext.smw.dropdown.css
* https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/res/smw/printer/ext.smw.tableprinter.css

[0] https://www.w3schools.com/css/css_rwd_intro.asp
[1] http://www.tutorialspoint.com/css/css_responsive.htm

kghbln:
I put this on my list. Hopefull I will not forget so poke in case ...

krabina:
documented at https://www.semantic-mediawiki.org/wiki/Help:Table_format and https://www.semantic-mediawiki.org/wiki/Demo:Table_formats_compared

PR #2721, opened 2017-09-30
API smwbrowse to add `article` search, refs 2696

mwjames:
This PR is made in reference to: #2696

- Extends the `smwbrowse` module to add `article` as type to support `action=smwbrowse&browse=article&params={ "limit": 10, "offset": 0, "search": "Main", "fullText": true, "fullURL": true }'`
- Removes the ` jquery.ui.autocomplet` module (MediaWiki has removed support for jquery widget)
- Uses the same autocomplete as implemented for the property input
- `Special:Browse` input box works on mobile now

```
This page is using the deprecated ResourceLoader module "jquery.ui.widget".  load.php:31:808
This page is using the deprecated ResourceLoader module "jquery.ui.core".
Please use "mediawiki.ui.button" or "oojs-ui" instead.  load.php:1:80
This page is using the deprecated ResourceLoader module "jquery.ui.position".
```
```
"Uncaught Error: Unknown dependency: jquery.ui.autocomplete" in mobile context as the module is not defined with target mobile
```

```sql
SELECT /* SMW\MediaWiki\Api\Browse\ArticleLookup::search  */  page_id,page_namespace,page_title  
FROM "page"   
WHERE page_title LIKE '%Lorem%' ESCAPE '`' OR page_title LIKE '%Lorem%' ESCAPE '`' OR page_title LIKE '%LOREM%' ESCAPE '`' OR page_title LIKE '%lorem%' ESCAPE '`'  
ORDER BY page_title,page_namespace 
LIMIT 21  
OFFSET 0
```
This PR includes:
- [x] CI build passed


kghbln:
Should `article` not rather be called `page` since the latter is a neutral name of what we are actually browsing? `articles` are only pages what MediaWiki calls content namespaces whereas Semantic MediaWiki considers pages with semantic links as "content" namespaces. Both must not be congruent. I know this is a nitpick but ...

mwjames:
> articles are only pages what MediaWiki calls content namespaces whereas Semantic MediaWiki considers pages with semantic links as "content" namespaces

I guess we have to rename this one then.

kghbln:
> I guess we have to rename this one then.

We would thus indeed have an unambiguous naming.

mwjames:
Could we make a follow up on this, I don't think my schedule for the weekend sprint can cover this.

kghbln:
> Could we make a follow up on this, I don't think my schedule for the weekend sprint can cover this.

Sure. I myself am going to the SMWCon Fall presentation tomorrow.

kghbln:
I have added the "requires follow-up" label to prevent this from being forgotten.

kghbln:
This is now documented with [`article`](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AAPI%3Asmwbrowse&type=revision&diff=64003&oldid=62670) and I still think it will be very cool if this could be changed to `page`.

kghbln:
@mwjames Do you think you could still give this a shot if it does not consume too much time?

PR #2724, opened 2017-09-30
Subobject to use a comparable content hash

### LINKED ISSUE #2723: Comparable hash for subobjects
### Setup and configuration

- SMW version: all

### Issue:

Anonymous subobjects describing the same "facts" but in different order produce a different resource ID (aka hash identifier). This should just not happen and bugged me for a while now!

For examples, the order of either `Foo,Bar` or `Bar,Foo` should NOT produce a different identifier since the subobject describes one of the same fact (the order of a statement doesn't matter at this point).

### Steps to reproduce:

```
{{#subobject:
 |Has text=Foo,Bar|+sep=,
}}

{{#subobject:
 |Has text=Bar,Foo|+sep=,
}}
```
```
{{#subobject:
 |Has number=123,345|+sep=,
 |Has text=Foo,Bar|+sep=,
}}

{{#subobject:
 |Has text=Foo,Bar|+sep=,
 |Has number=123,345|+sep=,
}}
```

mwjames:
This PR is made in reference to: #2723 

- Adds `$smwgUseComparableContentHash` as setting to help those who need more time for a migration but the default is `true` which means that it uses an internal sorting to ensure a comparable  hash when creating the subobject ID
-  `$smwgUseComparableContentHash` will be removed with 3.1, the setting is only introduced to ease migration
- This ticket is only related to anonymous subobjects 
- Content arguments are only internally sorted that are used for the hash, the content itself remains in the order of the input

This PR includes:
- [x] CI build passed

Fixes #2723

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1677,4 +1677,30 @@ return array(
 	'smwgChangePropagationProtection' => true,
 	##
 
+	##
+	# Subobject content hash !! BC setting ONLY !!
+	#
+	# Normalized content hash is enabled by default to ensure that a content
+	# declaration like:
+	#
+	# {{#subobject:
+	# |Has text=Foo,Bar|+sep=,
+	# }}
+	#
+	# yields the same hash as:
+	#
+	# {{#subobject:
+	# |Has text=Bar,Foo|+sep=,
+	# }}
+	#
+	# The setting is only provided to allow for a temporary backwards compatibility
+	# and will be removed with 3.1 at which point the functionality is enabled
+	# without any constraint.
+	#
+	# @since 3.0
+	# @default true
+	##
+	'smwgUseComparableContentHash' => true,
+	##
+
 );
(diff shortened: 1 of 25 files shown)


mwjames:
@kghbln FYI

mwjames:
@kghbln Any input on this before I press the merge button?

kghbln:
So two or more subobjects with the identical hash will be treated as one subobject, i.e. only one result is returned. This would indeed restore the expected behavior for multiple identical annotations.

`$smwgUseComparableContentHash` as a community service parameter will hopefully not be used much with `false`. However it would also serve as a failsafe in case something might have been overlooked.

mwjames:
> So two or more subobjects with the identical hash will be treated as one subobject,

This is already the case now. The issue is that currently for cited examples #2723 there are not identified as equivalent hence produce a different hash which means you get two subobjects that represent the same facts.

This PR is to correct that behavior so that the same hash is produced for an equivalent representations (where order doesn't matter).

```
{{#subobject:
 |Has text=Foo,Bar|+sep=,
}}

{{#subobject:
 |Has text=Bar,Foo|+sep=,
}}
```

kghbln:
> This is already the case now. The issue is that currently for cited examples #2723 there are not identified as equivalent hence produce a different hash which means you get two subobjects that represent the same facts.

I typed my understanding a bit misleading. This is indeed what I meant, i.e. those will be treated as being identical in the future, too.

PR #2735, opened 2017-10-03
Move definition of ask param `import-annotation` to `SMWQueryProcessor`

s7eph4n:
This PR is made in reference to: #1257

* Add ResultPrinter::supportsRecursiveAnnotation()
* Remove definition of ask param `import-annotation` from `ResultPrinter::getParamDefinitions()`
* Add definition of ask param `import-annotation` to `SMWQueryProcessor::getParameters()`
* Fix calls to `SMWQueryProcessor::getParameters()`

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/query/SMW_QueryProcessor.php
+++ b/includes/query/SMW_QueryProcessor.php
@@ -75,7 +75,7 @@ class SMWQueryProcessor implements QueryContext {
 	 * @return Processor
 	 */
 	public static function getValidatorForParams( array $params, array $printRequests = array(), $unknownInvalid = true, $context = null ) {
-		$paramDefinitions = self::getParameters( $context );
+		$paramDefinitions = self::getParameters( null, $context );
 
 		$paramDefinitions['format']->setPrintRequests( $printRequests );
 
@@ -442,7 +442,7 @@ class SMWQueryProcessor implements QueryContext {
 	 *
 	 * @return IParamDefinition[]
 	 */
-	public static function getParameters( $context = null ) {
+	public static function getParameters( $resultPrinter = null, $context = null ) {
 		$params = array();
 
 		$allowedFormats = $GLOBALS['smwgResultFormats'];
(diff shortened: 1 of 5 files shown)


s7eph4n:
`SMWQueryProcessor` needs some love generally, lots of static methods. `getResultPrinter()` creates a new ResultPrinter for each call. I took the lazy approach and just added another param to `SMWQueryProcessor::getParameters()`.

includes/query/SMW_QueryProcessor.php
```diff
@@ -442,7 +442,7 @@ static public function getResultPrinter( $format, $context = self::SPECIAL_PAGE
 	 *
 	 * @return IParamDefinition[]
 	 */
-	public static function getParameters( $context = null ) {
+	public static function getParameters( $resultPrinter = null, $context = null ) {
```
JeroenDeDauw: This changes makes code that calls the function with $context wrong in a somewhat subtle manner. For this reason normally new parameters are added to the end of the list. Any reason why to not do that here?

s7eph4n: I checked the calls from SMW and there is only one that actually uses `$context` (which I fixed). The reason I put `$resultPrinter` in front was simply that the method is called more often with that than with `$context`. I did not think of external callers, though. I'll turn the params around.

JeroenDeDauw:
+1 from my side

mwjames:
> ResultPrinter

I'll merge the PR but it is very likely that I have to change this once more pending the introduction of `RecursiveTextProcessor` that moves the `$wgParser` stuff into an isolated class where the `Parser` is injected as part of the `AskParserFunction` and if not available will fallback to `$wgParser` (during API and Special:Ask requests).

> SMWQueryProcessor needs some love generally, lots of static methods.
> creates a new ResultPrinter for each call.

See #2737.

PS: Just last week I was forced (means unwillingly-willingly needed to act in order to confirm a behaviour) to do some refactoring #2727.

PSS: There are enough integration tests in place to safeguard against regressions, so BREAK IT and make it a more reliable and enjoyable interactor.

PR #2751, opened 2017-10-08
ResultPrinter isolate $wgParser access

mwjames:

- Introduces `RecursiveTextProcessor` to isolate direct access to `wgParser` in `ResultPrinter`
- `RecursiveTextProcessor` carries the parser instance from the initial `#ask` request, and if no parser is available (as during `Special:Ask`) falls back to the global `wgParser` instance
- Fixes "/// TODO: explain problem (too much recursive parses)" and "// FIXME Parser should be injected into the ResultPrinter"
- Removes handling of `import-annotation` from the `AskParserFunction`

This PR includes:
- [x] CI build passed


mwjames:
@kghbln @s7eph4n Marking this as potentially regression tacit, in case of some unexpected results especially on those results that interact with a template.

PR #2767, opened 2017-10-14
Force update on template refresh, refs 2365

mwjames:
This PR is made in reference to: #2365

- If we somehow detect a template update is ongoing through the means of a `refreshLinks` job force an update even though the revision hasn't changed (and hereby the updater marker)

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/MediaWiki/Hooks/LinksUpdateConstructed.php
+++ b/src/MediaWiki/Hooks/LinksUpdateConstructed.php
@@ -88,6 +88,12 @@ class LinksUpdateConstructed implements LoggerAwareInterface {
 			$parserData->setOption( $parserData::OPT_FORCED_UPDATE, true );
 		}
 
+		// Update incurred by a template change and is signaled through
+		// the following condition
+		if ( $linksUpdate->mTemplates !== [] && $linksUpdate->mRecursive === false ) {
+			$parserData->setOption( $parserData::OPT_FORCED_UPDATE, true );
+		}
+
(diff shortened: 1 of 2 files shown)


mwjames:
Back-ported to 2.5.x with 848ed0c.

mwjames:
@kghbln FYI

mwjames:
A bit of a backstory:

- When content is altered (and hereby most likely relevant data for SMW) the SMW update process is run as part of the `LinksUpdate` event (aka hook)
- Since MW 1.28 (or 1.29) the hook is run asynchronously which means it can be scheduled as `EnqueueableDataUpdate` and run as job
- To avoid us running twice for the same revision we set an update marker #2365 and store it for an hour so the likelihood us running twice on the same revision within a short period of time is marginal slim (remember new content = new revision = new update)
- When a template change happens in rapid succession and/or with page still holding an update marker our general strategy is to suppress an update where the revision is known
- The only way to find out that pages require an update on a template refresh is by looking at `$linksUpdate->mTemplates !== [] && $linksUpdate->mRecursive === false` and in case the condition matches `OPT_FORCED_UPDATE` is set so that we skip our check on the update marker

kghbln:
> When a template change happens in rapid succession and/or with page still holding an update marker our general strategy is to suppress an update where the revision is known

Indeed, templates a usually edited frantically so this will reduce the number of jobs. Cool!

PR #2773, opened 2017-10-14
Additional checks for enabled namespaces

mwjames:

- While looking at [0], I found entries like the sample below which should not happen because `  "smw_namespace": "200"` (Print namespace) is disabled for SMW usage
- This is caused by a page transclusion [1] that hosts embedded queries with the `QueryDependencyLinksStore` not explicitly checking the  context page namespace prior creating an id reference

```
    "109866": {
        "smw_id": "109866",
        "smw_title": "User_manual",
        "smw_namespace": "200",
        "smw_iw": "",
        "smw_subobject": "_QUERY85faec9baa9a72381449338d9fb87801",
        "smw_sortkey": "User manual"
    },
```

[0] https://www.semantic-mediawiki.org/w/index.php?title=Special%3ASemanticMediaWiki&action=idlookup&id=&id=User+manual
[1] https://www.semantic-mediawiki.org/wiki/Print:User_manual

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/src/MediaWiki/Hooks/ParserAfterTidy.php
+++ b/src/MediaWiki/Hooks/ParserAfterTidy.php
@@ -23,7 +23,12 @@ class ParserAfterTidy extends HookHandler {
 	/**
 	 * @var Parser
 	 */
-	private $parser = null;
+	private $parser;
+
+	/**
+	 * @var NamespaceExaminer
+	 */
+	private $namespaceExaminer;
 
 	/**
 	 * @var boolean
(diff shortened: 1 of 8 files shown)


mwjames:
@kghbln FYI This maybe worth backporting but it would require a bit more manual testing to avoid a regression since I had to modify the `PropertyTableIdReferenceFinder::findAtLeastOneActiveReferenceById` method.

kghbln:
I switched smw.o to `"2.5.x@dev"` so a backport could more easily be tested.

mwjames:
Back-ported to 2.5.x with 4bc0d5f.

PR #2776, opened 2017-10-15
Track change propagation on categories, refs 2495

mwjames:
This PR is made in reference to: #2495, https://phabricator.wikimedia.org/T176540

- Track changes on a category when it alters its hierarchy structure (aka `_SUBC`) and hereby allows related entities to be reevaluated
- Adds `_SUBP`, `_SUBC` to the `$smwgChangePropagationWatchlist` setting
- Uses the `ArticleViewHeader` hook to display the status on the category page about of the change propagation
- Adds `ChangePropagationClassUpdateJob` to be able to count jobs related to category change propagations while `ChangePropagationUpdateJob` remains responsible for tracking property changes
- The category page is locked until the primary specification (see the image below that shows `Special:Browse` and the `Change propagation` property that holds the primary specification) has been updated and works similar to the lock on the property page, in an administrative emergency `smwgChangePropagationProtection` can be used to disable the lock
- To disable this functionality remove `_SUBC` from the `$smwgChangePropagationWatchlist` setting

image

image

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

Issue #2784, opened 2017-10-21
Property page, JSON schema, and content type


mwjames:
Moving discussion from #2769 to its own issue:

mwjames:
@JeroenDeDauw Can we move related discussions this this thread?

WolfgangFahl:
It would be great if the PlantUML extension would be installed and I could continue explaining .. as outlined in #1330

mwjames:
> It would be great if the PlantUML extension would be installed and I could continue explaining 

You can attached picture (or graphs) to this issue, no need to divert the discussion to a wiki page.

mwjames:
> Pierre Racine http://www.cef-cfr.ca/uploads/Membres/PierreRacine-SMWCon2013-SMW_Metaproperties.pdf had also an original idea on this ...

To be clear, the proposal will have nothing in common with PS or have a particular support for PF. It is about using a JSON schema to store declarative information in a structure and machine readable format that may or may not use its own UI interpreter to display the stored information.

WolfgangFahl:
@mwjames - I want to show live example which are created by a generator tool. I'd have to change the generator since it depends on using PlantUML. That's why quite a few useful things have not yet made it to http://www.semantic-mediawiki.org. I think it is more an issue of priorities.

mwjames:
Moving it to "Enhancements and features" for future deliberation as it unlikely to make the 3.0 release cut.

JeroenDeDauw:
Hi James. I think I never saw this. If you want me to look at a PR or issue, please mail me. You are doing too much stuff for me to check all of it :)

mwjames:
> Hi James. I think I never saw this. If you want me to look at a PR or issue, please mail me.

Thanks. Will do.

PR #2785, opened 2017-10-22
Move PropertyPage, ConceptPage

mwjames:
This PR is made in reference to: #2784 

- Just in case some activities in relation to #2784 are going to happen, I dusted off an old PR that moves the property and concept page to its own `SMW\Page` NS
- It contains some minor refactoring and injects some of the required services using a factory 
- Adds some simple unit tests to assert the general functionality

This PR includes:
- [x] CI build passed


mwjames:
@kghbln @s7eph4n 

I added a minor design change which can be switched to but by default it still uses the old `smw-table smw-property-page-results legacy` (legacy class). Removing `legacy` switches to the new design and since I'm not sure it makes much difference or whether the new aliment is an improvement or not it is hidden from the "normal" user. Feel free to suggest a better design.

## Legacy
image

## New
image

s7eph4n:
It makes it better aligned with the rest of the page, so I think it is a good move. :+1: 

N.B. 1: At some point we could think about a general rework of the styling, e.g. aligning _Pages with improper assignments_ and _Pages using..._, better use of "screen real estate", etc.
(A general SMW style overhaul might be a candidate for one of the [Outreach Programs](https://www.mediawiki.org/wiki/Outreach_programs).)

N.B. 2: MW supports Less since 1.22. Would it make sense to switch?

mwjames:
> N.B. 1: At some point we could think about a general rework of the styling, e.g. aligning Pages with improper

#2786

> N.B. 2: MW supports Less since 1.22. Would it make sense to switch?

#2787

kghbln:
Note that due to 6792db0 the new styling is default which  is indeed the better approach. PR now documented [here](https://www.semantic-mediawiki.org/wiki/Help:Property_page) and [there](https://www.semantic-mediawiki.org/wiki/Help:Concept_page)

PR #2207, opened 2017-01-22
Allow IRI export, ref 517, 2188

mwjames:
This PR is made in reference to: #517, #2188

This PR addresses or contains:

- Adds `smwgExportResourcesAsIri` as experimental setting (disabled by default) to allow exporting of IRIs as outlined in [0]

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

[0] https://www.w3.org/TR/rdf11-concepts/#section-IRIs

Issue #2781, opened 2017-10-21
RDF export escapes non-ascii URLs and property names


changwoo:
### Setup and configuration

- SMW version:2.5.4
- MW version:1.29
- PHP version:7.0.19
- DB system (MySQL, Blazegraph, etc.) and version:mariadb 10.2.9

### Issue:

I'm using non-ascii characters for SMW properties and it seems to work great. But in the exported RDFs  those non-ascii names are escaped using non-standard `-XX` form. Like this:

```
<swivt:page rdf:resource="http://localhost:8080/wiki/-EA-B0-80-EA-B3-B5_-28001-29_-28-ED-95-9C-EA-B5-AD-EC-96-B4-EA-B8-B0-EC-B4-88-EC-82-AC-EC-A0-84_27947-29"/>
```
```
<wiki:Property-3A-EA-B0-88-ED-80-B4-3A-EB-9D-BC-EC-9D-B4-EC-84-A0-EC-8A-A4 rdf:datatype="http://www.w3.org/2001/XMLSchema#string">CC BY-SA 2.0 KR</wiki:Property-3A-EA-B0-88-ED-80-B4-3A-EB-9D-BC-EC-9D-B4-EC-84-A0-EC-8A-A4>
```

But AFAIK they don't have to be escaped at all because,

- according to the RFC3986, the default URL encoding is UTF-8. So I think URLs don't have to be escaped. Or at least URLs should be encoded in the standard `%XX` form, not `-XX` form.
- And according to the XML spec, all the Unicode word characters can be used for tag names. Thus such characters don't have to be escaped.

mwjames:
There is some history on the encoding issue but since I'm pressed with time during my weekend sprint, I can only add some quick notes and may have to come back to this in a week or so.

Generally, the generated XML (RDF) is only expected to be consumed by a "machine" not a human therefore readability isn't really required as long as the generate XML is consistent over the encoded content.

> And according to the XML spec, all the Unicode word characters can be used for tag names. Thus such characters don't have to be escaped.

Not really, and if I recall correctly tags that contain a `:` are causing issues.

> according to the RFC3986, the default URL encoding is UTF-8. So I think URLs don't have to be escaped. 

See [0] on whether this can solve your issue on the UTF-8 aspect or not. 

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/DefaultSettings.php#L1210-L1224

changwoo:
[...]

PR #2788, opened 2017-10-23
Changes default setting for `smwgExportResourcesAsIri` (breaking)

kghbln:
This PR is made in reference to: #2207 

- Switches (BREAKING) the default to use IRIs for resource export
- Fixes test per [comment](#2788 (comment)) by @mwjames

This PR includes:
- [ ] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1215,12 +1215,10 @@ return array(
 	#
 	# @see https://www.w3.org/TR/rdf11-concepts/#section-IRIs
 	#
-	# This setting should be set TRUE with beginning of 3.x.
-	#
 	# @since 2.5
-	# @default false (to avoid any BC break for exsiting systems)
+	# @default true
 	##
-	'smwgExportResourcesAsIri' => false,
+	'smwgExportResourcesAsIri' => true,
 	##
 
 	##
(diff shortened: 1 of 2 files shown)


kghbln:
@mwjames Oops, would not have guessed that this will be failing. I am not sure why `true` is not returning the expected. Probably some extra fiddle is needed perhaps even in light of https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2781

mwjames:
> @mwjames Oops, would not have guessed that this will be failing. I am not
> sure why `true` is not returning the expected. Probably some extra fiddle is

There is a small misstep in the `EscaperTest` and the reason why the
setting isn't reset.

```
@@ -24,11 +24,11 @@ class EscaperTest extends \PHPUnit_Framework_TestCase {

 		$this->testEnvironment = new TestEnvironment();
 		$this->testEnvironment->addConfiguration(
'smwgExportResourcesAsIri', false );
 	}

-	protected function setU() {
+	protected function tearDown() {
 		$this->testEnvironment->tearDown();
 		parent::tearDown();
 	}

```

PS: Maybe changing the title of the PR to indicate what has changed
(making it easier when browsing the git log).

Cheers

kghbln:
@mwjames Looks like we are cool!

mwjames:
> @mwjames Looks like we are cool!

That needs a definite section in the release notes since it would make any non-ASCII resource export invalid (and hereby RDF and SPARQL). As to why, #2781 contains some clarification.

PS: I may consolidate the export settings at the later point but this is orthogonal to what is being addressed here.

kghbln:
[Documented on wiki.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgExportResourcesAsIri&type=revision&diff=57116&oldid=56130)

Issue #2798, opened 2017-11-04
Consolidate settings, config parameters


mwjames:
It mostly relates to Boolean type settings where a feature switch achieves the same thing and allows in future to add more if necessary without cluttering the config matrix.

## Tasks

- [x] Browse settings `smwgBrowseFeatures` (`smwgToolboxBrowseLink`, `smwgBrowseShowInverse`, `smwgBrowseShowAll`, `smwgBrowseByApi`)
- [x] Category settings `smwgCategoryFeatures` (`smwgUseCategoryHierarchy`, `smwgCategoriesAsInstances`, `smwgUseCategoryRedirect`)
- [x] Parser settings `smwgParserFeatures` (`smwgInlineErrors`, `smwgEnabledInTextAnnotationParserStrictMode`, `smwgDecodeTextAnnotationWithStripMarker`, `smwgShowHiddenCategories`)

Thinking about mashing `smwgLinksInValues` into the parser setting since this is purely parser related.

@kghbln Early warning before you go all nuts on the upcoming changes.

PS: I was planning to push some changes in regards to the config parameter consolidation but with Travis being broken, that ain't gonna happen.

kghbln:
Note to all. The deprecated configuration parameters will be removed with SMW 3.1.0!

PR #2802, opened 2017-11-05
Consolidate parser settings `smwgParserFeatures`, refs 2798

mwjames:
This PR is made in reference to: #2798

- Adds `smwgParserFeatures` and replaces:
  - `smwgInlineErrors` with `SMW_PARSER_INL_ERROR`
  - `smwgShowHiddenCategories` with `SMW_PARSER_HID_CATS`
  - `smwgEnabledInTextAnnotationParserStrictMode` with `SMW_PARSER_STRICT`
  - `smwgDecodeTextAnnotationWithStripMarker` with `SMW_PARSER_UNSTRIP`
- Default is defined with `SMW_PARSER_STRICT | SMW_PARSER_INL_ERROR | SMW_PARSER_HID_CATS`

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -281,14 +281,6 @@ return array(
 	##
 
 	###
-	# Should warnings be displayed in wikitexts right after the problematic input?
-	# This affects only semantic annotations, not warnings that are displayed by
-	# inline queries or other features.
-	##
-	'smwgInlineErrors' => true,
-	##
-
-	###
 	# Category hierarchy recognition
 	#
 	# Should a subcategory be considered a hierarchy element in the
@@ -994,20 +986,6 @@ return array(
 	##
 
 	###
-	# This option enables to omit categories (marked with __HIDDENCAT__) from
-	# the annotation process.
-	#
-	# If a category is updated of either being hidden or visible, pages need to
-	# be refreshed to ensure that the StoreUpdater can make use of the changed
-	# environment.
-	#
-	# @since 1.9
-	# @default true (true = legacy behaviour, false = not to show hidden categories)
-	##
-	'smwgShowHiddenCategories' => true,
-	##
-
-	###
 	# QueryProfiler related settings
 	#
 	# @note If these settings are changed, please ensure to run update.php/rebuildData.php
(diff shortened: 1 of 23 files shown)


kghbln:
https://www.semantic-mediawiki.org/wiki/Help:$smwgParserFeatures

kghbln:
@mwjames I am still a bit shaky about these bitwise operators. So to add `SMW_PARSER_LINV` and at the same time remove `SMW_PARSER_HID_CATS` one would do:

```php
$smwgParserFeatures = ( $smwgParserFeatures | SMW_PARSER_LINV ) & ~SMW_PARSER_HID_CATS;
```
or redifine the whole configuration parameter, which is probably the best approach if unsure.

When looking at the configuration parameter output on special page "SemanticMediaWiki" this is displayed as `"smwgParserFeatures": 42,`. Is this some internal representation?

mwjames:
> When looking at the configuration parameter output on special page "SemanticMediaWiki" this is displayed as "smwgParserFeatures": 42,. Is this some internal representation?

Constants such as `SMW_PARSER_LINV` or `SMW_PARSER_HID_CATS` are represented by a number [0] with PHP only knowing those numbers hence the 42 (as additive). The numbers are base 10 but to make the bit operation work those will be transformed to base 2 which is why we go from 2, 4, 8 ... as byte to make sure that a bit operation works with the corresponding bit mask (e.g. 1011010) allowing to find out which of the options are set since each has a different mask.

> $smwgParserFeatures = ( $smwgParserFeatures | SMW_PARSER_LINV ) & ~SMW_PARSER_HID_CATS;

`|` says OR `&` says AND, `~` means NOT so above translates into `$smwgParserFeatures` (with the default) OR `SMW_PARSER_LINV` (adding it to the mask) and NOT `SMW_PARSER_HID_CATS` removing the bit mask from `$smwgParserFeatures`.

When you have something like `A | B | C` its describes it as disjunctive option of `A OR B OR C` (which when represented by a number would mean A + B + C on the bit mask) and later you can probe whether A is part of that option description or not.

Does this make sense? For a more detailed introduction maybe consult [1].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Defines.php#L238-L247
[1] https://abdulapopoola.com/2016/05/30/understanding-bit-masks/

PR #2815, opened 2017-11-11
External identifier to support #nowiki, refs 2808

### LINKED ISSUE #2808: #nowiki should work for Properties of type External Identifier
according to 

https://www.semantic-mediawiki.org/wiki/Help:Displaying_information#URL

it is possible to query a Property of Type url with # nowiki

this should IMHO also work with external identifiers see:

see https://www.semantic-mediawiki.org/wiki/User:WolfgangFahl/Workdocumentation_2017-11-06_WF-b

for the current different behavior.

mwjames:
This PR is made in reference to: #2808

- Adds support for `#nowiki` as printout output formatter option

This PR includes:
- [x] CI build passed

Fixes #2808

### DIFF
--- a/src/DataValues/ExternalIdentifierValue.php
+++ b/src/DataValues/ExternalIdentifierValue.php
@@ -24,7 +24,7 @@ class ExternalIdentifierValue extends StringValue {
 	/**
 	 * @var string|null
 	 */
-	private $substitutedUri = null;
+	private $uri = null;
 
 	/**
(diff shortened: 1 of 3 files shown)


kghbln:
[Documented on wiki](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AType_External_identifier&type=revision&diff=57683&oldid=57024)

Issue #2810, opened 2017-11-07
Consider re-aligning MediaWiki versions supported w/ MediaWiki support cycle


legoktm:
In the [compatibility matrix](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/COMPATIBILITY.md#platform-compatibility), it lists support as version ranges, e.g. 2.5.x supports MW 1.23-1.29. 1.23 is the floor because its LTS, but it was supported much longer than 1.24 and 1.25. In fact, 1.23 has bug fixes that those two later releases never saw.

In addition, just like we strongly encourage people to upgrade their PHP version for security reasons, we should also be doing the same for MediaWiki - and continuing to provide support for EOL versions of MediaWiki works against that goal.

My naive proposal is to take the current compatibility table and just drop support for MW versions as upstream drops support for them. But I am not familiar with the needs/requirements of SMW users or really how the SMW development cycle looks like, so I'm sure that others have much better ideas than me. But my main point is that I don't think it's a good idea to support such old versions that no longer receive security fixes.

kghbln:
SMW 2.5 was released in March and at that time MW 1.23 was still with us. Breaking compatibility within a release series does not appeal to me personally and probably to others too. MW is not doing that either for a reason.

SMW is indeed aiming to follow the release circle MW is doing and as a matter of fact the SMW 2.5 series is approaching EOL too. The release cycle is somewhere between six and nine months.

However this project is solely supported by volunteers, so it may take more time to create a new release than anticipated. We all know already from e.g. MW 1.28 that it is complex and requires effort to do a cool release even for much stronger organizations than the SMW project. SMW 3.0 was originally scheduled for the for the third quarter of 2017 but along the way unanticipated issues have to be overcome etc, which have to be dealt with. Most recently PostgresQL is bellies up all of a sudden. Moreover this will be a big release requiring even more attention. As a result more time is needed.

Deliberately breaking compatibility with MW 1.24 to 1.26 while still supporting MW 1.23 does not appeal to me either. However, I understand that in an ideal world users of these unsupported versions should migrate to supported branches. What I will do is add in more encouragement to do so in our documentation.

[...]

PR #2818, opened 2017-11-11
Provide recommendation to always use supported versions

kghbln:
This PR is made in reference to: #2810 

- Provide recommendation to always use supported versions

This PR includes:
- [ ] CI build passed


kghbln:
@mwjames Note that I went "Verified" with this commit.

PR #2822, opened 2017-11-11
Move Csv to utility, add `merge` parameter to `format=csv`

mwjames:
This PR is made in reference to: #2804

- Moves csv operations (`fputcsv` etc.) to a utility class
- Adds a `merge` option where rows and column values with an identical subject identifier are combined       
  - such as lines containing `'Foo', '1', '2', '3'` `'Foo', '4', '5', '6'` will result in `'Foo', '1,4', '2,5', '3,6'`
  - situation can normally only appear when `mainlabel=-` is used that creates a specific first column and contains identical identifier

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -69,6 +69,7 @@
 	"smw-paramdesc-table-transpose": "Display table headers vertically and results horizontally",
 	"smw-paramdesc-rdfsyntax": "The RDF syntax to be used",
 	"smw-paramdesc-csv-sep": "The separator to use",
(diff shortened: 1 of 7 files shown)


mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2822

kghbln:
[Documented on wiki.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ACSV_format&type=revision&diff=57417&oldid=24962)

Issue #874, opened 2015-03-08
Do not filter when asked to sort


alex-mashin:
# Overview

The `{{#ask:}}` query removes all results without value for a property that is specified in the "sort" parameter.
# Steps to Reproduce

For example, if I have a page P1 in category C with a property `[[has X::value]]` and another page P2 in same category C without the "has X" property. They both use the same template and same semantic form.
# Actual Results

When I write:

```
{{#ask: [[Category:C]]
  | ?has X
  | sort=has X
  | order=asc
}}
```

it displays only page P1.

But:

```
{{#ask: [[Category:C]]
  | ?has X
}}
```

displays both P1 and P2 pages.
# Expected Results

it should have displayed both P1 and P2 in both cases.

Reproduced from https://phabricator.wikimedia.org/T24583

mwjames:
See also #540

s7eph4n:
I think #540 is different. In the #540 scenario pages are excluded from the result set if they have a property, but that property points to a non-existent page. Here in the #874 scenario pages are excluded, that do not have the property at all.

And for what it's worth, this bug her (#874) is an old issue that comes up again and again but does not have a (formal) solution: You just can not sort what you do not have. In a list of cities sorted by area, where do you put those, that do not have the "Has area" property set? So IMHO it is not a bug.

mwjames:
> I think #540 is different. In the #540 scenario pages are excluded from the result set if they have a property, but that property points to a non-existent page.

This is OK but I'd like to keep them connected since both are touching the `sort` parameter.

> So IMHO it is not a bug.

:-)

alex-mashin:
I think, it *is* a bug, for all practical purposes. Pages are not supposed to be missing from query results if even some printouts are empty, and sorting property is even less important than a printout. Usually sorting is meant to bring some data closer to user's attention but not to remove other data completely.

In a SQL query, NULL is a sortable value.

In believe, absent properties should behave like zeroes and empty strings when sorted.

s7eph4n:
> In believe, absent properties should behave like zeroes and empty strings when sorted.

Please sort by birth date:
* Lucan, 39 - 65
* Ovid, 43 B.C. - 17
* Ptolemy, 90 - 168
* Theocritus
* Virgil, 70 B.C. - 19 B.C.
* Homer, ~850 B.C.

alex-mashin:
[...]

PR #2823, opened 2017-11-11
Add SMW_QSORT_UNCONDITIONAL, refs 874

mwjames:
This PR is made in reference to: #874

- Original from https://github.com/mediawiki4intranet/SemanticMediaWiki/commit/53ba5309452f651c1a671a3c816883c475f4ab5e
- Experimental setting (`$GLOBALS['smwgQSortFeatures'] = $GLOBALS['smwgQSortFeatures'] | SMW_QSORT_UNCONDITIONAL;`) that allows to apply an unconditional sort which means that sorting on something like `Has page, Has text` will work even though not all subjects contain a `Has page` or `Has text` assignment
- `SPARQLStore` currently does not support this feature and would require some refactoring of the `ConditionBuilder` to make sort statements `OPTIONAL`
- Experimental status can only be lifted by the time the `SPARQLStore` supports the same feature
- Integration test `"q-0910.json"` shows a different sorting behavior for `postgres`

Some references for SPARQL on the `LEFT OUTER JOIN` discussion:

- https://www.w3.org/2008/07/MappingRules/StemMapping (see Optionals)
- https://arxiv.org/abs/1304.7799
- https://supportcenter.cambridgesemantics.com/semantic-university/sparql-vs-sql-intro (NULLs, OPTIONALS and LEFT OUTER JOINs)

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/src/Defines.php
+++ b/src/Defines.php
@@ -272,6 +272,7 @@ define( 'SMW_CAT_HIERARCHY', 8 ); // Support for category hierarchies
 define( 'SMW_QSORT_NONE', 0 );
 define( 'SMW_QSORT', 2 ); // General sort support
 define( 'SMW_QSORT_RANDOM', 4 ); // Random sort support
+define( 'SMW_QSORT_UNCONDITIONAL', 8 ); // Unconditional sort support
 /**@}*/
 
 /**@{
--- a/src/SQLStore/QueryEngine/OrderCondition.php
+++ b/src/SQLStore/QueryEngine/OrderCondition.php
@@ -47,6 +47,11 @@ class OrderCondition {
 	private $isSupported = true;
 
 	/**
+	 * @var boolean
+	 */
+	private $asUnconditional = false;
+
+	/**
 	 * @since 2.5
 	 *
 	 * @param QuerySegmentListBuilder $querySegmentListBuilder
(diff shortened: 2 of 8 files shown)


EmperorLucifer666:
Translate

kghbln:
Documented [here](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgQSortFeatures&type=revision&diff=63112&oldid=59401) and [there.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ASorting&type=revision&diff=63115&oldid=55379)

kghbln:
[Configured on sandbox](https://sandbox.semantic-mediawiki.org/w/index.php?title=SMW_Sandbox:Setup_information&diff=9214&oldid=9213)

[Example on sandbox](https://sandbox.semantic-mediawiki.org/wiki/Issue/2823) - Currently not working due to `SMWElasticStore`

mwjames:
> - Currently not working due to `SMWElasticStore`

As experimental feature, I wanted to avoid mingling it so the
ElasticStore has its own [0] setting to circumvent strictness (aka
restriction).

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/es/DefaultSettings.php#L2013-L2016

kghbln:
> I wanted to avoid mingling it so the ElasticStore has its own

Thanks a lot for the note which makes sense to me. I changed [configuration](https://sandbox.semantic-mediawiki.org/w/index.php?title=SMW_Sandbox:Setup_information&diff=9265&oldid=9215) to have the same beheviour for both backends.

Issue #2835, opened 2017-11-15
SMW gets confused in queries with a property named "category"


krabina:
It is generally not a very good idea to have a property called "category" in a wiki. However, SMW handles it well:
When you define the property called "Category" with the value "Test" [[Category::Test]] on a page and put this page  in the [[Category:Test]], SMW handles it well as can be seen in the Browse Properties page: there is an attribute "Category" and it is automatically assigned property "Categories" that shows the list of categories a page is in.

The only problem is the ask query: 
If you put the printout statment |?Category then it will show the lists of categories the pages of the query are in and not the property category

It can be seen here: https://sandbox.semantic-mediawiki.org/wiki/Category_Attribute

kghbln:
One slightly unrelated question: You are now the third person this week creating an issue without using the issue template which should be offered during the creation of the issue. Does it show up for you and something must be fixed here? When I try to create an issue I see it so I am very much confused.

krabina:
Yes it shows up, but as I created the issue on sandbox before I entered the issue, I thought that it is not necessary to indicate the Versions of the sandbox wiki there. You guys know better about the sandbox versions then I do. 
So in my opinion this is only necessary when filing an issue in an environment different from sandbox...

kghbln:
> Yes it shows up, but as I created the issue on sandbox before I entered the issue,

Good to know. Creating an example is indeed part of the process. That step succeeded.

> I thought that it is not necessary to indicate the Versions of the sandbox wiki there.

The software gets updated daily so yes it is still necessary. If one would like to get back to figuring out what the actual software used at the time the issue was reported it will be an extra effort.

>  So in my opinion this is only necessary when filing an issue in an environment different from sandbox...

No this is the mandatory part of the template. Anyways thanks for confirming that you saw the template. I will add more text to it to make it less interpretable.

kghbln:
As for your issue: You are in direct conflict with SMW syntax on printing categories. See [the docu](https://www.semantic-mediawiki.org/wiki/Help:Displaying_information#Printing_categories). So I am afraid that you will have to migrate to a different property name.

krabina:
OK, but it does not really make sense, that the property name shown in the factbos is "Categories" and die printout not to be used is ?Category 
There should be only one name to be avoided...

kghbln:
[...]

PR #2840, opened 2017-11-25
Add reserved name list for properties, refs 2835

mwjames:
This PR is made in reference to: #2835

- Adds the `smwgPropertyReservedNameList` setting to list reserved names either as simple string or translatable representation using `smw-property-reserved-...`
- The creation isn't blocked and a reserved property can still be used but the following message will appear on the property page for a matchable name.

image

image

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1564,6 +1564,26 @@ return array(
 	##
 
 	##
+	# Reserved property names
+	#
+	# Listed names are reserved as they may interfere with Semantic MediaWiki or
+	# MediaWiki itself.
+	#
+	# Removing default names from the list is not recommended (extending the list
+	# is supported and may be used to customize use cases for an individual wiki).
+	#
+	# The list can contain simple names or identifiers that start with
+	# `smw-property-reserved-` to link to a translatable representation that
+	# matches a string in a content language.
+	#
+	# @see #2835, #2840
+	#
+	# @since 3.0
+	##
+	'smwgPropertyReservedNameList' => array( 'Category', 'smw-property-reserved-category' ),
+	##
+
+	##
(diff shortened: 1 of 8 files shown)


mwjames:
@kghbln FYI

kghbln:
Documented:
* [Property naming](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AProperty_naming&type=revision&diff=57824&oldid=57726)
* [`$smwgPropertyReservedNameList`](https://www.semantic-mediawiki.org/wiki/Help:$smwgPropertyReservedNameList) / I documented as if one always has to add the name and the respective system message in pairs to make it easier.

mwjames:
> I documented as if one always has to add the name and the respective system message in pairs.

Let me clarify:

>  "smw-property-reserved-category" is the message key that holds the warning message in namespace "MediaWiki", i.e. MediaWiki:Smw-property-reserved-category that is shown to the user if the reserved name "Category"

"smw-property-reserved-category" doesn't hold a warning message but a translatable name that is matched against the property name. For example if the content language is `ja` then Category (as canonical form) is blocked but also `カテゴリ` since this is the reserved name in the corresponding language.

So instead of listing all possible names in `$smwgPropertyReservedNameList` for any known language, `smw-property-reserved-...` is a "convention" key allowing to use the translation system to help maintain a list.

The warning message that is displayed will be the same and is represented by `smw-property-name-reserved`.

kghbln:
Ah, thanks for noting. This makes a lot of sense! [Fluffed](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgPropertyReservedNameList&type=revision&diff=57846&oldid=57826)

PR #2841, opened 2017-11-25
Replace `smwgLinksInValues` setting, refs 2802

mwjames:
This PR is made in reference to: #2802

- Replaces `smwgLinksInValues` with `smwgParserFeatures` and flag `SMW_PARSER_LINV`
- `SMW_PARSER_LINV` represents the previous `SMW_LINV_OBFU`, an equivalent for `SMW_LINV_PCRE` has not been introduced as it was only added to avoid a BC and would create OOM in some case and is therefore removed entirely 
- Disabled by default (same as it was with `smwgLinksInValues`)

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -302,33 +302,6 @@ return array(
 	##
 
 	###
-	# InText annotation to support "links in value"
-	#
-	# SMW_LINV_OBFU (2.5+)
-	#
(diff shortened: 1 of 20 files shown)


mwjames:
@kghbln FYI

PR #2844, opened 2017-11-25
Ren `#-ia` to `#-raw`, refs 2325

mwjames:
This PR is made in reference to: #2325

- Use `#-raw` instead of `#-ia` (is kept as alias) which seems more intuitive on the expected outcome where text annotations remain in its raw format

This PR includes:
- [x] CI build passed


kghbln:
[Documented.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AAsk_syntax%2FAnnotation_markup_output_formatter&type=revision&diff=62818&oldid=62774)

PR #2847, opened 2017-11-26
Use different approach to update query dependencies, refs 1117

mwjames:
This PR is made in reference to: #1117

- Rely on `OutputPageCheckLastModified` and `RejectParserCacheValue` to disable the parser cache on-demand for a requested dependency because:
  - Tracking the query and dependent subjects require `ParserCachePurgeJob` to run effectively in order to prune the parser cache so that pages with embedded `#ask` are to request a new query execution that would generate a possible new result set
  - The problem is that `ParserCachePurgeJob` relies on the `Title::invalidateCache` to push all invalidations during the page save event as deferred request (queuing the job isn't really an option since we don't know when the job is run which could be to late given that some users don't carefully monitor the job queue) which may cause a surge in DB requests due to lag or an excessive invalidation process
  - This PR tries to mitigate the issue by:
    - `ParserCachePurgeJob` adds a `ex:mode` parameter with the `EXEC_DB` to reflect the current modus operandi while the new `EXEC_JOURNAL` is to use a cache instance in connection with the `RejectParserCacheValue` hook to act as invalidation controller
    - The `EXEC_JOURNAL` mode will not try to push invalidations in `ParserCachePurgeJob` via `Title` and instead store required dependency links temporarily in cache using the `DependencyLinksUpdateJournal` so that when a specific entity (aka page) is requested for viewing the journal will check whether the page has been listed and if so will return false to reject the parser cache which in turn will allow to execute a new `#ask` request
    - Using `EXEC_JOURNAL` has the advantage that in case of successive update requests (due to other or competing query dependency updates) only the cache entry is updated leaving the DB back-end "untouched" and hereby the page table for the the journal update. Only in the event that a marked page is viewed an actual `touched`/cache invalidation takes place.

This PR includes:
- [x] CI build passed

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -627,7 +627,7 @@
 	"smw-format-datatable-toolbar-export": "Export",
 	"smw-category-invalid-redirect-target": "Category \"$1\" contains an invalid redirect target to a non-category namespace.",
 	"smw-parser-function-expensive-execution-limit": "The parser function has reached the limit for expensive executions (see configuration parameter [https://www.semantic-mediawiki.org/wiki/Help:$smwgQExpensiveExecutionLimit <code>$smwgQExpensiveExecutionLimit</code>]).",
-	"smw-postproc-queryref": "The page was marked as to be refreshed due to some required post processing.",
+	"smw-postproc-queryref": "Semantic MediaWiki will initiate a page refresh due to some required query post processing.",
 	"apihelp-smwinfo-summary": "API module to retrieve information about Semantic MediaWiki statistics and other meta information.",
 	"apihelp-ask-summary": "API module to query Semantic MediaWiki using the ask language.",
 	"apihelp-askargs-summary": "API module to query Semantic MediaWiki using the ask language as list of conditions, printouts and parameters.",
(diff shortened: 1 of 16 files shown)


mwjames:
@kghbln There is currently nothing wrong with how the query dependencies are updated but as outlined they have some drawbacks and for the mentioned reason I created this PR. We should monitor any "hinkyness" that may suddenly emerge as part of the PR in connection with self-invoked query updates due to changed dependencies. We can always go back to the current working mode (`EXEC_DB`).

In short:

- `EXEC_DB`: `ParserCachePurgeJob` → find all required dependencies → run for each of them `Title::invalidateCache`  (changes the touched date on the DB side so that the parser cache is invalidated on a view request)
- `EXEC_JOURNAL`: `ParserCachePurgeJob` → find all required dependencies → store them in a special cache journal 
  - Now, when a page is viewed → check the journal → only run `Title::invalidateCache` for the requested page

The objective is to minimize the need to run specific queries or updates on the DB. Therefore users who still run `CACHE_DB` should seriously consider their infrastructure since using `CACHE_DB` will perform updates on the DB back-end which we try to avoid.

To reiterate, the preferred cache instance for running MediaWiki with SMW is `redis` (til a better system emerges).

PR #2859, opened 2017-12-09
Use `smw-property-predefined` as predefined property msg keys

mwjames:
This PR is made in reference to: #https://github.com/SemanticMediaWiki/SemanticCite/issues/58

- As proposed in https://github.com/SemanticMediaWiki/SemanticCite/issues/58#issuecomment-349050589
- Uses `smw-property-predefined` as convention key including the conversion of `_` to `-`

This PR includes:
- [x] CI build passed


mwjames:
@kghbln @Nikerabbit I made the necessary changes therefore merge this PR at your own discretion as the when it most convenient in terms of additional key changes etc..

kghbln:
@Nikerabbit Just give me a ping the day I should merge.

Nikerabbit:
Whenever you want. I am aware of this now :)

kghbln:
> Whenever you want. I am aware of this now :)

Cool. Done now!

i18n/en.json
```diff
@@ -419,8 +419,8 @@
...
 	"action-smw-admin": "access Semantic MediaWiki administration tasks",
-	"smw-pa-property-predefined-default": "\"$1\" is a predefined property.",
-	"smw-pa-property-predefined-common": "This property is pre-deployed (also known as [https://www.semantic-mediawiki.org/wiki/Help:Special_properties special property]) and comes with additional administrative privileges but can be used just like any other [https://www.semantic-mediawiki.org/wiki/Property user-defined property].",
+	"smw-property-predefined-default": "\"$1\" is a predefined property.",
+	"smw-property-predefined-common": "This property is pre-deployed (also known as [https://www.semantic-mediawiki.org/wiki/Help:Special_properties special property]) and comes with additional administrative privileges but can be used just like any other [https://www.semantic-mediawiki.org/wiki/Property user-defined property].",
 	"smw-pa-property-predefined_ask": "\"$1\" is a predefined property that represents meta information (in form of a [https://www.semantic-mediawiki.org/wiki/Subobject subobject]) about individual queries and is provided by [https://www.semantic-mediawiki.org/wiki/Help:Special_properties Semantic MediaWiki].",
```
Nikerabbit: Btw, did you forget to rename this? I see you renamed it in German bot not in English or qqq.

mwjames: No, those needed to be named so the integration tests pass for the particular language setting that tests the condition for the mentioned key and any other renames should follow independently (here or on twn).

kghbln: I guess this is about not renaming "smw-pa-property-predefined_ask" and "smw-pa-property-predefined_mdat" in "en.json" whereas it was done in "de.json".

mwjames: The integration test I was referring to requires the `de` version (instead of the standard English one) to ensure that switching to a non `en` user language produces an expected result.

kghbln: I have not issue with that but shouldn't the message keys be identical for each language. "en" is canonical for the message keys. Probably I am not thinking into the right direction.

mwjames: Of course and there should be renamed. Those keys should be the same for any language, I just didn't want to mingle that with the PR expect for when it was necessary as in case of the integration test.

kghbln: So we need a follow up pr for en since no renames will be done if it is not in en. The renames in all other languages will then be done "automatically" with the new code coming from twn. Is this just these two or are there other key using this pa.

mwjames: > So we need a follow up pr for en 

Yes.

> the new code coming from twn. Is this just these two or are there other key using this pa.

`smw-pa-property-predefined-...` to `smw-property-predefined-...` with any following `_` being transformed to `-`

kghbln: @mwjames Authored with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2880

@Nikerabbit Thanks for noting and your patience

PR #2864, opened 2017-12-09
Switch some property alias labels

mwjames:

- Define `Property description` (with `Has property description` as alias) and `Preferred property label`  (with `Has preferred property label` as alias) as output label for those declarative properties

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

PR #2873, opened 2017-12-09
Add support for `in:`expression in #ask syntax

mwjames:

- Adds the `in:` expression for the "lazy" user and is equivalent to `~*...*` so that in case of `[[Has text::in:foo]]` the condition is transformed into `[[Has text::~*foo*]]`
- Adds `NumberValueDescriptionDeserializer` so that
  - `[[Has number::in:100]]` creates a range of `[[Has number:: [[≥0]] [[≤100]] ]]` and
  - `[[Has number::in:-100]]` is established as `[[Has number:: [[≥-100]] [[≤0]] ]]`
- Using it on a date type such as `[[Has date::in:1970]]`  will create a range of `[[Has date:: [[≥1970]] [[<<1 January 1971 00:00:00]] ]]`

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -451,7 +451,7 @@ return array(
 	#
 	# @since 1.0
 	##
-	'smwgQComparators' => '<|>|!~|!|~|≤|≥|<<|>>|~=|like:|nlike:',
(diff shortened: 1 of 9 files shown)


mwjames:
@kghbln FYI

PR #2874, opened 2017-12-10
Special:Browse to add `SMW_BROWSE_SHOW_GROUP` (property group support)

mwjames:

- `Special:Browse` can easily become crowded when plenty of properties (with values) are required to be displayed and this PR tries to counter this by allowing to define property groups
- Adds `SMW_BROWSE_SHOW_GROUP`(enabled by default) to support a simple grouping feature in `Special:Browse` by which properties that are annotated with a group category are being displayed together
- Adds declarative property `Is property group` to identify a group category
- To modify the caption of a group (e.g. `Category:Foo bar`) the convention key `smw-property-group-label-` as in `smw-property-group-label-foo-bar` can be used to modify the group caption and if necessary allows to translate the group using the standard `smw-property-group-label-foo-bar/someLang` approach
- The group system is very simple, group membership is a 1:1 (no fancy property X is member of group A and B etc.), class hierarchies are not resolved (property X is member of category B with category B being a subcategory of A where category A is marked as `Is property group` etc.)

### How does it work?

- `SMW_BROWSE_SHOW_GROUP` has to be enabled 
- Select a property `Foo` and add a `[[Category:Foo group]]`
- Add `[[Is property group::true]]` to the `Category:Foo group` page
- That's it!
- Now, when browsing a subject that contains a `Foo` property, `Special:Browse` should display a group called `Foo group` and list property `Foo` under this group

image

image

This PR includes:
- [x] CI build passed

Fixes #


mwjames:
@kghbln FYI

kghbln:
Wow, great idea. This is actually a feature that I have been missing for a while.

kghbln:
- https://www.semantic-mediawiki.org/wiki/Help:Property_group
- https://www.semantic-mediawiki.org/wiki/Help:Special_property_Is_property_group

kghbln:
@mwjames Thanks for cranking this up. I added examples to the wiki so people can see directly what this is all about.

Issue #2736, opened 2017-10-07
Remove res/images folder


mwjames:
## Target release

3.0

## Issue

I'd like to see that the `res/images` folder is removed. It contains images that should be embedded as base64 in the relevant css file.

`ResourceLoader` has good support for embedded base64 images hence making this folder go away is a step to more cleanliness and avoids issue with setting relative paths.

## Tasks

- [x] Find out which images are still used and where
  - [x] browse_icon.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2869
  - [x] browse_iconsmall.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2869
  - [x] info.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2691
  - [x] note.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2691
  - [x] rdf_flyer.10.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2869
  - [x] rdf_flyer.12.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2869
  - [x] rdf_flyer.18.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2869
  - [x] search_icon.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2869
  - [x] searchgray_iconsmall.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2869, https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2886 
  - [x] smw_button.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2885
  - [x] sort_down.gif - unused - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2900
  - [x] sort_none.gif - unused - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2900
  - [x] sort_up.gif - unused - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2900
  - [x] warning.png - https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2691
  - [x] world.png - unused -> https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2872
- [x] Convert them (see [0]) and add them to the related CSS rule

[0] https://www.base64-image.de/

[...]

PR #2885, opened 2017-12-16
Update Setup.php

kghbln:
This PR is made in reference to: #2736 and https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2376

- Changes "smw_button.png" to a data URI

This PR includes:
- [x] CI build passed


kghbln:
@mwjames I am afraid this seems to be the only way to get rid of "smw_button.png" since `$wgFooterIcons` does not support a ["class" parameter](https://stackoverflow.com/questions/2182716/is-it-possible-to-set-the-equivalent-of-a-src-attribute-of-an-img-tag-in-css).

mwjames:
I'm currently cleaning up the `Setup`, let me get back on this once I'm done.

kghbln:
This was not nuclear science but I would not have come up with this. Thanks a lot. Will kill the files tomorrow.

PR #2889, opened 2017-12-16
Make subobject sortkey distinguishable

mwjames:
This PR is made in reference to: https://sourceforge.net/p/semediawiki/mailman/message/36135782/

- The email described a scenario where the sortkey of subobjects is not idempotent. The reason for this odd behaviour lies with the sortkey being not distinguished from any other subobject on the same page. One can verify that by looking at the `smw_object_ids` table and the `smw_sortkey` field. When a sort key is indistinguishable, the SQL engine will use a row sequence [0] (or any other defined by the `ORDER BY` clause) where the sortkeys are identical which then hereby depends on the storage sequence an entry is inserted into the table.
- The behaviour that a subobject sortkey does not include any subobject identifier has been verified at least for 1.9.2 so there is no regression on behalf of the sortkey storage.
- If you specify a `@sortkey` as part of a subobject, you would get a constant order no matter the position within the page (and hereby the storage order) since the `smw_sortkey` is augmented and not derived from the "root" subject.
- The PR will amend the sortkey for when `@sortkey` is not used and `p-0456.json` contains a test case for the described scenario from the mailing-l thread to verify that a constant order is produced independent of the storage order.

[0] https://stackoverflow.com/questions/6662837/how-mysql-order-the-rows-with-same-values

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/includes/SemanticData.php
+++ b/includes/SemanticData.php
@@ -158,6 +158,11 @@ class SemanticData {
 	protected $options;
 
 	/**
+	 * @var array
+	 */
+	protected $extensionData = [];
+
+	/**
 	 * This is kept public to keep track of the depth during a recursive processing
 	 * when accessed through the SubSemanticData instance.
 	 *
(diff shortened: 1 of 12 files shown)


mwjames:
@kghbln FYI

Some queries may result in a different outcome and I will demonstrate the change on following example.

Entity | Sortkey (Before) | Sortkey (After)
------------ | -------------| -------------
Foo | Foo | Foo
Foo#_44444 | Foo | Foo# 44444
Foo#NamedSubobject | Foo | Foo#NamedSubobject

Prior this PR, all three entities would be stored with the `Foo` sortkey and would make the following query to return all three objects `[[Has page::~Foo]]` since it would match `Foo` (using the `smw_sortkey`) to all three instances.

Starting with this PR the following query `[[Has page::~Foo]]` would only return the first entity as it contains an exact match to `Foo` and with subobjects storing additional information in a sortkey (to make them distinguishable) one would have to add `*` to match instances of `~Foo*`.

PR #2891, opened 2017-12-16
Special:Ask/Browse add flex (responsive) mode to div table

mwjames:

- Uses `display: flex;` on `DIV` tables so they can be transformed to create a responsive representation of content (`Special:Ask` and `Special:Browse`)
- Adds `.smwb-factbox .group-link:before` and `.smwb-ifactbox .group-link:after` to add a group icon (`▾`)
- During the responsive mode, group sections become distinguishable by colour when switching to a one-column representation

image

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/res/smw/ext.smw.table.css
+++ b/res/smw/ext.smw.table.css
@@ -61,3 +61,18 @@
 .smw-table-body {
 	display: table-row-group;
 }
+
+/**
+ * Responsive settings
+ */
+@media screen and (max-width: 800px) {
+	.smw-table-cell {
+		display: flex;
+		flex-direction: row;
+		flex-wrap: wrap;
+	}
+
+	.smw-table-row, .smw-table-header {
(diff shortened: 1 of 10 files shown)


mwjames:
@kghbln @s7eph4n Access via a mobile (small screen) device should get easier with this.

It might be worthwhile to investigate [0] whether the `TableResultPrinter` could be transformed in a way so that behaviour becomes similar.

[0] https://css-tricks.com/accessible-simple-responsive-tables/

kghbln:
> It might be worthwhile to investigate [0] whether the TableResultPrinter could be transformed in a way so that behaviour becomes similar.
> [0] https://css-tricks.com/accessible-simple-responsive-tables/

I very much favour the [Vertical scroll approach](https://css-tricks.com/accessible-simple-responsive-tables/#article-header-id-2) described.

PR #2892, opened 2017-12-16
Tidy (Setup)

mwjames:

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/GlobalFunctions.php
+++ b/src/GlobalFunctions.php
@@ -214,20 +214,19 @@ function smwfGetLinker() {
  * @codeCoverageIgnore
  */
 function enableSemantics( $namespace = null, $complete = false ) {
-	global $smwgNamespace;
 
 	// Apparently this is required (1.28+) as the earliest possible execution
 	// point in order for settings that refer to the SMW_NS_PROPERTY namespace
 	// to be available in LocalSettings
 	NamespaceManager::initCustomNamespace( $GLOBALS );
 
-	if ( !$complete && ( $smwgNamespace !== '' ) ) {
+	if ( !$complete && ( $GLOBALS['smwgNamespace'] !== '' ) ) {
 		// The dot tells that the domain is not complete. It will be completed
 		// in the Export since we do not want to create a title object here when
 		// it is not needed in many cases.
-		$smwgNamespace = '.' . $namespace;
+		$GLOBALS['smwgNamespace'] = '.' . $namespace;
 	} else {
-		$smwgNamespace = $namespace;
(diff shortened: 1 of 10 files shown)


src/GlobalFunctions.php
```diff
@@ -214,20 +214,19 @@ function smwfGetLinker() {
...
 function enableSemantics( $namespace = null, $complete = false ) {
-	global $smwgNamespace;
 
 	// Apparently this is required (1.28+) as the earliest possible execution
 	// point in order for settings that refer to the SMW_NS_PROPERTY namespace
 	// to be available in LocalSettings
 	NamespaceManager::initCustomNamespace( $GLOBALS );
 
-	if ( !$complete && ( $smwgNamespace !== '' ) ) {
+	if ( !$complete && ( $GLOBALS['smwgNamespace'] !== '' ) ) {
```
kghbln: @mwjames This change causes:
```
PHP Notice:  Undefined index: smwgNamespace in /../w/extensions/SemanticMediaWiki/src/GlobalFunctions.php on line 223
Notice: Undefined index: smwgNamespace in /../w/extensions/SemanticMediaWiki/src/GlobalFunctions.php on line 223
```

mwjames: Interesting, just a sec.

mwjames: Quick question, what happens if we change it back to?

```
function enableSemantics( $namespace = null, $complete = false ) {
	global $smwgNamespace;

	// Apparently this is required (1.28+) as the earliest possible execution
	// point in order for settings that refer to the SMW_NS_PROPERTY namespace
	// to be available in LocalSettings
	NamespaceManager::initCustomNamespace( $GLOBALS );

	if ( !$complete && ( $smwgNamespace !== '' ) ) {
		// The dot tells that the domain is not complete. It will be completed
		// in the Export since we do not want to create a title object here when
		// it is not needed in many cases.
		$smwgNamespace = '.' . $namespace;
	} else {
		$smwgNamespace = $namespace;
	}

	$GLOBALS['smwgSemanticsEnabled'] = true;

	return true;
}
```

kghbln: Silence. At least cron does not longer fire away these notices. The wiki seems to be fine, too.

mwjames: >Silence. At least cron does no longer fire away these notices. The wiki seems to be fine, too.

In that case #2894.

kghbln: I just updated the wiki manually to assure that continuous automatic updating does not stop. Thanks for fix!

PR #2933, opened 2018-01-06
Add smwgDefaultLoggerRole

mwjames:

- Adds the `smwgDefaultLoggerRole` admin setting to filter the amount (in terms of granularity) of information being logged when either `$wgDebugLogFile` or `$wgDebugLogGroups` is enabled 
- The default role is set to `$GLOBALS['smwgDefaultLoggerRole'] = 'production'` which means that only a very limited log output is generated, to show all SMW related loggable events the `developer` role is required, the `user` role will include extra events such executions of jobs 
- Use `LoggerAwareTrait` more consistently instead of `implements LoggerAwareInterface`

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -75,6 +75,27 @@ return array(
 	'smwgDefaultStore' => "SMWSQLStore3",
 	##
 
+	##
+	# Debug logger role
+	#
+	# A role (developer, user, production) defines the detail of information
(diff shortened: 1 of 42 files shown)


mwjames:
@kghbln By default this should minimize the log spam produced by SMW with the `[smw]` identifier. The next time we require some in-depth analysis, you most likely need to change the role setting to `developer`.

PR #2936, opened 2018-01-06
Rename experimental setting, refs 2461

mwjames:

- Renames it to a more general `smwgExperimentalFeatures` setting, allowing it to be used in a broader context

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1619,17 +1619,22 @@ return array(
 	##
 
 	##
-	# Entity store specific lookup features (internal use)
+	# Experimental settings
 	#
-	# - SMW_ESTORE_NONE applies no query or schema changes
+	# Features enabled are considered stable but for any unforeseen behaviour, the
+	# feature can be disabled to return to a previous working state avoiding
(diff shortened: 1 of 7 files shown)


kghbln:
@mwjames In the past you introduced explicit flags for disabling all features to a configuration parameter like e.g. `SMW_EL_NONE`. You removed this flag for this specific configuration parameter so disabling is done with `$smwgExperimentalFeatures = '';`. Out of curiosity: Is there a reason?

mwjames:
>In the past you introduced explicit flags for disabling all features to a configuration parameter like e.g. SMW_EL_NONE. You removed this flag for this specific configuration parameter so disabling is done with $smwgExperimentalFeatures = '';. Out of curiosity: Is there a reason?

Setting it to empty (aka '') or `false` is valid since the bit operation will still work with those values. Otherwise, I rephrased the intention (different from the previous objective) as "Features enabled are considered stable but for any unforeseen behaviour ..." which means disabling all features should not be necessary.

kghbln:
Thanks for the info!

Issue #1014, opened 2015-05-13
SMW 3.0 Remove depreciated function, class task list


kghbln:
List potential functions, classes, or marked `FIXED` comments to be removed with SMW 3.0 to make use of the semver "make incompatible API changes" [0] policy.
- It is expected that SMW 3.0 no longer supports MW 1.19 to 1.26 therefore remove all code that was introduced as compatibility layer 
- #698 / It is expected that SMW 3.0 requires at least PHP 5.4 (or PHP 5.5) therefore check callbacks/Closure for potential improvement in the `use` (local vs `$this`)
## List of
- [x] 1 Remove depreciated SEMANTIC_EXTENSION_TYPE / Flag used to indicate SMW defines a semantic extension type for extension credits, now provieded by MW core. [See also](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/SemanticMediaWiki.php#L116)
- [x] 2 Remove support for the "type" namespace #399
- [x] 3 Remove `SMW_conceptCache.php`, `SMW_dumpRDF.php`, `SMW_refreshData.php`, `SMW_setup.php` see #2789
- [x] 4 Remove deprecated functions from `SMW\DIProperty` (see `SMW\PropertyRegistry`)
- [x] 5 Remove deprecated functions from `SMW\DataValueFactory` (see `SMW\DataTypeRegistry`)
- [x] 6 Remove `Special:SemanticStatistics` (see `Special:Statistics` for providing the information)
- [x] 7 `SMWDataValue::prepareValue`, see https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1061#issuecomment-145354555
- [x] 8 Remove old $wgSpecialPageGroups, see #1214
- [x] 9 Replace usages of Revision::getRawText #1371

**These are definitely not used by SMW itself**:
- [x] Remove `ResultPrinter::m_params`
- [x] Remove `ResultPrinter::textDisplayParameters`
- [x] Remove `ResultPrinter::exportFormatParameters`
- [x] Remove `SMWQueryProcessor::processFunctionParams`
- [x] Remove `SMWQueryProcessor::getResultFromFunctionParams`
- [x] Remove `SMWQueryProcessor::getResultFromQueryString`
- [x] Remove `SMWResultArray::getNextObject`
- [x] Remove `SMWJsonResultPrinter` class_alias
- 
## Notes

[0] http://semver.org/

```
[...]

PR #2948, opened 2018-01-08
Update SemanticMediaWiki.php

kghbln:
This PR is made in reference to: #1014 

- Bumps the minimum MediaWiki version required to MW 1.27

This PR includes:
- [ ] CI build passed


kghbln:
@mwjames I am sure master has probably left MW 1.23 to 1.26 far behind. Thus this would just align this file with it.

kghbln:
Ah, does not make sense right now. Should be done later together with the testing environment. Forgot about this.

Issue #2951, opened 2018-01-10
Feature request: Natural Sorting of query results 


revansx:
### Setup and configuration

- SMW version:Any/All/Current
- MW version:Any/All/Current
- PHP version:Any/All/Current
- DB system (MySQL, Blazegraph, etc.) and version: Any/All

### Feature Request ###
Hi,

> I have users entering in reference IDs as a decimal structure (WBS) as 
> follows as values to a Semantic Property of type Text:

Under the assumption (due to missing context) that an entity contains annotations like [[Has text::1.4.4.8.3.1]] ... [[Has text::1.4.4.8.3.12]] and your query looks similar to:

```
{{#ask: [[SomeProperty:WithSomeCondition]]
 |?Has text
}}
```

It produces the mentioned order using the codified procedure in [0] for how to sort "column" values to an entity. As seen in [0] it uses `SORT_LOCALE_STRING` as flag for anything that is not numeric therefore the order is as you have described it.

```
{{#ask: [[SomeProperty:WithSomeCondition]]
 |?Has text|+order=asc
}}
```

You can modify the order using [1], yet of course it will not alter the sorting behaviour in terms of the codified `SORT_LOCALE_STRING` flag.

> Is there a way (data type or otherwise) that I can sort this as 
> desired? (Attempting to pre-define it is not practical.)

I'm presume you are looking for a natural sorting (`SORT_NATURAL`) [2] with 1.4.4.8.3.1 ... 1.4.4.8.3.12. At a glance, four lines of code should be able to extend the sort behaviour (without altering the current behavior) and if you create a ticket, I'll try to comment on it.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/2.5.x/src/SQLStore/RequestOptionsProcessor.php#L240-L245
[1] https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Printout_sorting_order
[2] http://php.net/manual/en/function.sort.php

Cheers

On 1/10/18, Evans, Richard K. (GRC-H000) <richard.k.evans@nasa.gov> wrote:
> I have users entering in reference IDs as a decimal structure (WBS) as 
> follows as values to a Semantic Property of type Text:
>
> 1.4.4.8.3.1
> 1.4.4.8.3.2
> 1.4.4.8.3.3
> 1.4.4.8.3.4
> 1.4.4.8.3.5
> 1.4.4.8.3.6
> 1.4.4.8.3.7
[...]

PR #2953, opened 2018-01-13
Support printout column nat sort (n-asc, n-desc), refs 2951

mwjames:
This PR is made in reference to: #2951 

- See #2951
- Adds support for `n-asc`and `n-desc` as `|+order=` parameter to command a natural sorting of __printout__ column values

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/Query/Result/ResultFieldMatchFinder.php
+++ b/src/Query/Result/ResultFieldMatchFinder.php
@@ -170,6 +170,12 @@ class ResultFieldMatchFinder {
 				$options->limit = trim( $limit );
 			}
 
+			// Expecting a natural sort behaviour (n-asc, n-desc)?
+			if ( strpos( $order, 'n-' ) !== false ) {
+				$order = str_replace( 'n-', '', $order );
+				$options->natural = true;
+			}
+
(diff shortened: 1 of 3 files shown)


mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/2953

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AAsk_syntax%2FPrintout_sorting_order&type=revision&diff=58260&oldid=58257) on wiki.

krabina:
When I wanted to try out this feature and improve the documentation, I encountered the following behaviour:
If you want a table to have two soring columns and one of them is not a property, but the mainlabel, this works:
` |sort=,Average rainy days`
or
` |sort=Average rainy days,`

as documented here:
https://www.semantic-mediawiki.org/wiki/Help:Sorting#Sorting_the_main_column

However, if you add an order parameter
`|order=asc,desc`
or 
`|order=n-asc,n-desc`
this does not work.

As the syntax is not very intuitive, it might me even better to have somehing like 
`|sort=mainlabel,Average rainy days`
or do a 
```
|?mainlabel=City
|sort=City,Average rainy days
```
both of which do not work currently (and would probably need a new feature request).

Issue #2713, opened 2017-09-30
PHPUnit 6.x fails


mwjames:
### Setup and configuration

- SMW version: all
- PHP version: 7.1

### Issue:

Requires https://phabricator.wikimedia.org/T177132 (MediaWiki core)

#### Fatal error: Class 'PHPUnit_ ...

The first issue one encounters is that PHPUnit doesn't declare any aliases so before it starts y'd be greeted with:

* `Fatal error: Class 'PHPUnit_Framework_TestCase' not found in ...`
* `Fatal error: Class 'PHPUnit_Framework_TestResult' not found in ...`
* `Fatal error: Class 'PHPUnit_Framework_Assert' not found in ...`

Adding the following aliases removes above error messages.

```
class_alias( 'PHPUnit\Framework\TestCase', 'PHPUnit_Framework_TestCase' );
class_alias( 'PHPUnit\Framework\TestResult', 'PHPUnit_Framework_TestResult' );
class_alias( 'PHPUnit\Framework\Assert', 'PHPUnit_Framework_Assert' );
```

#### Error: Call to undefined method

The next issue is:

`Error: Call to undefined method SMW\Tests\...::setExpectedException()` [0], doing a quick `git grep -l 'setExpectedException' | xargs sed -i 's/setExpectedException/expectException/g'` solves the issue and have the tests pass as before with:

```
Using PHP 7.1.1

Semantic MediaWiki: 3.0.0-alpha (7e9a979, SMWSQLStore3, mysql)
MediaWiki:          1.30.0-alpha (38999cc, MediaWiki vendor autoloader)
Site language:      en

Execution time:     2017-09-30 09:33
Debug logs:         Disabled
Xdebug:             Disabled (or not installed)

PHPUnit 6.3.0 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.1
Configuration: D:\xampp\htdocs\mw-master-phpunit\extensions\SemanticMediaWiki\phpunit.xml.dist

.............................................................   61 / 5544 (  1%)
...
......................................................        5544 / 5544 (100%)

Tests: 5544, Assertions: 15616, Warnings: 4, Skipped: 15.

```

Unfortunately, `expectException` is unknown before PHPUnit 5.2 therefore replacing it as of now is not possible.

[0] https://github.com/sebastianbergmann/phpunit/issues/2074#issuecomment-183472385

JeroenDeDauw:
[...]

PR #2968, opened 2018-01-27
Use `Maintenance::getDB` in `setupStore.php`, refs 2963

### LINKED ISSUE #2963: setupStore.php: Use wgDBadminuser instead wgDBuser
When setting $wgDBadminuser and -password in LocalSetttings.php the setupStore.php maintenance script, should use these instead of $wgDBuser and -password, as its done by native MediaWiki maintenance scripts, c.f. https://www.mediawiki.org/wiki/Manual:$wgDBadminuser .

### Steps to reproduce
configure remove CREATE TABLE, etc. rights from the mysql user your semantic mediawiki is using, and try to install/upgrade semantic mediawiki -> Tables does not get created.

mwjames:
This PR is made in reference to: #2963

- Adds `CallbackConnectionProvider`
- Uses `Maintenance::getDB` in `setupStore.php` which is expected to address #2963 (`$wgDBadminuser` issue) but has not explicitly tested whether this resolves it or not

This PR includes:
- [x] CI build passed

Fixes #2963

### DIFF
--- a/maintenance/setupStore.php
+++ b/maintenance/setupStore.php
@@ -5,6 +5,8 @@ namespace SMW\Maintenance;
 use SMW\Store;
 use SMW\StoreFactory;
 use Onoi\MessageReporter\MessageReporterFactory;
+use SMW\Connection\ConnectionManager;
+use SMW\Connection\CallbackConnectionProvider;
 use SMW\SQLStore\Installer;
 
 $basePath = getenv( 'MW_INSTALL_PATH' ) !== false ? getenv( 'MW_INSTALL_PATH' ) : __DIR__ . '/../../..';
@@ -110,6 +112,24 @@ class SetupStore extends \Maintenance {
 			$messageReporter = MessageReporterFactory::getInstance()->newObservableMessageReporter();
 			$messageReporter->registerReporterCallback( array( $this, 'reportMessage' ) );
 		}
+
+		// #2963 Use the DB handler from the Maintenance script which may access
+		// the `$wgDBadminuser` instead of the regular `$wgDBuser`
+		$callbackConnectionProvider = new CallbackConnectionProvider( function() {
+				return $this->getDB( DB_MASTER );
+			}
+		);
+
+		$connectionManager = new ConnectionManager();
+
+		$connectionManager->registerConnectionProvider(
+			DB_MASTER,
+			$callbackConnectionProvider
+		);
+
+		$store->setConnectionManager(
+			$connectionManager
+		);
 
 		$store->setMessageReporter(
 			$messageReporter
--- a/src/Connection/CallbackConnectionProvider.php
+++ b/src/Connection/CallbackConnectionProvider.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace SMW\Connection;
+
+use Closure;
+
+/**
+ * @license GNU GPL v2+
+ * @since 3.0
+ *
+ * @author mwjames
+ */
+class CallbackConnectionProvider implements ConnectionProvider {
+
+	/**
+	 * @var callable
+	 */
+	private $callback;
+
+	/**
+	 * @var mixed
+	 */
+	private $connection;
+
+	/**
+	 * @since 3.0
+	 *
+	 * @param callable $callback
+	 */
+	public function __construct( callable $callback ) {
+		$this->callback = $callback;
+	}
+
+	/**
+	 * @since 3.0
+	 *
+	 * @return mixed
+	 */
+	public function getConnection() {
+
+		if ( $this->connection === null ) {
+			$this->connection = call_user_func( $this->callback );
+		}
+
+		return $this->connection;
+	}
+
+	/**
+	 * @since 3.0
+	 */
+	public function releaseConnection() {
+		$this->connection = null;
+	}
+
+}
--- a/tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
+++ b/tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace SMW\Tests\Connection;
+
+use SMW\Connection\CallbackConnectionProvider;
+
+/**
+ * @covers \SMW\Connection\CallbackConnectionProvider
+ * @group semantic-mediawiki
+ *
+ * @license GNU GPL v2+
+ * @since 3.0
+ *
+ * @author mwjames
+ */
+class CallbackConnectionProviderTest extends \PHPUnit_Framework_TestCase {
+
+	private $conection;
+
+	public function setUp() {
+
+		$this->connection = $this->getMockBuilder( '\stdClass' )
+			->disableOriginalConstructor()
+			->getMock();
+	}
+
+	public function testCanConstruct() {
+
+		$callback = function() {};
+
+		$this->assertInstanceOf(
+			CallbackConnectionProvider::class,
+			new CallbackConnectionProvider( $callback )
+		);
+	}
+
+	public function getConnection() {
+		return $this->connection;
+	}
+
+	public function testGetConnectionFormCallback() {
+
+		$callback = function() {
+			return $this->connection;
+		};
+
+		$instance = new CallbackConnectionProvider(
+			$callback
+		);
+
+		$this->assertEquals(
+			$this->connection,
+			$instance->getConnection()
+		);
+	}
+
+	public function testGetConnectionFormStaticCallback() {
+
+		$instance = new CallbackConnectionProvider(
+			[ $this, 'getConnection' ]
+		);
+
+		$this->assertEquals(
+			$this->connection,
+			$instance->getConnection()
+		);
+	}
+
+}


mwjames:
@kghbln This passes the test suite which means it should be safe to use whether it addresses the `$wgDBadminuser` issue or not but it would be good to confirm the assumption somehow.

maintenance/setupStore.php
```diff
@@ -111,6 +113,24 @@ public function execute() {
...
+
+		$connectionManager->registerConnectionProvider(
+			DB_MASTER,
+			$callbackConnectionProvider
+		);
+
+		$store->setConnectionManager(
+			$connectionManager
+		);
+
```
JeroenDeDauw: Looks like this should be a single function. No need to have these two new variables in scope longer, or to access the ones already in scope. Plus this function is already very long, with lots of random details (thus mixed levels of abstractions) in it.

src/Connection/CallbackConnectionProvider.php
```diff
@@ -0,0 +1,55 @@
...
+	 * @param Closure $callback
+	 */
+	public function __construct( Closure $callback ) {
+		$this->callback = $callback;
+	}
+
+	/**
+	 * @since 3.0
+	 *
+	 * @return mixed
```
JeroenDeDauw: Should this not be a connection thing?

src/Connection/CallbackConnectionProvider.php
```diff
@@ -0,0 +1,55 @@
...
+	 * @var mixed
+	 */
+	private $connection;
+
+	/**
+	 * @since 3.0
+	 *
+	 * @param Closure $callback
+	 */
+	public function __construct( Closure $callback ) {
```
JeroenDeDauw: `callable` is less restrictive and has no downsides as far as I know

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
```diff
@@ -0,0 +1,48 @@
...
+			CallbackConnectionProvider::class,
+			new CallbackConnectionProvider( $callback )
+		);
+	}
+
+	public function testGetConnection() {
+
+		$connection = $this->getMockBuilder( '\stdClass' )
+			->disableOriginalConstructor()
+			->getMock();
```
JeroenDeDauw: huh?! This is the same as `$connection = new \stdClass()` no?

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
```diff
@@ -0,0 +1,48 @@
...
+
+	public function testCanConstruct() {
+
+		$callback = function() {};
+
+		$this->assertInstanceOf(
+			CallbackConnectionProvider::class,
+			new CallbackConnectionProvider( $callback )
+		);
+	}
```
JeroenDeDauw: IMO this does not add value and is thus harmful clutter. (Yes I realize I introduced the first tests like this in the codebase, this was a mistake)

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
```diff
@@ -0,0 +1,48 @@
...
+
+	public function testGetConnection() {
```
JeroenDeDauw: This is not a descriptive name. I suggest doing something like `testGetConnectionReturnsTheCallbacksReturnValue`.

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
JeroenDeDauw: Also beware that this PHPUnit syntax is not compatible with the latest versions of PHPUnit. I don't know if it is easy to switch to using a more recent version of PHPUnit (with SMW being a MW extension and all...), though if possible, doing this sooner rather than later will save migration work.

src/Connection/CallbackConnectionProvider.php
mwjames: No, it is a generic callback therefore you type. The invoker is responsible to make the right choice.

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
mwjames: This should work in any version, or? and if not what changed?

> to switch to using a more recent version of PHPUnit (with SMW being a MW extension and all...)

2713 has the details!

mwjames:
> Also beware that you did not test most of the new code (the new test only covers arguably the simplest addition)

That is not true,it is covered by the integration test.

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
mwjames: Feel free to changed it, it works for me.

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
mwjames: It works for me!

src/Connection/CallbackConnectionProvider.php
mwjames: Yes, but nothing is wrong with `Closure` and it doesn't impose a functional impairment.

JeroenDeDauw: It does: you can only provide Closures like this. You can't pass in 'globalFunctionName', 'SomeClass::staticFunction' or $instanceOfClassThatIsCallable

src/Connection/CallbackConnectionProvider.php
JeroenDeDauw: I understand that this function cannot guarantee the type that is returned. Though I think this does not really matter for the return type. The return type is about indicating what it should return. Of course this should match what the callback is supposed to return. Unless I am misunderstanding the code, in both cases this should be a connection, whatever a connection is. Weirdly enough the ConnectionProvider interface does not mention a specific type. Is that understanding wrong?

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
JeroenDeDauw: I'm not set on having this particular bit of code look one way or the other. What I am doing here with my time is giving you some advice, which you are of course free to ignore. Naming your test methods badly indeed "will work". Naming them more descriptive will work better, both for yourself, and to a bigger degree for other people. Not writing any tests also "will work". I recall giving advice to this one specific contributor several years ago that writing tests is a good idea, and at first getting a bunch of outraged "oh, so now I need to write tests for everything?!". I was happy to see this person changed their mind, wrote a lot of tests, and now encourages other people to do the same. If they had stayed at "it works for me", that'd have been a shame.

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
JeroenDeDauw: https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2713

JeroenDeDauw: Oh... actually this might indeed *work* in the latest PHPUnit. I'm not sure as I have not used this format recently. Which is because it does not make sense to use this form there. You'd just do `$connection = $this->createMock( '\stdClass' )`

At least you would if there was reason to use the mock API. Why not just create the instance like in my first comment?

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
mwjames: Let not mix functional relevancy (which is to use a unit test to verify that what is expected can actually be asserted) with cosmetic preference which the later is about. We do agree that descriptive names may help others to understand the code in question but for a class with one single responsibility we don't have go in circles here and keep discussing a dead horse.

src/Connection/CallbackConnectionProvider.php
mwjames: Right, yes that is the only valid functional argument I have seen in this thread that should be reconsidered to allow for something like [ $this, '...' ]. [0] requires PHP 5.4 which is fine given that min req. is PHP 5.5.

[0] http://php.net/manual/de/language.types.callable.php

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
mwjames: > harmful clutter

Disagree, this allows to check whether the constructor does work or not.

mwjames:
@kghbln Would be nice if you could test this with the `wgDBadminuser` user thing as I'd like to finish this up.

tests/phpunit/Unit/Connection/CallbackConnectionProviderTest.php
JeroenDeDauw: So you are saying that in classes with a single responsibility good naming is not important as long as the classes do their job? And also that I am wasting my time even bringing this up?

JeroenDeDauw:
> That is not true,it is covered by the integration test.

What integration test? One that was already there? That verify the correctness of the new code if it was already there and did not fail before the new code was added. In such a case you could refactor the code, accidentally break the new behavior, and still have the test passing.

kghbln:
> Would be nice if you could test this with the wgDBadminuser user thing as I'd like to finish this up.

I will switch sandbox today. Will it be possible to back-port this to 2.5.x once I confirmed it is working on master?

kghbln:
@mwjames The field test just failed. "setupStore.php" still uses user and password defined with `$wgDBuser` and `$wgDBpassword` respectively. `$wgDBadminuser` and `$wgDBadminpassword` also defined is being ignored.

mwjames:
> The field test just failed. "setupStore.php" still uses user and password defined

Well. after browsing the `Maintenance` class it became clear that we need to set `\Maintenance::DB_ADMIN` to force a preference for the `admin` related user.

kghbln:
> Well. after browsing the `Maintenance` class it became clear that we need to set `\Maintenance::DB_ADMIN` to force a preference for the `admin` related user.

No worries and thanks for the fix. I will do another field test after the next update on sandbox.

kghbln:
> I will do another field test after the next update on sandbox.

Works perfect now. Thanks a lot!

PR #2986, opened 2018-02-03
Use html format for _eid types in a reference, refs 2985

### LINKED ISSUE #2985: Reference Datatype displayed strangely
### Setup and configuration

https://www.semantic-mediawiki.org/wiki/Queen_Victoria

### Issue
I tried to create an example Property with type reference as:
https://www.semantic-mediawiki.org/wiki/Property:Has_Wikidata_Reference

and tried to assign
`[[Has Wikidata Reference::https://en.wikipedia.org/wiki/Queen_Victoria;Q9439]]
`

which is displayed as
qv

which looks strange to me since it has visible html code elements.

mwjames:
This PR is made in reference to: #2985

- When building a reference list, formatting of values need to be cared for to ensure that raw wikitext  (either as `[ ...]` or `[[ ... ]]` and in which case of the `_eid` type was `[https://www.wikidata.org/entity/...]` caused a double parse and ultimately broke the link representation) is transformed before embedded in the outer content representation 

This PR includes:
- [x] CI build passed

Fixes #2985

### DIFF
--- a/src/DataValues/ValueFormatters/ReferenceValueFormatter.php
+++ b/src/DataValues/ValueFormatters/ReferenceValueFormatter.php
@@ -5,6 +5,7 @@ namespace SMW\DataValues\ValueFormatters;
 use RuntimeException;
 use SMW\DataValueFactory;
 use SMW\DataValues\ReferenceValue;
+use SMW\DataValues\ExternalIdentifierValue;
 use SMW\DIProperty;
 use SMW\DIWikiPage;
 use SMW\Message;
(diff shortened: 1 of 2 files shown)


kghbln:
This could probably be backported?

mwjames:
> This could probably be backported?

The change is only one line and the test should work on 2.5.+ as well, so if you think it is worth it, feel free.

kghbln:
Back-ported with https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/9a8d1c8667f325739ce3da8478079a8166ca875e.

PR #2991, opened 2018-02-07
Update en.json

### LINKED ISSUE #2989: Special:Ask UX issues
I noticed the cool enticements to Special:Ask coming in 3.x when reading https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1365#issuecomment-362813504

On playing with it at https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Requ%C3%AAter I found some room to make it even moar awesome from a UX perspective:

* The `Empty` link has unclear behavior. I was not sure what it would do until I clicked it. Using the word "Clear" would be more... clear. "Clear all" or "Clear page" would be even clearer.

* The cheat sheet is displayed in a box that does not clock when you click outside of it, as is convention on the web. You need to actually move the cursor to the x at the top right and click it.

* It is not that obvious how to expand the parameter list or that one can do this to begin with. Having a "show more" link underneath the shortened list would help with that.

kghbln:
This PR is made in reference to: #2989 

- Tweak some system messages to be more intuitive

This PR includes:
- [x] CI build passed

Fixes #2989


JeroenDeDauw:
+1

PR #3001, opened 2018-02-11
Translated all propertyLabels to Spanish

ivanhercaz:
- Translation of the propertyLabels to Spanish

This PR includes:
- [ ] CI build passed

Note that I just not translate what it wasn't translated, I also fix two properties: `_NEWP` and `_PVUC`. I don't know if it could cause some problem in the current SMW installations that use it, so if there isn't a solution for that, I can undo both changes, although `_PVUC` I have been translating in TranslateWiki as "singularidad" and not "unicidad", I can change it there too.

Regards,
Iván

### DIFF
--- a/i18n/extra/es.json
+++ b/i18n/extra/es.json
@@ -75,6 +75,8 @@
         "_PPGR": "Es un grupo de propiedad"
     },
     "propertyAliases": {
+        "Es página nueva": "_NEWP",
+        "Tiene restricción de unicidad": "_PVUC",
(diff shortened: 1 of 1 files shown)


kghbln:
You could add the old versions to "propertyAliases"

ivanhercaz:
@kghbln, I added the propertyAliases. Is everything fine?

kghbln:
@ivanhercaz Thanks a ton!

PR #3017, opened 2018-02-17
Add `smwgCompactLinkSupport`

mwjames:

- Adds `$smwgCompactLinkSupport` setting (default: `true`) to generate "compact" (aka base64) links for pages like `Special:Ask`, `Special:Browse`, and `Special:SearchByProperty`
- Links will be similar to something like `Special:Browse/cl:OkV4YW1wbGUtMkZTMDAyNA` and depending on the size of the query string (especially for `Special:Ask` links) some space saving can be considered but is not the main objective
  - `Special:Ask/-5B-5BProperty:%2B-5D-5D-20-5B-5BCategory:Lorem-20ipsum-5D-5D/-3FHas-20description%3DDescription/-3FHas-20type/mainlabel%3D/limit%3D50/offset%3D50/format%3Dtable/class%3Ddatatable`
  -  `Special:Ask/cl:YzpFirEKwzAMRD-oGIcWLwEvaSgdOvQXlEQuAjsykjLk75u6Q-CGd-_OhcGF4S1cUWzvLy6Mv1w714Y7GH5Y9v7FguXQVHUr_5N3t8cT9JAL6ixUjXiN48nnwfaKvgCtGSbM0WcqZDF0nlNSbJRYClg0mDL6OYNqXMCg9S8`

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -303,6 +303,29 @@ return array(
 	##
 
 	###
+	# Compact infolink support
+	#
+	# Special:Browse, Special:Ask, and Special:SearchByProperty links can contain
+	# arbitrary text elements and therefore become difficult to transfer when its
+	# length exceeds a certain character length.
+	#
+	# A compact link will be encoded and compressed to ensure that it can be handled
+	# more easily when referring to it as an URL representation.
+	#
+	# It is not expected to be used as a short-url service, yet in some instances
+	# the generate URL can be comparatively shorter than the plain URL.
+	#
+	# The generated link has no security relevance therefore is is not
+	# cryptographically hashed or secure and should not be seen as such, it is
+	# foremost to "compact" an URL address.
+	#
+	# @since 3.0
+	# @default true
+	##
+	'smwgCompactLinkSupport' => true,
+	##
+
+	###
 	#
 	# - SMW_CAT_NONE
 	#
--- a/includes/SMW_Infolink.php
+++ b/includes/SMW_Infolink.php
@@ -1,19 +1,16 @@
 <?php
-/**
- * This file contains the SMWInfolink class.
- *
- * @author Markus Krötzsch
- * @author Jeroen De Dauw
- *
- * @ingroup SMW
- */
 
 /**
  * This class mainly is a container to store URLs for the factbox in a
  * clean way. The class provides methods for creating source code for
  * realising them in wiki or html contexts.
  *
- * @ingroup SMW
+ * @license GNU GPL v2+
+ * @since 1.0
+ *
+ * @author Markus Krötzsch
+ * @author Jeroen De Dauw
+ * @author mwjames
  */
 class SMWInfolink {
 
@@ -66,6 +63,11 @@ class SMWInfolink {
 	private $isRestricted = false;
 
 	/**
+	 * @var boolean
+	 */
+	private $isCompactLink = false;
+
+	/**
 	 * Create a new link to some internal page or to some external URL.
 	 *
 	 * @param boolean $internal Indicates whether $target is a page name (true) or URL (false).
@@ -80,6 +82,7 @@ class SMWInfolink {
 		$this->mTarget = $target;
 		$this->mStyle = $style;
 		$this->mParams = $params;
+		$this->setCompactLink( $GLOBALS['smwgCompactLinkSupport'] );
 	}
 
 	/**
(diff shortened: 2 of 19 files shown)


krabina:
I don't know it this is a good place to ask a question. The compact link support is a great feature, but I don't quite get some parts of it.
According to [1], the limitation of page titles is 255 bytes. As far as I understand, our "page titles" are more calls to a special page (no page with this title ever gets created), but I gess MW still sees it this way. This is reflected in the message I get, when I try a long ask query: "Invalid title". In the following message it indicates that a page title can only be 512 bytes long. I did not find reverences where 512 bytes comes from (there seems to be no configuratio option, also according to [1]. 

So the first problem is that I don't know how to influence that.

Second problem: different queries get handled differently
A) a short query works and gets sent get sent to
.../Spezial:Semantische_Suche/cl:YzotTzEO...
B) a medium sized query looks differently and gets sent to an empty Special:Ask (without the query being performed):
.../index.php?title=Spezial:Ask&cl:YzqdVUFOwzAQfA...
C) a very long query leads to a wiki page with the above message regarding page title length

I tried to recreate this on sandbox, but was only able to do this for A) and B): https://sandbox.semantic-mediawiki.org/wiki/Issue/3017
For C), you can see it here: http://stadtinzahlen.gemeindedaten.at/Spezial:Semantische_Suche/cl:YzqNk8FuwjAMhp9lD9C1AvUyqYchBgeGhjSJw26mddqI1EGJO6Q9yJ5gj7EbLzYHJlEWBkhVpPr_Yv9O4iQfJfloBoy1dRoflui2YLijOhlkHVWyLqzRrNdJPpYvTYaT6ewtGQyTwf5n96kUEhtdNsKOsNJInpHxgNAVJps46JBuhOe7LyJ0PfpUh87XWBuNFTq-lYotXMZjEwuBtSWP7UaJWF8V45JnqbjSI7Xo1nI7YXN2vCy_QectgemxU2xRk5gG9lto-tJo9-2khOjSZKg0RV82Bsumf2pL_YGrP-RFMW7rLBW39cpQsfg8G4qz9rQ4V6_t_6JxxlP52rkfJkJOLDyPF1cDYYDmttJKl8BykSJUMlQh_Aw-JJU3xTa0r33agiYDKzRFmCKjW81FnqVWKY9cZKmyrgUuSv-eNghiyxcbI3vSChV0hosnL4m2nav2fsjuZ2UtLQRHYykcwugUeM93qdIGCVosjg_md6jvpcQP

So I guess this could be a bug after all and also a question how to setup a wiki where C) doe not happen...

[1] https://www.mediawiki.org/wiki/Page_title_size_limitations

krabina:
btw, when I set $smwgCompactLinkSupport = false;
then even the longer queries in my wiki work. Seems to be a bug I guess.

mwjames:
Hi,

> So I guess this could be a bug after all and also a question how to setup a
> wiki where C) doe not happen...

My advice, if encountered an issue with the `smwgCompactLinkSupport`
feature, just disable it and you will have the behaviour as it was on
2.x.

If I ever have the time then I may revisit this feature because I know
the MW parser messes with the string (e.g. changing `__` to `_`) and
therefore can make those links loose information which ultimately
means we are unable to decode the string and result in a blank page.

Cheers

kghbln:
I just had this on a wiki. To me it appeared that the number of printout matters for this to work, i.e. it was broken with 27 printouts and after removing some to 24 it worked. In both cases `$smwgQMaxSize = 32;` was set. Anyways, I really like the idea behind it and getting it to work in all cases at one stage will be cool.

mwjames:
> I just had this on a wiki. To me it appeared that the number of printout
> matters for this to work, i.e. it was broken with 27 printouts and after
> removing some to 24 it worked. In both cases `$smwgQMaxSize = 32;` was set.
> Anyways, I really like the idea behind it and getting it to work in all
> cases at one stage will be cool.

Given the flakiness of this feature due to factors currently
undetermined (and I'm not planning to investigate the issue) we should
disable it (setting it to `false`) hereby restoring the 2.5.x
behaviour.

kghbln:
> Given the flakiness of this feature due to factors currently undetermined (and I'm not planning to investigate the issue) we should disable it (setting it to `false`) hereby restoring the 2.5.x behaviour.

Quite sad, but I concur. I created a separate issue #3855 to track this.

kghbln:
Documented:
* [here](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgCompactLinkSupport&type=revision&diff=69221&oldid=59119)
* [there](https://www.semantic-mediawiki.org/wiki/Help:Compact_links)

PR #3019, opened 2018-02-18
Add SMW_NS_RULE ns, refs 2273

mwjames:
This PR is made in reference to: #2273

This PR addresses or contains:

- This is a minimal framework that allows to define a rule set and rule interpreter (out of scope for this PR). For now it has been stripped form its original implementation to only contain a bare bone to manage a rule type, its definition, and representation (as I require a structureable content type for a different PR and using the MediaWiki NS was a non starter).
- A rule page can only be edited by someone with the `smw-ruleedit` right (smwcurator role has the right)
- Rules types are registered with `$smwgRuleTypes`
- I consider this experimental

A rule definition and its content page will look like

image

If the content violates the schema then the user will be notified with something that appears as

image

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

Fixes #

PR #3020, opened 2018-02-18
Add keyword (`_keyw`) type

mwjames:

- As the name suggests this PR adds a keyword type that is different from the standard text type by:
  - being restricted in length (150 for ASCII, 85 for UTF-8 to ensure that it fits into the 255 char `o_hash` field)
  - is normalized (lowercased, removed diacritics)
  - the original annotation value is kept but during a query request the normalized value is used to match the broadest possible result set which means whether you use `teSt`, `TEST`, or `test` the normalized version will always match the keyword `test` independent of any `SMW_FIELDT_CHAR_NOCASE` or the full-text index settings
  - allows to add a formatter rule (#3019, `[[Formatter rule:...]]`) which will self link any keyword to a `link_to` target (which can either be a `Special:SearchByProperty` or `Special:Ask`) to have immediate access to entities that have the same normalized keyword annotated
- The `Keyword` as property has been added to `smwgDataTypePropertyExemptionList` so that `Property:Keyword` is not automatically occupied and doesn't override any existing user properties with the same name

This PR includes:
- [x] CI build passed

## Why?

Users apply the `#lc` parser function to normalize texts and in a later phase format those via a template to help generate query links. Both functions `#lc` and templates can lengthen the parser time of a page therefore the keyword type and hereby any property defined as such should help minimize the work related to it.

## Formatter rule

Any keyword typed property can have its own formatter rule assigned which defines how links should be generated where #3017 was added to allow those links to become compact and easier to handle from a user point of view.

Link to `Special:Ask` with format `list`
```
{
    "description": "Specifies a formatting rule for a keyword type",
    "type": "LINK_FORMAT_RULE",
    "rule": {
        "link_to": "Special:Ask",
        "parameters": {
            "format": "list"
        }
    },
    "tags": [
        "formatter",
        "link formatter",
        "keyword"
    ]
}
```
Link to `Special:Ask` with format `table` and show `Has description` as printout.
```
{
    "description": "Specifies a formatting rule for a keyword type",
    "type": "LINK_FORMAT_RULE",
    "rule": {
        "link_to": "Special:Ask",
        "parameters": {
            "format": "table",
            "printouts": [
                "Has description"
            ]
        }
    },
    "tags": [
        "formatter",
        "link formatter",
        "keyword"
    ]
}
```
Link to `Special:SearchByProperty` with no parameters required (the link definition is clear by just being a property and a value).
```
{
    "description": "Specifies a formatting rule for a keyword type",
    "type": "LINK_FORMAT_RULE",
    "rule": {
        "link_to": "Special:SearchByProperty"
    },
    "tags": [
        "formatter",
        "link formatter",
        "keyword"
    ]
}
```

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -798,7 +798,8 @@ return array(
 	##
 	'smwgDataTypePropertyExemptionList' => array(
 		'Record',
-		'Reference'
+		'Reference',
+		'Keyword'
 	),
 	##
 
--- a/data/schema/rule/link-format-rule-schema.v1.json
+++ b/data/schema/rule/link-format-rule-schema.v1.json
@@ -71,9 +71,6 @@
                 "default": "",
                 "examples": [
                   "Has description"
-                ],
-                "enum": [
-                  "Has description"
                 ]
               }
             }
@@ -89,14 +86,15 @@
               "enum": ["Special:Ask"]
             }
           },
-          "required": ["parameters"]
+          "required": [ "parameters" ]
         },
         {
           "properties": {
             "link_to": {
-              "enum": ["Special:SearchByProperty"]
+              "enum": ["Special:SearchByProperty" ]
             }
-          }
+          },
+          "additionalProperties": false
         }
       ]
     },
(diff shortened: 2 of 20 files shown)


mwjames:
@kghbln This is where the minimal framework of #3017 comes into play without being a strict production rule but here `LINK_FORMAT_RULE` defines what sort of rules need to be enacted. Using JSON allows to define structured elements with a schema being able to enforce elements and types.

Of course, if the property has no formatter rule then it just displays the annotated value.

I needed this type for a better performance on the upcoming feature in order to decide when to use a term instead of a phrase_search with a keyword type always being normalized (it just means I can use a not_analyzed field without having to care about how the query value is formatted).

JeroenDeDauw:
Cool

krabina:
note: in the examples above, `LINK_FORMAT_RULE` must be changed to  `LINK_FORMAT_SCHEMA`

Issue #3022, opened 2018-02-23
Error when you search a rule and an internal error when you access via URL


ivanhercaz:
### Setup and configuration

- SMW version: 3.0.0-alpha
- MW version: 1.30.0
- PHP version: MariaDB
- DB system (MySQL, Blazegraph, etc.) and version: MariaDB 10.1.26-MariaDB-0+deb9u1

### Issue
I installed a new MediaWiki 1.30.0 with git to have it for tests of Semantic MediaWiki and another extensions. Then I made a clone of my SMW's fork, merged with the last version of this repo. When I activated SMW, it didn't works with my clone, so I changed the name of the directory, then I made a build of my fork and finally I move it to the MW's extension directory. It works fine!

If I try to create properties, concepts, whatever, I can. But if I try to create rules I can't. I tested it before to make the translations of the namespace (#3023) and then, but I get the same errors:

1. If I search one rule and it does not exist, it does not ask me if I want to create the rule. But it does not report any error.
![it does not ask me if I want to create the rule](https://user-images.githubusercontent.com/8690921/36603424-9ea4966c-18b2-11e8-9153-954a7c7d09a1.png)

2. At this situation I thought to try to create writting directly the rule in the URL, like `http://localhost/core/index.php/Rule:Test`. But it reports an internal error.

![internal error](https://user-images.githubusercontent.com/8690921/36603758-bd3baf4c-18b3-11e8-94df-149849562c39.png)

### Stack trace

```
[6f264221c56d6dc79ab01d9c] /core/index.php/Regla:Test Error from line 135 of /home/ivanhercaz/Apps/www/core/extensions/SemanticMediaWiki/src/Page/RulePage.php: Call to a member function get() on null

Backtrace:

#0 /home/ivanhercaz/Apps/www/core/extensions/SemanticMediaWiki/src/Page/Page.php(107): SMW\Page\RulePage->afterHtml()
#1 /home/ivanhercaz/Apps/www/core/includes/actions/ViewAction.php(68): SMW\Page\Page->view()
#2 /home/ivanhercaz/Apps/www/core/includes/MediaWiki.php(499): ViewAction->show()
#3 /home/ivanhercaz/Apps/www/core/includes/MediaWiki.php(293): MediaWiki->performAction(SMW\Page\RulePage, Title)
#4 /home/ivanhercaz/Apps/www/core/includes/MediaWiki.php(851): MediaWiki->performRequest()
[...]

PR #3023, opened 2018-02-23
adding Spanish translation for new namespaces and properties

ivanhercaz:

- Translation of new properties for the rule scheme
- Translation of the new namespace (Rule) for the rule scheme

This PR includes:
- [x] CI build passed

----
I translated the new strings in `i18n/extra/es.json` but I could not tested it because I can not create a rule, check #3022 to know the reason.

### DIFF
--- a/i18n/extra/es.json
+++ b/i18n/extra/es.json
@@ -72,7 +72,11 @@
         "_PPLB": "Etiqueta de propiedad preferida",
         "_EDIP": "Está protegida de edición",
         "_CHGPRO": "Propagación de cambio",
-        "_PPGR": "Es un grupo de propiedad"
+        "_PPGR": "Es un grupo de propiedad",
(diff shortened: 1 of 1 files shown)


ivanhercaz:
Well, I can not test if it works fine at all, but when I searched for a rule with the word `Regla:*` and it appears in the advanced search box, like in the screenshot below:
searching for a rule

mwjames:
>> Well, I can not test if it works fine at all, but when I searched for a rule with the word Regla:* and it appears in the advanced search box, like in the screenshot below:

Only users with the `smw-ruleedit` right can create a rule page with
`Special:Search` checking whether s(he) can actually create such page
and shows a link (red or blue) if it meets that constraint.

ivanhercaz:
@mwjames, I add another comment with more details in #3022. Thank you for you help and the merge.

PR #3024, opened 2018-02-24
Add `templatefile` result printer

mwjames:
This PR is made in reference to: https://www.semantic-mediawiki.org/wiki/Help:BEACON

- Adds `TemplateFileExportPrinter`  (`format=templatefile`)
- BEACON integration test based on the example 

This PR includes:
- [x] CI build passed

## Why?
SMW provides different specialized export formats (json, csv etc.) to accommodate various user needs and requirements but it is not always strategically meaningful (looking at the implementation cost, maintenance cost, expected audience etc.) to implement all sorts of export formats that the wide internet, some companies, or selected institutions do promote.  To serve those that require a custom format or want to pull data in a more exotic output form, 3.0 will provide a `format=templatefile` that enables users to define templates (incl. `introtemplate`, `outrotemplate`) and hereby its form on what needs where without burden SMW about specific export requirements.

## Example

https://sandbox.semantic-mediawiki.org/wiki/Issue/3024

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -615,7 +615,9 @@ return array(
 	'smwgQConceptCacheLifetime' => 24 * 60,
 	##
 
-	### Predefined result formats for queries
+	##
+	# Predefined result formats for queries
+	#
 	# Array of available formats for formatting queries. Can be redefined in
 	# the settings to disallow certain formats or to register extension formats.
 	# To disable a format, do "unset($smwgResultFormats['template'])," Disabled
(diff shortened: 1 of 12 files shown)


jaideraf:
Just noting that the name `templatefile` is also a parameter for [Word format](https://www.semantic-mediawiki.org/wiki/Help:Word_format).

kghbln:
> Just noting that the name templatefile is also a parameter for Word format.

I guess this should not conflict since one is a parameter to a result format whereas the other is a result format.

Format to be documented at: https://www.semantic-mediawiki.org/wiki/Help:Templatefile_format

krabina:
https://www.semantic-mediawiki.org/wiki/Help:Templatefile_format

PR #3049, opened 2018-03-03
Fix concept selection, refs 3043

### LINKED ISSUE #3043: "Special:Concepts" only list a small proportion of existing concepts
### Setup and configuration

- MediaWiki | 1.27.4 (317104c)23:34, 7 February 2018
- PHP | 7.0.25-0ubuntu0.16.04.1 (apache2handler)
- MySQL | 5.7.21-0ubuntu0.16.04.1

### Issue
"Special:Concepts" only list a small proportion of existing concepts starting with the first concepts by name.

### Example
["Special:Concepts"](https://www.semantic-mediawiki.org/wiki/Special:Concepts) in comparison to ["Special:AllPages"](https://www.semantic-mediawiki.org/wiki/Special:AllPages?from=&to=&namespace=108)

I do not think that this is a caching issue since "LocalSetting.php" was touched and the browser cache cleared. Perhaps something related to misermode?

mwjames:
This PR is made in reference to: #3043 

- To avoid an issue with possible selected subobjects as part of a `NamespaceDescription( SMW_NS_CONCEPT )`, directly query the `ID_TABLE`to to fetch entities that match `SMW_NS_CONCEPT`

This PR includes:
- [x] CI build passed

Fixes #3043

### DIFF
--- a/includes/specials/SpecialConcepts.php
+++ b/includes/specials/SpecialConcepts.php
@@ -3,7 +3,8 @@
 namespace SMW;
 
 use Html;
-use SMW\Query\Language\NamespaceDescription;
+use SMW\SQLStore\SQLStore;
+use SMW\Page\ListPager;
 use SMWPageLister;
 
 /**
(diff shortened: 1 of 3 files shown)


kghbln:
@mwjames May this be back-ported to 2.5.x once it was merged?

mwjames:
> @mwjames May this be back-ported to 2.5.x once it was merged?

No, because I reuse a class from 3.0 to fix the offset/limit display issue and I wasn't so keen on spending a lot of time on making it not depend on 3.0.

kghbln:
> No, because I reuse a class from 3.0 to fix the offset/limit display issue and I wasn't so keen on spending a lot of time on making it not depend on 3.0.

Thanks for the info. No worries since the fix is indeed not decisive but rather convenient for operations.

includes/specials/SpecialConcepts.php
```diff
@@ -42,27 +43,40 @@ public function __construct() {
...
-			if ( $subject->getSubobjectName() !== '' ) {
-				unset( $results[$key] );
-			}
-
-			// Avoid concepts that are mentioned in a query but doesn't
-			// exist as real page
-			if ( !$subject->getTitle()->exists() ) {
-				unset( $results[$key] );
-			}
+	public function getResults( $limit, $offset, $from, $until ) {
```
JeroenDeDauw: $from and $until are no longer used. This is odd. Is that intended?

includes/specials/SpecialConcepts.php
```diff
@@ -42,27 +43,40 @@ public function __construct() {
...
+			__METHOD__,
+			$options
+		);
+
+		foreach ( $res as $row ) {
+			$results[] = new DIWikiPage( $row->smw_title, SMW_NS_CONCEPT );
 		}
 
-		return array_values( $results );
+		return $results;
```
JeroenDeDauw: I'm a big fan of minimizing scope and mutability. In this case `$results` is in scope for quite some time. You could move it down a bit. It can also be avoided by just doing a map rather than the imperative form. Even if maps in PHP are not very elegant, still better than the foreach loop IMO.

```php
return array_map(
    function( SuchType $row ) {
        return new DIWikiPage( $row->smw_title, SMW_NS_CONCEPT );
    },
    $res
);
```

$res and $row are also not super names. Surely they represent things in the SMW domain?

JeroenDeDauw: Scope comment also applies to getting the `$connection`. I'd inline that and perhaps extract part of the query stuff into its own method.

mwjames: > $res and $row are also not super names. Surely they represent things in the SMW domain?

$res and $row are terms in connection with a DB result and while those names are debatable there are fairly common as references in MW [0].

> I'm a big fan of minimizing scope and mutability. In this case $results is in scope for quite some time.

We agree on that.

> return array_map(

This will not work because `$res` is an iterator which cannot be used directly with `array_map` [1, 2]. Of course, you could use `iterator_to_array` but then we resolve the iterator beforehand and that brings no benefit in terms of memory consumption compared to a `foreach` approach.

[0] https://github.com/wikimedia/mediawiki/search?utf8=%E2%9C%93&q=res&type=
[1] http://php.net/manual/en/function.array-map.php
[2] https://stackoverflow.com/questions/16273233/array-map-on-collection-with-array-interfaces

JeroenDeDauw: Oh did not realize this thing was an interator. And of course PHP sucks balls when it comes to iterators :\

> $res and $row are terms in connection with a DB result and while those names are debatable there are fairly common as references in MW [0].

MW really is not the place to look for how to write clean code :D

Issue #586, opened 2014-10-23
Support for NoSQL as graph database


mwjames:
The significance of SMW as product and software to deliver on its promise to "turn your wiki into a powerful and flexible knowledge management system" also depends on its ability to support different technical platforms to store and query statements made by its users through usage of various annotation approaches.

Currently, `SQLStore` and `SPARQLStore` are supported but what about using other technical means to store graph like data.
- [Cassandra vs MongoDB vs CouchDB vs Redis vs Riak vs HBase vs Couchbase vs OrientDB vs Aerospike vs Neo4j vs Hypertable vs ElasticSearch vs Accumulo vs VoltDB vs Scalaris comparison](http://kkovacs.eu/cassandra-vs-mongodb-vs-couchdb-vs-redis)
- [Introduction to Graph Databases](http://www.slideshare.net/maxdemarzi/introduction-to-graph-databases-12735789)
- [Using MongoDB as a high performance graph database](http://www.slideshare.net/robotrobot/mongodb-uk)
- [AllegroGraphto query MongoDB database via SPARQL 1.1 query engine](http://franz.com/agraph/support/documentation/current/mongo-interface.html#header2-47)
- [neo4j](http://neo4j.com/company/)
## Questions
- Would SMW need to support NoSQL as database platform to solve some of the shortcomings encountered with the `SQLStore`?
- What would be the benefit of supporting NoSQL?
- What would be the disadvantages of supporting NoSQL as graph database?
- Should SMW support text search feature (using ElasticSearch) before thinking about other Store interfaces?

WolfgangFahl:
For educational purposes I created an open source barebones triplestore implementation which might be helpful for discussion what a common Java API for backend triplestores could be. 
see https://github.com/BITPlan/org.sidif.triplestore
I intend to have Neo4J and Apache Jena Backends attached so targetting this java approach would open up the way to quite a few other backends.

Issue #2762, opened 2017-10-14
Provide and update documentation


mwjames:
## Target release

<s>3.0</s> Ongoing task

## Issue

Making this now a hard requirement for a release. As to why, see https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1365#issuecomment-334902073 and I'm explicitly expect @kghbln __NOT__ to pick this task.

This is a non-developer task, so anyone can contribute and help the project in moving forward so that by the time of a release the documentation is up-to-date.

Unrelated to the task itself but as reference on the documentation issue:

- https://github.com/SemanticMediaWiki/semantic-mediawiki.org/issues/33
- https://github.com/SemanticMediaWiki/semantic-mediawiki.org/issues/38

## Tasks

- [x] Check the [release notes](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/RELEASE-NOTES.md) for links to the `smw@org` wiki and provide necessary template work and place holders to avoid "This page doesn't exist" message
- [ ] Check the [wikidocu missing](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pulls?q=is%3Apr+is%3Aclosed+label%3A%22wikidocu+missing%22) task and fix require documentation

PS: I'm happy to answer specific questions to an unclear wording, or extend on the description of a specific PR.

kghbln:
I am doing a tour de force lately but I am still under the impression of loosing ground when it comes to documenting stuff.

jaideraf:
@kghbln in your previous message, I think you mean [this page for "wikidocu missing"](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues?utf8=%E2%9C%93&q=label%3A%22wikidocu+missing%22+).

mwjames:
My recent attempts with the documentation was to make sections shorter, provide more dedicated subpages, and add visual aids, to help users quickly grasp the "What it is about" and isn't overwhelmed by the sheer amount of text that has been the standard for the wiki for years.

[0] is such typical page (no offense to the various editors) where you have to scroll back and forth to catch what this page tries to convey mixing text and examples making it fairly difficult to see what the system has to offer and "How to" transfer the knowledge.

There should be common theme visible among the recent updates to:

- https://www.semantic-mediawiki.org/wiki/Help:Missing_redirect_annotations
- https://www.semantic-mediawiki.org/wiki/Help:Property_group
[...]

Issue #2274, opened 2017-02-18
Support for ElasticSearch


mwjames:
There are several implementations [0, 1] that use SOLR as basis for a better retrieval system in combination with SMW but maybe it is time that SMW-core provides a lean interface that allows to replicate its data into `ElasticSearch` more easily and may build a wrapper around ES DSL [2] to seamlessly integrate searches from and to `ElasticSearch`.

## Questions

* Would it make sense for SMW to support `ElasticSearch`?
* Any reason for not supporting `ElasticSearch` natively?
* What benefits would SMW gain by supporting `ElasticSearch`?
* Why `ElasticSearch`? Are there any other search engines that are more preferable (such as SOLR/Lucene)?

[0] https://packagist.org/packages/diqa/enhanced-retrieval
[1] https://www.mediawiki.org/wiki/Extension:SolrStore
[2] https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html

kghbln:
The current situation is a bit split between ElasticSearch and SOLR/Lucene. While MediaWiki as such integrates with ElasticSearch, all advanced implementations involving SMW which require more complexity use SOLR/Lucene [3,4].

The rationale of WMF for choosing ElasticSearch over SOLR/Lucene is described here: [5].

They may however be right about SOLR/Lucene being more painful so I'd opt for ElasticSearch. If other want to add to this a separate extension will be needed as it is already done now.

[3] [ExtendedSearch - BlueSpice](https://github.com/wikimedia/mediawiki-extensions-BlueSpiceExtensions/tree/master/ExtendedSearch)
[4] [SolrStore - Gesis](https://www.mediawiki.org/wiki/Extension:SolrStore)
[5] [Search infrastructure maintenance and overhaul - rationale](https://www.mediawiki.org/wiki/Search/Old#Rationale)

jaideraf:
> I'd opt for ElasticSearch.

Me too. Firstly, because we can have MediaWiki wikis using [Cirrus](https://www.mediawiki.org/wiki/Extension:CirrusSearch) (which requires ElasticSearch)  and SMW together (if SMW would integrate with Solr, one would have to install MW, SMW, ElasticSearch AND Solr, which is not desired at all). 

So, the challenge (and feature) would be to use ElasticSearch for both MW and SMW. 
[...]

PR #3054, opened 2018-03-03
Add ElasticStore, refs 2274

mwjames:
This PR is made in reference to: #2274, #586 

- Introduces the `ElasticStore` as interface between Semantic MediaWiki and Elasticsearch (aka ES) with the objective to provide ES as drop-in replacement for the SQL based __QueryEngine__ (not the entity or data store).
- Information about the how, the why, and any requirements can be found in the [readme](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/es/src/Elastic/README.md).
- It doesn't require or use any `CirrusSearch` specific elements or bindings, the implementation is genuine to avoid yet another MW dependency. Indices are separated and are not suppose to share any query context.

This PR includes:
- [x] CI build passed

Fixes: #586, #2274

## ES and SMW

For the last couple of weeks we enabled ES and this branch on the sandbox to verify its functionality in a live setting and @kghbln can share some experiences. A [short video](https://www.youtube.com/watch?v=o2LSYRMOKBI) demonstrates some aspects of the supported functionality.

image

### Target audience

- `SQLStore` small wikis, technical novice users, non-technical environments
- `SPARQLStore` RDF enthusiast, semantic web researchers
- `ElasticStore` enterprise users, wiki farms

## What's next?

This PR is tested with ES 6.1.x and 6.2.x (sandbox) and ES 5.6.6 (Travis) with the test suite which means that all test cases pass as they do with the `SQLStore` (or respectively `SPARQLStore`). Yet, this PR does not provide any tests on its own and before the PR can be merged it requires some effort to reach a coverage that is necessary for a merge.

As a result, SMW 3.0 will not be released in Q1/2018 and pending the progress of this PR and how good the documentation incorporates changes made in SMW 3.0 (see #2762) further postponement may be necessary.


kghbln:
https://www.semantic-mediawiki.org/w/index.php?title=Help:SMWElasticSearch

agcunha:
I'm an enterprise user, interested in continue using sparql store for graphs. AWS launches Neptune that will be a nice, fast and robust option for creating knowledge bases. In my point of view, key/value stores, like lucene used in Elastic or Solr, aren't the best architecture for knowledge representation. It could be used to develop search engines(eg: OpenSemanticSearch) because of the capabilities to process different formats of documents. OCR, text mining and index. What is the motivation to use Elastic in the architecture of SMW? I'm sorry if it's not the right channel to ask this.

mwjames:
> I guess #2274 is the better spot.

Yes (see https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2274#issuecomment-381318251).

This PR is about the technical implementation.

mwjames:
Some lessons learned from running the sandbox as live instance:

### ES related

- If you use the `ingest-plugin` and upgrade, avoid a "java.lang.IllegalArgumentException: plugin [ingest-attachment] is incompatible with version [6.2.4]; was designed for version [6.2.3]" by removing the plugin before doing the upgrade (remember to add it after the upgrade)

mwjames:
> After the most recent update on sandbox I get the following when rebuilding data:
> PHP Warning:  array_flip(): Can only flip STRING and INTEGER values! in /../w/extensions/SemanticMediaWiki/src/Elastic/QueryEngine/TermsLookup.php on line 504

The case for this is that I added support for `[[-Has subobject::+]]` but I didn't filter requests on something that looks like ` [[-Has number::+]]` (which only returns gibberish since `Has number` is not a resource aka page type) so by the next update this should be fixed. There will be an extra `p-0709` integration test to cover this scenario since we hadn't covered this before.

mwjames:
> Yet, this PR does not provide any tests on its own and before the PR can be merged it requires some effort to reach a coverage that is necessary for a merge.

I'm going to merge this PR even though it does not yet fulfill the stated requirement for the unit coverage but #3165 has been created to ensure that coverage does happen before the 3.0 release.

@kghbln This should make the administration of the master branch a bit easier and since I'm off for the next eight weeks, no changes or merges will happen from my side during that time.

Something may change later on but it has been two months+ since we enabled the es-branch on the sandbox without major incidents.

PR #3055, opened 2018-03-05
smw-property-predefined-rl-desc typo fixed

ivanhercaz:

- Typo fixed
- Translating the string [smw-property-predefined-rl-desc](https://translatewiki.net/w/i.php?title=Special:Translate&showMessage=smw-property-predefined-rl-desc&group=mwgithub-semanticmediawiki) i found what seems to me like a typo error, where is a "the" I think should be a "that". If I am wrong, just close this PR, but this "the" sounds strange to me.

This PR includes:
- [x] CI build passed

Regards,
Iván

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -693,7 +693,7 @@
 	"smw-rule-page-description": "Description:",
 	"smw-rule-page-type": "Type:",
 	"smw-rule-page-tag": "Tag:",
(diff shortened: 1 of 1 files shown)


kghbln:
@ivanhercaz You are right. Good find. Thanks a lot!

PR #3071, opened 2018-03-22
Use …::class feature in class_alias() instead of strings

thiemowmde:
This is not different in so far that a …::class still is a string, and class_alias() still does nothing but adding an entry to a string => string map.

But utilizing the …::class feature does have the benefit that IDEs, linters and such are aware that a class is referenced, and will include these lines into usage reports and such.

### DIFF
--- a/src/Aliases.php
+++ b/src/Aliases.php
@@ -5,90 +5,90 @@
  */
 
 // 3.0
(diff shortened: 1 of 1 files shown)


mwjames:
Thanks

PR #3076, opened 2018-03-24
Check magic words list

mwjames:
This PR is made in reference to: https://sourceforge.net/p/semediawiki/mailman/message/36246292/

- Ensures that different (hereby recursive) parse runs that return a magic words list and are merged (as in case of using the Cite extension) together don't override (such as with an empty list) an already registered list from a preceding parse that belong to the same process chain.

This PR includes:
- [x] CI build passed


kghbln:
Refs https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/3081

PR #3080, opened 2018-03-25
Add IDs warmUpCache

mwjames:

- Adds a `smw_hash` field to the `smw_objects_ids` table to contain a sha1 hash which is composed of title, namespace, iw, and the subobject identifier.
- Objective is to reduce queries required to match IDs for entities during a `webrequest` by collecting the sha1 hash and run a collective `WHERE IN` query before any access happens to find an ID. The warm up reduces the amount of queries necessary for finding entity IDs and may only marginal improve the lookup performance. 
- Requires to run `update.php` or `setupStore.php`. In case the `smw_hash` is `NULL`, the installer will compute and update the field during the setup.

For example, run one query instead of 7 individual ones.

```
SELECT smw_id,smw_title,smw_namespace,smw_iw,smw_subobject,smw_sortkey,smw_sort
FROM `smw_object_ids`
WHERE smw_hash IN (
'eac3684cbe0a08de88176820b2231f4a51de287f',
'670186d9da6d1f44ac46cedf92668143eb9d8b1a',
'17ad580112fdc0835b2b7d2494bf3ee0247b8807',
'68c3a3f3fbd9c847efa5f716a85193e414e31172',
'20b13a83a8986fabe4bccc84886e8c2ace31d302',
'aaf522f8c6262e768bcac3bd28be4ecf07a7c6b4',
'4db87b67e34bba23b746fe6595274567736f39c2')

1.5681ms | SMWSql3SmwIds::warmUpCache
```

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/includes/storage/SQLStore/SMW_SQLStore3_Readers.php
+++ b/includes/storage/SQLStore/SMW_SQLStore3_Readers.php
@@ -129,6 +129,14 @@ class SMWSQLStore3Readers {
 
 		$this->semanticDataLookup->unlockCache();
 
+		$this->store->smwIds->warmUpCache(
+			$semanticData->getProperties()
+		);
+
+		$this->store->smwIds->warmUpCache(
+			$semanticData->getPropertyValues( new DIProperty( '_INST' ) )
+		);
+
 		return $semanticData;
 	}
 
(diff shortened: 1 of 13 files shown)


kghbln:
Noted [here](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AMaintenance_script_populateHashField.php&type=revision&diff=67467&oldid=67462) and [there.](https://www.semantic-mediawiki.org/w/index.php?title=Help:Upgrade&curid=14499&diff=67466&oldid=67282)

Issue #3012, opened 2018-02-17
MW 1.31 Test failed with "ParserOptions::setEditSection was deprecated in MediaWiki 1.31."


mwjames:
[0] reported:

## Issue

```
There were 2 errors:
1) SMW\Tests\Parser\RecursiveTextProcessorTest::testRecursiveTagParse_WITH_IncompleteParser
Use of ParserOptions::setEditSection was deprecated in MediaWiki 1.31. [Called from SMW\Parser\RecursiveTextProcessor::recursiveTagParse in /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/src/Parser/RecursiveTextProcessor.php at line 262]
/home/travis/build/SemanticMediaWiki/mw/includes/debug/MWDebug.php:309
/home/travis/build/SemanticMediaWiki/mw/includes/debug/MWDebug.php:242
/home/travis/build/SemanticMediaWiki/mw/includes/GlobalFunctions.php:1181
/home/travis/build/SemanticMediaWiki/mw/includes/parser/ParserOptions.php:883
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/src/Parser/RecursiveTextProcessor.php:262
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Unit/Parser/RecursiveTextProcessorTest.php:231
/home/travis/build/SemanticMediaWiki/mw/maintenance/doMaintenance.php:94
```

[0] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/342762305

PR #3083, opened 2018-03-30
Don't use ParserOptions::setEditSection for MW 1.31+ (Backport to 2.5.x); Fix CI scripts

s7eph4n:
This PR is made in reference to: Deprecation of `ParserOptions::setEditSection()`
* [RELEASE-NOTES-1.31](https://github.com/wikimedia/mediawiki/blob/92cf49df5ca0c5dea25f5ae651996d79a2ce25af/RELEASE-NOTES-1.31#L124)
* https://github.com/wikimedia/mediawiki/commit/2791fb08619545b12e58dd576ee6a98c38082c64

- Wrap the call of `ParserOptions::setEditSection()` in an if block checking `ParserOutput::SUPPORTS_STATELESS_TRANSFORMS`
- Add the equivalent options array as parameter to the call of `ParserOutput::getText()`
- Remove Java download from Travis CI scripts

This PR includes:
- [ ] CI build passed

Fixes: #3012

### DIFF
--- a/includes/queryprinters/FeedResultPrinter.php
+++ b/includes/queryprinters/FeedResultPrinter.php
@@ -222,7 +222,12 @@ final class FeedResultPrinter extends FileExportPrinter {
 	protected function getPageContent( WikiPage $wikiPage ) {
 		if ( in_array( $this->params['page'], array( 'abstract', 'full' ) ) ) {
 			$parserOptions = new ParserOptions();
-			$parserOptions->setEditSection( false );
+
+			// FIXME: Remove the if block once compatibility with MW <1.31 is dropped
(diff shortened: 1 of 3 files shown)


mwjames:
CI failure relates to (according to Travis support):

>I've been looking into the details of the job that you shared with us and it seems that you stumbled upon the issue reported at: https://github.com/travis-ci/travis-ci/issues/9379
>Unfortunately, it looks like this problem is caused by a mismatch between the `oracle-java8-installer` package and a new Java 8 version that was released recently. The package was still pointing to the old version that was no longer available on Oracle's site and this is why the download failed until Oracle sorted this out yesterday.

I removed the download from Travis in master with https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/9510f7513c667b55b92336d8123fe4024ad0cee9.

s7eph4n:
So installing Java 8 is not required anymore?

mwjames:
> So installing Java 8 is not required anymore?

Exactly because Travis installs it by default using a different package which works for our services.

PR #3095, opened 2018-03-31
Add upgrade check

mwjames:

- Whenever we change the DB schema (new field, altered index etc.) we need to ensure that users do actually run `update.php` or `setupStore.php` and in order to do that we introduce `smwgUpgradeKey` which contains an arbitrary identifier that is used to validate the status by comparing the local result of `.smw.json` against the upgrade key.
- If `.smw.json` doesn't exists or contains an invalid key then the system stops before it can access any data or tables and shows a message to the user such as: "Error: The Semantic MediaWiki extension was enabled but the installation is incomplete ...".

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ build/release/build-*
 
 composer.phar
 composer.lock
-.smw.json
 
 !.*
 .idea/
+.smw.json
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -63,6 +63,24 @@ return array(
 	##
 
 	###
+	# Upgrade key
+	#
+	# This key verifies that a correct upgrade (update.php/setupStore.php) path
+	# was selected and hereby ensures a consistent DB setup.
+	#
+	# Whenever a DB table change occurs, modify the key value (e.g. `DB-Foo...`)
+	# to reflect the requirement for the client to follow the processes as
+	# outlined in the installation manual.
+	#
+	# Once the installer is run, the `.smw.json` will be updated and no longer
+	# causes an exception.
+	#
+	# @since 3.0
+	##
+	'smwgUpgradeKey' => 'DB-2018-03',
+	##
+
+	###
 	# CompatibilityMode is to force SMW to work with other extensions that may impact
 	# performance in an unanticipated way or may contain potential incompatibilities.
 	#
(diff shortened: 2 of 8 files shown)


mwjames:
@kghbln FYI

Just came across the issue during an API request on a different test wiki where I forgot to run the `update.php`.

```
[6c40097c18b75d7bb9b71940] /mw-29-00/index.php/Main_Page RuntimeException from line 303 of ...\extensions\SemanticMediaWiki\src\MediaWiki\Database.php: A database query error has occurred. Did you forget to run your application's database schema updater after upgrading?
Query: SELECT smw_id,smw_title,smw_namespace,smw_iw,smw_subobject,smw_sortkey,smw_sort FROM `smw_object_ids` WHERE smw_hash = '003b6a0985dc21e3dd947b1778cbbb63b75a4788'
Function: SMWSql3SmwIds::warmUpCache
Error: 1054 Unknown column 'smw_hash' in 'where clause' (localhost)
```

mwjames:
Just as a  note, the setting is only to be used by the project, a normal user should not override or manipulate the setting!

kghbln:
> Just as a note, the setting is only to be used by the project, a normal user should not override or manipulate the setting!

Indeed.

YOUR1:
Isn't it better to rename the upgrade_key field in to upgrade_hash, since it doesn't contain security information?

mwjames:
> Isn't it better to rename the upgrade_key field in to upgrade_hash in .smw.json, since it doesn't contain security information?

I'm not sure whether changing the name would invertible imply security or not while the name itself is a key for the upgrade and contains a sha1.

In case it makes a difference to some, feel free to send a PR that introduces a different field name but you have to ensure that you provide a migration in [0] to avoid invaliding previous information.

> I'm asking this because I was a bit confused whether or not to disallow public access to this file.

I'd say that you should apply the same policy as for the your `LocalSettings.php` and of course as mentioned in the documentation [1] the file "must be writable, persistent and accessible for Semantic MediaWiki throughout its operation".

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/SetupFile.php
[1] https://www.semantic-mediawiki.org/wiki/Help:Setup_information_file

kghbln:
I think this is a matter of documenting. I see the point but I do not think it is important enough if docu exists.

PR #3096, opened 2018-03-31
SMWSearch, display section title (subobject)

mwjames:

- Show the section title in case SMWSearch matches a subobject to indicate different entities
- This method is selected instead of trying to manipulate the result count to distinguish a page from a page with fragment

This PR includes:
- [x] CI build passed

image

### DIFF
--- a/src/MediaWiki/Search/SearchResult.php
+++ b/src/MediaWiki/Search/SearchResult.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace SMW\MediaWiki\Search;
+
+/**
+ * @ingroup SMW
+ *
+ * @license GNU GPL v2+
+ * @since 3.0
+ *
+ * @author mwjames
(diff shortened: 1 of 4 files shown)


mwjames:
@s7eph4n FYI

s7eph4n:
:+1: Makes sense. I wonder how this is done in MW. `getSectionTitle()` always returns `null` and nobody seems to instantiate it.

mwjames:
>  always returns null and nobody seems to instantiate it.

This why we need to extend SearchResult.

> I wonder how this is done in MW

https://github.com/wikimedia/mediawiki-extensions-CirrusSearch/blob/5629187dadb300408ac11212bdc1780babc3def7/includes/Search/Result.php#L288-L293

s7eph4n:
Aha, that's what I was looking for. Thx!

PR #3100, opened 2018-04-07
Account for changes to fixed tables, refs 3095

mwjames:
This PR is made in reference to: #3095

- Content of `smwgFixedProperties` and `smwgPageSpecialProperties` influence the shape of the DB therefore use it to compute the upgrade key

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/SQLStore/Installer.php
+++ b/src/SQLStore/Installer.php
@@ -200,7 +200,39 @@ class Installer implements MessageReporter {
 			return true;
 		}
 
-		return isset( $GLOBALS['smw.json']['upgradeKey'] ) && sha1( $GLOBALS['smwgUpgradeKey'] ) === $GLOBALS['smw.json']['upgradeKey'];
+		if ( !isset( $GLOBALS['smw.json']['upgradeKey'] ) ) {
+			return false;
+		}
+
+		return self::getUpgradeKey( $GLOBALS ) === $GLOBALS['smw.json']['upgradeKey'];
+	}
+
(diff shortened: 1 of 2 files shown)


mwjames:
@kghbln FYI

Just this morning I run into the issue when I tried adding `_MIME` which caused an internal DB exception since the upgrade key did not account for it and I forgot that the property requires a fixed table `smw_fpt_mime`. To avoid similar situations in future any change to `smwgFixedProperties` and/or `smwgPageSpecialProperties` will be part of the key and halt the processing if the upgrade didn't happen.

```
2018-04-07 08:59:38 TAURUS mw-30-00-elastic-test: [71109a33be8e5120cec3ae98] /mw-30-00-elastic-test/index.php/File:Cn200072h.pdf   RuntimeException from line 303 of D:\xampp\htdocs\mw-30-00-elastic-test\extensions\SemanticMediaWiki\src\MediaWiki\Database.php: A database query error has occurred. Did you forget to run your application's database schema updater after upgrading?
Query: SELECT  *  FROM `smw_fpt_mime`    WHERE s_id = '14531'
Function: SMW\SQLStore\PropertyTableRowDiffer::fetchCurrentContentsForPropertyTable
Error: 1146 Table 'mw-30-00-elastic-test.smw_fpt_mime' doesn't exist (localhost)
```

https://www.semantic-mediawiki.org/wiki/Help:Upgrade

kghbln:
Good find!

PR #3107, opened 2018-04-08
LinksProcessor to correctly recognize `::=`

mwjames:
This PR is made in reference to: #https://github.com/Wikifab/SemanticMediaWiki/commit/8b89a5a4da028f01a30a907f3e47db94c2fdef4b

- Ensures to correctly recognize `[[Foo::=Bar]]` (where `:=` is the legacy notation) and splits the string into `Foo` and `=Bar`

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/Parser/LinksProcessor.php
+++ b/src/Parser/LinksProcessor.php
@@ -142,6 +142,14 @@ class LinksProcessor {
 
 		if ( array_key_exists( 1, $semanticLink ) ) {
 
+			// Use case [[Foo::=Bar]] (:= being the legacy notation < 1.4) where
+			// the regex splits it into `Foo:` and `Bar` loosing `=` from the value.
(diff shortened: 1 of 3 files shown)


mwjames:
@pierreboutet I've seen your change and I went with a slightly different approach. I pushed the change myself since most of time when I ask contributors to provide unit and integration tests it becomes very often very quiet. We need those tests to ensure we have them on the regression radar where until now it wasn't tested and therefore not recognized as a misinterpretation.

mwjames:
https://sandbox.semantic-mediawiki.org/wiki/Issue/3107

kghbln:
Good find!

pierreboutet:
Hi !
Thanks for this pull request, i did this quickly for an urgent need, and i was thinking to do later a proper reintegration with tests, you have been much quicker than me ! :)

PR #3116, opened 2018-04-08
Add Excerpts, extend SearchResultSet

mwjames:

- Adds the `Excerpts` component to be injected as part of the query request into a `QueryResult` (the standard `SQLStore` doesn't use it, has been isolated from https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3054#issuecomment-379488988)
- Extends `SearchResultSet` that if an excerpt is available it is injected and used instead of the standard text provided by the `Special:Search`
- `SearchResultSet::termMatches` will return matches retrieved from the `QueryToken` instance (if those token related to a query context, not a filter context) and hereby allows the standard `Special:Search` highlighter to mark relevant words

This PR includes:
- [x] CI build passed

image

### DIFF
--- a/includes/storage/SMW_QueryResult.php
+++ b/includes/storage/SMW_QueryResult.php
@@ -4,6 +4,7 @@ use SMW\HashBuilder;
 use SMW\Query\PrintRequest;
 use SMW\Query\QueryLinker;
 use SMW\Query\ScoreSet;
+use SMW\Query\Excerpts;
 use SMW\Query\Result\InMemoryEntityProcessList;
 use SMW\SerializerFactory;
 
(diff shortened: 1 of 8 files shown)


mwjames:
@s7eph4n FYI The termMatches token extension is available for any search request that uses a query context (eg. like/nlike, ~, ~~ etc.). 

The excerpt function isn't currently used but allows to do so in future by simply injecting it where an excerpt should magically appear in `Special:Search`.
```
// Retrieved as part of the query request from a different back-end
$excerpts = new Excerpts()
$excerpts->addExcerpt( 'FooTitle', 'Text...' );

// Build the result object
$queryResult = new QueryResult( ... );

$queryResult->setExcerpts( $excerpts );
```

image

s7eph4n:
Nice! :)

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ASMWSearch&type=revision&diff=61067&oldid=55415)

PR #3123, opened 2018-04-14
Special:Search/SMWSearch cheat sheet

mwjames:

- To help users understand quickly what they can and cannot do, the `i` button provides a cheat sheet.

This PR includes:
- [x] CI build passed

image


mwjames:
@kghbln Of course, if you disable the note you won't see the `i` and therefore you are unable the call up the cheat sheet but then again, disabling the note signals that you have understood how it works.

kghbln:
> Of course, if you disable the note you won't see the i and therefore you are unable the call up the cheat sheet but then again, disabling the note signals that you have understood how it works.

I subscribe to this assumption. :) / [Documented.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ASMWSearch&type=revision&diff=61945&oldid=61798)

PR #3126, opened 2018-04-14
Add SMWSearch power profile form

mwjames:

- Adds a `smw` power form to `Special:Search` for when `$wgSearchType = 'SMWSearch';` is used

This PR includes:
- [ ] CI build passed

image

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -493,6 +493,8 @@
 	"smw-search-input": "Input and search",
 	"smw-search-help-input-assistance": "[https://www.semantic-mediawiki.org/wiki/Help:Input_assistance Input assistance] is provided for the input field and requires to use one of following prefixes:\n\n*<code>p:</code> to enable property suggestions (e.g. <code><nowiki>[[p:Has ...</nowiki></code>)\n\n*<code>c:</code> to enable category suggestions\n\n*<code>con:</code> to enable concept suggestions",
 	"smw-search-syntax": "Syntax",
+	"smw-search-profile": "Extended",
+	"smw-search-profile-tooltip": "Search functions in connection with Semantic MediaWiki",
 	"log-name-smw": "Semantic MediaWiki log",
 	"log-show-hide-smw": "$1 Semantic MediaWiki log",
 	"log-description-smw": "Activities for [https://www.semantic-mediawiki.org/wiki/Help:Logging enabled event types] that have been reported by Semantic MediaWiki and its components.",
--- a/res/smw/special/ext.smw.special.search.css
+++ b/res/smw/special/ext.smw.special.search.css
@@ -71,6 +71,14 @@
     margin-left: 10px;
 }
 
+#smw-searchoptions {
+    margin: 0;
+    padding: 0.5em 0.75em 0.75em 0.75em;
+    background-color: #f8f9fa;
+    border: 1px solid #c8ccd1;
+    border-top-width: 0;
+}
+
 /**
 + * .skin-chameleon specific styles
 + */
--- a/res/smw/special/ext.smw.special.search.js
+++ b/res/smw/special/ext.smw.special.search.js
@@ -68,6 +68,23 @@
 	// https://www.semantic-mediawiki.org/wiki/Help:SMWSearch
 	if ( mw.config.get( 'wgCanonicalSpecialPageName' ) == 'Search' && mw.config.get( 'wgSearchType' ) == 'SMWSearch' ) {
 		load( search );
+
+		// Copied from mediawiki.special.search.js in order
+		// to have the NS button to work without #powersearch
+		var $checkboxes = $( '#search input[id^=mw-search-ns]' );
+
+		$( document ).on( "click", "#mw-search-toggleall", function(){
+			$checkboxes.prop( 'checked', true );
+		} );
+
+		$( document ).on( "click", "#mw-search-togglenone", function(){
+			$checkboxes.prop( 'checked', false );
+		} );
+
+		// When saving settings, use the proper request method (POST instead of GET).
+		$( document ).on( "change", "#mw-search-powersearch-remember", function() {
+			this.form.method = this.checked ? 'post' : 'get';
+		} ).trigger( 'change' );
 	};
 
 } )( jQuery, mediaWiki );
--- a/src/MediaWiki/Hooks/HookRegistry.php
+++ b/src/MediaWiki/Hooks/HookRegistry.php
@@ -254,8 +254,10 @@ class HookRegistry {
 			'LoadExtensionSchemaUpdates' => 'newLoadExtensionSchemaUpdates',
 
 			'ExtensionTypes' => 'newExtensionTypes',
-			'SpecialSearchResultsPrepend' => 'newSpecialSearchResultsPrepend',
 			'SpecialStatsAddExtra' => 'newSpecialStatsAddExtra',
+			'SpecialSearchResultsPrepend' => 'newSpecialSearchResultsPrepend',
+			'SpecialSearchProfileForm' => 'newSpecialSearchProfileForm',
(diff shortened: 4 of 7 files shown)


mwjames:
@kghbln @s7eph4n This is a WIP and came as left-over while experimenting with ES on excerpts and score sorting.

The power form adds some sorting `Best Match, Most Recent...` in order for results to be sorted differently. Of course, "Best Match" only makes sense in connection with ES where we get scores and can sort them accordingly.

Questions:

- Do we want or need something like this in core?
- Currently, the profile is called `Extended` any better name for such profile?

mwjames:
> I am wondering if this could "also" be piggy-backed to the "Advanced" tab like 

Well, there is a side story to the "Advanced" tab which can be extended with a hook called `SpecialSearchPowerBox`  and as SIL has tried it is pretty much useless as soon as you click the paging >> button because the hook is missing a vital piece that would allow you to set your optional parameters (`SpecialSearch::setExtraParam`) so that they can be retained beyond the first page.

The "Advanced" tab and `SpecialSearchPowerBox` hook in MW-core could be fixed by allowing parameters to be transferred as in:

```php
@@ -269,17 +269,21 @@ class SearchFormWidget {
 		}
 
 		$showSections = [
 			'namespaceTables' => "<table>" . implode( '</table><table>', $namespaceTables ) . '</table>',
 		];
-		Hooks::run( 'SpecialSearchPowerBox', [ &$showSections, $term, $opts ] );
+		Hooks::run( 'SpecialSearchPowerBox', [ &$showSections, $term, $opts, &$parameters ] );
 
 		$hidden = '';
 		foreach ( $opts as $key => $value ) {
 			$hidden .= Html::hidden( $key, $value );
 		}
 
+		foreach ( $parameters as $key => $value ) {
+			$this->specialSearch->setExtraParam( $key, $value );
+		}
```

Above is not going to happen and your left with creating your own tab via `SpecialSearchProfileForm`, copy the whole NS select form from the "Advanced" tab (seriously what a waste), and add your own form parameters to function correctly.

s7eph4n:
I love it! Frankly, I am surprised that this is not a thing yet on the vanilla search.

>    Do we want or need something like this in core?

Yes, if you ask me. I think modularity is great, but users are already complaining about the complexity of the installation. So until we have a better way of installing extensions I'd go for including it directly.

>    Currently, the profile is called Extended any better name for such profile?

Would it be possible to not have it as a new profile, but as an additional option to all the other profiles? Sort order and search perimeter are orthogonal.

mwjames:
> Would it be possible to not have it as a new profile, but as an additional option to all the other profiles? Sort order and search perimeter are orthogonal.

I'd love to but see my comments on the issue that arises with `SpecialSearchPowerBox`.

kghbln:
Were do you did out all these cool ideas. Show me the place please. :) And the definition of the search templates is done in the ["rule" namespace!?](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3019)

s7eph4n:
Wow! This is great!
Would it make sense to find some name other than "Extended", something that hints more at the functionality?

kghbln:
> Would it make sense to find some name other than "Extended", something that hints more at the functionality?

Perhaps "Structured" as in structured content. This will definitively make people curious.

kghbln:
Placeholder page for the wikidocu:
* https://www.semantic-mediawiki.org/wiki/Help:Rule/Type/SEARCH_FORM_DEFINITION_RULE

kghbln:
> First I had it as MediaWiki... but since you mentioned it, I created a new rule type and the definition is now expected at Rule:Search-profile-form-definition. The name is fixed, well you could go all willy nilly and search for all forms with the SEARCH_FORM_DEFINITION_RULE type but that's just too ...

Great. Still I have to dig into the mechanics of the new rule system since I am currently not sure how to set up more than one search template per wiki.

> Currently it is maintained with "smw-search-profile": "Extended"

Thus everyone can adapt. "Extended" is fine with me too. The future will tell if there are other suggestions out there for the default naming. I do not expect any though.

mwjames:
> Great. Still I have to dig into the mechanics of the new rule system since I
> am currently not sure how to set up more than one search template per wiki.

```
{
    "type": "SEARCH_FORM_DEFINITION_RULE",
    "forms": {
        "My first form": [ // ---> Form
            "Property one",
            "Property two"
        ],
        "My second form": [ // ---> Form
            "Property One",
            {
                "Property three": { // Field with extra attributes
                    "autocomplete": true
                }
            }
        ],
        "My third form": [ // ---> Form
            "Property Foo",
            "Property Bar"
        ]
    }
}
```

If you call a form "Open" then this will create a predefined output.

If I really have time I may make an discovery on the
`SEARCH_FORM_DEFINITION_RULE` type but for the now all form
definitions are expected to be hosted in
`Rule:Search-profile-form-definition`.

PS: Comment "// ..." should be removed from the real JSON.

mwjames:
> Still I have to dig into the mechanics of the new rule system since I am currently not sure how to set up more than one search template per wiki.

https://sandbox.semantic-mediawiki.org/wiki/Rule:Search-profile-form-definition

Knock yourself out:

https://sandbox.semantic-mediawiki.org/w/index.php?search=%C2%A0&title=Sp%C3%A9cial%3ARecherche&profile=smw&fulltext=1&sort=best&smw-form=booksandjournals&hastitle%5B%5D=&hasauthor%5B%5D=&published%5B%5D=&hasgenre%5B%5D=Novel&haspublisher%5B%5D=&mediatype%5B%5D=&mimetype%5B%5D=&fileattachment.contenttype%5B%5D=&ns0=1&ns6=1&ns102=1&ns108=1

https://sandbox.semantic-mediawiki.org/w/index.php?search=%5B%5Bin%3AM%C3%BCnchen%5D%5D&title=Sp%C3%A9cial%3ARecherche&profile=smw&fulltext=1&sort=best&smw-form=mediaandfiles&hastitle%5B%5D=&hasauthor%5B%5D=&published%5B%5D=&hasgenre%5B%5D=&haspublisher%5B%5D=&mediatype%5B%5D=&mimetype%5B%5D=&fileattachment.contenttype%5B%5D=application%2Fpdf&ns0=1&ns6=1&ns102=1&ns108=1

https://sandbox.semantic-mediawiki.org/w/index.php?search=%5B%5Bin%3AM%C3%BCnchen%5D%5D&title=Sp%C3%A9cial%3ARecherche&profile=smw&fulltext=1&sort=best&smw-form=mediaandfiles&hastitle%5B%5D=&hasauthor%5B%5D=&published%5B%5D=&hasgenre%5B%5D=&haspublisher%5B%5D=&mediatype%5B%5D=&mimetype%5B%5D=&fileattachment.contenttype%5B%5D=&ns0=1&ns6=1&ns102=1&ns108=1

kghbln:
> Knock yourself out:

Thanks a lot for the kick-starting knock-out. :smile:

krabina:
This is great!
I'd prefer "Structured" over "Extended", but more importantly I would like so see alle Namespaces hidden from the user unless I want to show them. Every SMW installation has tons of namespaces that just confuse the user. It's nice to have the option to see them and filter for them, but the default should be: don't show the namespaces to start with.
One solution could be to hide the checkboxes via css and allow them to be extended/displayed if the users wants to.
Even better would be a way to configure this. In cases with heavy use of custom namespaces having the possibility is great, but maybe I want to only show them the custom (content) namespaces e. g. "Book:", "Journal:", "Media:" but not the "built in" namespaces like "Form:", "Widget:", "Template", "Property"....

mwjames:
> If I really have time I may make an discovery on the `SEARCH_FORM_DEFINITION_RULE` type but for the now all form definitions are expected to be hosted in `Rule:Search-profile-form-definition`.

See #3143.

>  I would like so see alle Namespaces hidden from the user unless I want to show them.

See #3143.

PR #3130, opened 2018-04-21
List format rework

### LINKED ISSUE #2488: Rework List format
SMW 3.0  is coming up. I am reworking the `list` format continually updating this issue with the implemented changes.

Here is what I will do:

**Separate separators for values, properties and result "rows".**
E.g. three separators (`sep=,`, `propsep=;`, `valsep=/`) could produce the following result:
> United States of America (Cities Washington / New York / Los Angeles; Neighbors Canada / Mexico), United Kingdom (Cities London / Glasgow / Manchester; Neighbors Ireland), France (Cities Paris / Marseille / Bordeaux; Neighbors Belgium / Luxemburg / Germany / Switzerland / Italy / Spain), and Germany (Cities Berlin / Frankfurt / Hamburg; Neighbors Poland / Czechia / Austria / Switzerland / France / Belgium / Luxemburg / Netherlands / Denmark)

This would also benefit the subformats (`ol`, `ul`, `template`, `tree` (in SRF))

**A simpler logic of when which separator is used.**
This is currently handled in [ListResultPrinter.php](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/includes/queryprinters/ListResultPrinter.php#L239) and is rather confusing. I'd just reduce this to a default value `, `, that may be overridden by parser function parameters. The default values should be defined in i18n messages.

**Remove the final list separator (", and").**
It has been gone for a while due to a bug and apparently nobody noticed.

**Remove `?` as prefix for template arguments (see #972).**
Apparently nobody remembers why it was introduced in the first place.

**Remove `template arguments` parameter.**
Without the `?` prefix for template parameters it is not needed anymore.

**Make the template format an alias of the list format.**
When the user supplies a parser function parameter `template`, then just assume they want it used, even for `format=list`. And inversely if `format=template`, but the parser function parameter `template` is missing or empty, just print a normal list.

**Rename/standardize parameters to templates.** E.g. all standard parameters start with a `#`:
* `smw-resultquerycondition` => `#querycondition`
* `smw-resultquerylimit` => `#querylimit`
* `smw-resultqueryoffset` => `#resultoffset`
* `userparam` => `#userparam`
* new: `#rowcount`
* new: `#rownumber`

**Add class attribute to HTML elements.**
To facilitate styling.

**General code clean-up and removal of legacy/deprecated functionality**
To make it easier to test the `list` format.

s7eph4n:
This PR is made in reference to: #2488

- Rework of the list format (incl. ol, ul, template, new: plainlist)
- see https://gist.github.com/s7eph4n/277e7804fe04954df7d1e15ae874b0d0

This PR includes:
- [x] CI build passed

Fixes #2488


mwjames:
I haven't found any major glitches on the quick review I did. Any issues can be fixed in a follow-up and might be easier to spot hereafter.

I'm guessing the most problematic will be the switch from template parameter `userparam` to `#userparam` where I can see that it is used by people to import arbitrary data into a template and now be surprised that it no longer returns expected results __BUT__ this work will be part of a major release so we are allowed to make that kind of change(s). [0] contains information about how to migrate and switch templates and related formats to conform with 3.0.

[0] https://gist.github.com/s7eph4n/277e7804fe04954df7d1e15ae874b0d0

mwjames:
> [0] https://gist.github.com/s7eph4n/277e7804fe04954df7d1e15ae874b0d0

Maybe want to link or copy to [0].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/technical/migration-guide-3.0.md

kghbln:
The biggest thing now will be to document on wiki and to create migration help.

PR #3132, opened 2018-04-22
Add a section to the migration guide; remove some leftover comment code

s7eph4n:
This PR is made in reference to: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3130#issuecomment-383343818

- Add a section to the migration guide
- remove some leftover comment code

This PR includes:
- [x] CI build passed

### DIFF
--- a/docs/technical/migration-guide-3.0.md
+++ b/docs/technical/migration-guide-3.0.md
@@ -49,3 +49,21 @@
 	return true;
 } );
 ```
+
+## Formats
+
+### List formats (incl. list, ol, ul, template)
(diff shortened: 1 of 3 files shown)


mwjames:
> remove some leftover comment code

I saw them earlier but then I thought you wanted to keep them as remainder for a follow-up.

s7eph4n:
We could, I guess. But it is generally bad practice and I figured if anybody really misses the feature, they would probably tell us. But I don't mind either way, we may also keep it.

PR #3138, opened 2018-04-26
Don't rely on $wgExtensionDirectory to find SMW's extension.json

### LINKED ISSUE #3133: extension.json and $wgExtensionDirectory relation
### Setup and configuration

- SMW version: 3.0.0-alpha
- MW version: 1.30.0
- PHP version: 7.1.16
- MySQL version: 5.6.39

### Issue
At my environment, I have two ways to deal with MW extensions: one way to extensions that are installed via `composer` and another way to extensions installed via `git` (the "Wikimedia way"):

SMW and other extensions installed via `composer` are placed at the standard `w/extensions/ ` directory.
General Wikimedia extensions (and skins) that I installed via `git` are placed outside the main MW directoy:

* /home/user/mw-REL1_28/extensions/
* /home/user/mw-REL1_28/skins/
* /home/user/mw-REL1_29/extensions/
* /home/user/mw-REL1_39/skins/
* /home/user/mw-REL1_30/extensions/
* /home/user/mw-REL1_30/skins/

So I use `$wgExtensionDirectory` (and `$wgStyleDirectory`) to tell MW that extensions and skins are not at the standard path.
This configuration serves more than one MW instance (let's say I have 2 MW instances pointing to `REL1_29` extensions, and 1 pointing to `REL1_30`, etc. - like a wiki farm) and eases the updating and the downgrading processes of MW versions, because I only have to change one setting (`$wgExtensionDirectory`) to the right directory.

For what I know, until SMW 2.5.6 these two methods of instalation were independent. `composer` installed extensions is always placed at `w/extensions/ ` directory. But, I think after #1732, MW is trying to find SMW `extension.json` where `$wgExtensionDirectory` is pointing to, causing the Fatal error below.

### Stack trace
**Fatal error**: Uncaught Exception: .../mw-REL1_30/extensions/SemanticMediaWiki/extension.json does not exist! in /w/includes/registration/ExtensionRegistry.php:99 
```
 #0 /w/includes/GlobalFunctions.php(120): ExtensionRegistry->queue('/home/jaideraf/...')
 #1 /w/extensions/SemanticMediaWiki/src/GlobalFunctions.php(220): wfLoadExtension('SemanticMediaWi...')
 #2 /w/LocalSettings.php(459): enableSemantics('wikincat.org')
 #3 /w/includes/WebStart.php(102): require_once('/home/jaideraf/...')
 #4 /w/index.php(40): require('/home/jaideraf/...')
 #5 {main} thrown in /w/includes/registration/ExtensionRegistry.php on line 99
```

### Steps to reproduce
* Set `$wgExtensionDirectory` in `LocalSettings.php` to something other than `{$IP}/extensions` (the default value).
* Install SMW 3.0.0-alpha.

legoktm:
While extension.json expects that all extensions will be in a single directory,
SemanticMediaWiki previously didn't require this. By passing the second
argument to `wfLoadExtension()`, we can specify exactly which extension.json
to read from, instead of relying upon `$wgExtensionDirectory`.

Fixes #3133.

I didn't test this.

This PR is made in reference to: #3133 and #1732 

- Bug fix for linked bug

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/GlobalFunctions.php
+++ b/src/GlobalFunctions.php
@@ -217,7 +217,7 @@ function enableSemantics( $namespace = null, $complete = false ) {
 	global $smwgNamespace;
 
 	// #1732 + #2813
-	wfLoadExtension( 'SemanticMediaWiki' );
+	wfLoadExtension( 'SemanticMediaWiki', dirname( __DIR__ ) . '/extension.json' );
 
 	// Apparently this is required (1.28+) as the earliest possible execution
 	// point in order for settings that refer to the SMW_NS_PROPERTY namespace


jaideraf:
@kghbln Yes. I confirm. This solves the issue. :+1:

kghbln:
Great, thanks a lot. This appears to be a minimalistic solution to a great problem. :)

mwjames:
>  By passing the second
argument to wfLoadExtension(), we can specify exactly which extension.json
to read from, instead of relying upon $wgExtensionDirectory.

Thanks.

> @kghbln Yes. I confirm. This solves the issue. +1 Thanks @legoktm.

Here as well, I rely on the feedback so I have not tried this myself.

PR #3143, opened 2018-04-28
Hide namespace section, auto-discovery, refs 3126

mwjames:
This PR is made in reference to: #3126

- Follow-up to https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3126#issuecomment-384204421 and https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3126#issuecomment-383365361

This PR includes:
- [x] CI build passed

> If I really have time I may make an discovery on the `SEARCH_FORM_DEFINITION_RULE` type but for the now all form definitions are expected to be hosted in `Rule:Search-profile-form-definition`.

You no longer are bound by the fixed name of  `Rule:Search-profile-form-definition`, any `Rule:...` with the type of `SEARCH_FORM_DEFINITION_RULE` will be auto-discovered and merged into one single form definition during the build process. It goes without saying that if form some reason you have two `Rule:...` pages that refer to the same form name then fields are merged!

> I would like so see alle Namespaces hidden from the user unless I want to show them.

Last week when we deployed the profile, I could see that we have a bunch of namespaces that are  irrelevant for structured searches and only clutter the interface. You may use them on a different profile but it seems less likely to be useful in combination with SMW.

image

To improve the situation:

### `namespaces`

Since we had to copy the namespace form from core in order to provide a common component within the search interface we are also under the control for the `smw` profile of what is being shown and how. First, a hide/show button is added to control the ability to toggle the display of the namespace form. If no form has been selected, the namespace section will be shown but as soon as the user selects a form this will be hidden and the user can choose to have it displayed by using the provided button.

image

The above mentioned show/hide will alter the display but not alter the appearance of which namespaces should be shown. To reduce the amount of selectable namepaces from a form, you can now add `namespaces`/`hidden` to a form definition allowing it to hide the mentioned namespace for any `smw` related form.

```
    "namespaces": {
        "hidden": [
            "SMW_NS_PROPERTY",
            "SMW_NS_PROPERTY_TALK",
            "SMW_NS_CONCEPT",
            "SMW_NS_CONCEPT_TALK",
            "SMW_NS_RULE",
            "SMW_NS_RULE_TALK",
            "NS_MEDIAWIKI",
            "NS_MEDIAWIKI_TALK",
            "NS_TEMPLATE",
            "NS_TEMPLATE_TALK",
            "NS_PROJECT",
            "NS_PROJECT_TALK",
            "NS_HELP_TALK",
            "NS_USER_TALK",
            "NS_CATEGORY_TALK"
        ],
        "preselect": {
            "Media": [
                "NS_FILE"
            ]
        }
    },
```
The `preselect` is there to support a preselection for certain namespace on a particular form so that for example in case of the a `Media` form, the `NS_FILE` is checked when the user selects that form without having him/her to explicitly mark the namespace as the form context most likely requires a search to include this namespace.

### DIFF
--- a/res/smw/special/ext.smw.special.search.css
+++ b/res/smw/special/ext.smw.special.search.css
@@ -170,6 +170,34 @@
   margin-left: unset;
 }
 
+#smw-search-togglensview {
+    float: right;
+}
+
+.ns-list-off {
+    display:none;
+}
+
+.ns-list-on {
+    display:block;
+}
+
+#smw-togglensview {
+    margin-left: 1em;
+    text-align: center;
+    text-decoration: none;
+    display: inline-block;
+    opacity: 0.6;
+}
+
+#smw-togglensview:hover {
+    opacity: 1
+}
+
+#smw-togglensview:focus{
+    outline: none;
+}
+
 #smw-searchoptions {
     margin: 0;
     padding: 0.5em 0.75em 0.75em 0.75em;
@@ -185,12 +213,6 @@
     margin-bottom: 0.5em;
 }
 
-/**
-+ * .skin-chameleon specific styles
-+ */
-.skin-chameleon .smw-search-help-proximity {
-	margin-top: 10px;
-}
 
 /**
  * Responsive settings (#see smw.table.css)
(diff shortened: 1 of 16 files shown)


mwjames:
@kghbln FYI

PR #3145, opened 2018-04-28
SMWSearch simplified term parser, refs 3126

mwjames:

- Experimental feature hat simplifies the SMWSearch input term parsing and is only relevant for those who use a full-text integration and should only be enabled by users that think it doesn't dilute the syntax understanding of #ask for their users.

This PR includes:
- [x] CI build passed

### term.parser

By default, the __big__ `Special:Search` input field in context with SMW only understands well defined (meaning terms that are isolated and structured with `[[` `]]` ) terms such  `[[Has text::foo bar]]` or `[[in:some text within any article]]` to start a search request with the SMW query engine.

Yet, when searching a user that uses `Special:Search` is most likely less concerned with forms and structures and all s(he) wants is to search quick (and easy) and adding extra `[[` `]]` around each term seems soporific therefore to relax the input process the `term.parser` can work with a simplified input to build the final #ask conform syntax.

Simple | #ask conform
------------ | -------------
`(in:foo bar \|\| phrase:some text that appears in the same order) && category:Loose text` |  `<q>[[in:foo bar]] \|\| [[phrase:some text that appears in the same order]]</q> && [[Category:Loose text]]`

Without the `term.parser` enabled the left search query is required to be rewritten to represent the exact #ask elements in order to express the query intent.

The `term.parser`  will recognize the following `'in:', 'phrase:', 'not:', 'category:', '&&', 'AND', '||', 'OR', ')', '('` expressions to build a query that adheres the #ask syntax.

The `'in:', 'phrase:', 'not:'` expression are almost always only meaningful in the context of the full-text search and while not limited to `Special:Search` (or the smw profile) there where introduced as part of SMW 3.0 to broaden the expressiveness of a search request.

### DIFF
--- a/src/MediaWiki/Search/QueryBuilder.php
+++ b/src/MediaWiki/Search/QueryBuilder.php
@@ -169,6 +169,10 @@ class QueryBuilder {
 			$data = $this->data;
 		}
 
+		if ( isset( $data['term.parser'] ) && $data['term.parser'] ) {
+			$term = $this->term_parser( $term );
+		}
+
 		$form = $this->request->getVal( 'smw-form' );
 
 		if ( ( $data = $this->fetchFieldValues( $form, $data ) ) === [] && trim( $term ) ) {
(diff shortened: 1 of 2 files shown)


mwjames:
To clarify:

Expression | #ask | Note
------------ | ------------- | -------------
`in: ...`  |  `~~* ... *` | means find anything that contains ...
`phrase: ...` |  `~~" ... "` | means find anything  that contains ... in the exact same order
`not: ...` | `!~~...` | to not match any entity that contains ...
`category:...` | `[[Category:...]]` | simple short cut
`(` |  `<q>` | indicates the start of a subquery
`)` |  `</q>` |  indicates the end of a subquery

`~~`generally expresses a wide proximity intent where a specific property is unknown (and in case of ES it can expand the search radius to fields that have not been annotated or processed by SMW prior the query request), so for instance when searching for `Berlin` and `London` in the `City` category without knowing which property has Berlin and London being assigned to we would write (it uses the widest possible intent to find some matches where Berlin and London are part of an annotation):

- `in:berlin in:london category:City` (term.parser enabled) or
- `in:berlin && in:london && category:City`  (term.parser enabled) or
- `[[in:berlin]] [[in:london]] [[Category:City]]` (standard request)

Of course, if we know the specific property we would use `[[Has city::Berlin]] [[Has city::London]] [[Category:City]]`

To make use of the simple syntax as described above, one would need to add:

```
    "type": "SEARCH_FORM_DEFINITION_RULE",
    "term.parser": true,
    "forms": { ...
```

The `in:` expression can also be combined with a property and depending on the type context behaves differently.

Expression | #ask
------------ | -------------
`[[Has text::in:some foo]]` | `[[Has text::~*some foo*]]`
`[[Has number::in:99]]` | `[[Has number:: [[≥0]] [[≤99]] ]]`
`[[Has number::in:-100]]` | `[[Has number:: [[≥-100]] [[≤0]] ]]`
`[[Has date::in:2000]]` | `[[Has date:: [[≥1.1.1999]] [[≤31.12.2000]] ]]`

Issue #540, opened 2014-09-16
Sorting on Pages will exclude non-existent pages from result


s7eph4n:
1. Create some pages containing attributes of type Page, e.g. with '[[knows page::foo]]'.
2. Create some, but not all of the referenced pages
3. Query for the pages created in step 1 using 'sort=knows page'

Expected: All pages created in 1
Observed: Only those pages are included in the query result, where the page referenced by 'knows page' exists.
- wikimedia/mediawiki-core@66d3eb5 (MediaWiki 1.24alpha)
- SemanticMediaWiki@f877274  (2.1 alpha)

mwjames:
@s7eph4n Somehow I'm unable to observe the describe issue:

- Created page Issue540/A with `[[knows page::Issue540-1]]`, and Issue540/B, Issue540/C both with `[[knows page:: Issue540-2]]`
- Reference page `Issue540-1` was created but left `Issue540-2` as is (red linked).

Executing (on the current master) the following queries did not limit (or better discriminated) the output in the described manner.

```
{{#ask: [[knows page::Issue540-1]] 
 |?Knows page
 |sort=Knows page
}}

{{#ask: [[knows page::Issue540-1]] 
 |?Knows page
}}
```
Above queries does return `Issue540/A`.

```
{{#ask: [[knows page::Issue540-2]] 
 |?Knows page
 |sort=Knows page
}}

{{#ask: [[knows page::Issue540-2]] 
 |?Knows page
}}
```
Above queries does return `Issue540/B` and `Issue540/C`.

mwjames:
> Observed: Only those pages are included in the query result, where the page referenced by 'knows page' exists.

I couldn't replicate the issue [0] or maybe I misunderstood the premises but based on the results retrieved from [0] I will close this issue since the described observation couldn't be made.

[0] https://sandbox.semantic-mediawiki.org/wiki/Issue/540

PR #3149, opened 2018-04-29
Test sorting on Pages will not exclude non-existent pages from result (#540)

s7eph4n:
This PR is made in reference to: #540

This PR tests that sorting on a Pages type attribute will not exclude pages referencing non-existent pages from the result

This PR includes:
- [x] CI build passed


mwjames:
Thanks.

PR #3152, opened 2018-04-29
Special:Ask add extra debug query parameter (score_set, q_engine)

mwjames:

- Allows to set some query parameters which are useful when debugging or getting extra information from a selected query engine
  - `q_engine=sql_store`
  - `score_set=show` (only works where a score set is available such as ES)

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/src/MediaWiki/Specials/SpecialAsk.php
+++ b/src/MediaWiki/Specials/SpecialAsk.php
@@ -293,6 +293,16 @@ class SpecialAsk extends SpecialPage {
 					$result .= is_string( $debug ) ? $debug : '';
 				}
 			}
+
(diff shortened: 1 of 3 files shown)


mwjames:
@kghbln For example, if we wanted to switch to the SQLStore query engine without changing `$GLOBALS['smwgDefaultStore']` we now simply add `&q_engine=sql_store` to the URL and it will use the `SMWSQLStore` to answer the query. `&q_engine=sql_store` is only expected to work with `Special:Ask` and should only be used as debug method.

Adding `&score_set=show` (only available with ES) will add a simple list of scores and entities at the end of the `Special:Ask` output.

image

kghbln:
**Examples**
- [q_engine=sql_store](https://bit.ly/2AQdQRa) - stort = SQL
- [q_engine=](https://bit.ly/2AP0JzD) - default store = ES

PR #3157, opened 2018-05-12
Add custom prefix map to term.parser, refs 3145

mwjames:
This PR is made in reference to: #3145

- As outlined in #3145 `term.parser` adds a simplified syntax to the `Special:Search` input field, the PR extends the parser to recognize custom prefixes so that a set of properties can be collected under a specific search prefix.
- For example, by adding prefix assignments (as demonstrated below)
  - A `keyword:foobar` input will be transformed into a #ask conform expression of `<q>[[Has keyword::foobar]] OR [[Has keywords::foobar]] OR [[Keyword::foobar]]</q>` allowing the shorten the input sequence without a user being require to remember particular properties.
  - Various language related properties are used for different purposes but during a quick search a user wants to find certain documents that match any of the possible language properties by simply referring to `lang:foo in:some documents with xyz`

This PR includes:
- [x] CI build passed

## Example

```
    "term.parser": {
        "keyword": [
            "Has keyword",
            "Has keywords",
            "Keyword"
        ],
        "abstract": [
            "Has abstract",
            "Has description"
        ],
        "lang": [
            "Language code",
            "Document language",
            "File attachment.Content language",
            "Has interlanguage link.Page content language"
        ]
    },
```

### DIFF
--- a/res/smw/special/ext.smw.special.search.css
+++ b/res/smw/special/ext.smw.special.search.css
@@ -213,6 +213,9 @@
     margin-bottom: 0.5em;
 }
 
+#mw-searchoptions h4, #smw-search-togglensview, #mw-search-togglebox {
+    margin-top: 2px !important;
+}
 
 /**
  * Responsive settings (#see smw.table.css)
--- a/src/MediaWiki/Search/Form/FormsBuilder.php
+++ b/src/MediaWiki/Search/Form/FormsBuilder.php
@@ -141,7 +141,9 @@ class FormsBuilder {
 			$html[] = "<option value='$k' $selected>$name</option>";
 		}
 
-		$link = ( $link !== '' ? $link . ' ⦁ ' : '' ) . Html::element(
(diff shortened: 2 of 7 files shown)


mwjames:
@kghbln This PR extends on #3145.

Most of the time when using `Special:Search` one wants to find something quick and any extra minute spend on finding out which syntax or property to use is forgone therefore the `term.parser` makes it possible to use a "lighter" syntax (as outlined here and in #3145) so that an input can follow a more "human" readable approach such as `(in:foobar || phrase:foo bar) lang:fr` being equivalent to `<q>[[in:foobar]] || [[phrase:foo bar]]</q><q>[[Language code::fr]] OR [[Document language::fr]] OR [[File attachment.Content language::fr]] OR [[Has interlanguage link.Page content language::fr]]</q>`.

The `term.parser` does only work on the `Special:Search` input and only in combination with the smw extended profile.

PR #3162, opened 2018-05-12
Support {{DISPLAYTITLE}} in format=feed, refs 1741

### LINKED ISSUE #1741: Format "feed" - emit page title rather than page name
When using e.g. `|format=feed |type=rss` in a query the actual name of the page is emitted which is indeed the expected behaviour. When using the [Semantic Title](https://www.mediawiki.org/wiki/Extension:Semantic_Title) extension it may be desirable not to get the page name but rather the page title set via that extension.

Is something like this doable from the technical point of view? If yes, support for this will be cool. @cicalese @mwjames 

Opening this question/feature request here since this format is provided by Semantic MediaWiki. It may very well be that also the Semantic Title extension needs to cater for something like this too.

mwjames:
This PR is made in reference to: #1741 

- Use the display title instead of the subject title, if available.

This PR includes:
- [X] CI build passed

Fixes #1741

### DIFF
--- a/includes/datavalues/SMW_DV_WikiPage.php
+++ b/includes/datavalues/SMW_DV_WikiPage.php
@@ -606,7 +606,9 @@ class SMWWikiPageValue extends SMWDataValue {
 
 		$preferredCaption = $this->getDisplayTitle();
 
-		if ( $preferredCaption === '' ) {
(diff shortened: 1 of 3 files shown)


mwjames:
@kghbln FYI

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AFeed_format&type=revision&diff=66531&oldid=66517)

PR #3167, opened 2018-05-13
Add RemoteRequest

mwjames:

- https://www.semantic-mediawiki.org/wiki/Help:Remote_request

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -725,18 +725,44 @@ return array(
 	'smwgResultFormatsFeatures' => SMW_RF_TEMPLATE_OUTSEP,
 	##
 
-	### Predefined sources for queries
+	###
+	# Handling of `RemoteRequest` features
+	#
+	# - SMW_REMOTE_REQ_SEND_RESPONSE allows Special:Ask to respond to remote requests in
+	# combination with $smwgQuerySources and the `RemoteRequest`.
+	#
+	# - SMW_REMOTE_REQ_SHOW_NOTE shows a note for each remote requests so users are aware
+	# that results retrieved from an external source.
+	#
+	# If `$smwgQuerySources` contains no entries then a remote request to a source
+	# is not supported and only sources that are available through the setting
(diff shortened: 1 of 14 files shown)


kghbln:
Docu of [remote requests](https://www.semantic-mediawiki.org/wiki/Help:Remote_request) and [`$smwgRemoteReqFeatures`](https://www.semantic-mediawiki.org/wiki/Help:$smwgRemoteReqFeatures)

kghbln:
Self-answering some of my questions after reading through the docu page I found:

> 1. Is this feature still experimental?

yes

> 2. What happens if ...

Removing `MW_REMOTE_REQ_SEND_RESPONSE` from `$smwg­Remote­Req­Features` however this only applies to special page "Ask" and not to inline queries. So it appears only to be partially possible.

> 3. I assume that the query source needs to be semantic, too?

yes

> 4. I assume that the API ...

Not sure yet.


Moreover the [example I created on sandbox](https://sandbox.semantic-mediawiki.org/wiki/My_second_external_query) shows a result but only a fraction of the expected result set.

mwjames:
I'm starting the review the backlog!

Your changes in [0] to `smwgQuerySources` override each other, if you assign `$smwgQuerySources = [ ... ]` you override any previous set values so either you combine them such as:

```
$smwgQuerySources = [
 'smwo' => ...
'smworg' =>
]
```
or you explicitly make an entry ` $smwgQuerySources['smwo'] = ....`

You can see that [1] only contains one entry because you mixed up the settings.

[0] https://sandbox.semantic-mediawiki.org/wiki/SMW_Sandbox:Setup_information
[1] https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Requ%C3%AAter

kghbln:
> I'm starting the review the backlog!

Welcome back! One thing at a time.

Thanks for the infos which I will incorporate into the docu.

PR #3178, opened 2018-06-26
Avoid use of deprecated "json" module

### LINKED ISSUE #3177: The "json" module is deprecated
### Issue

SemanticMediaWiki (latest master) registers a module `ext.jquery.jStorage` which depends on the `json` module from MediaWiki core which has been deprecated since MediaWiki 1.29.

https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/dd07f85b69fe2b58733017725982fb77bea67ec2/res/Resources.php#L65

In MediaWiki 1.29, the JavaScript run-time for Grade A was made to require native support for ES5, including `JSON.parse` and `JSON.stringify`. As such, the `json` polyfill no longer made sense.

The module was kept as an empty no-op serving only a deprecation warning.

I intend to remove the module in the current release cycle (for MW 1.32).

Krinkle:
The jquery.json module was removed in MW 1.25 (deprecated in MW 1.24).
Given SMW requires MW 1.27, logic for that is now redundant.

In MW 1.29, the json module was replaced with a deprecated dummy, given MW's
JS feature test now confirms native support in browsers. The dummy will be
removed in MW 1.32.

Fixes #3177.

### DIFF
--- a/res/Resources.php
+++ b/res/Resources.php
@@ -57,12 +57,11 @@ return array(
 		'targets' => array( 'mobile', 'desktop' )
 	),
 
-	// MW 1.24+ Fix Uncaught Error: Unknown dependency: jquery.json
-	// Introducing a mega-hack
+	// Avoid "Warning: Use of the json module is deprecated since MediaWiki 1.29"
 	// jStorage was added in MW 1.20
 	'ext.jquery.jStorage' => $moduleTemplate + array(
 		'scripts' => 'jquery/jquery.jstorage.js',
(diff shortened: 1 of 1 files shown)


kghbln:
@Krinkle Thank you for authoring the change.

@JeroenDeDauw Since James is still in his break it will be great if you could do the honors of reviewing and merging. Thanks a bunch!

JeroenDeDauw:
Thanks for the patch @Krinkle !

PR #3182, opened 2018-06-28
List format: Add 'class' parameter 

s7eph4n:
This PR is made in reference to: #3136 

This PR addresses or contains:
- List format: Add 'class' parameter 

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

Fixes #3136

PR #3183, opened 2018-06-28
List format: Fix xss vulnerability

s7eph4n:
F'up #3182

This PR is made in reference to: #3182

- List format: Fix xss vulnerability

This PR includes:
- [ ] CI build passed

### DIFF
--- a/src/Query/ResultPrinters/ListResultPrinter/ListResultBuilder.php
+++ b/src/Query/ResultPrinters/ListResultPrinter/ListResultBuilder.php
@@ -173,7 +173,7 @@ class ListResultBuilder {
 	 * @return string
 	 */
 	private function replaceVariables( $subject ) {
-		return str_replace( [ '$START$', '$CLASS$' ], [ $this->get( 'offset' ) + 1, $this->get( 'class' ) ], $subject );
+		return str_replace( [ '$START$', '$CLASS$' ], [ htmlspecialchars( $this->get( 'offset' ) + 1 ), htmlspecialchars( $this->get( 'class' ) ) ], $subject );
 	}
 
 	/**


JeroenDeDauw:
htmlspecialchars is for in the body of tags, not attribute values or in the opening tag itself.

This string is not affected by htmlspecialchars, so the XSS is still there

> onclick='alert(\'XSS\')'

This code also violates the practice of escaping directly in the HTML building code. Perhaps that is OK, but then some XSS specific tests are definitely in order.

s7eph4n:
> htmlspecialchars is for in the body of tags, not attribute values or in the opening tag itself.

Do you have a link to read up on that? Judging from the existence of the [ENT_QUOTES and ENT_COMPAT flags](https://secure.php.net/manual/en/function.htmlspecialchars.php) I would assume that it is intended to be used also for attribute values. WMF devs [seem to agree](https://github.com/wikimedia/mediawiki/blob/master/includes/parser/Sanitizer.php#L1130).

> This string is not affected by htmlspecialchars, so the XSS is still there

Could you explain? How exactly would you inject the JS code here so that it becomes active?

> This code also violates the practice of escaping directly in the HTML building code.

I am not sure what you mean. How should the code be changed so that it escapes directly in the HTML building code (without spreading the html code chunks all over the place instead of keeping them defined in one place)?

JeroenDeDauw:
> Do you have a link to read up on that? Judging from the existence of the ENT_QUOTES and ENT_COMPAT flags I would assume that it is intended to be used also for attribute values. WMF devs seem to agree.

No, just my memory. The last time that I build HTML like this was long ago, so perhaps I misremember.

> Could you explain? How exactly would you inject the JS code here so that it becomes active?

Something like this? https://3v4l.org/l8FYr

JeroenDeDauw:
Merging as this is already more secure

mwjames:
On the topic of tests and XSS, see [0].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/JSONScript/TestCases/p-0704.json

PR #3176, opened 2018-06-26
Remove use of deprecated wfBCP47() global function

cscott:
This function was deprecated in MW 1.31 and is slated to be hard-deprecated in MW 1.32.

Issue #1365, opened 2016-01-12
3.0.0 release


JeroenDeDauw:
Tracking issue for the 3.0 release.
- Release notes
- [Open issues for 3.0](https://github.com/SemanticMediaWiki/SemanticMediaWiki/milestones/SMW%203.0)
- [3.0 issues already closed](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues?q=milestone%3A%22SMW+3.0%22+is%3Aclosed)
- Changes since 2.5
- Contributors since 2.5 https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/2694#issuecomment-333473589

Breaking changes TODO list:
* https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1014 (includes PHP version bump https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/698)

## [rc.1](https://semver.org/#spec-item-11) target: 2018-09-21, done: 2018-09-23

- [x] Shape rel notes
  - [x] Add missing commits tagged with [releasenotes](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pulls?q=is%3Apr+is%3Aopen+label%3Areleasenotes)
- [x] Update INSTALL.md - not just version number
- [x] Update composer.json
- [x] Update version number
  - [x] extension.json
  - [x] COMPATIBILITY.md
- [x] Create tag
- [x] Update `semantic-mediawiki.org`
  - [x] Run `composer update`
  - [x] Touch `LocalSettings.php`
  - [x] Update the SMW configuration
  - [x] Check [Special:Version](https://www.semantic-mediawiki.org/wiki/Special:Version)
- [x] Ping testers group
- [x] Announce via mail
- [x] Do a release tweet on Twitter

## [rc.2](https://semver.org/#spec-item-11) target: 2018-10-02, done: 2018-10-02

- [x] Shape rel notes
  - [x] Add missing commits tagged with [releasenotes](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pulls?q=is%3Apr+is%3Aopen+label%3Areleasenotes)
- [x] Update version number
  - [x] extension.json
- [x] Create tag
- [x] Update `semantic-mediawiki.org`
  - [x] Run `composer update`
  - [x] Touch `LocalSettings.php`
  - [x] Update the SMW configuration
[...]

PR #3184, opened 2018-07-03
Backport for compat with dev MW

JeroenDeDauw:
Change by @cscott in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3176

### DIFF
--- a/src/Localizer.php
+++ b/src/Localizer.php
@@ -222,7 +222,12 @@ class Localizer {
 	 * @return string
 	 */
 	public static function asBCP47FormattedLanguageCode( $languageCode ) {
-		return wfBCP47( $languageCode );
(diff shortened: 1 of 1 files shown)


kghbln:
I am not sure if we will do a 2.5.7 but if we do this can very well be part of it.

JeroenDeDauw:
Well we really should get to [releasing 3.0 then](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1365) since if we wait much longer there will be no SMW release that fully works with the latest MW release.

kghbln:
After answering here I realized that Stephan back-ported some other stuff. So I will do a farewell release for the 2.5.x series.

Still this back-port avoids deprecation notices. I understand that it will still be possible to use the old code with 1.32 though I am not sure what the difference between a deprecation and a hard-deprecation is. Anyways it is good to have it in here.

Issue #2318, opened 2017-03-07
Migrating code to PHP 5.4+ short arrays


kghbln:
Now that SMW dropped support for PHP 5.4 and earlier the arrays could probably be migrated to use the short syntax, i.e. `array( )` to `[ ]`. Is this something still worth considering for the 2.5.0 release.

I think there is some software out there that helps to automatically convert this.

@mwjames @JeroenDeDauw

mwjames:
> Is this something still worth considering for the 2.5.0 release?

No.

>  be migrated to use the short syntax, i.e. array( ) to [ ]

I only see this as syntactic sugar which brings neither expressiveness nor performance advantages henceforth I'm more or less indifferent of its usage or forced adaptation.

kghbln:
> I only see this as syntactic sugar which brings neither expressiveness nor performance advantages henceforth I'm more or less indifferent of its usage or forced adaptation.

Well, no need to hurry then. Basically we are just talking about reduced file sizes then. I would have expected some other internal improvements tied to this short array syntax but reality ... :)

JeroenDeDauw:
Get https://github.com/FriendsOfPHP/PHP-CS-Fixer (it has a phar, so no real installation required)

    php php-cs-fixer.phar fix src/ --fixers=short_array_syntax

kghbln:
> Get https://github.com/FriendsOfPHP/PHP-CS-Fixer (it has a phar, so no real installation required)

Thanks for the pointer.

JeroenDeDauw:
I could run this on the repo. @mwjames you fine with that? Merge conflicts with open PRs likely...

kghbln:
> I could run this on the repo. @mwjames you fine with that? Merge conflicts with open PRs likely...

You could wait till Monday.

JeroenDeDauw:
What does Monday have to do with anything?

kghbln:
> What does Monday have to do with anything?

On Monday James usually does not have a pull or so in the pipeline so the chances of conflicts are low.

JeroenDeDauw:
Oh. I see you're becoming an expert in The Ways Of James :)

JeroenDeDauw:
https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2336

https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2335

https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2337

[...]

PR #3195, opened 2018-07-10
Convert to PHP short array syntax for /tests/

kghbln:
This PR is made in reference to: #2318 

- Converts the `tests` directory to PHP 5.4+ short array syntax

This PR includes:
- [ ] CI build passed


mwjames:
Please wait with the change until last otherwise I get into trouble
with my PRs and have to spent a lot of time rebasing them.

Cheers

kghbln:
> Please wait with the change until last otherwise I get into trouble with my PRs and have to spent a lot of time rebasing them.

Ok, so just ping as soon as you are ready. I have seen you use short arrays in your PRs but did not think of things you may have prepared in the meantime.

mwjames:
> Ok, so just ping as soon as you are ready. I have seen you use short arrays
> in your PRs but did not think of things you may have prepared in the
> meantime.

I do appreciate the initiative, it is just that those type of PRs
cause a lot of pain because it requires rebasing. Yesterday, I run
into the same issue due to some import optimization that happened
while being away.

kghbln:
> I do appreciate the initiative, it is just that those type of PRs cause a lot of pain because it requires rebasing. Yesterday, I run into the same issue due to some import optimization that happened while being away.

Fair enough, though the mentioned merge e.g. fixed broken builds quite some time ago. So sometimes it is not so easy to avoid. This one however is definitively avoidable to make things easier.

PR #2595, opened 2017-08-06
Special:Admin/Vertical tabs as nav help, refs 2044, 2268

mwjames:
This PR is made in reference to: #2044, #2268

This PR addresses or contains:

- Space is scarce therefore introduce a vertical navigation to allow for a more targeted browsing effort when a specific task needs to be accomplished
- Easily extendable in case an additional task needs to be added to `Special:SemanticMediaWiki`
- Based on https://www.w3schools.com/howto/howto_js_vertical_tabs.asp

image

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

PR #3218, opened 2018-07-15
Special:Admin switch to horizontal tabs, refs 2595, 3216

mwjames:
This PR is made in reference to: #2595, #3216

- Switches tabs from vertical to horizontal

This PR includes:
- [ ] CI build passed

### Example
image


kghbln:
> Slight visual change from before plus the overview contains now all environment settings we expect for issue reported, so users can just copy-and-paste the JSON formatted output such as:

Very nice!

REM self: Also update issue template after docu on wiki.

kghbln:
> Slight visual change from before plus the overview contains now all environment settings we expect for issue reporting, so users can just copy-and-paste the JSON formatted output such as:

@mwjames I think we should have a (section) header here: "Setup information"

PR #3221, opened 2018-07-16
Convert log-show-hide-{$type} messages to logeventslist-{$type}-log

prtksxna:
This PR is made in reference to: https://phabricator.wikimedia.org/T199657

- A follow-up after https://gerrit.wikimedia.org/r/c/mediawiki/core/+/428871
- i18n message changes

This PR includes:
- [x] CI build passed

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -517,6 +517,7 @@
 	"smw-search-profile-tooltip": "Search functions in connection with Semantic MediaWiki",
 	"log-name-smw": "Semantic MediaWiki log",
 	"log-show-hide-smw": "$1 Semantic MediaWiki log",
+	"logeventslist-smw-log": "Semantic MediaWiki log",
 	"log-description-smw": "Activities for [https://www.semantic-mediawiki.org/wiki/Help:Logging enabled event types] that have been reported by Semantic MediaWiki and its components.",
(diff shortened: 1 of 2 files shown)


kghbln:
@prtksxna Thanks a lot for the pull!

mwjames:
The old messages should remain since I suspect the current PR only
takes into account MW 1.32. The next release of SMW is still planned
to support MW 1.27 so removing those message will cause trouble for
non MW 1.32 environments.

Of course, if the new messages are also available for MW 1.27 then
there are no objections.

prtksxna:
> Of course, if the new messages are also available for MW 1.27 then
there are no objections.

I don't think these changes will be backported, so yes, in that case we need to keep the old ones. Give me some time to update this PR? I have raised your concern on the main task as well - https://phabricator.wikimedia.org/T199657#4426776

kghbln:
>  Give me some time to update this PR?

Yes. :)

prtksxna:
Updated PR to not remove old messages.

kghbln:
> Updated PR to not remove old messages.

Cool, thanks a lot!

kghbln:
Back-ported to 2.5.x with 954ff207d58e7ca24dc84d843293c86a9281539b

PR #3274, opened 2018-07-29
Update ResultIterator

mwjames:

This PR includes:
- [x] CI build passed

[skip ci]

### DIFF
--- a/src/Iterators/ResultIterator.php
+++ b/src/Iterators/ResultIterator.php
@@ -35,25 +35,26 @@ class ResultIterator implements Iterator, Countable, SeekableIterator {
 	/**
 	 * @var boolean
 	 */
-	public $isResultWrapper = false;
+	public $numRows = false;
 
 	/**
 	 * @since 2.5
 	 *
-	 * @param ResultWrapper|array $res
+	 * @param Iterator|array $res
 	 */
 	public function __construct( $res ) {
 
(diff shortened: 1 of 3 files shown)


src/Iterators/ResultIterator.php
```diff
@@ -35,25 +35,26 @@ class ResultIterator implements Iterator, Countable, SeekableIterator {
...
 	 * @since 2.5
 	 *
-	 * @param ResultWrapper|array $res
+	 * @param Iterator|array $res
 	 */
 	public function __construct( $res ) {
 
-		if ( $res instanceof ResultWrapper ) {
-			$this->isResultWrapper = true;
+		if ( !$res instanceof Iterator && !is_array( $res ) ) {
```
JeroenDeDauw: Probably more appropriate to check for `Traversable` instead of `Iterator`. Then you can also change the type hint in the doc to `iterable` (supertype of `Traversable` and `array`)

mwjames:
> Probably more appropriate to check for `Traversable` instead of `Iterator`

Good point.

>  type hint in the doc to `iterable` (supertype
> of `Traversable` and `array`)

Doc might be fine, the iterable pseudo-type was only added with PHP
7.1 so any actual type hint would have to wait til 7.1 becomes
mandatory.

PR #3289, opened 2018-08-05
Combine different value lists, support JSON, refs 2295

mwjames:
This PR is made in reference to: #2295

- The content of the "Allows value list" may become lengthy and challenging to edit via wikitext hence this PR adds support for JSON as format. The value list is using a simple key, value assignment with the key (e.g. "PD") being used as constraint value with the assigned value argument is there only to described the key and has for the validation no specific relevance.

This PR includes:
- [x] CI build passed

### Example

`MediaWiki:Smw_allows_list_other_license.json` contains:

```
{
    "CC Zero": "Public Domain Dedication",
    "PD": "Public Domain"
}
```

`Property:Has license` with an assignment of `[[Allows value list::other license.json]]`.

### DIFF
--- a/src/DataValues/AllowsListValue.php
+++ b/src/DataValues/AllowsListValue.php
@@ -50,9 +50,9 @@ class AllowsListValue extends StringValue {
 			foreach ( $allowsListValueParser->getErrors() as $error ) {
 				$this->addErrorMsg( $error );
 			}
+		} else {
+			parent::parseUserValue( $value );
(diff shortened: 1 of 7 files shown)


kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ASpecial_property_Allows_value_list&type=revision&diff=66637&oldid=57295)

PR #3292, opened 2018-08-05
`Allows value` to support bounded intervals, ranges for number, quantity type

mwjames:

- Adds supports for `<`, `>`, and `...` in the `Allows value` declaration
- The listed notations are only valid for a property of `[[Has type::Number]]` and `[[Has type::Quantity]]`
- A simple range can be defined using `>`/`<` as in `[[Allows value::>100]]` or in combination `[[Allows value::>100]]` `[[Allows value::<200]]`
- Defining various intervals should be done using the bounded range `...` indicator such as `[[Allows value::1...20]]` `[[Allows value::100...200]]`
- `[[Allows value::1...20]]` is equivalent to `[[Allows value::>0]]` `[[Allows value::<21]]` yet using `>`/`<` more than once per property specification is not supported therefore the `...` notation should be used to create numeric intervals

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/DataValues/ValueValidators/AllowsListConstraintValueValidator.php
+++ b/src/DataValues/ValueValidators/AllowsListConstraintValueValidator.php
@@ -7,6 +7,7 @@ use SMW\DataValues\ValueParsers\AllowsListValueParser;
 use SMW\PropertySpecificationLookup;
 use SMW\Message;
 use SMWDataValue as DataValue;
+use SMWNumberValue as NumberValue;
 use SMWDIBlob as DIBlob;
 
 /**
@@ -124,6 +125,8 @@ class AllowsListConstraintValueValidator implements ConstraintValueValidator {
 			array_keys( $allowedValueList ), 0 , 10 )
 		);
 
+		$allowedValueList = str_replace( [ '>', '<' ], [ '%3C', '%3E' ], $allowedValueList );
+
 		$dataValue->addErrorMsg(
 			array(
 				'smw_notinenum',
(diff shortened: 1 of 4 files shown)


mwjames:
@kghbln FYI

krabina:
wikidocu available at https://www.semantic-mediawiki.org/wiki/Help:Special_property_Allows_value

krabina:
I think that  ≥ operator should be allowed.  If you use [[Allows value::>0]] ths means that a value greater than zero must be annotated. However, according to the open world assumption of SMW, there is a difference, if NO values is assigend or zero is assigned. 
My usecase: users enter values in forms. If they omit a value, this just means they did not enter anything. If they enter "0" this means that the value of the property should be zero.  [[Allows value::>0]] and [[Allows value::0]] at the same time also does not work.

With [[Allows value::≥0]] one could easily define a value range of positive numbers.

mwjames:
> I think that  ≥ operator should be allowed.  If you use [[Allows value::>0]]
> ths means that a value greater than zero must be annotated. However,

This is a feature request and for anyone interested in extending the
functionality, the check happens here [0].

Extending [0] shouldn't be that difficult and while I see the appeal
to cram in the additional check into [0] it starts to become difficult
to maintain this class as more options may be added in future and
you'll end up with a convoluted class I've seen so often in the MW
environment.

Instead, use the new constraint interface (\#3746) and introduce
something like a `non_negative_integer` constraint [1, 2] so users can
create a schema (or combine different once) and attach them to a
property.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/DataValues/ValueValidators/AllowsListConstraintValueValidator.php#L180-L196
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/architecture/extending.constraint.md
[2] https://www.w3.org/TR/xmlschema11-2/#nonNegativeInteger

Cheers

PR #3293, opened 2018-08-05
Add tanslation page annotation, refs 2300

### LINKED ISSUE #2300: Status of page being `Is translated`?
@Nikerabbit Is it possible that the `Translate` extension can set a page props [0] `ParserOutput::setProperty` property of pages that are (have been) translated via the extension?

It would helpful to other processes or extensions (in this case SMW) to distinguish "real" content from translated content (as a copy of the original) during parsing. If the page props is set then the `ParserAfterTidy` hook can be used to set a SMW property of its own to reflect the status of such page.

[0] https://www.mediawiki.org/wiki/Manual:Page_props_table

mwjames:
This PR is made in reference to: #2300

- Adds support for `translate-translation-page` and `_TRANS`
- `_TRANS` related value annotations are stored as separate subobject, the property chain notation can be used to get access to the values such as:
  - `Translation.Language` (languagecode)
  - `Translation.Translation source` (sourcepagetitle)
  - `Translation.Translation group` (messagegroupid)
-  `_TRANS` is set as default in `'smwgPageSpecialProperties' => array( '_MDAT', '_TRANS' ),` so translation related annotations should work out-of-the box given that the Translation extension sets `translate-translation-page`

This PR includes:
- [x] CI build passed

Fixes #2300

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -847,20 +847,26 @@ return array(
 	'smwgJobQueueWatchlist' => [],
 	##
 
-	### List of enabled special page properties.
-	# Modification date (_MDAT) is enabled by default for backward compatibility.
-	# Extend array to enable other properties:
+	###
+	# List of enabled special page properties.
+	#
+	# - `_MDAT` Modification date is enabled by default for backward compatibility.
+	# - `_TRANS` Add annotations (language, source etc. ) when a page is
+	#   indentified as translation page (as done by the Translation extension)
+	#
+	#  Extend array to enable other properties:
 	#     $smwgPageSpecialProperties[ => '_CDAT',
 	# Or:
 	#     array_merge( $smwgPageSpecialProperties, array( '_CDAT' ) ),
 	# Or rewrite entire array:
 	#     	'smwgPageSpecialProperties' => array( '_MDAT', '_CDAT' ),
+	#
 	# However, DO NOT use `+=' operator! This DOES NOT work:
 	#     $smwgPageSpecialProperties += array( '_MDAT' ),
 	#
 	# @since 1.7
 	##
-	'smwgPageSpecialProperties' => array( '_MDAT' ),
+	'smwgPageSpecialProperties' => array( '_MDAT', '_TRANS' ),
 	##
 
 	###
--- a/i18n/extra/en.json
+++ b/i18n/extra/en.json
@@ -129,7 +129,10 @@
             "_CONT_LANG": "Content language",
             "_CONT_TITLE": "Content title",
             "_CONT_DATE": "Content date",
-            "_CONT_KEYW": "Content keyword"
+            "_CONT_KEYW": "Content keyword",
+            "_TRANS": "Translation",
+            "_TRANS_SOURCE": "Translation source",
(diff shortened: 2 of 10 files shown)


mwjames:
@kghbln FYI

Currently `_TRANS` as set as default whether that should remain so can be discussed before the release. I'm not sure which Translate version does actually send the `translate-translation-page` extension content so maybe ask someone that could know in terms of a corresponding semver version number.

Nikerabbit:
> I'm not sure which Translate version does actually send the translate-translation-page

MLEB 2018.01 and newer, Translate 2018-01-30 and newer. We don't currently use semver.

kghbln:
> MLEB 2018.01 and newer, Translate 2018-01-30 and newer. We don't currently use semver.

@Nikerabbit Thanks for being faster than my ping.

@mwjames Thus we are talking about MW 1.29+

kghbln:
> Currently _TRANS as set as default whether that should remain so can be discussed before the release.

I do not think that this should be enabled by default since this feature depends on an extra extension.

kghbln:
> Good point, in this case removing `_TRANS` from the `smwgPageSpecialProperties` setting in the `DefaultSettings.php` should be sufficient.

Moreover this would avoid having to do extra code voodoo to determine if a version of the "Translate" extension is installed which supports this feature.

>  Maybe add a note about the version required to make `_TRANS` usable and avoid confusion.

Indeed.

mwjames:
It appears that the sandbox has the required Translation extension
installed, so adding annotation values [0] and related queries [1, 2]
seems to work.

[0] https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Parcourir/cl:OlRyYW5zbGF0ZUNhcHRpb24tMkZkZS0yM3RyYW5zLmRl

[1] https://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial:Ask&cl=YzpdTkEKg0AMfFBZs7R4EXIppace-4GoURfjLiQrfr9rpVCEOcxkwswMSRfK2GqiPlMrDBKWkPHqfWFxRhKBialnNbQpbWBM2k1CLQtWVeWuflg1T6yFKdsq2aATsvKe9BtZjC3M4cdt2Y6m3UdIWrKRrIM0DMYZPbj6XvBWiiaUQ4pNc3H1owDc7fl3P8nqRXFcadxbutQznjQsFOKx_AM

[2] https://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial:Ask&cl=YzpNjsEKgzAQRL9IN1i8CLmU0lOP_YHVbExwk8BuxN9vrBQKcxiY4c34IgmrnaWgqzgzAce8WWSGQOhI1GooByihLIFxJrZ933eD8bvUQNKckO5cFRZGbfUiX1ALjrjFn9d0XPwzt1CksS3qAsV7pWpNG06x2sEY6MZ701swK2ONJfcvzOuO6wlaiqNpctSNjybobs-_IiSM-Xr5AQ

kghbln:
> It appears that the sandbox has the required Translation extension installed, so adding annotation values [0] and related queries [1, 2] seems to work.

Affirmative!

PR #3294, opened 2018-08-05
Support #show label generation when template is used, refs 502

### LINKED ISSUE #502: Template with Named Arguments not working with #show
I'm using templates with named arguments to query lists of data. Usually, I use #ask to query that data. But to show some examples (e.g. how the template looks like), I occasionally use #show.
However, it seems that #show doesn't support named arguments at all. When using `named args=yes`, no more parameters are passed and no variables are set (numbered as well as the named arguments).
I created an example on the Referata scratchpad:
http://scratchpad.referata.com/wiki/NamedArgumentPage

mwjames:
This PR is made in reference to: #502

This PR includes:
- [x] CI build passed

Fixes #502

### DIFF
--- a/src/Query/Processor/ParamListProcessor.php
+++ b/src/Query/Processor/ParamListProcessor.php
@@ -73,6 +73,7 @@ class ParamListProcessor {
 
 		$serialization = [
 			'showMode'   => $showMode,
+			'templateArgs' => false,
 			'query'      => '',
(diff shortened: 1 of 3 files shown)


mwjames:
@kghbln This should fix your [0] example.

[0] https://sandbox.semantic-mediawiki.org/wiki/Issue/502

PR #3309, opened 2018-08-11
Process `#-` in print request, refs 481

### LINKED ISSUE #481: Bug with 'further results' when template uses 'named args'
I have a semantic query that uses a template with 'named args' as its result format. When a user clicks on 'further results', the results are malformed, containing no link to the original pages. Going to 'Edit query' reveals that for reasons unknown to me, the parameter "|?#=Page" ("Page" being the name used for the template) has been changed into "?=Page#-". 

MW 1.22.7/ SMW 1.9.2 (SB)

mwjames:
This PR is made in reference to: #481

- Moves the handling of `#` into the `PrintRequest` class and removes special treatment of it from different places 
- To limit the possibilities of a regression in terms of `#` as mainlabel/subject identifier, add `f-0211.json` and `s-0029.json` to cover known uses case of `... |?#=Foo ...`

This PR includes:
- [x] CI build passed

Fixes #481

### DIFF
--- a/includes/query/SMW_Query.php
+++ b/includes/query/SMW_Query.php
@@ -482,9 +482,8 @@ class SMWQuery implements QueryContext {
 		}
 
 		foreach ( $this->getExtraPrintouts() as $printout ) {
-			$serialization = $printout->getSerialisation();
-			if ( $serialization !== '?#' ) {
-				$serialized['printouts'][] = $serialization;
+			if ( ( $serialisation = $printout->getSerialisation() ) !== '' ) {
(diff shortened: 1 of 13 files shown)


mwjames:
Should fix https://sandbox.semantic-mediawiki.org/wiki/Issue/481

@kghbln I tried to be vigilant of not introducing any regression in regards to the usage of "mainlabel=", "mainlabel=-" in combination with the "?# ..." but some extra testing in the field would be appreciated.

PR #3318, opened 2018-08-12
Add smwgPostEditUpdate, refs 3310

mwjames:
This PR is made in reference to: #3310

- Follow-up to #3310 where instead of a socket connection via a special page it uses an API request to invoke extra "work" after an update has been completed as part of an independent transaction
- Uses the `smwtask` API interface with `task=run-joblist` to run listed jobs from the queue without the job scheduler and hereby circumvents the randomness of its execution
- Builds on the existing `PostProcHandler` (used to direct `@annotation` post processing of queries via the API)
- Adds `smwgPostEditUpdate` as __internal__ setting (meaning a normal users shouldn't care for it, yet an administrator can regulate those listed jobs if the necessity arises) to regulate which and how many jobs should be executed as part of a post-edit event with the default being:
  - 'SMW\FulltextSearchTableUpdateJob' => 1,
  - 'SMW\ParserCachePurgeJob' => 5 (5 means that it attempts to execute 5 parked jobs from the queue to increase the likelihood that entities get refreshed in time that have been detected by the `QueryDependencyLinksStore` )

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1260,6 +1260,28 @@ return array(
 	##
 
 	###
+	# Regulates task specific settings for the postEdit process.
+	#
+	# The main objective is to defer secondary updates until after the GET request
+	# has been finalized so that resource requirements are part of an API request
+	# (and not a GET) and hereby ensures that a client remains responsive
+	# independent of the update workload.
+	#
+	# `job.task` specifies jobs that should be executed on a post-edit to run in a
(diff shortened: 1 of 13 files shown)


mwjames:
The CI shows random timeout errors, given that the first run did pass, I'll merge this so it can be part of the weekend sprint and is available on the sandbox during the week.

PR #3393, opened 2018-09-02
MW 1.31+ Highlighter avoid <div> in <span> (causes an extra inline <p> element)

mwjames:

- Appeared with MW 1.31+ (changes to tidy or something else) 
- The `<span><div> ... </div></span>` element structure contains in MW 1.31+ an extra `<p>` element within the those outlined elements and hereby  distorts to entire display (see example)
- Mitigate the issue by relying on `<span><span> ... </span></span>` 

This PR includes:
- [x] CI build passed

## Example

https://sandbox.semantic-mediawiki.org/wiki/Berlin

image
image


mwjames:
@kghbln FYI

kghbln:
Works. :)

```
<span class="smw-highlighter smwttinline" data-type="3" data-state="inline" data-title="Einheitenumrechnung" data-hasqtip="1" aria-describedby="qtip-1"><span class="smwtext">891.85 km²</span></span>
```

PR #3415, opened 2018-09-08
Special:Ask add basic compact view, refs 3180

### LINKED ISSUE #3180: "Special:Ask": Allow to collapse or expand the "ask" form
### Setup and configuration

- MediaWiki | 1.31.0 (63a4e21)13:06, 20. Jun. 2018
- PHP | 7.0.27-0+deb9u1 (apache2handler)
- MariaDB | 10.1.26-MariaDB-0+deb9u1
- Semantic MediaWiki | 3.0.0-alpha (dd07f85) 08:18, 25. Jun. 2018

### Issue
Then using a set query as [shown here](https://sandbox.semantic-mediawiki.org/wiki/SetQueryTemplate) you get to [special ask](https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Requ%C3%AAter/cl:LTVCLTVCSGFzLTIwa2V5d29yZDo6c2V0LTVELTVEL21haW5sYWJlbD0vb2Zmc2V0PTAvZm9ybWF0PWxpc3Qvc2VhcmNobGFiZWw9U2V0) when clicking on a keyword. Depending on the skin used as well as the client used (mobile) one will have to do a lot of scrolling until the results are actually visible on screen. Thus I suggest to merge the grey bar and the "ask" form and make it collapsible. In a new line only "Results 1 – 1    (Previous 200 | Next 200)   (20 | 50 | 100 | 250 | 500)   (JSON | CSV | RSS | RDF)" taken from the "ask" form is shown and right blow the results. I guess having the top section collapsed by default will be nice.

bildschirmfoto vom 2018-06-27 15-51-34

mwjames:
This PR is made in reference to: #3180

- Adds a basic compact view for logged-in users that have enabled the related user preference
- "Thus I suggest to merge the grey bar and the "ask" form and make it collapsible", collapsible was not considered as a feature given that it would require some effort to rearrange the HTML instead a registered user can choose the enable the basic compact view 

This PR includes:
- [x] CI build passed

Fixes #3180


JeroenDeDauw:
Really cool UI improvement. Does this just affect the page when there are search results, or also before a query gets executed? If only after query execution, then I suspect it makes sense to use the compact view by default.

I am a bit confused by the difference between "compact view" and "basic compact view". What about "default view", "compact view" and "minimal view"? For reference, this is what Gmail has:

image

kghbln:
> Really cool UI improvement.

+2

> Really cool UI improvement. Does this just affect the page when there are search results, or also before a query gets executed? If only after query execution, then I suspect it makes sense to use the compact view by default.

From looking at it I'd say you get either default view before and after or compact view before and basic compact view after. I am not sure though.

> What about "default view", "compact view" and "minimal view"?

I am for having this as a default to make sure that anonymous users get the "basic compact view" right way which was the rationale for creating the feature request.

mwjames:
> I am a bit confused by the difference between "compact view" and "basic compact view". 
> I am for having this as a default to make sure that anonymous users get the "basic compact view" right way which was the rationale for creating the feature request.

After some deliberation, I changed the UI so that the user preference is no longer required yet still gain enough space without the need to distinguish between "compact view" and "basic compact view". This is achieved by relying on tabs to hide auxiliary information.

image
 
image

PR #3416, opened 2018-09-08
Property page add <section> support, refs 3308

mwjames:
This PR is made in reference to: #3308

- Since #3308 content on a property page is condensed with tabs allowing to mask auxiliary information to keep the page clean and focused, yet user content can still contain amounts of text that may hinder users from selecting (seeing) the essential facts
- PR adds the `<section>` tag by which users can define parts of a text (on the property page marked with a special class) that should be displayed on a "secondary" tab
- Why `section` and not `div` because:
  - HTML `<section>` [0] tag is not officially supported by the MW Parser
  - Doing a preg_replace on `div` elements is tedious, using `<section>` keeps it simple and allows to keep custom design elements without doing unnecessary DOM parsing in PHP

This PR includes:
- [x] CI build passed

Fixes #

[0] https://www.w3schools.com/tags/tag_section.asp

## Example

image

image

image

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -781,6 +781,7 @@
 	"smw-property-tab-redirects": "Synonyms",
 	"smw-property-tab-subproperties": "Subproperties",
 	"smw-property-tab-errors": "Improper assignments",
+	"smw-property-tab-specification": "... more",
 	"smw-concept-tab-list": "List",
 	"smw-concept-tab-errors": "Errors"
 }
--- a/res/smw/ext.smw.page.css
+++ b/res/smw/ext.smw.page.css
@@ -300,6 +300,10 @@
 	padding-top: 8px;
 }
 
+.smw-property-specification {
+	display: none;
+}
+
 /**
  * Tabs
  */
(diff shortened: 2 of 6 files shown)


mwjames:
Just saw that `LabeledSectionTransclusion` uses `setHooks( 'section' )` to create`<section>` elements and as it happens is installed on the sandbox so what's the alternative?
-  `<smw/section>...</smw/section>` tag to create the HTML `<section>`
- or ...?

The objective was to genuinely create HTML `<section>` elements which in case of the property namespace would get and extra fixed class.

kghbln:
> ... so what's the alternative?

I'd go for `<smwsection>...</smwsection>` or `<sectionsmw>...</sectionsmw>`. I think having a slash somewhere in there is too unusual.

kghbln:
@mwjames Probably you already have a pull for this in the pipeline?

mwjames:
> @mwjames Probably you already have a pull for this in the pipeline?

I gave it some thought and we could check if the `<section>` is already registered, or use a different name but in the end I leave to the user that enable/disable support for the `<section>` tag by  adding `smwgSupportSectionTag`. It is set to true by default and ff for some reason that collides with some other extension,set it to false and SMW stops invoking the hook.

kghbln:
I am not sure if renaming the tags would have been a better solution rather to make operators decide on one or the other extension however let's leave it at this.

PR #3420, opened 2018-09-08
TermParser support compact form in:( ... ), has:..., refs 3157

mwjames:
This PR is made in reference to: #3157

- Only applies to the Special:Search and the completion box (#3419) input where quick matches are the focus and any removal of unnecessary keystrokes helps to find things faster without compromising on the intend of the search
- Adds support for `in:( ... )` to condense a collective condition set so that something like`in:ham in:cheese` (or `in:ham && in:cheese` or as raw condition `[[~~*ham*]] [[~~*cheese*]]`) can be simply expressed as `in:(ham && cheese)` 
- Adds `has:` (and `has:( ... )`) has fixed prefix to transform something like `has:Foo bar` into the equivalent of `[[Foo bar::+]]`

This PR includes:
- [x] CI build passed

### DIFF
--- a/res/smw/special.search/search.input.js
+++ b/res/smw/special.search/search.input.js
@@ -32,7 +32,7 @@
 				};
 
 				// Disable (hide) the MW's search input highlighter
-				if ( context.val().search( /\[|\[\[|in:|not:|phrase:|::/gi ) > -1 ) {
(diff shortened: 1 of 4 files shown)


mwjames:
@kghbln A quick reminder on how convenient the use of the `has:` prefix is in connection with the completion search (of course only when `SMWSearch` is enabled) especially when you quickly try to find a subject that "has" the property annotated.

image

PR #3429, opened 2018-09-15
Special:Ask support different form submit methods

mwjames:

- Filling the Special:Ask form and submitting it for a search on a GET request fails to attach a `#...` anchor which is the reason why any such request jumps to the top after a response which kind of sucks and is annoying
- Adds `$smwgSpecialAskFormSubmitMethod` to switch settings in case the following settings doesn't suite a user group with:
  - SMW 3.0 switches to a POST submit method (`SMW_SASK_SUBMIT_POST`) allowing for any submitted  request to jump directly to a dedicated anchor but loses the ability to copy the URL (a HTTP POST request doesn't create any copyable URL) from the address bar (use the bookmark button)
  - `SMW_SASK_SUBMIT_GET` use GET as it was the default til SMW 3.0-
  - `SMW_SASK_SUBMIT_GET_REDIRECT` is a compromise between GET and POST by being an actual GET but redirects shortly after a request and setting an anchor to jump to a dedicated target, URL remains copyable but it requires two requests one GET and one REDIRECT to achieve the same outcome as a POST request would

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1978,6 +1978,26 @@ return array(
 	##
 
 	##
+	# Special:Ask form submit method
+	#
+	# - SMW_SASK_SUBMIT_POST uses `post` as submit method, allows to jump
+	#   directly to the search result but will not produce any copyable URL
+	#   string (use the result bookmark button instead)
+	#
+	# - SMW_SASK_SUBMIT_GET uses `get` as submit method and was the default
+	#   until 2.5, is not able to jump the search result directly after a submit
+	#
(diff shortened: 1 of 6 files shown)


mwjames:
> a HTTP POST request doesn't create any copyable URL

For some general differences between POST and GET, see [0] and includes:

- GET requests should never be used when dealing with sensitive data
- GET requests have length restrictions
- GET requests is only used to request data (not modify)
- POST requests are never cached
- POST requests do not remain in the browser history
- POST requests cannot be bookmarked
- POST requests have no restrictions on data length

[0] https://www.w3schools.com/tags/ref_httpmethods.asp

PR #3431, opened 2018-09-15
Move Rule NS to smw/schema NS, refs 3019

mwjames:

- Since #3019, the focus shifted and the SMW_NS_RULE namespace with the apparent "rule" semantics is too narrow to allow for adaption in future
- The PR moves all parts from #3019 (content handler etc.) to a more wider namespace SMW_NS_SCHEMA (keeping SMW_NS_RULE for now) where each schema type can be described individually
  - `smwgRuleTypes` -> `smwgSchemaTypes`
    - `LINK_FORMAT_RULE` -> `LINK_FORMAT_SCHEMA`
    - `SEARCH_FORM_DEFINITION_RULE` -> `SEARCH_FORM_SCHEMA`
- I would have preferred to use the literal representation of `Schema:`... as prefix for the namespace but it is already in use by the`EventLogging` extension and it is possible that a user has created a custom NS with that common name, so to avoid any issues with the prefix, it is now called `smw/schema:`

This PR includes:
- [x] CI build passed


mwjames:
@kghbln One of the things I wanted to change before the 3.0 release to avoid having to carry some legacy after it is released. SMW_NS_SCHEMA (smw/schema) has a wider focus and definition space and allows to create specific rule types later without necessarily be bound now to some narrow content interpretation.

kghbln:
@mwjames I wonder if it may be preferable to call the new namespace "Schema_(SMW)" instead of "smw/schema" however I do not know if brackets are prohibited here. I suspect not but one never knows.

mwjames:
> I do not know if brackets are prohibited here. I suspect not but one never knows.

Should not but those brackets feel aesthetically unbalanced when referring to some like `Schema_(SMW):Search profile` compared to `smw/schema:Search profile`.

>I wonder if it may be preferable to call the new namespace "Schema_(SMW)" instead of 

I went through various iterations on the prefix name (even looked for a word that would have a similar connotation) but in the end I just found "smw/schema" the closest compared to `Schema`.

kghbln:
> Should not but those brackets feel aesthetically unbalanced when referring to something like Schema_(SMW):Search profile compared to smw/schema:Search profile.

Fair enough but I still feel that "smw/schema" is too unusual. What about "SMW-Schema"?

> I went through various iterations on the prefix name (even looked for a word that would have a similar connotation) but in the end I just found "smw/schema" the closest compared to Schema.

I just did a little brain storm and the only thing I came up alternatively was "Structure" as in content structure or data structure but probably you thought about this one, too.

mwjames:
Some extra features that came with this PR:

- Follows the general SMW UI content representation
- Quick navigation on tags and types went from top to bottom
- `SMW_NS_SCHEMA` NS supports both JSON and YAML (given that ` Symfony\Component\Yaml\Yaml` is available due to PHP not providing any native YAML support) content
- Content is stored using its own `CONTENT_MODEL_SMW_SCHEMA` (`smw/schema`) model
- Schema validation (if available) happens for each type individually via `$smwgSchemaTypes[...][' validation_schema']` and `JsonSchema\Validator`

### Example

image

kghbln:
Special:Notification gives the following probably because of the notification sent prior to the change for the property: 

```
[4b792a8e9d4e42477cc3cd06] /wiki/Sp%C3%A9cial:Notifications SMW\Exception\PredefinedPropertyLabelMismatchException from line 94 of /../w/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Property.php: There is no predefined property with "_FORMAT_RL".
```
Backtrace:
```
#0 /../w/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Property.php(488): SMW\DIProperty->__construct(string, boolean)
#1 /../w/extensions/SemanticNotifications/src/ChangeNotification/ChangeNotificationPresentationModel.php(117): SMW\DIProperty::newFromUserLabel(string)
#2 /../w/extensions/Echo/includes/formatters/SpecialNotificationsFormatter.php(48): SMW\Notifications\ChangeNotification\ChangeNotificationPresentationModel->getBodyMessage()
#3 /../w/extensions/Echo/includes/formatters/EchoEventFormatter.php(63): SpecialNotificationsFormatter->formatModel(SMW\Notifications\ChangeNotification\ChangeNotificationPresentationModel)
#4 /../w/extensions/Echo/includes/DataOutputFormatter.php(171): EchoEventFormatter->format(EchoEvent)
#5 /../w/extensions/Echo/includes/DataOutputFormatter.php(136): EchoDataOutputFormatter::formatNotification(EchoEvent, User, string, Language)
#6 /../w/extensions/Echo/includes/special/SpecialNotifications.php(59): EchoDataOutputFormatter::formatOutput(EchoNotification, string, User, Language)
#7 /../w/includes/specialpage/SpecialPage.php(522): SpecialNotifications->execute(NULL)
#8 /../w/includes/specialpage/SpecialPageFactory.php(568): SpecialPage->run(NULL)
#9 /../w/includes/MediaWiki.php(288): SpecialPageFactory::executePath(Title, RequestContext)
#10 /../w/includes/MediaWiki.php(861): MediaWiki->performRequest()
#11 /../w/includes/MediaWiki.php(524): MediaWiki->main()
#12 /../w/index.php(42): MediaWiki->run()
#13 {main}
```

mwjames:
> Special:Notification gives the following probably because of the notification sent prior to the change for the property:

This one is fine since `_FORMAT_RL` no longer exists (renamed to `_FORMAT_SCHEMA`) but some old data still has a reference to `_FORMAT_RL` which only applies to the sandbox until those old data are purged.

Of course we could add a try and catch in `ChangeNotification/ChangeNotificationPresentationModel.php(117): SMW\DIProperty::newFromUserLabel(string)` to catch future occurrences.

mwjames:
> Schema validation (if available) happens for each type individually via $smwgSchemaTypes[...][' validation_schema']

In case the validation schema (in the example below the assigned `search-form-schema.v1`) detects some discrepancies then errors are displayed similar to: 

image

mwjames:
> I went through various iterations on the prefix name (even looked for a word that would have a similar connotation) but in the end I just found "smw/schema" the closest compared to Schema.

So, what's all the fuss with that renaming? 

Now, that the namespace contains a more common name it is easier to create addition such as the `VEGA_SCHEMA`. In case of the `VEGA_SCHEMA` it enables to store individual vega schemata [0] to create powerful interactive charts and graphs (given that one has a vega result printer) using the vega grammar.

```
'smwgSchemaTypes' => [
	...
	'VEGA_SCHEMA' => [
		//'validation_schema' => __DIR__ . '/data/schema/vega-schema.v1.json',
		'group' => SMW_SCHEMA_GROUP_VEGA,
		'type_description' => 'smw-schema-description-vega-schema'
	],
],
```

image

image

image

The example produced by the printer are not available to the public but it demonstrates what can be done using the `smw/schema` namespace and its content handler.

[0] https://vega.github.io/vega/

kghbln:
> Now, that the namespace contains a more common name it is easier to create addition such as the VEGA_SCHEMA. In case of the VEGA_SCHEMA it enables to store individual vega schemata [0] to create powerful interactive charts and graphs (given that one has a vega result printer) using the vega grammar.

This is a real big carrot for the future. Great perspective I believe!

kghbln:
@mwjames Namespace Rule 112 and Rule_talk 113 are still being exposed to the end user as whereas "smw/schema" is at 114 and 115 respectively. Since this new namespace was not part of any official release it should probably be ditched so nobody event gets a chance using it.

kghbln:
Also to mention when documenting: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3673

kghbln:
> @mwjames Namespace Rule 112 and Rule_talk 113 are still being exposed to the end user as whereas "smw/schema" is at 114 and 115 respectively. Since this new namespace was not part of any official release it should probably be ditched so nobody event gets a chance using it.

I think we should kick the "Rule" namespace.

kghbln:
[`$smwgSchemaTypes`](https://www.semantic-mediawiki.org/w/index.php?title=Help:$smwgSchemaTypes&oldid=74822) documented for 3.0.0

kghbln:
@mwjames Could you add a [link.format.md](https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/master/src/Schema/docs) to document the `LINK_FORMAT_SCHEMA` also allowing for easy embedding of docu on the wiki. This will be great.

kghbln:
@mwjames Just documended the ["schema link" special property](https://www.semantic-mediawiki.org/wiki/Help:Special_property_Schema_link), however I am not really sure what it holds. I figured probably the link to `SPECIAL_ASK` as specified with this [schema](https://sandbox.semantic-mediawiki.org/wiki/smw/schema:Keyword_link_formatter_rule) but this does not seem to be it.

PR #3440, opened 2018-09-16
Property page, move usage count display to tab

mwjames:

- Moves the count display from the top indicator to an individual tab

This PR includes:
- [x] CI build passed

### Example

image

### DIFF
--- a/res/smw/ext.smw.page.css
+++ b/res/smw/ext.smw.page.css
@@ -304,6 +304,37 @@
 	display: none;
 }
 
+.usage-count, .item-count {
+	display: inline-block;
+    padding: 2px 5px;
+    font-size: 12px;
+    font-weight: 600;
+    line-height: 1;
+    background-color: rgba(27,31,35,0.04);
+    border-radius: 4px;
+    color: #bbb;
+    margin-left: 10px;
+}
+
+#tab-smw-property-value:checked + label.nav-label .usage-count,
+#tab-smw-property-subp:checked + label.nav-label .item-count,
+#tab-smw-property-errp:checked + label.nav-label .item-count,
+#tab-smw-property-redi:checked + label.nav-label .item-count,
+#tab-smw-property-spec:checked + label.nav-label .item-count {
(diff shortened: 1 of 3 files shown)


kghbln:
@mwjames This does not [(yet) work for special properties](https://sandbox.semantic-mediawiki.org/wiki/Attribut:Boolean)

mwjames:
> @mwjames This does not [(yet) work for special
> properties](https://sandbox.semantic-mediawiki.org/wiki/Attribut:Boolean)

I've seen this before which relates to predefined properties (and
appears in connection with the content language) and is unrelated to
the current PR.

kghbln:
> I've seen this before which relates to predefined properties (and appears in connection with the content language) and is unrelated to the current PR.

Ok. I guess this is on your radar. :)

mwjames:
>Ok. I guess this is on your radar. :)

I believe that to be fixed with #3458.

kghbln:
> I believe that to be fixed with #3458.

Affirmative. Thanks a lot!

kghbln:
@mwjames Another thing I just realize. The information about a property being user defined or deployed by the software went missing. I guess this could still be a useful information in the old spot. Moreover we still have the info on the page being protected in this area.

mwjames:
> Another thing I just realize. The information about a property being user defined or deployed by the software went missing. I guess this could still be a useful information in the old spot.

I thought about it but for predefined properties you always have the intro explaining that it is in fact a predefined property so creating an extra indicator seems somewhat redundant.

> Moreover we still have the info on the page being protected in this area, which is good.

The protection is displayed as indicator on the usual place because the state information is separate from the property page.

kghbln:
> I thought about it but for predefined properties you always have the intro explaining that it is in fact a predefined property so creating an extra indicator seems somewhat redundant.

Fair enough, after a second thought I agree. :)

> The protection is displayed as indicator on the usual place because the state information is separate from the property page.

Right. I just mentioned it to convey that the spot would still be useful to put stuff there.

PR #3441, opened 2018-09-16
rebuildData.php add "--revision-mode", refs 1698, 3390

### LINKED ISSUE #1698: Can we make `rebuildData` to operate faster?
For users with a page volume of 20k+(https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1690#issuecomment-228883175, https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1688#issue-161915395) doing a simple `rebuildData` run posses an inherent problem of time vs effort.

The bottleneck for the `rebuildData` script is not running an insert transaction to the DB backend (with either `SQLStore` or `SPARQLStore` active) and it is also not its own `InTextAnnotationParser` (which only takes a fraction of time to process a raw text to filter `[[ ... :: ...]]` annotations and transform them into `DataValues`) but the necessary reliance on the MediaWiki's `Parser` to:
- expand templates and provide access to annotation values in context of that target page that is embedding the template
- resolve the `Parser::Marker` that hides any #parser function (including `#subobject`, `#set`, and `#ask`) after the pre-processing
- process all registered `Parser` functions registered by other extensions (which may add or manipulate the text)

Most time is spend waiting on the `Parser::parse` to return a processed document in form of a `ParserOutput` instance which then can be "send" to the `StoreUpdater` to finalize the DB transaction.
## How about just using the raw text?

As outlined above, using `InTextAnnotationParser` alone is insufficient as this would only result in filtering `[[ ... :: ...]]` annotations, leaving things like `#subobject`, `#set`, and `#ask` hidden from the processing since those need to be parsed by the `Parser::parse` to allow for a correct content interpretation.

Parsing the raw text without involving the `Parser` would further dilute value annotations that are bound to the outcome of a MediaWiki specific context such as `Modification date`, `{{DEFAULTSORT:...}}`, `{{DISPLAYTITLE:...}}` or registered extension that add on-the-fly information during the parse process.
## What about the `--procs` option?

Apparently `runJobs` provides an option to run processes in parallel using the `ForkController` (which requires the PHP posix and pcntl extensions and are only available on Linux/Unix derivatives) and as commented in other threads (https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1690#issuecomment-228896038 and https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/927#issuecomment-89262404) due to this limitation, I personally cannot provide any insides to this option as far as it concerns performance or execution.
## See also
- https://www.semantic-mediawiki.org/wiki/Help:RebuildData.php#Quick_and_slow_progress
- https://www.semantic-mediawiki.org/wiki/Incremental_and_continues_data_rebuild

mwjames:
This PR is made in reference to: #1698, #3390

- #3390 tracks the revision ID for an associated entity hereby allows to identify which content elements in MW relates to the annotation body created in SMW
- Running in  "--revision-mode" will use the revision information and compares the latest title/page revision ID with that of the associated revision ID in SMW hereby allowing to make some assumptions about the content state including:
  - If both revision IDs match then it is assumed that no content divergence occurred.
  - The wiki content (including those embedded annotations) should match on what is stored in SMW for that particular entity.
  - As a result, further processing (especially the parsing of content which is the most costly operation during a data rebuild) of that entity is skipped.
- Skipping the processing of entities where no content divergence is expected should speed-up the rebuild process.
- "--force-update" is added as additional parameter to "force" an update under any circumstance 

This PR includes:
- [x] CI build passed

Fixes #1698

### DIFF
--- a/includes/storage/SQLStore/SMW_Sql3SmwIds.php
+++ b/includes/storage/SQLStore/SMW_Sql3SmwIds.php
@@ -740,6 +740,31 @@ class SMWSql3SmwIds {
 	/**
 	 * @since 3.0
 	 *
+	 * @param string $title
+	 * @param integer $namespace
+	 * @param string $iw
+	 */
+	public function findAssociatedRev( $title, $namespace, $iw = '' ) {
+		$connection = $this->store->getConnection( 'mw.db' );
+
+		$row = $connection->selectRow(
+			self::TABLE_NAME,
+			'smw_rev',
+			[
+				"smw_title =" . $connection->addQuotes( $title ),
+				"smw_namespace =" . $connection->addQuotes( $namespace ),
+				"smw_iw =" . $connection->addQuotes( $iw ),
+				"smw_subobject =''"
+			],
+			__METHOD__
+		);
+
+		return $row === false ? 0 : $row->smw_rev;
+	}
+
+	/**
+	 * @since 3.0
+	 *
 	 * @param integer $sid
 	 * @param integer $sid
 	 */
--- a/maintenance/rebuildData.php
+++ b/maintenance/rebuildData.php
@@ -90,8 +90,11 @@ class RebuildData extends \Maintenance {
 		$this->addOption( 'dispose-outdated', 'Only Remove outdated marked entities (including pending references).', false );
 
 		$this->addOption( 'skip-properties', 'Skip the default properties rebuild (only recommended when successive build steps are used)', false );
-		$this->addOption( 'shallow-update', 'Skip processing of entitites that compare to the last known revision date', false );
+		$this->addOption( 'shallow-update', 'Skip processing of entities that compare to the last known revision date', false );
 		$this->addOption( 'property-statistics', 'Execute `rebuildPropertyStatistics` after the `rebuildData` run has finished.', false );
+
+		$this->addOption( 'force-update', 'Force an update even when an associated revision is known', false );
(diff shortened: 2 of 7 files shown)


kghbln:
@mwjames So this (https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/294774/) kills the option to use this flag in case you are in miser mode including #3390 Basically: What I get out of it is that if you run SMW do not use miser mode.

mwjames:
> (https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/294774/) kills the
> option to use this flag in case you are in miser mode including #3390
> Basically: What I get out of it is that if you run SMW do not use miser
> mode.

This particular change you cited doesn't have any influence on the
functioning of SMW or SESP.

kghbln:
@mwjames I though that smw is relying on the revisions being exposed to the client too.

mwjames:
> @mwjames I though that smw is relying on the revisions being exposed to the
> client too.

Judging from the change cited [0] this is about `{{REVISIONID}} magic
word in miser mode` and is about the magic word. We don't rely on that
magic word, we use the revision object itself to fetch the ID.

[0] https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/294774/

kghbln:
Apparently I have totally misinterpreted this change. I do not know why since it is pretty clear when looking at this. I was already pretty much in grump. Paradise regained here, at least here.

Issue #3134, opened 2018-04-22
XML: error parsing RDF document


jaideraf:
### Setup and configuration

- SMW version: 3.0.0-alpha
- MW version: 1.30.0
- PHP version: 7.0.27
- MariaDB: 10.1.26

### Issue
XML informs error at parsing:
`<property:Datenqualität/Herkunft rdf:resource="&wiki;Der_Datensatz_wurde_basierend_auf_der_ÖK50-2C_Stand_2011_digitalisiert._Es_wurden_alle_Waldbestände_für_die_Gemeinde_Kopfing_erfasst."/>`

Should we escape "/" by `&#47;` or `&#x2F; `or `&sol;`? [0]

### Stack trace
[See sandbox](https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Export_RDF/Lorem_ipsum_Export).

[0] https://en.wikipedia.org/wiki/Slash_(punctuation)#Encoding

mwjames:
> <property:Datenqualität/Herkunft rdf:resource="&wiki;Der_Dat

As far as I can recall that should happen here [0].

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Exporter/ExpResourceMapper.php#L239-L243

kghbln:
Well, considering the age of the code it is at least not a regression for 3.0.0

jaideraf:
Maybe it is not a blocker but, given sandbox is testing 3.0.0-rc.1, this issue is not over yet, [look](https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Export_RDF/Lorem_ipsum_Export).

mwjames:
>Maybe it is not a blocker but, given sandbox is testing 3.0.0-rc.1, this issue is not over yet, look.
> https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Export_RDF/Lorem_ipsum_Export

The issue here is with `[[Datenverantwortliche Stelle – E-Mailkontakt::poststelle.magistratsabteilung33@wien.gv.at]]` where the first `–` is not a normal dash but a Unicode symbol hence it is not recognized as dash in the ASCII format.

jaideraf:
Thanks for the explanation @mwjames

kghbln:
Added at different spots to make sure this does not get overlooked:
[...]

PR #3478, opened 2018-09-24
Adds a "dash" to the array of invalid characters

kghbln:
This PR is made in reference to: #3134 

- Adds a "dash" to the array of invalid characters for usage in property names

This PR includes:
- [ ] CI build passed

Fixes #

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1799,11 +1799,11 @@ return array(
 	# Listed characters are categorized as invalid for a property label and will
 	# result in an error.
 	#
-	# @see #1568, #1638
+	# @see #1568, #1638, 3134
 	#
 	# @since 2.5
 	##
-	'smwgPropertyInvalidCharacterList' => array( '[', ']' , '|' , '<' , '>', '{', '}', '+', '%', "\r", "\n" ),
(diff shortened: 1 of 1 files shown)


kghbln:
@mwjames Everything on go here? :)

mwjames:
I guess you took the right "dash" because from just looking at it you
can't tell the difference.

+1

kghbln:
> I guess you took the right "dash" because from just looking at it you can't tell the difference.

Hopefully since this was the point of the "exercise". It is definitively a dash and not the minus. There are many dashs endash, emdash and even larger dashes. However with the one added being the most popular apart from just using the minus. Anyways sandbox will soon tell.

mwjames:
The invalid character list prevents the creation of annotations,
meaning that you can create a property page with a name that contains
those invalid characters but you won't be able to make any annotation
to it.

kghbln:
> The invalid character list prevents the creation of annotations, meaning that you can create a property page with a name that contains those invalid characters but you won't be able to make any annotation to it.

I was just about to ask. So it worked.

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help:$smwgPropertyInvalidCharacterList&curid=12133&diff=64609&oldid=64593)

PR #3556, opened 2018-12-16
Also run tests with PHP 7.3

JeroenDeDauw:


kghbln:
Thanks a lot. Since it passes ...

mwjames:
> Thanks a lot. Since it passes ...

It did not ([0], see the results of `DB=sqlite; MW=master; PHPUNIT=6.5.* `). 

We have to **revert** the PR because the selected CI job didn't pass [1] prior this PR making the error analysis even more harder now that it mingles with issues that appeared with MW 1.33+.

First order of business is to pass all tests before changing to PHP 7.3.

[0] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/builds/468542813
[1] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/builds/467388862

kghbln:
> We have to revert the PR because the selected CI job didn't pass [1] 

Oops, yeah go ahead. However it showed that it is generally working.

> First order of business is to pass all tests before changing to PHP 7.3.

Indeed. This is why I labelled it 3.1.0 and not 3.0.1

JeroenDeDauw:
Welcome back James! :)

PR #3585, opened 2019-01-06
Fix "strpos(): Non-string needles ...", refs 3579, 3556

mwjames:
This PR is made in reference to: #3579, #3556

This PR addresses or contains:

This PR includes:
- [ ] Tests (unit/integration)
- [x] CI build passed

Fixes #3579

Issue #3579, opened 2019-01-06
PHP 7.3 throws deprecation warning "strpos(): Non-string needles will be interpreted as strings in the future ..."


mwjames:
### Setup and configuration

- SMW version: all
- MW version: all
- PHP version: 7.3+
- DB system (MySQL, Blazegraph, etc.) and version: n/a

### Issue

PHP 7.3 (refs #3556) throws "strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior..." as deprecation warning.

### Stack trace

https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/475665183

```30) SMW\Tests\Integration\JSONScript\JsonTestCaseScriptRunnerTest::testCaseFile with data set "p-0711.json" ('/home/travis/build/SemanticMe...1.json')
strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/SemanticDataValidator.php:419
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/SemanticDataValidator.php:304
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/SemanticDataValidator.php:255
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/JSONScript/ParserTestCaseProcessor.php:134
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/JSONScript/ParserTestCaseProcessor.php:92
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/JSONScript/JsonTestCaseScriptRunnerTest.php:377
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/JSONScript/JsonTestCaseScriptRunnerTest.php:213
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/JsonTestCaseScriptRunner.php:215
[...]

PR #3591, opened 2019-01-11
Enable testing for PHP 7.3

kghbln:
This PR is made in reference to: #3579, #3556, #3585

- Enable testing for PHP 7.3

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ matrix:
   include:
     - env: DB=mysql; MW=REL1_29; PHPUNIT=5.7.*
       php: 7.1
+    - env: DB=mysql; MW=master; PHPUNIT=6.5.*
+      php: 7.3
     - env: DB=mysql; MW=REL1_28; FUSEKI=2.4.0; PHPUNIT=4.8.*
       php: 5.6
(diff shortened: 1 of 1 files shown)


mwjames:
Just as a note, MW master is now voting which means as soon as something erupts on the MediaWiki front the CI will fail.

JeroenDeDauw:
Yeah I realized when merging. I think it is good to notice this. If it happens and the issue can't be dealt with then the thing can always be disabled.

kghbln:
Wasn't this already the case before with sqlite? I actually just swapped sqlite for mysql which appeared to be a bit more realistic to me. Anyways I guess we have no issues adapting this further.

mwjames:
> Wasn't this already the case before with sqlite?

No. See how the `allow_failures` section in the .travis.yml no longer matches relevant entries in the `include` section.

kghbln:
> No. See how the allow_failures section in the .travis.yml no longer matches relevant entries in the include section.

Ah, yeah got it. This however means that the respective line in that section is superfluous since it is not a specified environment any longer.

PR #3596, opened 2019-01-12
Add `smwgConfigFileDir`, refs 3506, 3563

### LINKED ISSUE #3506: 3.0.0-rc.1 is cannot be used because .smw.json cannot be created
### Setup and configuration

- SMW version: 3.0.0-rc.1
- MW version: e53f26bde1d4a7763f243fa8a04f2a51131e995f (master)
- PHP version: 7.1.20-1+ubuntu16.04.1+deb.sury.org+1 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: 10.0.36-MariaDB-0ubuntu0.16.04.1

### Issue
After running update.php and deploying new code, I get the message "missing upgrade key" as shown in https://www.semantic-mediawiki.org/wiki/Help:Upgrade

Output of update.php:
```
Setting upgrade key ...
   ... done.
```
But no file is created. This is because SMW assumes it has write permissions to the code directory, which is not the case when I run update.php. This information should be stored in the database itself or in a location that is writable.

### Steps to reproduce
Switch to different user which does not have a write permission to SMW directory before running update.php (or remove write permissions recursively from the directory)

### LINKED ISSUE #3563: New upgrade key causes problems in farm
### Setup and configuration

- SMW 3.0.0
- MW 1.27.4
- PHP 7.0:
- MySQL 5.7.24

### Issue
The upgrade key is stored in .smw.json in the extension folder. If SMW is used in a farm each update of one wiki invalidates the updates of all other wiki's in the same farm. This makes SMW 3.0.0 unusable in such farms.

mwjames:
This PR is made in reference to: #3506, #3563

- Adds `smwgConfigFileDir` to specify the location (`$smwgConfigFileDir = __DIR__ . "/../.config/{$wgDBname}";`) that is expected to be writable and persistent (by default uses the SMW extension root)
- Raises an `FileNotWritableException` when the file cannot be written
- Using the wiki-id as internal identifier (#3563) in the file to distinguish between different instances
- `smwgUpgradeKey` uses a new schema 'smw:2018-09-01' to identify core changes with other extensions being able to alter the key by adding their own `sesp:..` identifier when modify the DB structure (adding fixed tables etc.)

This change when backported **will cause the upgrade key to be regenerated** now that includes the wiki-id (and new schema) as factor for comparing values.

This PR includes:
- [x] CI build passed

Fixes #3506, Fixes #3563

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -34,6 +34,41 @@ return [
 	##
 
 	###
+	# Configuration directory
+	# @see #3506
+	#
+	# The maintained directory needs to be writable in order for configuration
+	# information to be stored persistently and be accessible for Semantic
+	# MediaWiki throughout its operation.
+	#
+	# You may assign the same directory as in `wgUploadDirectory` (e.g
+	# $smwgConfigFileDir = $wgUploadDirectory;) or select an entire different
+	# location. The default location is the Semantic MediaWiki extension root.
+	#
+	# @since 3.0
+	##
+	'smwgConfigFileDir' => __DIR__,
+	##
+
+	###
+	# Upgrade key
+	#
+	# This key verifies that a correct upgrade (update.php/setupStore.php) path
+	# was selected and hereby ensures a consistent DB setup.
+	#
+	# Whenever a DB table change occurs, modify the key value (e.g. `smw:20...`)
+	# to reflect the requirement for the client to follow the processes as
+	# outlined in the installation manual.
+	#
+	# Once the installer is run, the `.smw.json` will be updated and no longer
+	# cause any exception.
+	#
+	# @since 3.0
+	##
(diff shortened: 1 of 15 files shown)


kghbln:
@mwjames I guess it is save to back-port to 3.0.x?

mwjames:
As explained in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3605#issuecomment-453787418, the sha1 is compromised of different keys/values. The `smwgUpgradeKey` is only one component to allow a developer to force an update when s(he) changes the shape of tables/indices directly.

The #3597 has changed the error message so that `smwgUpgradeKey` is displayed directly to the end-user to improve clarity in case of support questions.

PR #3605, opened 2019-01-12
Conditional create full-text table (default MyISAM)

### LINKED ISSUE #3604: Create `MySIAM` full-text table conditionally
### Setup and configuration

- SMW version: 2.5+
- MW version: all
- PHP version: all
- DB system (MySQL, Blazegraph, etc.) and version: all

refs #1481, #2661, #3559 

### Issue

Now and then it happens that users struggle [n.1, n.2, n.3] to understand the documentation and end-up having trouble mixing table types especially when running the DB with a transaction model [0, 1].

`smw_ft_search` is classified as non-transactional because it uses `MyISAM` [2, 3, 4] by default and you have to adapt the `smwgFulltextSearchTableOptions` setting to switch to InnoDB.

#### Why `MySIAM`?

`smwgFulltextSearchTableOptions` has a note that explains "MySQL version 5.5 or later with only MyISAM and InnoDB storage engines to support full-text search (according to sources) ... MariaDB full-text indexes can be used only with MyISAM and Aria tables, from MariaDB 10.0.5 with InnoDB tables ..." therefore the default setting is `MySIAM` as the lowest common denominator.

[0] https://dev.mysql.com/doc/refman/8.0/en/innodb-transaction-model.html
[1] https://dev.mysql.com/doc/refman/8.0/en/replication-gtids.html
[2] "... When using GTIDs, updates to tables using nontransactional storage engines such as MyISAM cannot be made in the same statement or transaction as updates to tables using transactional storage engines such as InnoDB."
[3] https://dev.mysql.com/doc/refman/5.6/en/replication-gtids-restrictions.html
[4] https://www.percona.com/blog/2013/05/21/replication-in-mysql-5-6-gtids-benefits-and-limitations-part-1/

#### Notes
[n.1] https://www.mediawiki.org/wiki/Topic:Tu4rane5czgppcjn

"... my mediawiki DB into Azure Database for MySql (their new hosted MySql offering). Got blocked on the use of MyISAM fulltext search on smw_ft_search. Can this be migrated to InnoDB soon? Latest MySQLs have support for fulltext in InnoDB ..."

[n.2]  https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/3558#issuecomment-448619267

```
Wikimedia\Rdbms\DBQueryError from line 1457 of /../wiki/includes/libs/rdbms/database/Database.php: A database query error has occurred. Did you forget to run your application's database schema updater after upgrading?
Query: DELETE FROM `smw_ft_search` WHERE s_id = '501'
Function: SMW\SQLStore\PropertyTableIdReferenceDisposer::cleanUpSecondaryReferencesById
Error: 1785 Statement violates GTID consistency: Updates to non-transactional tables can only be done in either autocommitted statements or single-statement transactions, and never in the same statement as updates to transactional tables. (192.168.236.1)
```
[n.3] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/3559#issue-393714480

```
Since it complains about MyISAM
ALTER TABLE smw_ft_search ENGINE=InnoDB;
```

mwjames:
This PR is made in reference to: #3604

- `Installer` checks if `smwgEnabledFulltextSearch` is enabled (which by default is not) and the `Setup` routine will compare the `smwgEnabledFulltextSearch` value as part of the upgrade key so it is ensured once you enable the option that you run `update.php`/`setupStore.php` before able to continue.

This PR includes:
- [x] CI build passed

Fixes #3604

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -65,7 +65,7 @@
        <var name="smwgEntityLookupCacheType" value="hash"/>
        <var name="smwgEnabledDeferredUpdate" value="false"/>
        <var name="smwgImportReqVersion" value="false"/>
-       <var name="smwgEnabledFulltextSearch" value="false"/>
+       <var name="smwgEnabledFulltextSearch" value="true"/>
        <var name="smwgEnabledQueryDependencyLinksStore" value="true"/>
        <var name="smwgQueryResultCacheType" value="hash"/>
        <var name="benchmarkPageCopyCount" value="1000"/>
--- a/src/MediaWiki/Specials/Admin/EntityLookupTaskHandler.php
+++ b/src/MediaWiki/Specials/Admin/EntityLookupTaskHandler.php
@@ -269,23 +269,7 @@ class EntityLookupTaskHandler extends TaskHandler {
 				);
 
 				$formattedRows[$id] = (array)$row;
-
-				$row = $connection->selectRow(
-						SQLStore::FT_SEARCH_TABLE,
-						[
-							's_id',
-							'p_id',
-							'o_text'
-						],
-						[
-							's_id' => $id
-						],
-						__METHOD__
-				);
-
-				if ( $row !== false ) {
-					$references[$id][SQLStore::FT_SEARCH_TABLE] = (array)$row;
-				}
+				$this->addFulltextInfo( $id, $references );
 			}
 		}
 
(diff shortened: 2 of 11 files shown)


mwjames:
Of course it should be obvious that this doens't work retrospect meaning that once the table has been created the PR will not magically remove the it just because `smwgEnabledFulltextSearch` isn't disabled. The conditional creation based on `smwgEnabledFulltextSearch` will only work for new installs or where all tables have been dropped prior.

mwjames:
```
$components = [
	$vars['smwgUpgradeKey'],
	$vars['smwgFixedProperties'],
	$vars['smwgEnabledFulltextSearch'],
	$pageSpecialProperties
];
```

Those keys are used to compute the sha1 which we compare with the one stored in `.smw.json`. Anytime a value changes so will the sha1 indicating that the DB structure has been  (or is going to be) modified hereby making an intervention mandatory.

Issue #3517, opened 2018-10-10
Special:Ask fails with Records field-properties having `Coresponds to::` units on them


ankostis:
### Setup
- Core & extensions download & install date: Sept 2018
- MW version: 1.3.10
- DB (MySQL etc.): mysql
- PHP version: 7.2.10
- SMW version: 2.5.8
- SRF version: 2.5.6
- Browsers and versions (if applicable): Chrome latest

### Issue
When: 
- semantically-searching pages that contain a `[[Has type:: Record || Reference]]` property (e.g `Ref prop`), AND 
- that `Ref prop` has as the 1st  `[[Has fields::]]` property one with `[[Has type::Quantity]]` (e.g. `Field prop`), WHICH
- has at least one unit `[[Corresponds to::]]`, AND
- the query-display results specify that field with the construct `?Ref prop.Field prop`

THEN the search http-request fails, and you get this this stack printed (e.g. inside the `Semantic:Ask` page):

```
[2adc3b5803483af8961ddae0] /wiki/index.php?title=Special%3AAsk&q=%5B%5BCategory%3AEntities%5D%5D%5B%5BFD009%5D%5D&po=%3FRef prop.Field prop%3DLLL%0D%0A&eq=yes&p%5Bformat%5D=datatables&sort_num=&order_num=ASC&p%5Blimit%5D=&p%5Boffset%5D=&p%5Blink%5D=all&p%5Bsort%5D=&p%5Bheaders%5D=show&p%5Bmainlabel%5D=&p%5Bintro%5D=&p%5Boutro%5D=&p%5Bsearchlabel%5D=...+further+results&p%5Bdefault%5D=&p%5Bclass%5D=sortable+wikitable+smwtable&p%5Btheme%5D=bootstrap&eq=yes 
TypeError from line 162 of /var/lib/wiki/extensions/SemanticMediaWiki/src/DataValueFactory.php: 
Argument 2 passed to SMW\DataValueFactory::newDataValueByItem() must be an instance of SMW\DIProperty or null, 
instance of SMWDIBlob given, called in /var/lib/wiki/extensions/SemanticMediaWiki/src/Serializers/QueryResultSerializer.php on line 119

Backtrace:

#0 /var/lib/wiki/extensions/SemanticMediaWiki/src/Serializers/QueryResultSerializer.php(119): SMW\DataValueFactory->newDataValueByItem(SMWDINumber, SMWDIBlob)
[...]

PR #3608, opened 2019-01-12
QueryResultSerializer to handle `_qty` on chained property, refs 3517

mwjames:
This PR is made in reference to: #3517

- Fixes the "TypeError: Argument 2 passed to SMW\DataValueFactory::newDataValueByItem() must be an instance of SMW\DIProperty or null, instance of SMWDIBlob given, called in ...\SemanticMediaWiki\src\Serializers\QueryResultSerializer.php on line 133" caused by the property chain
- Adds integration test with the example from #3517 that uncovered the error
- However it does not resolve the `DataTables` format (initially reported) in how it should interpret the values from the API and display them which is an entire different matter in SRF

This PR includes:
- [x] CI build passed

Fixes #


kghbln:
Back-ported with b17526d6ba398f3060aade0fe7d5128e79503547

PR #3611, opened 2019-01-13
Add `populateHashField.php`

mwjames:

- During an upgrade some elements of a table may require to be updated and pending its importance has to happen before any access to that table occurs but due to the sheer size of the table it could hog resources making the `update.php` run for a considerable time
- This PR provides a mechanism to postpone such tasks and trigger a message block about outstanding tasks
- The PR includes an example implementation for the `smw_hash` conversion task. The initial update of `smw_hash` (#3080) only happens once when switching to 3.0 and with a set execution threshold of `200000` (meaning missing converted rows) the conversion will be part of the normal `update.php` / `setupStore.php` routine, hereafter the process is postponed (the status is saved in `.smw.json`) to allow administrators to quickly run the upgrade while a message will start to appear to remind users and administrators that the upgrade hasn't finished yet and requires immediate action.
- The message cannot be turned off unless actions follow and remove the necessary obstacles. In case of `smw_hash`, the `populateHashField.php` script is provided to finalize the conversion and set the ` "populate.smw_hash_field_complete": true`.
- As previously noted (https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3596#issuecomment-453788765), `.smw.json` will contain more than the upgrade key allowing the system to store state information hereby providing a method to check its software, upgrade, and components.

This PR includes:
- [x] CI build passed

## Example

image

```
{
    "mw-31-00": {
        "upgrade_key": "a10c426c369eddc1bb5bffda68e565c430d2d96a",
        "populate.smw_hash_field_complete": true
    },
    "mw-31-00-sunittest_": {
        "upgrade_key": "01bae0a6f5c40524001a7fa4912de253b5a2f6da",
        "populate.smw_hash_field_complete": true
    }
}
```

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -31,6 +31,7 @@ return [
 	'smwgExtraneousLanguageFileDir' => __DIR__ . '/i18n/extra',
 	'smwgServicesFileDir' => __DIR__ . '/src/Services',
 	'smwgResourceLoaderDefFiles' => [ 'smw' => __DIR__ . '/res/Resources.php' ],
+	'smwgMaintenanceDir' => __DIR__ . '/maintenance',
 	##
 
 	###
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -833,5 +833,7 @@
 	"smw-ask-tab-extra": "Extra",
 	"smw-ask-tab-debug": "Debug",
 	"smw-ask-tab-code": "Code",
-	"smw-helplink-concepts": "https://www.semantic-mediawiki.org/wiki/Help:Concepts"
+	"smw-helplink-concepts": "https://www.semantic-mediawiki.org/wiki/Help:Concepts",
+	"smw-install-incomplete-intro": "The installation (or upgrade) of <b>Semantic MediaWiki</b> has not been finalized and an administrator should run the following tasks to prevent data inconsistencies before users continue to create or alter content.",
+	"smw-install-incomplete-populate-hash-field": "The <code>smw_hash</code> field population was skipped during the setup, the [https://www.semantic-mediawiki.org/wiki/Help:populateHashField.php populateHashField.php] script is required to be executed."
 }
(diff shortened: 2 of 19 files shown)


mwjames:
> execution threshold of 200000

`200000` is more a less a house number I picked so that by normal server hardware standards it should finish the task within 2 minutes tops and hereafter switches to the `populateHashField.php` tasks.

This is mainly of interested for users who have a huge `smw_object_ids` table (with millions of rows) making the `update.php` to monopolize the time required to finish the table update.

We may trust people to run `populateHashField.php` when we tell them to do so but we cannot be sure and reminding them in a persistent way (== annoy users) is the only guarantee we have to ensure administrators actually do follow the task required.

kghbln:
I will "rename" the file from "update key file" to "setup key file" in the docu since this appears to be more appropriate in the light of this enhancement.

kghbln:
>  "setup key file" 

Second thoughts, now "setup information file".

mwjames:
> I will "rename" the file from "update key file" to "setup key file"

Fine by me, I hadn't any name in mind so I generally refer to it as `.smw.json`. For technical reasons that matches the `$GLOBALS['smw.json']` key that holds the content of the file once loaded by `Installer:loadSchema`.

kghbln:
> Fine by me ...

Fair enough. I just figured something a bit more descriptive and more intuitive is good for the end users "marketing wise".

mwjames:
> docu since this appears to be more appropriate in the light of this enhancement.

At first it just contained the upgrade key but I needed a place that to store certain information without relying on a DB connection (also we cannot manipulate `LocalSettings.php` to set some variables) and since `.smw.json` was already in place it seemed a natural choice to extent its application.

mwjames:
> the main page of the wiki which will be accessible: 

To be clear, the message will be shown an every page to really "annoy" users so that administrator gets his act together and runs the job. It is also a warning to the end-user that something hasn't been finished and they should avoid data intensive work.

kghbln:
Back-ported with 8f1177a3d92d5d00fffeb35ce8950647f52b751a

kghbln:
Just upgraded a rather big instance:

```
Checking 'smw_hash' field consistency ...
   ... missing 344933 rows ...
   ... updating document no.       14077114 (100%)
   ... writing the status to the setup information file ... 
   ... done.
```
For 4 Cores, 10 GB RAM it took 7 minutes + another 5 minutes for the first 200000 rows.

kghbln:
@mwjames Not sure if related to the changes connected with this feature, but the size of the database connected to the wiki mentioned in my previous post dropped from 3.3 GB to 2.9 GB. Talking about more efficient data storage. :)

PR #3612, opened 2019-01-13
Add `FieldType::TYPE_ENUM` support

mwjames:

- Adds `FieldType::TYPE_ENUM` to support ENUM as type for fields used by SMW
- Only Postgres and MySQL support ENUM natively, for SQLite we use a simple text field [0]

This PR includes:
- [x] CI build passed

[0] https://stackoverflow.com/questions/5299267/how-to-create-enum-type-in-sqlite


mwjames:
> @mwjames Should or could this be back-ported?

No.

PR #3617, opened 2019-01-13
SemanticDataLookup, use `DISTINCT` for non subject items, refs 2928, 3531

### LINKED ISSUE #3531: EntityLookup->getPropertyValues( null, ..., limit=15 ) fails to return all property values
### Setup and configuration

- SMW version: 3.0.0 (71df044)
- MW version: 1.31.1 (a7064a0)
- PHP version: 7.1.16 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: 5.5.56-MariaDB

### Issue

I have a form created with [PageForms](https://termipankki.nikerabb.it/w/index.php?title=Lomake:Nimitys&action=edit) which has a field defined as `{{{field|sanaluokka|input type=combobox|values from property=Sanaluokka}}}`. When editing [a page](https://termipankki.nikerabb.it/w/index.php?title=Nimitys:kielellinen_merkki&action=formedit)  the dropdown for the field Sanaluokka lists only one option: `adjektiivi`. When viewing [the property page](https://termipankki.nikerabb.it/wiki/Ominaisuus:Sanaluokka), it is clear that there are other options. I have been tracing the thing. The call chain looks like this:

1. PFFormPrinter->formHTML
1. PFFormField::newFromFormFieldTag
1. PFValuesUtils::getAllValuesForProperty 
1. PFValuesUtils::getSMWPropertyValues
1. SMWSQLStore3->getPropertyValues
1. SMW\SQLStore\EntityStore\NativeEntityLookup->getPropertyValues
1. SMWSQLStore3Readers->getPropertyValues
1. SMW\SQLStore\EntityStore\CachingSemanticDataLookup->fetchSemanticData
1. SMW\SQLStore\EntityStore\SemanticDataLookup->fetchSemanticData

For some reason SMW\SQLStore\EntityStore\SemanticDataLookup->fetchSemanticData is generating a query like this
```
SELECT o_blob AS v0, o_hash AS v1 FROM `smw_di_blob` WHERE (p_id='522') LIMIT 15
```
while I think the correct query looks like this:
```
SELECT o_blob AS v0, o_hash AS v1 FROM `smw_fpt_pval` WHERE (s_id='522')
```

I assume this is because $isSubject is false instead of true. $dataItem given in this case is `SMW\DIProperty (Sanaluokka)` instead of `SMW\DIWikiPage (Property:Sanaluokka)`.

I am not sure whether this a bug in SMW or PageForms, but what is the correct way to call `SMWSQLStore3->getPropertyValues` to get list of the all used property values efficiently?

Currently the call parameters are: 
1. NULL
1. SMW\DIProperty (Sanaluokka)
1. SMW\RequestOptions (limit=15)

This looks right according to the [documentation](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/f85cddc2416f5480cde4e31419450f4685ca51b9/src/EntityLookup.php#L49-L64).

This leads me to believe there is a bug on either on `SMW\SQLStore\EntityStore\SemanticDataLookup->fetchSemanticData` or `SMWSQLStore3Readers->getPropertyValues`, which is mostly hidden by the fact that without limit parameter, the query returns all matching rows, which returns correct results in the end, but in an inefficient way.

### Steps to reproduce
1. Create a string property
1. Use the property on 20 different pages with always a different value
1. Create a form which autocompletes a field based on that property values
1. Set `$wgPageFormsMaxAutocompleteValues` to a small value like 5

Observe how only 5 options of the total 20 options are shown.

Possible work around: `$wgPageFormsMaxAutocompleteValues = 99999999999999` (so big number that the limit parameter becomes meaningless and scans all rows anyway.

mwjames:
This PR is made in reference to: #2928, #3531

- #2928 caused a regression with the case described in #3531

This PR includes:
- [x] CI build passed

Fixes #3531

### DIFF
--- a/src/SQLStore/EntityStore/SemanticDataLookup.php
+++ b/src/SQLStore/EntityStore/SemanticDataLookup.php
@@ -220,6 +220,8 @@ class SemanticDataLookup {
 					$query->condition( $query->eq( 'p_id', $extraCondition['p_id'] ) );
 				}
 			}
+		} else {
+			$requestOptions = new RequestOptions();
 		}
 
 		$valueCount = 0;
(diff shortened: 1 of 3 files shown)


kghbln:
Back-ported with 9152f948271393131de7515ffd73417579f6a4ed

kghbln:
> Back-ported with 9152f94

I guess some of the previous tidies need to be backported too or this backport reverted. Not sure if this needs to be in 3.0.x

mwjames:
> I guess some of the previous tidies need to be backported too or this backport reverted. 

Should be #3585.

kghbln:
> Should be #3585.

Thanks for the pointer. Done!

Issue #3640, opened 2019-01-19
Attachment links


mwjames:
### Setup and configuration

- SMW version: all

### Feature

The "Media" in MediaWiki always gets a bit neglected with MW-core file upload and metadata retrieval to be rudimentary at best but nothing we can do about besides making sure users annotate files and images accordingly and have something like SESP to annotate the EXIF data (if available).

What SMW can do to improve the interaction with attachments (files, images) on a per page basis is to have:
- Automatically annotate embedded files and images (`_ATTCH_LINK`)
- Use the existing infrastructure of the `Factbox` to show an additional tab that lists those attachment links as a compact list of links

This will allow a user to uncover attachments easily without having to sift through the raw text.

### Example

![image](https://user-images.githubusercontent.com/1245473/51431412-442f6a00-1c20-11e9-90fe-331e466fda1c.png)

kghbln:
[Documented](https://www.semantic-mediawiki.org/wiki/Help:Attachment_links)

Issue #2552, opened 2017-07-21
Display formats for Page properties: `link`, `nolink`, `pagename`, `fullpagename`


s7eph4n:
Currently (SMW <3.0) a printout for a Page property produces a link with the full page name (incl. namespace) by default. This may be modified using the `link` parameter of the `#ask` function.

To have more influence on the printout format (without resorting to the `template` format) it might be helpful to have display formats for Page type printouts, e.g.
```
|?foo #fullpagename
|?bar #pagename
|?baz #link
|?quok #nolink
```

It should be possible to combine the keywords. The full specification for the current default would then be
```
|?foo #link fullpagename
```

These printout level display formats should override whatever would be produced by default or according to the `link` parameter of the `#ask` function.

mwjames:
* Before working on the this issue, moving all formatting related part to its own `WikiPageValuFormatter` is obligatorily using the `DataValueServiceFactory` (see existing DV examples)
* Can be combined with #2470

mwjames:
This issue been categorized and selected as "Enhancements and features" and since it has not seen any update within the last 90 days, it has now been closed and moved to the "Enhancements and features" board. In case the issue finds a sponsor or is actively worked on it can be reopened.

Issue #2470, opened 2017-05-20
Link to image page instead of an image itself


mwjames:
Relates to http://wikimedia.7.x6.nabble.com/Query-result-how-to-display-links-to-images-instead-of-the-image-itself-td5075002.html

`SMWWikiPageValue` would require some changes to allow for `#-l` to generate only links in cases of an image.

The change to `SMWWikiPageValue` below is a proposal, unfortunately the current class is untested and it would require some refactoring effort before additional conditions are likely to be introduced. In case someone wants to pick-up the work, see https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/CONTRIBUTING.md#feature-requests.

## Example
```
{{#ask: [[Category:Image]]
 |?#-l
 |format=table
 |mainlabel=-
}}
```
```
{{#ask: [[Category:Image]]
 |?#-l
 |format=table
}}
```
```
{{#ask: [[Category:Image]]
 |?Has image#-l
 |format=table
}}
```

## Changes

```
@@ -247,11 +247,11 @@ class SMWWikiPageValue extends SMWDataValue {
 			$this->m_outformat == '-' || !$this->isValid() ||
 			$this->m_caption === '' ) {
 			return $this->m_caption !== false ? $this->m_caption : $this->getWikiValue();
 		}
 
-		if ( $this->m_dataitem->getNamespace() == NS_FILE && $this->m_dataitem->getInterwiki() === '' ) {
+		if ( $this->m_dataitem->getNamespace() == NS_FILE && $this->m_dataitem->getInterwiki() === '' && $this->m_outformat !== '-l' ) {
 			$linkEscape = '';
 			$options = $this->m_outformat === false ? 'frameless|border|text-top|' : str_replace( ';', '|', \Sanitizer::removeHTMLtags( $this->m_outformat ) );
 			$defaultCaption = '|' . $this->getShortCaptionText() . '|' . $options;
 		} else {
 			$linkEscape = ':';
@@ -306,11 +306,11 @@ class SMWWikiPageValue extends SMWDataValue {
 		}
 
 		$caption = $this->m_caption === false ? $this->getShortCaptionText() : $this->m_caption;
 		$caption =  \Sanitizer::removeHTMLtags( $caption );
 
-		if ( $this->getNamespace() == NS_MEDIA ) { // this extra case *is* needed
[...]

PR #3650, opened 2019-01-20
WikiPageValue to support `noimage` output option

mwjames:
This PR is made in reference to: #3640 

- We need the `noimage` option for #3640 to show simple links
- In the past, users have ask that there would want to have a possibility to suppress an image display and instead create simple links for an image, `#noimage` (e.g. `?Has page#noimage`) is introduced to support that request (refs #2552, #2470, https://sourceforge.net/p/semediawiki/mailman/message/35848786/)

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

PR #3656, opened 2019-01-24
Update COMPATIBILITY on master

kghbln:

- Tidies and extends COMPATIBILITY for SMW 3.1.0

This PR includes:
- [x] CI build passed

### DIFF
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -109,13 +109,13 @@ Note that HHVM is only required if you do not use PHP.
 	</tr>
 	<tr>
 		<th>SMW 3.1.x</th>
-		<td><strong><a href="https://php.net/supported-versions.php">7.1.0</a></strong> - latest</td>
-		<td>No support</td>
-		<td><strong><a href="https://www.mediawiki.org/wiki/Version_lifecycle">1.31.0</a></strong> - ?</td>
+		<td><strong><a href="https://php.net/supported-versions.php">7.0.0</a></strong> - latest</td>
+		<td>No support<sup>[t.0]</sup></td>
+		<td><strong><a href="https://www.mediawiki.org/wiki/Version_lifecycle">1.31.0</a></strong> - 1.33.x</td>
 	</tr>
 	<tr>
 		<th>SMW 3.0.x</th>
-		<td><strong><a href="https://php.net/supported-versions.php">5.6.0</a></strong> - latest</td>
+		<td><strong><a href="https://php.net/supported-versions.php">5.6.0</a></strong> - 7.2.x</td>
 		<td>3.5.0 - latest</td>
 		<td><strong><a href="https://www.mediawiki.org/wiki/Version_lifecycle">1.27.0</a></strong> - 1.31.x</td>
 	</tr>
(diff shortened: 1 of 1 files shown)


mwjames:
> @mwjames @JeroenDeDauw @s7eph4n I know that supporting PHP 7.0 is evil however I believe we should adhere to the 1.31.x LTS branch when it comes to PHP support.

Following this logic we then could only switch to the another PHP
version once we switch to the next LTS as lower bound which would what
MW 1.36 (2 years+)?

I could agree on:

- SMW 3.1, MW 1.31 ... PHP 7.0 (but not for SMW 3.2)
- SMW 3.2, MW 1.31 ... requires PHP 7.1 (even though MW 1.31 only
requires PHP 7.0)

PS: PHP 7.0 is already EOL. PHP 7.1 comes with security support til 1 Dec 2019.

Cheers

kghbln:
Fair enough. I was mostly focusing on SMW 3.1.0.

The reason for me to favour PHP 7.0 here is Debian. The current stable is serving PHP 7.0 and the next major release is expected by Mid 2019 and serving PHP 7.3. Give people another quarter to upgrade the OS which means that a lot of people will not be able to use software requiring PHP 7.1 before September 2019. Since SMW 3.1.0 will be reality in Q2.2019 people will probably not use it and even wait for SMW 3.2.0 somewhere to be released in Q4.2019. I do not think we should create software which will most likely be skipped by most users.

MW 1.35 will be the next LTS to be released in June 2020 and I guess that this one will probably still support PHP 7.1 since WMF is currently planning to move to PHP 7.0 (they use Debian too). So my argument was not good here. Let's perhaps focus on OS support and the most popular ones are Debian and Ubuntu. I think MediaWiki will somehow be aboard with a stable version for all situations.

To cut a long story short. I think the following may be great:
- SMW 3.1, MW 1.31 ... PHP 7.0 for Debian 9 (PHP 7.0), Ubuntu 16.04 (PHP 7.0)
- SMW 3.2, MW 1.31 ... PHP 7.2 for Debian 10 (PHP 7.3), Ubuntu 18.04 (PHP 7.2)

docs/COMPATIBILITY.md
```diff
@@ -109,14 +109,14 @@ Note that HHVM is only required if you do not use PHP.
...
+		<td><strong><a href="https://php.net/supported-versions.php">7.0.0</a></strong> - latest</td>
+		<td>No support<sup>[t.0]</sup></td>
+		<td><strong><a href="https://www.mediawiki.org/wiki/Version_lifecycle">1.31.0</a></strong> - 1.33.x</td>
 	</tr>
 	<tr>
 		<th>SMW 3.0.x</th>
-		<td><strong><a href="https://php.net/supported-versions.php">5.6.0</a></strong> - latest</td>
-		<td>3.5.0 - latest</td>
+		<td><strong><a href="https://php.net/supported-versions.php">5.6.0</a></strong> - 7.2.x</td>
+		<td>3.5.0 - 3.30.x</td>
```
JeroenDeDauw: Updated the wrong line

JeroenDeDauw: Oh this is about HHVM... nvm then

kghbln: Yeah, and 3.30 is a hard ending since this is the last version supporting php

JeroenDeDauw: To be honest I'd just drop the HHVM stuff and let people that want to use it (maybe no one) figure out compat based on the PHP versions. That is what I did in https://github.com/JeroenDeDauw/Maps/blob/master/INSTALL.md#platform-compatibility-and-release-status

kghbln: Fair enough, I will use the Maps' INSTALL.md as an inspiration to redo the table and merge them into one excluding HHVM

kghbln: So I reworked the whole document. I is a bit more radical that perhaps expected but I believe the document gained in clarity. I kicked everything below the 1.9.0 release to make the whole thing more sane. Ancient info is still available on wiki.

mwjames: I think you add a class attribute to the table so that when the content is included into a wiki page it use that class instead of the vanilla "i-am-not-a-table-and-i-look-like-...".

mwjames:
Now, that you kicked the HHVM part (?) you should at least have a sentence explaining the circumstances as to "We did support it from ... to ... but since ... it is no longer tested hence we no longer validate its working state ..." (or something like that).

docs/COMPATIBILITY.md
kghbln: Yeah, I guess you saw me fiddle with the compat page the other day. :)

kghbln:
> Now, that you kicked the HHVM part

Done. I guess this can now be squashed and merged. Further tweaking can still be done.

PR #3682, opened 2019-02-02
Remove `IsFileCacheable` hook

mwjames:
This PR is made in reference to: #2847, https://sourceforge.net/p/semediawiki/mailman/message/36531972/

- Introduced as part of #2847 to "Disallow the file cache to avoid skipping the ArticleViewHeader hook on Article::tryFileCache" in connection with `$smwgEnabledQueryDependencyLinksStore` but since we switched on using `RejectParserCacheValue` and no longer have to rely on the `ArticleViewHeader` hook to evict the parser cache we can remove the `IsFileCacheable` hook usage.
- `IsFileCacheable` hook allows an extension to disable file caching on pages

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/MediaWiki/Hooks/HookListener.php
+++ b/src/MediaWiki/Hooks/HookListener.php
@@ -146,24 +146,6 @@ class HookListener {
 	}
 
 	/**
-	 * Hook: Allow an extension to disable file caching on pages
-	 *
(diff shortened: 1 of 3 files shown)


Henryfunk:
Beautiful! Thanks!

PR #3696, opened 2019-02-09
Highlighter, decode <, > in content

mwjames:

- The MW parser injects `<p>` (I suspect that to be MW 1.32+ with remex) elements where `span` contains other elements hereby disrupting entire passages of content as seen below (it worked before MW 1.32 !!)
- To avoid issues with the MW parser, encode `<` and `>` and have the tooltip to decode those to restore the content

This PR includes:
- [x] CI build passed

## Example

https://sandbox.semantic-mediawiki.org/wiki/Berlin

image


kghbln:
@mwjames I guess this may be back-ported.

mwjames:
> @mwjames I guess this may be back-ported.

No, because the decoding part that happens on the JS side relies on
the new `smw.tippy.js` which we won't backport.

kghbln:
> No, because the decoding part that happens on the JS side relies on

Indeed, very true. I actually should have come up with this myself.

PR #3697, opened 2019-02-09
[ES] Monitor replication (per entity)

mwjames:
This PR is made in reference to: #3687 

- After recent incidents on the sandbox whether the replication of entities (aka wikipage) to ES was successful or not, this PR adds the `indexer.monitor.entity.replication` (default `true`) setting to allow to actively monitor the replication status for a page that is being viewed
- The monitoring will distinguish two cases:
  - The page wasn't replicated at all
  - The `modification date` differs (the one stored in the DB and ES) indicating a potential discrepancy of data
- The status monitoring (checking) is deferred using the new `check-es-replication` API task

This PR includes:
- [x] CI build passed

### Missing entity

image

### Modification date discrepancy
image

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -2111,7 +2111,13 @@ return [
 			'job.recovery.retries' => 5,
 
 			// Number of max. retries before the file ingest
-			'job.file.ingest.retries' => 3
+			'job.file.ingest.retries' => 3,
+
+			// Compare and report the status of the replication on a page view
+			// about entities hereby allowing users to immediately comprehend a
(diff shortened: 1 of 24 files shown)


mwjames:
@kghbln For the last couple of hours I have sat down to make the replication more transparent so that users will get information about the status and is guided towards resolving those issues. Of course, if the ES instance has some issues like space or memory it cannot be solved from within SMW but the user would have to contact the administrator instead.

mwjames:
Placeholder at https://www.semantic-mediawiki.org/wiki/Help:Replication_monitoring

PR #3733, opened 2019-02-23
Support for enforced property parent type inheritance, refs 3528

### LINKED ISSUE #3528: Subproperty not inheriting parent property type
### Setup and configuration

- SMW version: 3.0.0
- MW version: 1.31.1
- PHP version: 7.0.32-0ubuntu0.16.04.1 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: 5.7.24-0ubuntu0.16.04.1

### Issue

When adding `[[Subproperty of::...]]` to a property page and based on another page with `[[Has type::Number]]`, I would have expected the child property to inherit the type info from the parent. However, in order to get the property to avoid showing up as a link, as well as to be available in queries, I needed to add `[[Has type::Number]]` to the subproperty page as well.

If this is not a bug, I'd like to request it as a feature.

Thanks for the really awesome tool!

mwjames:
This PR is made in reference to: #3528

- Adds `smwgMandatorySubpropertyParentTypeInheritance` as setting (default `false`) to enforce that a subproperty inherits the same type as its parent property
- When disabled (which is the default, legacy behaviour), a warning will appear on the property page to inform the user about the mismatch
- When enabled and the user added a different type (in references to the parent property) an error message will be displayed and the annotated type is forcibly removed (the page content isn't altered which is why the error message appears to ensure the user removes the type or modifies it to match the parent property type)
- https://www.semantic-mediawiki.org/wiki/Help:Mandatory_parent_type_inheritance

This PR includes:
- [x] CI build passed

Fixes #3528


kghbln:
**Done**
- [docu](https://www.semantic-mediawiki.org/wiki/Help:Mandatory_parent_datatype_inheritance)

**To do**
- [parameter](https://www.semantic-mediawiki.org/wiki/Help:$smwgMandatorySubpropertyParentTypeInheritance)

PR #3738, opened 2019-02-24
Support `in.maintenance_mode`, refs 3596, 3715

mwjames:
This PR is made in reference to: #3596, #3715

- Adds `in.maintenance_mode` to provide a better feedback to users when the upgrade is ongoing meaning when the setup/update script runs hereby communicates that things are going to take a moment instead of just displaying the error message

This PR includes:
- [x] CI build passed

## No maintenance activities

image

## Setup/update in-progress

image


mwjames:
@kghbln I just switched a test wiki (2.5 to 3.x) with 4 mil rows (in the `smw_object_ids` table) and due to the smw_id/s_id conversion to  (int 11, `#3360`) it took about an hour (well the hardware wasn't much up-to-date) anyway, instead of continuing to show for an hour the error message it now shows the the second screen when the actual update is in progress.

Issue #3743, opened 2019-02-27
Introduce Property:Implies namespace


Larivact:
### Setup and configuration

- SMW version: 3.0.1

### Issue
It is not uncommon to have a property of datatype Page that is only meant for one namespace (for example `Has user` or `Has file`). There are currently two issues with such properties: redundant syntax and the possibility of improper values.

`[[Has user::User:Name]]` is redundant. While templates can help with this, there is no way to prevent redundancy when using the `+sep` parameter to allow something like `{{Foo|users=Bar, Baz}}` (at least not without an additional extension providing something like a `#arraymaptemplate` parser function).
Because the namespace is not implied it can be set to something improper. While this can be remedied with `Allows pattern`, it would be nicer if improper values were impossible.

I therefore propose the introduction of a `Implies namespace` special property for properties of the `Page` datatype. The implication mechanism would simply prefix the implied namespace when it is missing. Ideally both implied and explicit syntax would work for both annotating and querying.

ankostis:
Wouldn't be better to be more generic, and have a validation template or rule(like #2273) that either returns nothing(ok), or something(invalid-message)?

kghbln:
Thanks for suggesting an enhancement.

mwjames:
>Wouldn't be better to be more generic, and specify with that new special property a validation template or rule(like #2273) that either returns nothing(ok), or something(invalid-message)?

I concur that introducing a more generic constraint check model is the way forward instead of introducing different special properties to cater for various use cases (unless they support a very broad scenario). So, while I understand the redundancy argument, we should avoid manipulating data by some hidden magic (unit and unit conversion is different in that respect) and instead make sure a user is informed about inconsistency of his/her data and can taken appropriate action to remove those obstacles that hinders the data storage.

#3746 contains an example and approach most likely to be available.

mwjames:
> I therefore propose the introduction of a Implies namespace special property for properties of the Page datatype. 

Thanks for the general proposal and your engagement with the project.

As outline above, it is fairly unlikely that we will support "Implies namespace special property" in SMW-core (for reasons stated). Of course any custom datatype can implement such a support in an extension if the implementer requires additional restriction or supplement functionality.

PR #3746, opened 2019-03-02
Add `PROPERTY_CONSTRAINT_SCHEMA`, refs 3743

mwjames:
This PR is made in reference to: #3431, #3743

- Adds `PROPERTY_CONSTRAINT_SCHEMA` as a new schema type
- Introduces an approach by assigning a `[[Constraint schema::...]]` to a property hereby allows to define additional constraints without relying on other special properties as long as the check behaviour doesn't alter the value
- Currently, I'm a bit fuzzy on the expected schema notation, for now the example shown in the lower section parses and validates against the ` data/schema/property-constraint-schema.v1.json`

This PR includes:
- [x] CI build passed

Fixes #

## Tasks

- [x] Write tests
- [x] Clarify schema notation

## Example 

See #3743 for a request to restrict annotations to only the NS_USER namespace.

```
{
    "type": "PROPERTY_CONSTRAINT_SCHEMA",
    "constraints": {
            "allowed_namespaces": [
                "NS_USER"
            ]
    },
    "tags": [
        "property constraint"
    ]
}
```
image

image


mwjames:
> Currently, I'm a bit fuzzy on the expected schema notation

@s7eph4n @kghbln @krabina @jamesmontalvo3 @jaideraf 

I'd like to solicit some feedback (see the rational here [0]) on the proposed constraint schema approach as to:

- Do we have other known use cases that would fit the schema?
- Does the syntax/notation is expressive enough for SMW advanced users to be able to maintain such constraint schemata
- ...

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/3743#issuecomment-468921870

mwjames:
Instead of defining a `PROPERTY_CONSTRAINT_SCHEMA ` as a type for the `smw/schema` namespace, one could create a `smw/constraint` namespace and host all of those constraint related JSON definitions there.

jaideraf:
First, I think this type of restriction (only values from a specific namespace) can be achieved by using the `Allows pattern` special property [0] with a regular expression like `^User:.*$`. It is not what the issue creator wished, however, I think the solution proposed here have almost the same effect as using the `Allows pattern` solution: a message saying that that value violates the [namespace ou regular expression] constraint.

On the other hand, restricting values from a SMW query (or Concept or Category) would be interesting (like Page forms and Semantic Forms Select do with `|values from concept=`, `|values from category=` and `|query=`). 

> Do we have other known use cases that would fit the schema?

I think cardinality constraint and domain and range restrictions would fit as another “rule”.

#### Property Cardinality Restrictions [1]

* Property X must have 3 values assigned (`cardinality = 3`)
* Property Y must have at least 2 values assigned (`minCardinality = 2`), it can have more values – without limit.
* Property Z can have a maximum of 5 values assigned (`maxCardinality = 5`)
* Property W can have a minimum of 3 values assigned and maximum of 10 values assigned (`minCardinality = 3` and `maxCardinality = 10`)

#### Domain and Range Restrictions [2]

* Property (or Predicate) X must have a Subject that belongs to Category (or Concept, or Namespace  - or Class) Y (`rdfs:domain = Y`).
* Property (or Predicate) X must have a value (or Object) that belongs to Category (or Concept, or Namespace  - or Class) Z (`rdfs:range = Z`).
* Property (or Predicate) X must have a Subject that belongs to Category (or Concept, or Namespace  - or Class) Y and a value (or Object) that belongs to Category (or Concept, or Namespace  - or Class) Z (`rdfs:domain = Y` and `rdfs:range = Z`).

#### xsd:nonNegativeInteger [3]

I think `xsd:nonNegativeInteger` is already achived by using `[[Allows value::>0]]`

[0] https://www.semantic-mediawiki.org/wiki/Help:Special_property_Allows_pattern 
[1] https://www.w3.org/TR/2012/REC-owl2-primer-20121211/
[2] https://www.w3.org/TR/rdf-schema/#ch_properties
[3] https://www.w3.org/TR/2012/REC-owl2-primer-20121211/#Datatypes

mwjames:
@jaideraf Thanks

> First, I think this type of restriction (only values from a specific namespace) can be achieved by using the Allows pattern special property [0] with a regular expression like ^User:.*$.

Yes, that can be done using `Allows pattern`. The #3746 was merely a trigger to search for a more generic approach as to how to define constraints.

As for `Allows pattern` or any other content stored in the `MediaWiki` namespace, storing SMW administrative and operative content in that NS has always felt inappropriate (we don't own the NS and we have no control over it especially if someone decides to cast additional restriction on that NS) where we cannot run any validation to check for its consistency.

As I see it, there are three broad levels of constraints:

- `annotation_constraint` which is limited to a single value annotation without any dependency on any other value of the same property within the same entity space
- `property_constraint` which represents interdependencies of values to the same property instance within the same entity space (aka subject); for example, local cardinality would be such a constraint where you restrict value occurrences within the same entity
- `entity_constraint` is a broad constraint for a property beyond the individual entity, the uniqueness constraint [0] comes to mind

> Property Cardinality Restrictions 

The question is are those restrictions local (per entity) or global (wiki-wide). A per entity cardinality should be relatively cheap to check as you work with an in-memory representation, a wiki-wide check on the other hand will be rather expensive since you have to query the store (use the DB) in order to find existing entities with matching characteristics.

Given that you assign a constraint to a property you end up with something like:

Property:Foo → `[[Property constraint::Constraint:Foo]]` → `Constraint:Foo` is defined with:
```
    "constraints": {
        "cardinality_constraints": {
            "cardinality": 3,
            "min_cardinality": 2,
            "max_cardinality": 5
        }
    }
```

> Domain and Range Restrictions [2]

Good examples, not sure the effort and how best to express those rules with a simple language.

> On the other hand, restricting values from a SMW query (or Concept or Category) would be interesting (like Page forms and Semantic Forms Select do with |values from concept=, |values from category= and |query=).

Unless you run a `ElasticStore` where queries are cheap (because I have seen people run pathological queries like `[[Foo::!Bar]]` (get me all except ...) without any other restrictions allowing such queries to run on each value check would be expensive and risky. You have to remember each check has to run for each value annotation that uses a property with a constraint set therefore I would be hesitant to allow users to define such constraint sets through a `#ask` query.

[0] https://www.semantic-mediawiki.org/wiki/Help:Special_property_Has_uniqueness_constraint

PS: The `Rule` namespace (in the beginning of SMW 3 dev) was designed to host those kind of definitions but since I didn't have enough time to iron out the "language", `smw/schema` was used as a replacement to allow for a bit more flexibility when introducing different schemata as JSON.

mwjames:
This PR only provides a single constraint implementation (`allowed_namespaces` [0, 1]) to showcase how to use the `Constaint schema` with the newly introduced interfaces. [2] describes how to introduce new constraints.

[0] https://sandbox.semantic-mediawiki.org/wiki/Issue/3746
[1] https://sandbox.semantic-mediawiki.org/wiki/smw/schema:Constraint:User_namespace
[2] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/architecture/extending.constraint.md

PR #3749, opened 2019-03-03
Add schema group import, refs 2874

mwjames:
This PR is made in reference to: #2874

- Adds `PROPERTY_GROUP_SCHEMA` for isolating the management of group assignments via a schema
- `import/smw.groups.json`/`schema.json` define the import for creating a group for schema related properties

This PR includes:
- [x] CI build passed

Fixes #

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -2058,6 +2058,11 @@ return [
 			'validation_schema' => __DIR__ . '/data/schema/search-form-schema.v1.json',
 			'group' => SMW_SCHEMA_GROUP_SEARCH_FORM,
 			'type_description' => 'smw-schema-description-search-form-schema'
+		],
+		'PROPERTY_GROUP_SCHEMA' => [
+			'validation_schema' => __DIR__ . '/data/schema/property-group-schema.v1.json',
+			'group' => SMW_SCHEMA_GROUP_PROPERTY,
+			'type_description' => 'smw-schema-description-property-group-schema'
 		]
 	],
 	##
--- a/data/import/groups/schema.json
+++ b/data/import/groups/schema.json
@@ -1,0 +1,22 @@
+{
+    "type": "PROPERTY_GROUP_SCHEMA",
+    "manifest_version": 1,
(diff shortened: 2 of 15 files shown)


mwjames:
https://www.semantic-mediawiki.org/wiki/Help:Schema/Type/PROPERTY_GROUP_SCHEMA

mwjames:
As with the `schema.json` import example (contains group definition for `_SCHEMA*` related properties), the intent is to have a deployment mechanism where properties can be grouped (SESP properties etc.) without a user having to manually create categories and annotated them with `Is property group`.

A  `Is property group` definition to a category (and hereby property) takes precedence over any group definition found in `PROPERTY_GROUP_SCHEMA`, so a property listed as part of a `PROPERTY_GROUP_SCHEMA` where a listed property itself is assigned to category and that category `Is property group`, the property will be a member of the category assigned group.

kghbln:
[Examples](https://sandbox.semantic-mediawiki.org/wiki/Special:SearchByProperty/:Schema-20type/PROPERTY-5FGROUP-5FSCHEMA)

mwjames:
> Examples

+1 but I have been scratching my head why the validation schema didn't kick in since you were using an old schema and then it occurred to me that the `justinrainbow/json-schema` (A library to validate a json schema) is missing from the sandbox [0]. `justinrainbow/json-schema` is part of MW so I'm not sure why this is missing but we need throw an error on those schemata that require a validation instead of silently ignoring it.

[0] https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:Version
[1] https://github.com/wikimedia/mediawiki/blob/master/composer.json#L73

kghbln:
> since you were using an old schema

This shortly came to my mind since e.g. manifest was missing. Since I did not see this in the other examples and the setup worked ....

> the `justinrainbow/json-schema` (A library to validate a json schema) is missing from the sandbox

Probably because both MW and SMW only do a "require-dev" for it. Will probably not hurt to put it in the "require" section on the SMW end.

PR #3751, opened 2019-03-03
Declare [?, *, !] as invalid property characters, refs 2282

mwjames:
This PR is made in reference to: #2282

- Extends the list of invalid characters (`?`, `*`, `!`) for a property name which means that creating annotations that contain `[[Foo!::...]]`,`[[F*oo::...]]`, `[[?Foo::...]]`, or any combination of it will no longer be excepted
- `?`, `*`, `!` are part of the query language and can easily be confused or mixed when expressing  the intention
- Property names should be clear, simple, expressive to convey the intention using special characters should be avoided as they dilute the representation, of course users are free the modify `smwgPropertyInvalidCharacterList` but it is not recommended

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1825,11 +1825,15 @@ return [
 	# Listed characters are categorized as invalid for a property label and will
 	# result in an error.
 	#
-	# @see #1568, #1638, 3134
+	# @see #1568, #1638, #3134
 	#
 	# @since 2.5
 	##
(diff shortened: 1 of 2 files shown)


mwjames:
@kghbln FYI

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgPropertyInvalidCharacterList&type=revision&diff=68625&oldid=68624)

PR #3792, opened 2019-03-16
Recognize error type

mwjames:
This PR is made in reference to: #3746

- Adds the `ProcessingError` interface, allowing to distinguish certain error types (`constraint`, `restriction` etc.) and store them using the `_ERR_TYPE` property to make it possible the query errors by type
- Adds `_ERR_TYPE` as predefined property 

This PR includes:
- [x] CI build passed


kghbln:
[Documented](https://www.semantic-mediawiki.org/wiki/Help:Special_property_Processing_error_type)

@mwjames Restrictions errors could probably be triggered by "Allows value", "Allows pattern" and "Uniqueness constraint". Etc. types could be "query" for query errors ("Has query").

I do not think that restriction errors are triggered. Currently we only have constraint errors being identified by type?

mwjames:
> I do not think that restriction errors are triggered. Currently we only have
> constraint errors being identified by type?

Yes, only constraint errors are identified. You may create a tracking
issue to investigate whether tracking other error types is possible or
not hereby making them discoverable or better distinguishable.

Issue #3728, opened 2019-02-21
Introduce Property:External formatter interwiki


Larivact:
### Setup and configuration

- SMW version: 3.0.1

### Issue

It would be handy if you could use existing interwiki prefixes with [Datatype "External identifier"](https://www.semantic-mediawiki.org/wiki/Help:Type_External_identifier), so that you don't have to duplicate the URL with [Property:External formatter URI](https://www.semantic-mediawiki.org/wiki/Help:Special_property_External_formatter_URI), which could be achieved with a new `External formatter interwiki` special property.

Example: `[[External formatter interwiki::bug]]`

kghbln:
Thanks for suggesting.

mwjames:
>It would be handy if you could use existing interwiki prefixes with Datatype "External identifier", so that you don't have to duplicate the URL with Property:External formatter URI, which could be achieved with a new External formatter interwiki special property.

I understand the proposal, I'm unsure whether it is worth the effort
or not in creating an additional formatter just to avoid maintaining
an URL with a place holder.

In case someone wants to work on it:

- Avert any performance issues with the interwiki table lookup (cache
the lookup etc.)
- The interwiki table lookup needs to be isolated from the rest of SMW
by adding a new class (e.g. `InterwikiLookup`) to the `SMW\MediaWiki`
namespace to avoid SMW objects "talking" directly to a MediaWiki table
- Add necessary unit and integration tests

PS: I will review any code (given above points have been followed)
that tries to tackle the request but I won't implement it myself.

Larivact:
I just had a brilliant idea how this can be solved without any code changes:

    [[External formatter uri::{{fullurl:bug:$1}}]]

If you confirm my account request, I can add it as a tip to the wiki.

kghbln:
> I just had a brilliant idea how this can be solved without any code changes:

That's even better.

> If you confirm my account request, I can add it as a tip to the wiki.

Done. :)

Larivact:
Thanks, [added it to the wiki](https://www.semantic-mediawiki.org/wiki/Help:Special_property_External_formatter_URI#Reusing_interwiki_prefixes). Closing this.

PR #3818, opened 2019-03-17
External identifier to support multi substitutes using {...}, refs 3728

mwjames:
This PR is made in reference to: #3728 

- For example, an external canonical URL can have different parameters to represent different version of the content as in case of Wikipedia and in order to pinpoint the version additional parameters are necessary.
- Support multi substitutes using `{...}` to declare the parameter substitution scheme
- For a property `wikipedia` with:
  - `[[Has type::External identifier]]`
  - `[[External formatter uri::https://en.wikipedia.org/w/index.php?title=$1&oldid=$2]]`
  - `[[wikipedia::Truid Aagesen{837787373}]]` (you can also add a space in between `Truid Aagesen {83 ...}` for better readability), `{...}` identifies the 2nd...n parameters and will resolve as:
    - `https://en.wikipedia.org/w/index.php?title=Truid_Aagesen&oldid=837787373`

This PR includes:
- [x] CI build passed


mwjames:
@kghbln FYI

PR #3823, opened 2019-03-20
Use local copy of 'jquery.async'

Krinkle:
The repo has its own copy of this module defined in `/res/Resources.php` which is good as the one in core is been deprecated, to be removed in the next release (https://phabricator.wikimedia.org/T209699)

Update this line to refer to that module instead so that things keep working :)


mwjames:
Thanks.

PR #3841, opened 2019-03-23
UnitConverter to use EntityCache

mwjames:

- Remove reference to `CachedPropertyValuesPrefetcher` and only rely on `PropertySpecificationLookup` and `EntityCache`, the `EntityCache` invalidation for an entity (like a property page) will invalidate all sub keys associated with an object

This PR includes:
- [x] CI build passed


src/Services/DataValueServices.php
```diff
@@ -48,6 +49,26 @@
  */
 return [
 
+	/**
+	 * UnitConverter
+	 *
+	 * @return callable
+	 */
+	'UnitConverter' => function( $containerBuilder ) {
```
JeroenDeDauw: I think the MW services style approach for object graph construction, which is used here, is not so good maintenance wise. No time to describe this now, shoot me a mail if you want some more explanation

PR #3903, opened 2019-04-07
Add `PluggableDataType`

mwjames:

- Until recently (#3899) it was not possible for custom types to control the construction of corresonding DV instances leaving out any possibility for dependency injection or the necessity to rely on global state.  The new `PluggableDataType` allows for custom types to define relevant fields together with how a DV instance should be constructed.

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/PluggableDataType.php
+++ b/src/PluggableDataType.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace SMW;
+
+/**
+ * @license GNU GPL v2+
+ * @since 3.1
+ *
+ * @author mwjames
+ */
+interface PluggableDataType {
+
+	/**
+	 * Returns the type_id and is by convention for an third-party extension
+	 * indicated by two underscores such as `__some_type`.
+	 *
+	 * @since 3.1
+	 *
+	 * @return string
+	 */
+	public function getTypeId();
+
+	/**
+	 * Returns the class name or callable associated with the type_id.
+	 *
+	 * @since 3.1
+	 *
+	 * @return string|callable
+	 */
+	public function getClass();
+
+	/**
+	 * Returns the DataItem::TYPE_...
+	 *
+	 * @since 3.1
+	 *
+	 * @return string
+	 */
+	public function getItemType();
+
+	/**
+	 * Returns a label or false for types that cannot be accessed by users.
+	 *
+	 * @since 3.1
+	 *
+	 * @return string
+	 */
+	public function getLabel();
+
+	/**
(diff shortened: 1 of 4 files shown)


mwjames:
https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/e88c4d94a8466ee69bd50cb4344fae6e11a91d28/docs/technical/code-snippets/register.custom.datatype.md contains detail on the `PluggableDataType` usage.

src/PluggableDataType.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @return string
+	 */
+	public function getLabel();
+
+	/**
+	 * Returns possible aliases for a type
+	 *
+	 * @since 3.1
+	 *
+	 * @return []
```
JeroenDeDauw: string[]

tests/phpunit/Unit/DataTypeRegistryTest.php
```diff
@@ -81,6 +81,25 @@ public function testRegisterDatatypeWithCallable() {
 		);
 	}
 
+	public function testRegisterPluggableDataType() {
+
+		$pluggableDataType = $this->getMockBuilder( '\SMW\PluggableDataType' )
+			->disableOriginalConstructor()
+			->getMock();
```
JeroenDeDauw:
```php
$this->createMock( PluggableDataType::class );
```

src/PluggableDataType.php
```diff
@@ -0,0 +1,90 @@
...
+	public function getTypeId();
+
+	/**
+	 * Returns the class name or callable associated with the type_id.
+	 *
+	 * @since 3.1
+	 *
+	 * @return string|callable
+	 */
+	public function getClass();
```
JeroenDeDauw: I'm having some problems understanding what this is about, which suggests the naming is not super.

JeroenDeDauw:
Kinda painful to see new extension interfaces without scalar and return types. Maybe worth waiting till 3.2 so these can be used. Unfortunately you can't add them after the fact without creating a breaking change.

JeroenDeDauw:
@mwjames what do you think? Can we push this to 3.2? (That release can be in a few months rather than in a year if we want)

mwjames:
> what do you think? Can we push this to 3.2? 

Fine by me.

JeroenDeDauw:
Hehe, 3.2 will be released in 5 days anyway :D

image

PR #3905, opened 2019-04-07
Add `purgeEntityCache.php`

mwjames:

- Adds `purgeEntityCache.php` to purge all cache entries (including associates) that used the `EntityCache` interface 

This PR includes:
- [x] CI build passed

Fixes #


kghbln:
@mwjames I believe that this one should not be run on a regular basis. However a bit more information about this script will be nice. Anyways I guess the `--report-runtime` and `--with-maintenance-log` options could be useful here.

mwjames:
> @mwjames I believe that this one should not be run on a regular basis. However a bit more information about this script will be nice

This is an expert script which should only be executed by someone who wants to purge all entity cache entries at once so that requests normally relying on a cache entry are fetched from the DB.

kghbln:
> This is an expert script which should only be executed by someone who wants to purge all entity cache entries at once so that requests normally relying on a cache entry are fetched from the DB.

So I assumed right. :) - [Documented](https://www.semantic-mediawiki.org/wiki/Help:Maintenance_script_purgeEntityCache.php)

PR #3981, opened 2019-05-04
Add `must_exists` constraint, refs 3746

mwjames:
This PR is made in reference to: #3746

- Adds the `must_exists` constraint and is only relevant to the `Page` type where its checks whether the value exists or not before it can be used as annotation element for a property that has the constraint attached.

This PR includes:
- [x] CI build passed

## Example

```json
{
    "type": "PROPERTY_CONSTRAINT_SCHEMA",
    "constraints": {
        "must_exists": true
    },
    "tags": [
        "property constraint",
        "must exists"
    ]
}
```

### DIFF
--- a/data/schema/property-constraint-schema.v1.json
+++ b/data/schema/property-constraint-schema.v1.json
@@ -50,6 +50,9 @@
 				},
 				"non_negative_integer": {
 					"$ref": "#/definitions/non_negative_integer"
+				},
+				"must_exists": {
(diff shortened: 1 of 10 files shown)


mwjames:
- https://sandbox.semantic-mediawiki.org/wiki/Attribut:Page_exists
- https://sandbox.semantic-mediawiki.org/wiki/Issue/3981

PR #3968, opened 2019-04-28
Display `constraint` errors using an indicator

mwjames:
This PR is made in reference to: #3746, #3792

This PR addresses or contains:

- Errors of type `constraint` can be quickly accessed via an indicator to signal to a user that some data require attention 

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

## Example

image

PR #3989, opened 2019-05-05
Extend `ErrorLookup` to scan subobjects, cache lookup, refs 3968

mwjames:
This PR is made in reference to: #3968, #3987 

- Extends `ErrorLookup` to also scan on subjects when `smwgCheckForConstraintErrors` is set to `SMW_CONSTRAINT_ERR_CHECK_ALL` (default setting)
- Adds caching to `ConstraintErrorIndicatorProvider` so that repeated page views will avoid active DB reads
- Default lookup limit is set to 20

This PR includes:
- [x] CI build passed

## Example

image


kghbln:
https://www.semantic-mediawiki.org/wiki/Help:$smwgCheckForConstraintErrors

PR #4042, opened 2019-05-25
Support `#` as formatting directive for no link, refs 1855, 4037

### LINKED ISSUE #4037: Use @@@ syntax for property links without actually linking.
### Feature request
The [@@@ syntax to generate a property link](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1855) feature is great, but there are several use cases where in your wiki you don't want users to visit property pages.

Therefore it would be great to be able to use the feature of having the preferred property label and property description showing up in a tooltip, but without the possibility to actually click on it an got to the property page. 

This could be achieved by the syntax
`[[Foo::@@@|@@@]]` (where `|@@@]]` indicates a no link)

A workaround is to use something like
`{{#show:Property:Foo|?has preferred property label|+lang={{USERLANGUAGECODE}}}}`
to display the correct name of the poperty and then also do the tooltip in a similar way which is a lot of coding and querying in templates...

This is what it can look like when done manually:
grafik

As there is no link present, users don't know that they can hover over the property names. Therefore it might be better to have it undlined as with the Semantic Glossary extension.
grafik

mwjames:
This PR is made in reference to: #4037

- Defines `#` to describe a no link state in connection with `@@@` as in `[[Foo::@@@|#]`

This PR includes:
- [x] CI build passed

Fixes #4037

### DIFF
--- a/includes/datavalues/SMW_DV_WikiPage.php
+++ b/includes/datavalues/SMW_DV_WikiPage.php
@@ -274,7 +274,18 @@ class SMWWikiPageValue extends SMWDataValue {
 		if ( is_null( $linked ) || $linked === false ||
 			$this->m_outformat == '-' || !$this->isValid() ||
 			$this->m_caption === '' ) {
-			return $this->m_caption !== false ? $this->m_caption : $this->getWikiValue();
+			$text = $this->m_caption !== false ? $this->m_caption : $this->getWikiValue();
+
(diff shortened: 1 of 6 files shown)


krabina:
documented here: https://www.semantic-mediawiki.org/wiki/Help:Property_links

PR #3277, opened 2018-08-04
Tidy (TraversalPropertyLookup), refs 2461

mwjames:
This PR is made in reference to: #2461 

This PR addresses or contains:

This PR includes:
- [ ] Tests (unit/integration)
- [x] CI build passed

PR #4063, opened 2019-06-01
Add prefetch cache and lookup, refs 3722

### LINKED ISSUE #3722: Prefetch lookup
When Semantic MediaWiki is resolving a query request (i.e. resolving the parser function `{{#ask ... }}` or `{{#show: ... }}`) it has two distinct tasks to accomplish:
- First, build a query condition in a corresponding language (SQL, ES-DSL, SPARQL) that is understood by the assigned `QueryEngine` where it returns a list of subjects that match the condition (similar to finding rows in a table that match a `WHERE` condition)
- Secondly, build a `QueryResult` object that contains the matched subjects and provide information on the requested printouts (equivalent of columns displayed in a table)

## Issue

The performance of the first task (i.e. retrieving and sorting results that match the condition) mostly depends on the selected `QueryEngine` (and can vary to quite a degree especially when complex disjunctive queries or negation conditions return a large pool of possible results and need to be sorted) where any optimizations is more or less independent of how Semantic MediaWiki is functioning or the `SQLStore` is used as primary data store.

The second task of retrieving printouts for each column and each matched subject is currently designed in a way that each such request creates a dedicated SQL SELECT statement in the `SQLStore` to only fetch those items that are required for the display and means that a result set of 200 subjects with a request for three different properties (e.g `{{#ask: [[__condition__]] |?Foo |?Bar |?Foobar }}`) requires 200 x 3 SELECT requests to fetch all relevant data.

The approach to be able to selectively retrieve data for a single subject does make sense when viewing a page or compare the data during the pre/post storage process but for a bulk (or list) operation it is a fairly sub-optimal solution.

## Solution

To significantly reduce the amount of DB queries required when resolving printrequests yet produce the same result output a refactoring of `SemanticDataLookup` (and `PropertySubjectsLookup` for inverse lookups) is necessary to introduce a so-called prefetch mode (or lookup) to be able to “bulk” load all values at once for a list of subjects (using the `WHERE IN` construct).

Some preliminary tests showed:

Current select method
```
Queries (424)
MediaWiki: 1.31.0
PHP: 7.1.1
Time: 3.40497
Memory: 26.71 MB (Peak: 26.75 MB)
```

Prefetch lookup
```
Queries (30)
MediaWiki: 1.31.0
PHP: 7.1.1
Time: 2.93601
Memory: 26.54 MB (Peak: 26.58 MB)
```

## Notes

image

mwjames:
This PR is made in reference to: #3722

- As outline in #3722 the prefetch mode is assumed to significantly reduce DB queries when retrieving values for #ask printouts
- Adds the `SMW_QUERYRESULT_PREFETCH` flag to `smwgExperimentalFeatures` in order to make the prefetch lookup the default mode and in case some issues occurs allow to easily remove the flag hereby switches back to the legacy mode (#4056, i.e. the mode that was used until now)

This PR includes:
- [x] CI build passed

Fixes #3722

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1873,9 +1873,12 @@ return [
 	# to the desired target state and hereby automatically retires the related
 	# setting.
 	#
+	# - SMW_QUERYRESULT_PREFETCH to use the prefetch method to retrieve row
+	# related items for a `QueryResult`.
+	#
 	# @since 3.0
 	##
-	'smwgExperimentalFeatures' => false,
+	'smwgExperimentalFeatures' => SMW_QUERYRESULT_PREFETCH,
 	##
 
 	##
--- a/includes/storage/SMW_QueryResult.php
+++ b/includes/storage/SMW_QueryResult.php
@@ -122,6 +122,9 @@ class SMWQueryResult {
 
 		$itemFetcher = new ItemFetcher( $store, $this->mResults );
 
+		// Used temporarily to allow switching back while testing
+		$itemFetcher->setPrefetchFlag( $GLOBALS['smwgExperimentalFeatures'] );
+
 		// Init the instance here so the value cache is shared and hereby avoids
 		// a static declaration
 		$this->resultFieldMatchFinder = new ResultFieldMatchFinder(
--- a/src/Defines.php
+++ b/src/Defines.php
@@ -204,6 +204,7 @@ define( 'SMW_RF_TEMPLATE_OUTSEP', 2 ); // #2022 Enable 2.5 behaviour for templat
 /**@{
   * Constants for $smwgExperimentalFeatures
   */
+define( 'SMW_QUERYRESULT_PREFETCH', 2 );
 /**@}*/
 
 /**@{
--- a/src/MediaWiki/Connection/Query.php
+++ b/src/MediaWiki/Connection/Query.php
@@ -294,7 +294,11 @@ class Query {
 	 * @param string $value
 	 */
 	public function option( $key, $value ) {
-		$this->options[$key] = $value;
+		if ( $value === null ) {
+			unset( $this->options[$key] );
+		} else {
+			$this->options[$key] = $value;
+		}
 	}
 
 	/**
--- a/src/Query/Result/ItemFetcher.php
+++ b/src/Query/Result/ItemFetcher.php
@@ -29,7 +29,7 @@ class ItemFetcher {
 	private $store;
 
 	/**
-	 * @var prefetchCache
+	 * @var PrefetchCache
 	 */
 	private $prefetchCache;
 
@@ -51,7 +51,7 @@ class ItemFetcher {
 	/**
 	 * @var boolean
 	 */
-	private $prefetch = false;
+	private $prefetch = true;
 
 	/**
 	 * @since 3.1
@@ -61,6 +61,15 @@ class ItemFetcher {
 	public function __construct( Store $store, array $dataItems = [] ) {
 		$this->store = $store;
 		$this->dataItems = $dataItems;
+	}
+
+	/**
+	 * @since 3.1
+	 *
+	 * @param int $features
+	 */
+	public function setPrefetchFlag( $features ) {
+		$this->prefetch = ( (int)$features & SMW_QUERYRESULT_PREFETCH ) != 0;
 	}
 
 	/**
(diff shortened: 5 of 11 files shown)


mwjames:
I used a special data set (~3500 pages) to create an assortment of data and queries to monitor and benchmark the change.

Legacy mode
Console (1) Queries (772) Debug log (53) RequestPHP includes (857) MediaWiki: 1.31.0 PHP: 7.1.1Time: 5.89055 Memory: 32.25 MB (Peak: 32.29 MB)

Prefetch mode
Console (1) Queries (65) Debug log (52) RequestPHP includes (859) MediaWiki: 1.31.0 PHP: 7.1.1Time: 4.64988 Memory: 32.58 MB (Peak: 32.62 MB)

@kghbln We can import the data (NFL statistics data fetched from [0]) into the sandbox, if we wanted to but the export XML file is about 11.5 MB of size.

[0] https://github.com/ryurko/nflscrapR-data

kghbln:
@mwjames I changed the virtual host and configuration of the wiki to allow for files up to 20M. The restrictive setup was an inheritance for the old server which is no longer necessary on the new server. The warning is at 2M.

Adding or removing the `SMW_QUERYRESULT_PREFETCH` flag does not require to run "setupStore.php"?!

mwjames:
> Adding or removing the SMW_QUERYRESULT_PREFETCH flag does not require to run "setupStore.php"?!

Running `setupStore.php` is not required because this PR doesn't change the DB (i.e. tables or indices), it changes how the SQL is executed by collecting all subjects of a #ask query into a list and runs a WHERE IN (instead of single selects for each subject) to return a collective list of values (== doing a prefetch) to temporarily push them into a `PrefetchCache` which then is used as lookup resource. It is why the comment above notes "Queries (65)" and differs from "Queries (772)" in the legacy mode for a single #ask query.

> @kghbln We can import the data (NFL statistics data fetched from [0]) into the sandbox, if we wanted to but the export XML file is about 11.5 MB of size.

[mw-31-00-ql-20190601190838.zip](https://github.com/SemanticMediaWiki/SemanticMediaWiki/files/3244697/mw-31-00-ql-20190601190838.zip)

kghbln:
> Running "setupStore.php" is not required because ...

Thanks for confirming. [Now documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgExperimentalFeatures&type=revision&diff=69833&oldid=69830)

mwjames:
> Thanks for confirming. Now documented.

As for `smwgExperimentalFeatures`, `SMW_SQLSTORE_TRAVERSAL_PROPERTY_LOOKUP` matured with #3277 and is therefore no longer experimental (== removed the flag).

kghbln:
> As for `smwgExperimentalFeatures`

Ouch, one hour of docu time lost. :( I will change the page accordingly. Basically this configuration parameter was introduced in 3.0.0 without a setting.

mwjames:
> Ouch, one hour of docu time lost. :( I

Sorry about that, `smwgExperimentalFeatures` is a general purpose setting to allow to add feature flags without introducing different settings each time a new feature rolls in.

kghbln:
> The NFL data set uses an extra NS.

Since 3000 was already taken I witched to 3300. Moreover the data was unusable because of diverging content languages. My first attempt to fix this made things even worse. Anyways I will revert and start a fresh attempt.

kghbln:
> Anyways I will revert and start a fresh attempt.

Oh, what a mess. Well this happens if you work on three things in parallel. Will to the rest tomorrow.

kghbln:
This is starting to be a disaster. ReplaceText replaces until 1000 occurrences and MassEditRegex until 300 (the latter I did only realize after I hit 2012). So I will be reverting for a 5th time. If the server script provided by ReplaceText does not do the job then I may conclude that we cannot use the data on the wiki since I cannot change the page language.

mwjames:
> This is starting to be a disaster. ReplaceText replaces until 1000
> occurrences and MassEditRegex until 300 (the latter I did only realize after
> I hit 2012). So I will be reverting for a 5th time. If the server script

What about changing the XML file directly (instead of doing it
on-wiki) by replacing the string in file:

- [[Category:NFL:Data]] [[Category:Season:2009]]
- [[Category:NFL:Data]] [[Category:Season:2010]]
- [[Category:NFL:Data]] [[Category:Season:2011]]
- [[Category:NFL:Data]] [[Category:Season:2012]]

with

{{interlanguagelink:en|{{FULLPAGENAME}} }} [[Category:NFL:Data]]
[[Category:Season:2009]] ...

and reimport the content?

kghbln:
> What about changing the XML file directly

This would have been the last resort. However the "replaceAll.php" script provided by ReplaceText proved to be up for the job, i.e. it is possible to edit a much higher number of pages (guess that only memory is the limit here). I believe the data is now on we wiki in the right language mapping.

PR #4078, opened 2019-06-08
Optimize #show to use specialized query lookup, refs 3588

### LINKED ISSUE #3588: Default values for SMW {{#show:}} calls reliably do not work on non-existing pages if the page we ask data from has any incoming properties.
### Setup and configuration

- SMW version: 3.0.0
- MW version: 1.31.1
- PHP version: 7.0.30-0+deb9u1 (apache2handler)
- DB system (MySQL, Blazegraph, etc.) and version: 10.1.37-MariaDB-0+deb9u1

### Issue
Default values for SMW {{#show:}} calls reliably do not work if the page we ask data from has any incoming properties. An example can be seen here: https://wiki.liquipedia.space/wiki/index.php/SMW_Default_Value
```
==Mentionned Page==
{{#show:SMW Mentionned Page|?has test|default=no data}}
==Not Mentionned Page==
{{#show:SMW Not Mentionned Page|?has test|default=no data}}
```
Both `SMW Mentionned Page` And `SMW Not Mentionned Page` page do not exist, but `SMW Mentionned Page` has incoming data as another page has the page title set as a property: https://wiki.liquipedia.space/wiki/index.php?title=SMW_Mentioning_Page&action=edit
See also https://wiki.liquipedia.space/wiki/index.php/Special:Browse?title=Special%3ABrowse&article=SMW+Mentionned+Page and https://wiki.liquipedia.space/wiki/index.php/Special:Browse?title=Special%3ABrowse&article=SMW+Not+Mentionned+Page

In https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Query/ResultPrinters/ResultPrinter.php#L326 `$results->getCount()` will return 1 for `SMW Mentionned Page`, even though the page does not exist at all.

### Steps to reproduce
I included a minimal testcase above, relevant URLs are 
* https://wiki.liquipedia.space/wiki/index.php/SMW_Default_Value
* https://wiki.liquipedia.space/wiki/index.php?title=SMW_Mentioning_Page&action=edit
* https://wiki.liquipedia.space/wiki/index.php/Special:Browse?title=Special%3ABrowse&article=SMW+Mentionned+Page
* https://wiki.liquipedia.space/wiki/index.php/Special:Browse?title=Special%3ABrowse&article=SMW+Not+Mentionned+Page

mwjames:
This PR is made in reference to: #3588

- `#show` always expects a single entity (subject) as condition and instead of building a specific SQL (or any other query related DSL) condition and run it through the `QueryEngine`, use the fact that it is a single subject, circumvent the query building (via the `QueryEngine`), and immediately return a `QueryResult` object
- `#show` may be used to make a judgement about whether the requested entity "exists" or not (#3588), the `SingleEntityQueryLookup` will check if an associated revision exists and in case it does not it will return an empty `QueryResult` making the assumption (and so the `default=...` printout) about an existential statement always to apply (this will only be done for `#show`)
- Using `#show` in connection with `SMW_SHOWPARSER_USE_CURTAILMENT` should reduce the amount of queries required now that the `QueryEngine` is no longer used to resolve the query condition
- Adds `SMW_SHOWPARSER_USE_CURTAILMENT` as `smwgExperimentalFeatures` flag (== default) to curtail on how `#show` query results are retrieved
  - The flag is added to revert to the previous behaviour state in case of some unforeseen behaviour change that hasn't been tested
  - The flag should be removed by the next release given that the feature has been declared as stable and the functionality as been promoted to be the default behaviour

This PR includes:
- [x] CI build passed

Fixes #3588

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -1876,9 +1876,13 @@ return [
 	# - SMW_QUERYRESULT_PREFETCH to use the prefetch method to retrieve row
 	# related items for a `QueryResult`.
 	#
+	# - SMW_SHOWPARSER_USE_CURTAILMENT to use a short cut and circumventing the
+	# `QueryEngine` and directly access the DB since `#show` will always only
+	# request an output for one particular entity.
+	#
 	# @since 3.0
 	##
-	'smwgExperimentalFeatures' => SMW_QUERYRESULT_PREFETCH,
+	'smwgExperimentalFeatures' => SMW_QUERYRESULT_PREFETCH | SMW_SHOWPARSER_USE_CURTAILMENT,
 	##
 
 	##
--- a/includes/query/SMW_QueryProcessor.php
+++ b/includes/query/SMW_QueryProcessor.php
@@ -376,11 +376,20 @@ class SMWQueryProcessor implements QueryContext {
 			$query->setLimit( 0 );
 		}
 
-		$res = self::getStoreFromParams( $params )->getQueryResult( $query );
+		$querySourceFactory = ApplicationFactory::getInstance()->getQuerySourceFactory();
+		$source = $params['source']->getValue();
+
+		if ( $source === '' && $context === self::CURTAILMENT_MODE ) {
+			$querySource = $querySourceFactory->newSingleEntityQueryLookup();
+		} else {
+			$querySource = $querySourceFactory->get( $source );
+		}
+
+		$res = $querySource->getQueryResult( $query );
 		$start = microtime( true );
 
 		if ( $res instanceof SMWQueryResult && $query->getOption( 'calc.result_hash' ) ) {
-			$query->setOption( 'result_hash', $res->getHash( 'quick' ) );
+			$query->setOption( 'result_hash', $res->getHash( SMWQueryResult::QUICK_HASH ) );
 		}
 
 		if ( ( $query->querymode == SMWQuery::MODE_INSTANCES ) ||
@@ -414,10 +423,6 @@ class SMWQueryProcessor implements QueryContext {
 
 			return $result;
 		}
-	}
-
-	private static function getStoreFromParams( array $params ) {
-		return ApplicationFactory::getInstance()->getQuerySourceFactory()->get( $params['source']->getValue() );
 	}
 
 	/**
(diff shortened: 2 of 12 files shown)


mwjames:
@kghbln FYI

With this PR, there will be a deliberate behaviour change between `#ask` and `#show` which is that the "existential statement" about an entity (`{{#show: Foo |?Bar |default=Foo not exists}}`) can be relied upon when using `#show`, `#ask` is not expected to be used for such applications.

kghbln:
To cut it really short: For `#show` the printout to `default=` will always be shown whereas `#ask` will only print in case of a result.

> The flag should be removed by the next release given that the feature has been declared as stable and the functionality as been promoted to be the default behaviour

I personally would wait for a year minimum, i.e. SMW 3.4.x or even better until SMW 4.x since we can never know how people use `#show`. I can imagine cases there `#show` is used to either fill a table field etc. or show nothing if nothing was actually found for it.

mwjames:
> For #show the printout to default= will always be shown whereas #ask will only print in case of a result.

No.

If the subject in question (e.g. `Foo`) exists,  a possible default will __not__ be set but if `Foo` doesn't exists in terms of an associated revision then the result will and the selected result printer may handle the default output. `#ask` remains and has changed its behaviour while `#show` now has an explicit check [0] to make things consistent on the previous assumed behaviour independent of the selected result printer. The reason why #3588 was working in the past was how the result printer interpreted the results which could vary with the implementation and since the list printers were refactored in 3.0, #3588 was no longer consistent with previous assumed behaviour.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/SQLStore/Lookup/SingleEntityQueryLookup.php#L71-L75

kghbln:
> No.

Ah, yes, now I get it. Currently four parties in parallel. Yeah, this is nice!

mwjames:
> terms of an associated revision then the result printer may handle the default output. #ask remains and has not changed its behaviour while #show now

Previously, `{{#show: Foo |default=...}}` and `{{#ask: [[Foo]] |default=...}}` would behave the same (they used the same application logic), now `{{#show: Foo |default=...}}` can behave differently to `{{#ask: [[Foo]] |default=...}}` given that `#show` explicitly checks whether `Foo` has an associated revision or not.

Hence using a query to answer an existential question (`{{#show: Foo |default=...}}`), `#show` is preferred over `#ask` as outlined above.

kghbln:
Partially [documented](https://www.semantic-mediawiki.org/w/index.php?title=Help:$smwgExperimentalFeatures&diff=80818&oldid=69835)

PR #4114, opened 2019-06-29
[ES] Add `rebuildElasticMissingDocuments.php`, refs 3697

mwjames:
This PR is made in reference to: #3697

- Adds `rebuildElasticMissingDocuments` script to find missing entities (aka documents) from Elasticsearch and schedule `smw.update` jobs for those identified subjects

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -2195,7 +2195,7 @@ return [
 			// possible discrepancy between the stored on-wiki data and the data
 			// replicated to Elasticsearch.
 			'monitor.entity.replication' => true,
-			'monitor.entity.replication.cache.lifetime' => 3660,
+			'monitor.entity.replication.cache_lifetime' => 3660,
 		],
 		'query' => [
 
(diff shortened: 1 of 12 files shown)


mwjames:
@kghbln FYI

mwjames:
The difference between the `rebuildData` and `rebuildElasticIndex` script is that `rebuildElasticMissingDocuments` will only compare IDs and schedule a `smw.update` job if something is missing on the Elasticsearch side. This allows the script to be faster in comparison to the other two especially when the `smw_object_ids` table has a lot of rows.

mwjames:
This is part of the replication monitoring [0] effort to ensure that both (SQL and Elasticsearch) systems are in sync given that both rely on an active HTTP communication and can therefore be disrupted.

[0] https://www.semantic-mediawiki.org/wiki/Help:Replication_monitoring

kghbln:
From reading this I assume that this one can be run via cron on a daily basis?

mwjames:
I depends how fast the servers are and the size of the `smw_object_ids` table because the script requires to load the entire `smw_objects_ids` table and compare each ID (in terms of MDAT and associated revision ID) with the ES backend stored information. The comparison is fairly quick but when you have 5 Million rows in the `smw_objects_ids` table and schedule the script for every hour then you do more harm than good.

> From reading this I assume that this one can be run via cron on a daily basis?

I'd say, if the scripts runs less than 5 min then running it on a daily basis shouldn't be an issue, yet if it takes more time then the wiki administrator should use a normal maintenance window and make it part of their task portfolio.

kghbln:
Thanks for the insight! We document accordingly.

PR #4137, opened 2019-07-13
Add `updateQueryDependencies.php`, refs 1117

mwjames:
This PR is made in reference to: #1117

- Adds the `updateQueryDependencies.php` script to run updates on all entities that embedded queries hereby allows to fill (or update) the query links table especially when enabling `smwgEnabledQueryDependencyLinksStore`
- https://www.semantic-mediawiki.org/wiki/Help:UpdateQueryDependencies.php

This PR includes:
- [x] CI build passed

### DIFF
--- a/maintenance/updateQueryDependencies.php
+++ b/maintenance/updateQueryDependencies.php
@@ -1,0 +1,186 @@
+<?php
+
+namespace SMW\Maintenance;
+
+use Onoi\MessageReporter\MessageReporter;
+use SMW\ApplicationFactory;
+use SMW\SQLStore\SQLStore;
+use SMW\DIWikiPage;
+use SMW\DIProperty;
+use SMW\Setup;
+use Title;
+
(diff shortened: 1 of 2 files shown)


mwjames:
@kghbln FYI. This isn't script you run on a regular basis (because it uses a full parse similar to what the `rebuildData.php` script does) only if someone query dependencies are missing or you enable `smwgEnabledQueryDependencyLinksStore`.

kghbln:
Thanks for the info! Even with `smwgEnabledQueryDependencyLinksStore` enabled you would only run the script in case one detects issues, right?

mwjames:
> enabled you would only run the script in case one detects issues, right?

Yes.

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AUpdateQueryDependencies.php&type=revision&diff=70170&oldid=70025)

PR #4170, opened 2019-07-27
Check extension registration is complete, refs 1732

mwjames:
This PR is made in reference to: #1732

- #1732 introduced `extension.json` has method to register SMW with MW
- Users maybe unfamiliar with it or misinterpreting the installation instruction and forgetting to add `enableSemantics` to `LocalSettings.php` therefore this PR will inform users on `Special:Version` about the discrepancy of having SMW installed but not enabled
- `smwgIgnoreExtensionRegistrationCheck` (default `false`) is provided so wiki-farm admins can disable the check (e.g. only selected wikis are allowed to have SMW enabled etc.)

This PR includes:
- [x] CI build passed

## Example

extension-not-loaded

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -104,6 +104,18 @@ return [
 	##
 
 	###
+	# Allows to ignore the check for whether the extension was correctly enabled
+	# or not. It will display a error message on `Special:Version` in case it was
+	# not.
+	#
+	# To ignore the check and suppress the error, set the value to `true`.
+	#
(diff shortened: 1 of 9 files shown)


mwjames:
https://www.semantic-mediawiki.org/wiki/Extension_registration

kghbln:
Also documented here: https://www.semantic-mediawiki.org/wiki/Help:$smwgIgnoreExtensionRegistrationCheck

PR #2474, opened 2017-05-21
QueryProcessor::getProcessedParams fix limit param

mwjames:
This PR is made in reference to: #

This PR addresses or contains:

- Fix `limit` when the value is greater as the `$smwgQMaxLimit`, `$smwgQMaxInlineLimit` where it is reset to the default value despite the global limitation

```
{{#ask: [[Modification date::+]]
 |?Modification date
 |format=csv
 |limit=5000 --> resets to default 100 but should reset to $smwgQMaxInlineLimit
}}

{{#ask: [[Modification date::+]]
 |?Modification date
 |format=csv
 |limit=222 --> is within the $smwgQMaxInlineLimit
}}
```

This PR includes:
- [ ] Tests (unit/integration)
- [x] CI build passed

PR #4172, opened 2019-07-28
Stop using long deprecated ParamProcessor stuff

JeroenDeDauw:
This PR is made in reference to: https://github.com/JeroenDeDauw/ParamProcessor/issues/40

Not tested and expecting at least one thing to blow up

`ResultPrinter` still uses `Param` though that is for another PR

### DIFF
--- a/includes/query/SMW_QueryProcessor.php
+++ b/includes/query/SMW_QueryProcessor.php
@@ -1,11 +1,10 @@
 <?php
 
 use ParamProcessor\Options;
-use ParamProcessor\Param;
 use ParamProcessor\ParamDefinition;
+use ParamProcessor\ProcessedParam;
 use ParamProcessor\Processor;
 use SMW\ApplicationFactory;
-use SMW\Message;
 use SMW\Parser\RecursiveTextProcessor;
 use SMW\Query\Deferred;
 use SMW\Query\PrintRequest;
@@ -14,6 +13,8 @@ use SMW\Query\Processor\DefaultParamDefinition;
 use SMW\Query\QueryContext;
 use SMW\Query\Exception\ResultFormatNotFoundException;
 use SMW\Query\ResultFormat;
+use SMW\Query\ResultPrinter;
+use SMW\Query\ResultPrinters\NullResultPrinter;
 
 /**
  * This file contains a static class for accessing functions to generate and execute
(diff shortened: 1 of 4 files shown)


includes/query/SMW_QueryProcessor.php
```diff
@@ -52,26 +53,31 @@ public static function setRecursiveTextProcessor( RecursiveTextProcessor $recurs
...
+		}
+
+		// TODO: ????
 		// Negates some weird behaviour of ParamDefinition::setDefault used in
 		// an individual printer.
 		// Applying $smwgQMaxLimit, $smwgQMaxInlineLimit will happen at a later
 		// stage.
 		if ( isset( $params['limit'] ) && isset( $parameters['limit'] ) ) {
-			$parameters['limit']->setValue( (int)$params['limit'] );
+//			$parameters['limit']->setValue( (int)$params['limit'] );
```
mwjames: The reason is outlined in #2474, if you think the cited issue no longer appears with the new `ParamDefinition` I'm fine with removing this line. Please confirm that the cited use case (as in #2474) continues to work otherwise this will introduce a regression.

JeroenDeDauw: Thanks for linking. Now I get it.

There is no unexpected behavior on the ParamProcessor side. If validation for a parameter fails, it will be set to the default value. So if you set lower and/or upper bounds for a numeric parameter, and the value is out of bounds, it will go to the default and not to the nearest valid value.

If that is the behavior you want then simply do not set bounds in the parameter definition and handle this validation yourself (outside of ParamProcessor). If for whatever reason that is not possible, you can do something similar to what you did here.

Can't check now: why are bounds defined in the parameter definition? Are they always the same or does it depend on result printer?

JeroenDeDauw:
I put a new version of the "workaround" in place now. I suspect the bounds can just be removed for the limit parameter instead though can't find where they are set in the place.

JeroenDeDauw:
@mwjames some more deprecations incoming ;) https://github.com/JeroenDeDauw/ParamProcessor#release-notes

JeroenDeDauw:
@mwjames I'm currently writing a new result format so if you have time to review this 
image

JeroenDeDauw:
Yeah this is real ugh. I'm kinda forced to use the deprecated stuff in my new code now :(

mwjames:
> Yeah this is real ugh. I'm kinda forced to use the deprecated stuff in my
> new code now :(

If someone can confirm that the cited limit issue (#... see above) did
not create an opening for any regression (we don't have any
integration test for this so I cannot verify whether this got broken
or not) then this can be merged.

JeroenDeDauw:
Just tried it. Seems no change with master for limits exceeding their limit.

includes/query/SMW_QueryProcessor.php
```diff
@@ -52,26 +53,35 @@ public static function setRecursiveTextProcessor( RecursiveTextProcessor $recurs
...
+		foreach ( $processingResult->getParameters() as $parameter ) {
+			$parameters[$parameter->getName()] = $parameter;
+		}
+
+		// Set limit parameter to its original value.
+		// This is to allow for special handling of invalid values rather than just setting the parameter to its default.
+		// $smwgQMaxLimit and $smwgQMaxInlineLimit will be applied at a later point.
 		if ( isset( $params['limit'] ) && isset( $parameters['limit'] ) ) {
-			$parameters['limit']->setValue( (int)$params['limit'] );
+			$parameters['limit'] = new ProcessedParam(
```
mwjames: I felt particular uneasy about this alteration knowing the limit setting has caused a regression in the past therefore I devised an integration test #4187 that would fail if the limit where causing some unexpected changes due to use of `ParamDefinition::setDefault`.

JeroenDeDauw: As expected from the Hero Of The Tests ;)

PR #4177, opened 2019-07-31
Providing missing semicolon

jaideraf:
According to css-validator: https://jigsaw.w3.org/css-validator/


JeroenDeDauw:
Thank you!

Issue #125, opened 2014-01-26
New well designed result printer mechanism


JeroenDeDauw:
Right now the query results are tightly bound to the store, making it ~~nearly~~ impossible to test logic in consumers (ie result printers).
- [Original tread on the dev list](http://sourceforge.net/mailarchive/forum.php?thread_name=CAMhmagA8YDPw6mv%3D2NV1Mpn8qvS8J_LTHGdo1G4hWzPQ_idZ6Q%40mail.gmail.com&forum_name=semediawiki-devel)
- [Some conceptual hacking](https://github.com/JeroenDeDauw/CheckedTable/tree/master/src/SMW/Query/Result)

This item needs to be tackled before we can make big changes affecting result printers, such as adoption of the new data values libraries.

mwjames:
Nothing has been done to resolve this task within the last 2 years, moving this two [0] where interested parties can still pick this up.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/projects/4

JeroenDeDauw:
I was about to create a new issue to talk about an experiment I am doing to create a sane representation of a query result. Where sane includes:

* No services like the SMW Store required for construction
* No services exposed
* Immutable
* Good naming
* Dedicated objects and not arrays in arrays

GitHub showed me this issue. And surprise, my experiment came up with the exact same structure as I apparently already found 6 years ago. New experiment: https://github.com/ProfessionalWiki/ModernTimeline/tree/master/src/ResultFacade

Modifying the existing result printer interface is a big breaking change, and modifying it this much would be a ton of work. Plus likely a bunch of printers depend on complexity that is better avoided for others. So I suggest we create a new much simpler interface that works for 95% ish of formats. Such an interface would not have these methods https://github.com/ProfessionalWiki/ModernTimeline/blob/530bf26052aa8d3a2e33ab487c4f12523f24bf15/src/ModernTimelinePrinter.php#L94-L118 and be invoked with a query result like in the two experiments.

JeroenDeDauw:
Progress: https://github.com/ProfessionalWiki/ModernTimeline/tree/37e16eb77bb7ea6972fd7fd869c072b16b2ddaec/src/ResultFacade

This now includes a draft for the new interface fulfilling these design improvements:

* Dependency injection possible
* Simple query result object
* Query result can be extended later without breaking changes (both data and logic)
* No exposure to legacy ParamProcessor classes
[...]

PR #4178, opened 2019-07-31
Fix errors caused by calling methods not in interface

JeroenDeDauw:
The ResultPrinter interface does not have this method,
hence it cannot be called without first checking it exists


JeroenDeDauw:
@kghbln this is not really a bug since all current result printers have the methods defined

JeroenDeDauw:
So there is a software design issue in that some code violates the interface contract, which leads to a bug if you add new code that just follows the contract and does nothing else. But no user visible bug, which is what the GH issue tag is for.

kghbln:
> this is not really a bug since all current result printers have the methods defined

Fair enough though the title suggested that there is an issue happening. So "code quality" is the label to go I think.

kghbln:
> But no user visible bug, which is what the GH issue tag is for.

Indeed.

kghbln:
In case of merging it will be cool if "Squash and merge" will be used to avoid confusion.

mwjames:
> So there is a software design issue in that some code violates the interface contract, which leads to a bug if you add new code that just follows the contract and does nothing else.

Yes, existing printers inherit its signature from `\SMW\Query\ResultPrinters\ResultPrinter`. If you wanted to develop only against an interface then you would need to add new interfaces to cover some of the expect methods such as:

- `DeferrablePrinter` to cover `isDeferrable`
- Method `getDefaultSort` should be covered by the `ResultPrinter` interface (knowing we extend the contract is a BC but no one uses interface besides you as far as I know).
- Method `supportsRecursiveAnnotation` and `setRecursiveTextProcessor` could be covered by something like `RecursablePrinter` (better name??!!)

JeroenDeDauw:
If implementing the current interface is not enough then I wonder why it exists in the first place. My impression is that it used to be enough, but then things just got added to the base class that everything happens to implement and no one noticed.

Having dedicated interfaces for different types of printers is in line with my thoughts here: https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/125 I'd start with minimal interface for "normal printers" and then maybe also provide new alternatives for export printers and whatnot.

JeroenDeDauw:
> If implementing the current interface is not enough then I wonder why it exists in the first place. My impression is that it used to be enough, but then things just got added to the base class that everything happens to implement and no one noticed.

To be more explicit: In my mind either implementing this interface should be sufficient or it should not be there

mwjames:
>  but then things just got added to the base class that everything happens to implement and no one noticed.

Well, yes because along the way we needed to fix a couple of things. So, for example, the text processor was added to be able to use the `Parser` instance from the #ask call without exposing the `Parser` object directly to printer and track any recursive parsing which is needed when #ask is part of a template or embedded processing chain and a printer itself requires additional parsing on the text output. 

Deferred printer as the name suggests allows to defer the process and has been design in a way that it impacts existing printers to a minimum while allowing some to support that feature.

Default sort was necessary because there was a hidden expectation on the a default sort and has been made explicit.

JeroenDeDauw:
These methods can still be added to the interface right? That is probably better than what I did here.

mwjames:
> These methods can still be added to the interface right? That is probably better than what I did here.

Yes, which is why instead of extending the `ResultPrinter` interface add new once as proposed above and check against them without breaking the `ResultPrinter` contract by adding those methods to its interface.

JeroenDeDauw:
In my mind the `ResultPrinter` contract was already broken. The code relies on new methods. By not adding them to the interface this break is hidden but still there. So I would not consider adding the missing methods a BC break.

PR #4180, opened 2019-07-31
Add to return type docs in ItemJournal

JeroenDeDauw:


src/Query/Result/ItemJournal.php
```diff
@@ -4,7 +4,7 @@
 
 use SMW\DIProperty;
 use SMW\DIWikiPage;
-use SMWDataItem as DataItem;
```
mwjames: Why was this changed?

src/Query/Result/ItemJournal.php
```diff
@@ -20,20 +20,13 @@
  */
 class ItemJournal {
 
-	/**
-	 * @var array
-	 */
 	private $dataItems = [];
```
mwjames: While for private methods comments and type hinting can be relaxed, I'd like to keep comments on properties (private and public) indifferent of whether strict typing for methods and properties are used in future or not.

src/Query/Result/ItemJournal.php
JeroenDeDauw: The alias makes the code ever so slightly less easy to understand.

src/Query/Result/ItemJournal.php
JeroenDeDauw: In this case it might have been better to make the doc more specific, so I agree with you it is not clear this removal was good.

OTOH many fields get set in he constructor and only there. This is true both for services and value objects. For these cases docs on the fields are IMO redundant (modern tools don't even need them) and better avoided. Especially if each one adds 4 lines, which means you might not even see the constructor on your first screen.

PR #4153, opened 2019-07-20
Improve error handling

kghbln:
This PR is made in reference to: #4150 

This PR addresses or contains:
- Improves error handling

This PR includes:
- [ ] Tests (unit/integration)
- [ ] CI build passed

Fixes #4150

PR #4190, opened 2019-08-03
Avoid "Uncaught Exception: It was attempted to load SemanticMediaWiki twice", refs 1732

mwjames:
This PR is made in reference to: #1732

- Despite the fact that the documentation explicitly states not to invoke `wfLoadExtension( 'SemanitcMediaWiki' )` in `LocalSettings.php` users will be tempted to add this line anyway. So, in order to avoid "Uncaught Exception: It was attempted to load SemanticMediaWiki twice" shown to a user, check via `ExtensionRegistry` whether it was registered or not and in case it was ignore the call in `enableSemantics`. 
- `enableSemantics` remains required in `LocalSettings.php`

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/GlobalFunctions.php
+++ b/src/GlobalFunctions.php
@@ -244,8 +244,13 @@ function swfCountDown( $seconds ) {
 function enableSemantics( $namespace = null, $complete = false ) {
 	global $smwgNamespace;
 
-	// #1732 + #2813
-	wfLoadExtension( 'SemanticMediaWiki', dirname( __DIR__ ) . '/extension.json' );
+	// Avoid "Uncaught Exception: It was attempted to load SemanticMediaWiki
+	// twice" in case users added `wfLoadExtension` manually to the
+	// `LocalSettings.php`
+	if ( !ExtensionRegistry::getInstance()->isLoaded( 'SemanticMediaWiki' ) ) {
+		// #1732 + #2813
+		wfLoadExtension( 'SemanticMediaWiki', dirname( __DIR__ ) . '/extension.json' );
+	}
 
 	// #4107
 	define( 'SMW_EXTENSION_LOADED', true );


mwjames:
@kghbln FYI

kghbln:
In other words: The user will get a user friendly information page: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4153#issuecomment-515703149 / [Documented.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AInstallation%2FUsing_Composer_with_MediaWiki_1.25%2B&type=revision&diff=70534&oldid=69104)

mwjames:
> In other words: The user will get a user friendly information page: #4153 (comment) / Documented.

No.

It was that (to fail and show an error message) or just avoid calling `wfLoadExtension` within `enableSemantics` which is sufficient because extension are lazy loaded from the queue after `LocalSettings.php` has been finalized meaning that `enableSemantics` can do work its magic either way.

mwjames:
S(he) will get an error screen in case `enableSemantics` isn't used but something like:

```
wfLoadExtension( 'SemanitcMediaWiki' );
enableSemantics( ... )
```
will not create any error but adding it after will produce the error since we have no control over what happens after `enableSemantics`.

```
enableSemantics( ... )
wfLoadExtension( 'SemanitcMediaWiki' );
```

kghbln:
So this was about helping if both lines were added?!

kghbln:
> S(he) will get an error screen in case enableSemantics isn't used but something like:

So yes. :) Will tweak the docu accordingly.

mwjames:
> So this was about helping if both lines were added?!

Kind of, I figured users will do (out of habit or for other reasons) :

```
wfLoadExtension( 'Maps' );
wfLoadExtension( 'SemanitcResultFormats' );
wfLoadExtension( 'SemanitcMediaWiki' );

enableSemantics( ... );
```

mwjames:
... and if you only do:

```
wfLoadExtension( 'Maps' );
wfLoadExtension( 'SemanitcResultFormats' );
wfLoadExtension( 'SemanitcMediaWiki' );
```

You get some kind of error that says `SemanitcMediaWiki` is required and `enableSemantics` is missing.

kghbln:
Yes, indeed, everything what is theoretically possible will also happen sooner or later.

Note that I changed from Maps to Mermaid since the former is a pretty bad example here and adds more confusion since it is the only extension I know that requires registration and "invocation" at the same time.

ruediklein:
Sorry for an Amateur Question, but I just installed MediaWiki, then the SemanticBundle on a Server within our Company domain and am at the step of Enable Semantics:
require_once __DIR__ . '/extensions/SemanticBundle/SemanticBundle.php';
# enableSemantics(' http://sntmucespalo01/' );
# enableSemantics(' http://sntmucespalo01/Playbook' );
enableSemantics(' http://sntmucespalo01/Playbook', true );
# enableSemantics();

Everytime I try to enableSemantics with any of the above Options I get:
Fatal error: Uncaught Exception: It was attempted to load SemanticMediaWiki twice, from C:\FS1\XAMPP\htdocs\Playbook/extensions/SemanticMediaWiki/extension.json and C:\FS1\XAMPP\htdocs\Playbook\extensions\SemanticMediaWiki/extension.json. in C:\FS1\XAMPP\htdocs\Playbook\includes\registration\ExtensionProcessor.php:462 Stack trace: #0...

It appears, as if the different Forward and backward slashes (wherever they may originate) are interpreted as two calls? I am not using wfLoadExtension at all.

At the Moment, the result is, that the install is not usable with SemanticMediawiki. Do you have any idea where to look for a solution?

JeroenDeDauw:
If multiple people are running into this it could be worth the backport.

ruediklein:
I now know, that the Installation guidelines for semanticbundle are off: The require for the semanticbundle includes a call to semanticmediawiki, which then is double with the enablesemantics.
I commented out the call to the semanticmediawiki in the semanticbundle and moved the require to after the enablesemantics.
It now seems to work as intended.

JeroenDeDauw:
Huh... wfLoadExtension should not be called for SMW?! https://www.semantic-mediawiki.org/wiki/Help:Installation/Quick_guide#Enable @kghbln ???

kghbln:
> . wfLoadExtension should not be called for SMW?!

Yes, this is correct. It is imperative not to add the `wtfLoadExtension( … );` call and I am happy that it is like this.

kghbln:
@ruediklein Sorry that you had a hard time to figure this out. The Semantic Bundle indeed needs to be fixed.

PR #4227, opened 2019-08-11
Add support for `order_map`, refs

### LINKED ISSUE #4226: Introducing the concept of `sequence map` for annotation values
In general, stored values are idempotent and have no particular retrieval order (i.e. meaning the order of values retrieved from a back-end is not guaranteed to be in the order as to the point there were stored) unless a particular order attribute is declared and even then it would retrieve values in a lexicographical order (or if given in a random order).

## Issue

Sometimes we want to retrieve values of a property (e.g. `Has author` etc.) in the order they appeared during the input, yet as outlined above, a DB back-end doesn't guarantees such order when stored as list of values.

## Solution

SMW 3.1 will introduce the concept of an "sequence map" where a new auxiliary table with a blob field holds information about the encoded position or sequence (i.e. input order) of annotated values for properties that enabled this option.

The "sequence map" will follow FIFO meaning that the first appearance of the value in a list will define its position even though the same value has been annotated in different positions.

To recall, duplicate values are not stored in SMW therefore declaring something like  `[[Has number::42]] [[Has number::1]] [[Has number::42]]` only leads to `[[Has number::42]]` and `[[Has number::1]]` as representation with duplicate values being discarded.

With that in mind, an input with something like:
- `[[Has number::42]] [[Has number::1]] [[Has number::42]] [[Has number::1001]]` will produce a map that matches to
- `[[Has number::42]] [[Has number::1]] [[Has number::1001]]`

### Example

image

## Use case(s)

- `Has authors` containing a list of authors and reflect its position as that of the source ("If there is more than one author, arrange them in the same order as found in the source." see https://library.shsu.edu/research/guides/handouts/APA_Citation_Basics.pdf)

## Technical notes

- A new table `smw_object_aux` is introduced to hold auxiliary information about a subject
- The field `smw_seqmap` will contain the sequence mapping for all properties of a particular subject that have the `sequence_map` option enabled for a property
- The sequence map consists of md5 hash values of the annotation values together with its associated position in a list (aka. array)
- By default, no sequence map is stored for any property unless there is a profile schema assigned to it
- Any property that wants to make use of the sequence mapping has to assign a `[[Profile schema::...]]` to its declaration with a schema being defined as something like:

```
{
    "type": "PROPERTY_PROFILE_SCHEMA",
    "profile": {
        "sequence_map": true
    },
    "tags": [
        "property profile"
    ]
}
```

mwjames:
This PR is made in reference to: #4226

- ~~Unit tests are missing~~, integration test added to ensure DB related queries do work and pass on all systems
- An order map isn't available by default, a property that requires an order map has to assign a `Profile schema` with something like `[[Profile schema::MyProfileSchema]]` where the `MyProfileSchema` schema defines that `sequence_map` is enabled
- Any existing properties that want to use this feature need to run `rebuildData.php` as it requires to build up the new table (`smw_object_aux`) and field (`smw_seqmap`) from the ground

This PR includes:
- [x] CI build passed

Fixes #4226

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -71,7 +71,7 @@ return [
 	#
 	# @since 3.0
 	##
-	'smwgUpgradeKey' => 'smw:2019-01-19',
+	'smwgUpgradeKey' => 'smw:2019-08-18',
 	##
 
 	###
@@ -2144,7 +2144,14 @@ return [
 			'type_description' => 'smw-schema-description-class-constraint-schema',
 			'change_propagation' => [ '_CONSTRAINT_SCHEMA' ],
 			'usage_lookup' => '_CONSTRAINT_SCHEMA'
-		]
+		],
+		'PROPERTY_PROFILE_SCHEMA' => [
+			'group' => SMW_SCHEMA_GROUP_PROFILE,
+			'validation_schema' => __DIR__ . '/data/schema/property-profile-schema.v1.json',
+			'type_description' => 'smw-schema-description-property-profile-schema',
+			'change_propagation' => '_PROFILE_SCHEMA',
+			'usage_lookup' => '_PROFILE_SCHEMA'
+		],
 	],
 	##
 
--- a/data/import/groups/schema.json
+++ b/data/import/groups/schema.json
@@ -11,7 +11,8 @@
                 "_SCHEMA_TAG",
                 "_SCHEMA_LINK",
                 "_FORMAT_SCHEMA",
-                "_CONSTRAINT_SCHEMA"
+                "_CONSTRAINT_SCHEMA",
+                "_PROFILE_SCHEMA"
             ]
         }
     ],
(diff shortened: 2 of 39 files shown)


mwjames:
#### Why using a `Profile schema` and not a new defined declarative property?

The `Profile schema` is a newly defined declarative property but rather than containing a single value assignment for an optional declaration, it takes a schema which allows to create different profiles and reassign them to other properties without having to create new declarative properties every time we'd like to introduce a new option or feature.

For example, in future we may introduce a `fulltext` key that isolates the use of the SQL based full-text index instead of applying it to all properties as it is right now (by default, if enabled). To avoid creating dozen of new declarative properties to "shape" a property profile, we introduce one declarative property that links to a schema which itself is guarded by a JSON schema to only allow specified attributes to appear.

#### Performance and size

By default, no order maps are stored and in the event that some properties use this feature the order map is a map of hashed values in a list (== array) which is compressed before storing it in the DB. Further more, resolving a sizeable query (limit > 500) with printouts where the feature is enabled for referenced properties, the loading of an order map happens as prefetch (== bulk) restricting the load on the DB when resolving said queries.

kghbln:
[Documented]https://www.semantic-mediawiki.org/wiki/Help:Special_property_Profile_schema)

PR #3644, opened 2019-01-20
Introduce `DependencyLinksValidator`, refactor update logic, refs 1117, 3537

mwjames:
This PR is made in reference to: #1117, #2283, #3537

This PR addresses or contains:

- Introduces `DependencyLinksValidator` to validate a temporal attribute (see below) together with a refactored and entire new update logic which makes the following classes obsolete:
  - `ParserCachePurgeJob`
  - `SpecialDeferredRequestDispatcher`
  - `DependencyLinksUpdateJournal`, and
  - `EntityIdListRelevanceDetectionFilter`
- Setting `smwgQueryDependencyAffiliatePropertyDetectionList` is removed as well as it no longer  serves any function (only used in `EntityIdListRelevanceDetectionFilter`)

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

Fixes #3537

## Notes

The `ParserCache` is important and should not be disabled arbitrarily, yet we have to reject the cache if an embedded query has been detected to be potentially outdated in order for the MediaWiki `Parser` to initialize a re-parse hereby forcing an update on subsequent embedded queries including their result sets and dependencies.

The rejection of the cache happens via the `RejectParserCacheValue` hook (since 3.0) and the new approach will rely on the same hook but with a new `DependencyLinksValidator` class. The `DependencyLinksValidator` will be responsible to detect “… whether any embedded query for subject X contains archaic dependencies (those entities that are part of a query either as result subject, property, condition, or printrequest) that were updated recently ...”

To make a judgement about a temporal attribute (aka "recently") a new field is required therefore the field `smw_touched` is added to track the time of the “latest touched” for each entity. This field is updated during the general data processing (or when a query is updated) hereby avoids any post-processing or time lag due to job queuing. The detection will happen on a local entity for a limited set of matches.

```
Subject X
  -> contains Query Y (last touched 12:00)
     -> contains result subject Foo (last touched 12:05)
        -> since Foo is "younger" it could contain new/altered assignments
           -> presents the likelihood of being outdated therefore makes
              it a (or contain) archaic dependency to X/Y
```

```
28	SELECT v.smw_id,v.smw_touched FROM "smw_fpt_ask"
INNER JOIN smw_object_ids AS p ON ((s_id=p.smw_id))
INNER JOIN smw_object_ids AS v ON ((o_id=v.smw_id))
[...]

PR #4230, opened 2019-08-11
Avoid overriding `smw_rev`, `smw_touched` during setup, refs 3390, 3644

mwjames:
This PR is made in reference to: #3390, #3644

- If a predefined property exists then also check for `smw_rev`, `smw_touched`.

This PR includes:
- [x] CI build passed

Replication monitoring in ES would show a revision in ES but not in SMW after running `setupStore.php` and the reason is that those fields were overridden.

image


mwjames:
@kghbln [0] is such a case as described above.

image

[0] https://sandbox.semantic-mediawiki.org/wiki/Attribut:Modification_date

Issue #4260, opened 2019-08-24
`ProcessedParam` in `QueryProcessor` produces wrong offset/limit representation


mwjames:
### Setup and configuration

- SMW version: 3+
- MW version: all

### Issue

`QueryProcessor::getProcessedParams` produces an incorrect `ProcessedParam` representation for an offset/limit parameter when the raw parameter is an integer value. The issue doesn't appear when values are cast as string.

### Steps to reproduce

Some preliminary research on the internal representation. When used as integer:

```

int(20) -->limit
int(500) -->offset

object(ParamProcessor\ProcessedParam)#574 (5) {
  ["value":"ParamProcessor\ProcessedParam":private]=>
  int(50)
  ["name":"ParamProcessor\ProcessedParam":private]=>
  string(5) "limit"
  ["wasSetToDefault":"ParamProcessor\ProcessedParam":private]=>
  bool(true)
  ["originalValue":"ParamProcessor\ProcessedParam":private]=>
  NULL
  ["originalName":"ParamProcessor\ProcessedParam":private]=>
  NULL
}
object(ParamProcessor\ProcessedParam)#575 (5) {
  ["value":"ParamProcessor\ProcessedParam":private]=>
  int(0)
  ["name":"ParamProcessor\ProcessedParam":private]=>
  string(6) "offset"
  ["wasSetToDefault":"ParamProcessor\ProcessedParam":private]=>
  bool(true)
  ["originalValue":"ParamProcessor\ProcessedParam":private]=>
  NULL
  ["originalName":"ParamProcessor\ProcessedParam":private]=>
  NULL
}
```

As for using strings as parameter values, `ProcessedParam` contains a corresponding value representation.

```

string(2) "20" -->limit
string(3) "500" --> offset

object(ParamProcessor\ProcessedParam)#570 (5) {
  ["value":"ParamProcessor\ProcessedParam":private]=>
  int(20)
  ["name":"ParamProcessor\ProcessedParam":private]=>
  string(5) "limit"
  ["wasSetToDefault":"ParamProcessor\ProcessedParam":private]=>
  bool(false)
  ["originalValue":"ParamProcessor\ProcessedParam":private]=>
  string(2) "20"
  ["originalName":"ParamProcessor\ProcessedParam":private]=>
  string(5) "limit"
}
object(ParamProcessor\ProcessedParam)#571 (5) {
  ["value":"ParamProcessor\ProcessedParam":private]=>
  int(500)
[...]

PR #4261, opened 2019-08-24
QueryProcessor, offset, limit to yield a correct ProcessedParam representation, refs 4260

mwjames:
This PR is made in reference to: #4172, #4260

- Adds tests for the #4260 issue which introduced an inconsistency about how limit/offset produces different results depending on the value type (string vs. integer).
- It adds a unit test for `QueryProcessor::getProcessedParams` and an integration test with `Special:Ask` where the issue was uncovered.

This PR includes:
- [ ] CI build passed

Fixes #

### DIFF
--- a/tests/phpunit/Unit/Query/Processor/QueryProcessorTest.php
+++ b/tests/phpunit/Unit/Query/Processor/QueryProcessorTest.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace SMW\Tests\Query\Processor;
+
+use SMWQueryProcessor as QueryProcessor;
+
+/**
+ * @covers SMWQueryProcessor
+ * @group semantic-mediawiki
+ *
+ * @license GNU GPL v2+
+ * @since 3.1
+ *
+ * @author mwjames
+ */
+class QueryProcessorTest extends \PHPUnit_Framework_TestCase {
+
+	/**
+	 * @dataProvider limitOffsetParamsProvider
+	 */
+	public function testGetProcessedParams_YieldCorrectProcessedParamValue( $params, $key, $expected ) {
+
+		$processedParam = QueryProcessor::getProcessedParams(
+			$params
+		);
+
+		$this->assertEquals(
+			$expected,
+			$processedParam[$key]->getValue()
+		);
+	}
+
+	public function limitOffsetParamsProvider() {
+
+		yield 'limit-string' => [
+			[ 'limit' => '12' ],
+			'limit',
+			12
+		];
+
+		yield 'limit-integer' => [
+			[ 'limit' => 12 ],
+			'limit',
+			12
+		];
+
+		yield 'offset-string' => [
+			[ 'offset' => '42' ],
+			'offset',
+			42
+		];
+
+		yield 'offset-integer' => [
+			[ 'offset' => 42 ],
+			'offset',
+			42
+		];
+	}
+
+}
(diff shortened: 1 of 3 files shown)


mwjames:
@JeroenDeDauw The test fails for obvious reasons because `ProcessedParam` is broken. Please have a look at the unit test which shows the inconsistency. `offset-integer` fails and the `limit-integer` would also fail if weren't for the special handling of limit parameter at the end of the `QueryProcessor::getProcessedParams` method.

As outlined in #4260, you can see the `ProcessedParam` object stack which should help narrow down the issue together with the provided unit test.

tests/phpunit/Unit/Query/Processor/QueryProcessorTest.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
+namespace SMW\Tests\Query\Processor;
+
+use SMWQueryProcessor as QueryProcessor;
```
JeroenDeDauw: FYI: this caused me some moments of confusion

JeroenDeDauw:
I've been looking at this now and it is not clear to me there is a regression. What makes you think the behavior used to be different?

JeroenDeDauw:
No change compared to 1.0.0 from 2013 as far as I can tell

JeroenDeDauw:
Some extra context: Parameter Processor has two modes: string inputs and non-string inputs. When using the former, which is the default, and the one SMW has always used, then everything that is not a string will be considered invalid and set to the default when there is a default value. This is expected behavior and has not changed since at least 2013. Hence this cannot be changed now since it would be a breaking change.

Is there a change in SMW from the user level? The newly added test here would also have failed 6 years ago.

mwjames:
> I've been looking at this now and it is not clear to me there is a
> regression. What makes you think the behavior used to be different?

Whether it is a regression or not, I cannot tell since I have no way
of knowing since no tests have been in place that can verify the
behaviour one way or the other way. I was talking about an
inconsistency which I know now exists (and is testable) and is the
reason I made a blocker for the 3.1 release.

I doubt that casting an integer to string just so that ProcessParam is
happy and can actually make an integer representation (see the object
trace) is something I understand. It is even more confusing that the
param is defined as integer which means integer to integer should
always work but currently only string to integer works.

```
		$params['limit'] = [
			'type' => 'integer',
			'default' => $vars['smwgQDefaultLimit'],
			'negatives' => false,
		];

		$params['offset'] = [
			'type' => 'integer',
			'default' => 0,
			'negatives' => false,
			'upperbound' => $vars['smwgQUpperbound'],
		];
```

JeroenDeDauw:
I'm not changing this in Param Processor as bugfix and I do not think it makes sense to regard this as a bug in SMW if the only place where it fails is a test that provides a type of input the SMW user cannot provide.

mwjames:
> I'm not changing this in Param Processor as bugfix and I do not think it
> makes sense to regard this as a bug in SMW if the only place where it fails
> is a test that provides a type of input the SMW user cannot provide.

This is going in circles, it feels like I'm talking to a WMF guy here.
I said in my email, I found this issue by actually using Special:Ask
which after the fact was the reason I engaged in creating the tests to
verify the behaviour and prove that this is something that creates
unexpected results.

You find the behaviour consistent then this a fine but don't ask me
about any release where this isn't fixed.

The tests are going to be merged and a release will only happen where
these tests pass.

JeroenDeDauw:
Please describe the behavior difference you saw in Special:Ask then. At the moment I have no idea what changes from a user perspective.

Issue #4294, opened 2019-09-18
PHP 7.4 deprecation warnings


mszabo-wikia:
### Setup and configuration

- SMW version: 3.1.0-rc.1, git: 5c4837e,
- MW version: 1.34.0-alpha, git: 7612ebe,
- PHP version: 7.4.0RC1
- DB system (MySQL, Blazegraph, etc.) and version: MySQL 5.7.27

### Issue
The first PHP 7.4 release candidate is out. This new version of PHP brings several new deprecations that affect Semantic MediaWiki, including the deprecation of accessing array offsets via curly brackets[1][2] and unparenthesized nested ternary operators[3]. Semantic MediaWiki is using some of these deprecated features, which cause deprecation warnings.

---
[1] https://github.com/php/php-src/blob/PHP-7.4/UPGRADING#L362
[2] https://wiki.php.net/rfc/deprecate_curly_braces_array_access
[3] https://github.com/php/php-src/blob/PHP-7.4/UPGRADING#L350

### Stack trace
Executing the Semantic MediaWiki PHPUnit test suite results in various warnings, examples follow:

```
Deprecated: Unparenthesized `a ? b : c ? d : e` is deprecated. Use either `(a ? b : c) ? d : e` or `a ? b : (c ? d : e)` in /usr/src/myapp/extensions/SemanticMediaWiki/tests/phpunit/Utils/JSONScript/JsonTestCaseFileHandler.php on line 380

Call Stack:
    0.0015     407504   1. {main}() /usr/src/myapp/tests/phpunit/phpunit.php:0
    0.0238     671968   2. require('/usr/src/myapp/maintenance/doMaintenance.php') /usr/src/myapp/tests/phpunit/phpunit.php:136
    1.0265   15686736   3. PHPUnitMaintClass->execute() /usr/src/myapp/maintenance/doMaintenance.php:99
    1.0579   16402872   4. MediaWikiPHPUnitCommand->run() /usr/src/myapp/tests/phpunit/phpunit.php:97
    1.0579   16402872   5. MediaWikiPHPUnitCommand->handleArguments() /usr/src/myapp/extensions/SemanticMediaWiki/vendor/phpunit/phpunit/src/TextUI/Command.php:159
    1.1184   16861472   6. MediaWikiPHPUnitCommand->handleBootstrap() /usr/src/myapp/extensions/SemanticMediaWiki/vendor/phpunit/phpunit/src/TextUI/Command.php:788
[...]

PR #4296, opened 2019-09-18
Avoid offset access warnings in DIProperty

mszabo-wikia:
This PR is made in reference to: #4294

- Avoid offset access warnings (`Trying to access array offset on value of type int`)[1] on PHP 7.4+ in `DIProperty` by casting the corresponding parameters to strings.

---
[1] https://github.com/php/php-src/blob/PHP-7.4/UPGRADING#L362

This PR includes:
- [ ] CI build passed

### DIFF
--- a/includes/dataitems/SMW_DI_Property.php
+++ b/includes/dataitems/SMW_DI_Property.php
@@ -464,6 +464,9 @@ class DIProperty extends SMWDataItem {
 	 * @return DIProperty object
 	 */
 	public static function newFromUserLabel( $label, $inverse = false, $languageCode = false ) {
+		// [#4294] Convert the passed-in label to a string value
+		// Avoids warning "Trying to access array offset on value of type X" on PHP 7.4+
+		$label = strval( $label );
 
 		if ( $label !== '' && $label{0} == '-' ) {
 			$label = substr( $label, 1 );
@@ -501,6 +504,9 @@ class DIProperty extends SMWDataItem {
 	}
 
 	private function newDIWikiPage( $dbkey, $subobjectName ) {
+		// [#4294] Convert the passed-in dbkey to a string value
+		// Avoids warning "Trying to access array offset on value of type X" on PHP 7.4+
+		$dbkey = strval( $dbkey );
 
 		// If an inverse marker is present just omit the marker so a normal
 		// property page link can be produced independent of its directionality


mszabo-wikia:
PHP 7 failed with what seems to be a Travis hiccup before actually executing any tests:
```
  [Composer\Downloader\TransportException]                                     
  The "https://packagist.org/p/provider-2019-04%246e9d3eb372b3e43a14ff5ba6e6c  
  b037890da61f8f3288e468ac3a1a1bbbf27c1.json" file could not be downloaded (H  
  TTP/1.1 404 Not Found)                    
```

JeroenDeDauw:
I'm not keen on these casts. The method docs say it should get a string. So with this change we just silence the issue even though clearly something is going wrong. Hard failing might be better.

Hard fail is what would happen if we had a string type check to the parameter itself. Which presumably is not far down the line since we are about to switch to PHP 7.1. This would make the casts superfluous as well.

I rather see the root cause (whatever it is) fixed and a string type be added to the parameter.

mszabo-wikia:
Yeah, we can safely add the string parameter type IMHO. The errors indicated that the value was an int instead of a string, so the parameter type should properly convert it into a numeric string as the project doesn't use `strict_types`.

JeroenDeDauw:
Do you have a stack trace so we can figure out what code is incorrectly passing an integer?

mszabo-wikia:
@JeroenDeDauw I've extracted the stack trace for every test that failed for me with a PHP Notice. Here it is: https://gist.github.com/mszabo-wikia/bf2745970f308a24c908c1b4734acb37

This was generated using xq from PHPUnit's JUnit report XML: `xq -r '..|.testcase? | select(. != null) | if (. | type) == "array" then (. | flatten | .[]) else . end | select(.error."@type"? == "PHPUnit\\Framework\\Error\\Notice") | .error."#text"'`.

JeroenDeDauw:
Thanks a bunch!

12205 lines wtf :D Also booleans passed. Needs some investigation.

mszabo-wikia:
Yeah I think there are lots of duplicates due to a lot of code paths passing through DIProperty :) So hopefully it will quickly get a lot smaller once that class is fixed.

JeroenDeDauw:
@mszabo-wikia I'll close this since it does not deal with the root cause. I'm not sure how to fix that properly but it is tracked at https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4308

Thanks for your efforts here!

PR #4302, opened 2019-09-19
Guard clause in AllowsListConstraintValueValidator

JeroenDeDauw:


kghbln:
I guess this could be back-ported to 3.1.x?

JeroenDeDauw:
No point, this is just a refactor. Which means it should normally NOT go into a patch release.

PR #4313, opened 2019-09-25
WIP Design draft for new result format interface

JeroenDeDauw:
Progress towards https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/125

Based on work in the ModernTimeline extension.

https://github.com/ProfessionalWiki/ModernTimeline/tree/fe1ca3e2686752ab7e67f69586b4bfac0b91b459/src/ResultFacade

No dedicated tests in this draft yet, though most code tested in
ModernTimeline via integration and system tests.

### DIFF
--- a/src/ResultFormats/ResultFormat.php
+++ b/src/ResultFormats/ResultFormat.php
@@ -1,0 +1,40 @@
+<?php
+
+declare( strict_types = 1 );
+
+namespace SMW\ResultFormats;
+
+/**
+ * @private Package private
+ */
+class ResultFormat {
+
+	private $name;
+	private $nameMessageKey;
+	private $parameterDefinitions;
+	private $constructionFunction;
+
+	public function __construct( string $name, string $nameMessageKey, array $parameterDefinitions, callable $presenterBuilder ) {
+		$this->name = $name;
+		$this->nameMessageKey = $nameMessageKey;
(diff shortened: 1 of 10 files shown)


krabina:
this is somehting for @s7eph4n

s7eph4n:
I like the approach.

I am not sure though if I am a fan of that and-and-and-construct. One issue for me is that from their title it is not immediately obvious what those methods do. And generally, what's the advantage over just requiring four parameters?

JeroenDeDauw:
@s7eph4n this is a [fluent interface](https://en.wikipedia.org/wiki/Fluent_interface) which you'll have seen before in places such as PHPUnits mock builder. Positional parameter lists have the disadvantage that (a) they are positional, so swapping parameters breaks things and (b) they are not named so hard to understand without looking at the function definition. Also some advantages when dealing with optional parameter and adding new parameters. And some disadvantages in that the required parameters are less obvious. All the usual fluent interface + builder pattern stuff really.

Naming can be changed and in the end how exactly the registry takes the info is an implementation detail. I'm much more interested in what info it takes.

JeroenDeDauw:
@mwjames thoughts?

kghbln:
@mwjames It will be super cool if you could have a peep on this and comment. My general impression as a user is that result formats need kinda more love. This may be the way to ....

Issue #4319, opened 2019-10-02
Error: 1305 SAVEPOINT wikimedia_rdbms_atomic1 does not exist caused by SMW\SQLStore\QueryEngine\HierarchyTempTableBuilder::buildTempTable


FO-nTTaX:
### Setup and configuration

- SMW version: `3.1.0`
- MW version: `1.31.3`
- PHP version: `7.3.9-1+0~20190902.44+debian9~1.gbpf8534c (fpm-fcgi)`
- DB system (MySQL, Blazegraph, etc.) and version: `10.4.8-MariaDB-1:10.4.8+maria~buster`

### Issue
I have consistently been seeing errors of the type 
```
Error: 1305 SAVEPOINT wikimedia_rdbms_atomic1 does not exist (localhost)

PHP message: DB Error: A database query error has occurred. Did you forget to run your application's database schema updater after upgrading?
```
The error happens when I save certain pages with SMW calls, and happens in the deferred updates that get queued from WikiPage::doModify(). Semantic MediaWiki seems to be creating and working with a templorary table called `smw_new` which it creates, inserts to, truncates, inserts to again and then drops. According to the [MariaDB documentation](https://mariadb.com/kb/en/library/sql-statements-that-cause-an-implicit-commit/) and the [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/implicit-commit.html) calling `TRUNCATE TABLE` on a temporary table still implicitly commits, and therefor flushes the `SAVEPOINT`, which then causes the issue with the missing savepoint when `RELEASE SAVEPOINT` gets called. This is a problem, because all of this happens inside a transaction created by MediaWiki Core. The `TRUNCATE TABLE` in question seems to be this one here: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/54a9495e2804509941442471df3666906e140e83/src/SQLStore/QueryEngine/HierarchyTempTableBuilder.php#L183
Changing this `TRUNCATE TABLE` to `DELETE FROM` seems to fix the issue, as `DELETE FROM` does not come with an implicit `COMMIT`.

### Stack trace
The stack trace points to the MediaWiki database classes, so I doubt it is helpful, but here it is (line numbers might be a bit off because I added some error logging to the file)
```
/includes/libs/rdbms/database/Database.php (1459) __construct
/includes/libs/rdbms/database/Database.php (1429) makeQueryException
[...]

PR #4320, opened 2019-10-03
Use DELETE FROM instead of TRUNCATE TABLE

JeroenDeDauw:
To avoid committing open transactions, as per
https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4319

### DIFF
--- a/src/SQLStore/QueryEngine/HierarchyTempTableBuilder.php
+++ b/src/SQLStore/QueryEngine/HierarchyTempTableBuilder.php
@@ -180,7 +180,7 @@ class HierarchyTempTableBuilder {
 
 			// empty "new" table
 			$db->query(
-				'TRUNCATE TABLE ' . $tmpnew,
+				'DELETE FROM ' . $tmpnew,
 				__METHOD__
 			);
 


mwjames:
This needs to be reverted!

On the matter of DELETE FROM vs TRUNCATE issue, I remember that we had
a similar discussion last year and the initial ticked missed some
context because:

- The issue can only appear in a replication environment where there a
statement or mixed replication is used, row based replication are not
suppose to replicate temporary tables (see the MySQL docs for that)
hence this issue should not appear there, if it does then it means MW
is doing something it shouldn't do.
- TRUNCATE was selected (has been this at least since SMW 1.8+)
because out of performance concerns (see Google search or [0] as
example)
- Some figures about the performance impact are necessary here since
the code path that generates those SQL statements are "hot".
- If you cannot get around the issue of temporary tables being
replicated then instead of arbitrarily changing the SQL, you introduce
a setting so users can switch to `DELETE FROM` if it deemed necessary,
yet `TRUNCATE` remains the default.

JeroenDeDauw:
Alright, reverted

notr1ch:
Hello, I'm the admin of the system in question referenced in #4319. We are not using replication and I don't see how that would affect the issue. I've confirmed the order of the statements that are being passed to the MySQL server - a transaction is started by Mediawiki core and then implicitly closed by the use of `TRUNCATE TABLE` in SMW. While it may be rare and require a certain configuration of SMW and the associated data, this bug triggers multiple times per day under our setup without the `DELETE FROM` change. You can verify the behavior with the MySQL command line:

```
MariaDB [test]> BEGIN;
Query OK, 0 rows affected (0.00 sec)

MariaDB [test]> SAVEPOINT test1;
Query OK, 0 rows affected (0.00 sec)

MariaDB [test]> CREATE TEMPORARY table test2 (id int unsigned not null);
Query OK, 0 rows affected (0.00 sec)

MariaDB [test]> TRUNCATE test2;
Query OK, 0 rows affected (0.00 sec)

MariaDB [test]> RELEASE SAVEPOINT test1;
ERROR 1305 (42000): SAVEPOINT test1 does not exist
```

If `DELETE FROM` performance is unsatisfactory, dropping and re-creating the temporary table may be a suitable alternative (this is essentially what `TRUNCATE TABLE` does under the hood) as DROP / CREATE for temporary tables does not implicitly commit a transaction.

mwjames:
> not using replication and I don't see how that would affect the issue. I've
> confirmed the order of the statements that are being passed to the MySQL
> server - a transaction is started by Mediawiki core and then implicitly

Temporary tables should not cause a transaction since there are not
expected (depending on the mode) to be replicated and are discarded at
the end of a session.

> If `DELETE FROM` performance is unsatisfactory, dropping and re-creating the
> temporary table may be a suitable alternative (this is essentially what
> `TRUNCATE TABLE` does under the hood) as DROP / CREATE for temporary tables
> does not implicitly commit a transaction.

There are documented differences (how rows are locked etc.) and I'm
unable to determine whether those concerns are real or not since the
specific code path has been in place since at least SMW 1.8.

> closed by the use of `TRUNCATE TABLE` in SMW. While it may be rare and
> require a certain configuration of SMW and the associated data, this bug
> triggers multiple times per day under our setup without the `DELETE FROM`

Well, I'm unable to observe the issue and I know which queries would
trigger a call to `HierarchyTempTableBuilder` and ultimately run those
queries in question.

> `TRUNCATE TABLE` does under the hood) as DROP / CREATE for temporary tables
> does not implicitly commit a transaction.

Doing `DROP / CREATE for temporary tables` may create overhead which
is why a different strategy was selected in the first place.

Bottom line, if someone can guarantee that `DELETE FROM` doesn't pose
any significant performance degradation then we can go ahead make the
change otherwise we should introduce a setting where this can be
switched on and off but leaving the status quo

notr1ch:
The issue is not temporary tables causing a transaction - the transaction is already opened by MW core before SMW code is executed. Once SMW hooks are called, SMW code creates temp tables as part of its processing and at some point calls `TRUNCATE TABLE` on one of the temp tables, which according to https://dev.mysql.com/doc/refman/8.0/en/implicit-commit.html causes the existing transaction opened by MW core to be committed. When the SMW hooks are finished and control is passed back to MW core, errors are triggered since the transaction that MW core expected to still be open was implicitly committed by SMW.

PR #4322, opened 2019-10-08
Fix missing use declaration; minor clean-up

s7eph4n:

- SMW\Query\QueryContext was used but not pulled in in ResultPrinter.
- minor clean-up and optimization.

This PR includes:
- [x] CI build passed

Fixes #


JeroenDeDauw:
so ... setQueryContext calls where broken before?

JeroenDeDauw:
If the method was definitely broken (causing fatal error on execution) during the last 2 years then we can remove it

mwjames:
If it isn't used then we may kick it (could have been used in past
where its usage got removed and is no longer necessary after some
refactoring).

kghbln:
Back-ported with b75e468910ca9be0a2dcd50f3fe3df6ed1fd847b

PR #4324, opened 2019-10-10
Clean TransactionHandler

JeroenDeDauw:


mwjames:
> Yes they do provide benefit, just like tests do.

We may agree on adding type hints optionally and voluntarily (and in
my case I won't bother) to new classes but please refrain from
changing existing classes, especially don't make any mass changes.

mwjames:
> What makes you so convinced types are useless?

I haven't said useless, I have said I don't want to be bothered
(semantics are different here).

> As to merge hell, that is a valid concern. (I hate merge conflicts.) In my
> experience that does not happen though. We can easily test if it does. We
> can experiment and see what happens. Assuming we do not make some mass
> changes that are hard to revert the max cost of this is low. Again, why not
> give this a try?

I do have some old changes that haven't made into 3.1 and there
probably a couple of new ones I'll write while traveling and any
change to existing files will require me to resolve those conflicts
due to some non functional change (removing comments, adding type
hints etc.) that happen in between and only add to my already existing
time constraint. So, while you may have a preference for adding type
hints (where I don't) adding them to new classes only should help
avoid creating merge conflicts in the first place.

JeroenDeDauw:
Which old changes are you talking about? The 3 PRs on GitHub? Or is there more?

JeroenDeDauw:
> I haven't said useless, I have said I don't want to be bothered

But you can be bothered to add the same info in a docblock? Cmon man, just give them a try, they are super useful. Especially if you use proper tools.

PR #4334, opened 2019-10-17
Supress error "Constant MW_ENTRY_POINT already defined"

### LINKED ISSUE #4331: Constant MW_ENTRY_POINT already defined
### Setup and configuration

- MediaWiki | 1.34.0-rc.0 (366bf66)22:03, 15 October 2019
- PHP | 7.2.19-0ubuntu0.18.04.2 (apache2handler)
- MariaDB | 10.1.41-MariaDB-0ubuntu0.18.04.1
- Semantic MediaWiki | 3.2.0-alpha (c0b7a8f) 17:35, 14 October 2019

### Issue
When accessing ["Special:SemanticMediaWiki"](https://sandbox.semantic-mediawiki.org/wiki/Sp%C3%A9cial:SemanticMediaWiki) the following notice is shown:

```
Notice: Constant MW_ENTRY_POINT already defined in /../w/maintenance/Maintenance.php on line 23
```

kghbln:
This PR is made in reference to: #4331 

- Code changes as suggested by @mwjames 

This PR includes:
- [x] CI build passed

Fixes #4331

### DIFF
--- a/src/MediaWiki/Specials/Admin/MaintenanceTaskHandler.php
+++ b/src/MediaWiki/Specials/Admin/MaintenanceTaskHandler.php
@@ -168,7 +168,9 @@ class MaintenanceTaskHandler extends TaskHandler {
 		$scripts = [];
 
 		foreach ( $files as $script ) {
-			require_once $script[0];
+			// Supress error "Constant MW_ENTRY_POINT already defined" due to
+			// https://github.com/wikimedia/mediawiki/commit/b7ce7aacb0fb6803d9135f465e9cf6b48912883e
+			@require_once $script[0];
 
 			// Auto-discover the class name!
 			$classes = get_declared_classes();


kghbln:
Once this passes it can be squashed and merged.

mwjames:
> Once this passes it can be squashed and merged.

+1

Henryfunk:
I am still seeing this error message on a local test site running MediaWiki 1.34.1. How about changing Maintenance.php so that it tests if the constant is already defined before attempting to (re)define it. It seems such an obvious fix that I thought someone would have done it long ago. Is there any reason not to do it?

mwjames:
>  It's not part of SMW of course, but still? Evidently if works in SMW with "index" (or whatever) as entry point?

I don't know what you mean by that but SMW doesn't owns the `MW_ENTRY_POINT` definition nor the `Maintenance.php` class hence there can be no modification to either of them. Anyway, I'm unable to confirm your claim of "... still seeing this error message on a local ...".

Henryfunk:
I meant simply: why hide the error message instead of fixing the error?

mwjames:
> I meant simply: why hide the error message instead of fixing the error?

We can't which why we hide the error given that we have no influence as the when `MW_ENTRY_POINT` becomes a defined constant. This was introduced in https://github.com/wikimedia/mediawiki/commit/b7ce7aacb0fb6803d9135f465e9cf6b48912883e.

Henryfunk:
mwjames, thanks! Some odd decision there, but never mind, I have fixed it with a test of the constant before it's defined in my "own" Maintenance.php.

mwjames:
> Some odd decision there, but never mind, I have fixed it with a test of the constant before it's defined in my "own" Maintenance.php.

wikimedia/mediawiki@b7ce7aa added `define( MW_ENTRY_POINT, 'cli' );` without checking whether is has already been defined or not, now, when we try to auto-discover which maintenance scripts are available (to avoid having to hard-code a list because there is no one who wanted to do this) we have to load the script once at which point it runs through the said `define ...` but given it has already be defined somewhere else in the MW stack it will issue an error.

We cannot influence the error message of constants being defined more than once nor can we change the `Maintenance` class so all we can do is to suppress the error to allow us to retrieve the actual class name.

Under assumption class names always match the script name (#4659), there are no other files in the `maintenance` folder, and all of the them belong to the `Maintenance` class you could simply do:

```
@@ -166,26 +166,10 @@ class MaintenanceTaskHandler extends TaskHandler implements ActionableTask {
 
 		$files = $this->fileFetcher->findByExtension( 'php' );
 		$scripts = [];
 
 		foreach ( $files as $script ) {
-			// Supress error "Constant MW_ENTRY_POINT already defined" due to
-			// https://github.com/wikimedia/mediawiki/commit/b7ce7aacb0fb6803d9135f465e9cf6b48912883e
-			@require_once $script[0];
-
-			// Auto-discover the class name!
-			$classes = get_declared_classes();
-			$class = end( $classes );
-
-			$reflectionClass = new \ReflectionClass( $class );
-
-			if (
-				!$reflectionClass->getParentClass() ||
-				$reflectionClass->getParentClass()->getName() !== 'Maintenance' ) {
-				continue;
-			}
-
 			$name = basename( $script[0] );
 			$section = '';
 ```

But here you would make a lot of assumptions without verifying that is actual true what you stipulate.

Henryfunk:
Thanks, I see the problem. Well, I guess people who don't want to hack Maintenance.php will just have to live with it :-)

PR #4335, opened 2019-10-18
Test with PHP 7.4

JeroenDeDauw:


JeroenDeDauw:
Great. PHPunit 6.x does not support PHP 7.4 and PHPUnit 7+ can't be installed here.

@mwjames what's the deal with having this special PHPUnit setup anyway? Do you know? (It's possible I introduced this way back for god knows what reason. I don't recall.) For Maps and co I do not have this and am not running into the problems seen here. Running tests with PHP 7.4 just fine.

mwjames:
> @mwjames what's the deal with having this special PHPUnit setup anyway? Do
> you know? (It's possible I introduced this way back for god knows what
> reason. I don't recall.) For Maps and co I do not have this and am not
> running into the problems seen here. Running tests with PHP 7.4 just fine.

There is no special PHPUnit setup, we need MediaWiki to be initialized
before and have some of the config setup prior running PHPUnit via
`phpunit.php`.

As a reminder, you cannot simply run the vendor/phpunit here since
SemanticMediaWiki isn't a library and we need some of the MediaWiki
infrastructure for tests to be functional, after all SemanticMediaWiki
depends on MediaWiki.

We only use a peeled set (no `MediaWikiTestCase` or any derivative
work of it) and mostly rely on PHPUnit itself but for the DB setup we
need some additional setup but also here we avoid direct reference to
`MediaWikiTestCase` and instead have `DatabaseTestCase` for those that
require an active DB instance.

As for PHP 7.4, I played with it last year (by the time it was PHPUnit
7.3) and encountered the [0] issue, now given the unhumourous
response, I stayed away from phab ever since.

For example, in the past I tried to isolate unit tests a bit better
but once you run across a `Title` instance in the stack even though it
isn't part of the actual test it will hit the DB sooner or later and
if the DB isn't setup prior the test, the whole test run will fall
apart like a house of cards.

From where I'm standing, I'd say we are fairly independent of
MediaWiki when it comes to designing tests and use cases, of course in
some instances we have to abide the limitations set forth by it as in
[1] and other times we cannot do much about it [0] since we don't own
the interfaces and are unable to work around it.

[0] https://phabricator.wikimedia.org/T204420
[1] https://github.com/SemanticMediaWiki/SemanticScribunto/commit/c8692d4fec324ed864f2c010509506c9d19a1a78

JeroenDeDauw:
You know I agree on decoupling from MediaWiki and that I appreciate the work you did there so far.

What I meant with "special PHPUnit setup" is the stuff like `PHPUNIT=6.5.*` in the Travis config and the code using those values. In Maps I don't have that (removed some years ago) and am not having the problem I ran into with this PR. Maybe this problem is not caused by the special code but it does seem probable without further investigation.

PR #3829, opened 2019-03-23
Add `_CONSTRAINT_SCHEMA` property, refs 3746

mwjames:
This PR is made in reference to: #3746

This PR addresses or contains:

- Adds `_CONSTRAINT_SCHEMA` property as part of #3746 

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

PR #4338, opened 2019-10-19
[DNM] Fixes alias being identical to name

kghbln:
This PR is made in reference to: #3829 

- Fixes alias being identical to name

This PR includes:
- [ ] CI build passed

### DIFF
--- a/i18n/extra/en.json
+++ b/i18n/extra/en.json
@@ -200,7 +200,7 @@
             "Schema definition": "_SCHEMA_DEF",
             "Schema link": "_SCHEMA_LINK",
             "Formatter schema": "_FORMAT_SCHEMA",
-            "Constraint schema": "_CONSTRAINT_SCHEMA",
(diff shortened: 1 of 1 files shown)


kghbln:
Actually I found many more which does not make sense to me except for these are needed to map additional aliases to a name. Do not think that this is likely but ...

kghbln:
@mwjames There is probably a reason for doing this, i.e. having alias and name being indentical?

mwjames:
> @mwjames There is probably a reason for doing this, i.e. having alias and name being identical?

As far as I remember that was required by the old lang object handling (1.8x) to allow traversing the lookup. It could be that we still rely on it somewhere in the `Lang.php` when building the mapping between those literals. It shouldn't be necessary but I would have to confirm that which I cannot guarantee to happen anytime soon.

kghbln:
Thanks for the info. I just wanted to avoid unnecessary pull request rot.

Issue #4345, opened 2019-11-01
DisplayTitle in query results showing previous value after change


gesinn-it:
### Setup and configuration
- SMW version: 3.1.0
- MW version: 1.31.5
- PHP version: 7.1
- $smwgQueryProfiler = true;
- $smwgEnabledQueryDependencyLinksStore = true;
- $wgMainCacheType = CACHE_MEMCACHED;
- $wgParserCacheType = CACHE_MEMCACHED;
- $wgMessageCacheType = CACHE_MEMCACHED;
JobQueue empty and jobs run via cron.

### Issue
DisplayTitle (randomly/every 2nd time?) in query results showing previous (t-1) value after change.
Trying to reproduce the issue, I've created two pages in SMW sandbox:
* https://sandbox.semantic-mediawiki.org/wiki/Test_DTITLE
* https://sandbox.semantic-mediawiki.org/wiki/Test_DTITLE_Query

### Steps to reproduce
Change the DisplayTitle of a page (Test_DTITLE). Check the result on another page that queries this page (Test_DTITLE_Query). The current DisplayTitle should be shown in the query result. This seems to work only randomly/every 2nd time. Even after waiting a few minutes the DisplayTitle does not show the current value.

This happens on one of my test wikis with the config above but also on SMW sandbox.

![Update-DisplayTitle-fails](https://user-images.githubusercontent.com/10398316/68036728-8cf37380-fcc6-11e9-98b7-cd9275c8d2bd.gif)

kghbln:
> This seems to work only randomly/every 2nd time.

It tested several times. This indeed happens for every 2nd change. Pretty strange.

gesinn-it:
@kghbln do you have access to the sandbox to share the configuration for cache, runJobs (cron?) and $smwgEnabledQueryDependencyLinksStore (which I think is relevant).

kghbln:
```php
### Cache settings

## Backend
$wgObjectCaches['redis'] = [
	'class' => 'RedisBagOStuff',
	'servers' => [
		'127.0.0.1:6379'
		],
	'password' => $sgRedisKey,
	'persistent' => true,
	];
$wgJobTypeConf['default'] = [
	'class' => 'JobQueueRedis',
	'redisServer' => '127.0.0.1:6379',
	'redisConfig' => [
		'connectTimeout' => 2,
		'password' => $sgRedisKey,
		'compression' => 'gzip',
		],
[...]

PR #4346, opened 2019-11-04
Add Integration Test to repeated change of DISPLAYTITLE propagation.

SbstnS:
This PR is made in reference to: #4345


kghbln:
I am not sure if the json testing framework works for a dynamic environment. My strong guess is that it does not.

gesinn-it:
@kghbln in our test environment we could reproduce #4345 with the JSONScript tests. At the moment we expect that the test including the query fails. That's what travis has reported. In an older version of SMW (f5a10cd326fc97155b28d358b44aedb9ca0c9a66) all tests passed.

jaideraf:
@gesinn-it-gea this pull is still needed? Could you resolve the conflicts?

PR #4358, opened 2019-11-20
Update ResultPrinter.php

zoranzoki21:
Remove getParameters() function which is deprecated since 1.8


JeroenDeDauw:
Hi @zoranzoki21. Thanks for the effort. Unfortunately this method is still used, both in SMW itself and in several extensions to SMW. We cannot get rid of it easily. For current efforts on improving the result printers, please see https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4313

PR #4362, opened 2019-12-06
Fix types and imports in SharedServicesContainer

JeroenDeDauw:
Not comprehensive

### DIFF
--- a/src/Services/SharedServicesContainer.php
+++ b/src/Services/SharedServicesContainer.php
@@ -7,31 +7,36 @@ use Onoi\BlobStore\BlobStore;
 use Onoi\CallbackContainer\CallbackContainer;
 use Onoi\CallbackContainer\ContainerBuilder;
 use SMW\CacheFactory;
+use SMW\Connection\ConnectionManager;
+use SMW\Constraint\ConstraintErrorIndicatorProvider;
+use SMW\ConstraintFactory;
 use SMW\ContentParser;
 use SMW\DataItemFactory;
+use SMW\DependencyValidator;
+use SMW\DisplayTitleFinder;
+use SMW\Elastic\ElasticFactory;
+use SMW\EntityCache;
 use SMW\Factbox\FactboxFactory;
 use SMW\HierarchyLookup;
 use SMW\InMemoryPoolCache;
 use SMW\IteratorFactory;
 use SMW\Localizer;
-use SMW\MediaWiki\Database;
-use SMW\Connection\ConnectionManager;
 use SMW\MediaWiki\Connection\ConnectionProvider;
+use SMW\MediaWiki\Database;
 use SMW\MediaWiki\Deferred\CallableUpdate;
 use SMW\MediaWiki\Deferred\TransactionalCallableUpdate;
-use SMW\MediaWiki\JobQueue;
+use SMW\MediaWiki\IndicatorRegistry;
 use SMW\MediaWiki\JobFactory;
+use SMW\MediaWiki\JobQueue;
 use SMW\MediaWiki\ManualEntryLogger;
 use SMW\MediaWiki\MediaWikiNsContentReader;
 use SMW\MediaWiki\PageCreator;
 use SMW\MediaWiki\PageUpdater;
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
@mwjames the reason I'm looking at this code is that I'm fixing a custom extension (written by someone else) that uses 

    ApplicationFactory::getInstance()->getQueryFactory()->newQueryCreator();

The newQueryCreator method was removed in SMW 3.0. The new way to get this service is 

    ApplicationFactory::getInstance()->create('QueryCreator');

That follows the MediaWiki core service container pattern. I do not think this is a good pattern or that it should be refactored to. I much rather have a ->getFluffService() than a ->get('FluffService'). The former is much more user friendly, especially if it has a return type. Just an FYI for future work.

mwjames:
> The newQueryCreator method was removed in SMW 3.0. The new way to get this
> service is
>
>     ApplicationFactory::getInstance()->create('QueryCreator');
>
> That follows the MediaWiki core service container pattern. I do not think
> this is a good pattern or that it should be refactored to. I much rather
> have a ->getFluffService() than a ->get('FluffService'). The former is much
> more user friendly, especially if it has a return type. Just an FYI for
> future work.

There is a note that clearly states that using `create` or singleton`
is for internal use only and no extension should rely on them to
return any services. Even if they return services (which are for
internal use only), those services can change anytime without notice,
classes not available via an individual accessible method in a factory
are especially not meant for public consumption.

Furthermore, the mentioned class `QueryCreator` is marked as @private
which means it is not supposed to be accessible outside of SMW-core.
As with all things, users can do what they want but they are not
expected to see any support for such classes, no guarantee is given
and can disappear or change their signature within releases without
prior notice.

If a developer thinks s(he) needs a particular service they can open a
ticket and request clarification or advice but classes marked as
private (especially those from the outset [0]) are not considered to
provide any public accessible endpoint.

As for the particular mentioned class, it was born out of the need to
encapsulate some "stuff" from the `QueryProcessor` where the
`QueryProcessor` remains the only public interface (well it should be
a factory but I don't have the time for this) to build a query object
from a string representation.

> have a ->getFluffService() than a ->get('FluffService'). The former is much
> more user friendly, especially if it has a return type. Just an FYI for
> future work.

SMW has been using a container builder much longer than MediaWiki
thought about services or the locator pattern. One difference to the
MW service is that the SMW container builder checks for types and
raises an exception which means services are type safe even when
called via `create` or `singleton`.

PS: Any class marked as `@private` is not for public use or
consumption and I will be more vigilant to mark relevant (which are
most) classes in SMW accordingly in future.

PSS: Classes in `SMW\MediaWiki*` are all considered private and should
not be accessed and relied upon in any extension.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1889/files#diff-83669ea76bf05a131dfff52e6e37c31aR11

JeroenDeDauw:
I did not notice this class is private. How do you recommend I fix this extension then? The below code did not use anything marked as private, but unfortunately the `newQueryCreator` method was removed. I'm not sure why it was removed and did not see another way to get this service from SMW now.

    ApplicationFactory::getInstance()->getQueryFactory()->newQueryCreator();

mwjames:
> I did not notice this class is private. How do you recommend I fix this
> extension then? The below code did not use anything marked as private, but
> unfortunately the `newQueryCreator` method was removed. I'm not sure why it
> was removed and did not see another way to get this service from SMW now.

`newQueryCreator` was removed because `QueryCreator` is (and was from
the beginning) private and to avoid anyone assuming that this is
accessible the direct factory methods was removed.

Why is someone relying on this class? The user should use
`QueryProcessor` as public interface to the query, the only consumer
of the `QueryCreator` is the `QueryProcessor` which is why it was
marked private from the start.

JeroenDeDauw:
I don't know. And after dealing with list related bullshit for 4 hours (see https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4363) I can't bring myself to care either. It works now, which is good enough in this case.

Issue #4363, opened 2019-12-07
Current default ask format no longer has default parameter


JeroenDeDauw:
It did in SMW 2.5 and earlier. This is a breaking change.

Caused by the plainlist related changes.

JeroenDeDauw:
This used to output "0" and now outputs nothing:

```
[[SomeDate::2000/01/02]]
{{#ask: [[{{PAGENAME}}]] |?SomeDate= |mainlabel=- |default=0 }}
```

JeroenDeDauw:
Interestingly enough the `plainlist` format is also not outputting anything even though it does have a `default` parameter.

```
{{#ask: [[{{PAGENAME}}]] |?SomeDate= |mainlabel=- |default=hello|format=plainlist }}
```

JeroenDeDauw:
Example: https://sandbox.semantic-mediawiki.org/wiki/Plainlist_default_value

Interestingly enough it worked fine the first time I saved it. After a third save the "hello" went away. Something is very wrong.

JeroenDeDauw:
@s7eph4n

JeroenDeDauw:
This piece of code appears to have gotten lost: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/2.5.8/includes/queryprinters/ListResultPrinter.php#L190-L193

kghbln:
Moved in the fix commit into sandbox for a sneak preview: https://sandbox.semantic-mediawiki.org/wiki/Issue/4363 Appears that "plainlist" is now working, however "list" still fails. Will add a test to the commit.

JeroenDeDauw:
Thanks Karsten. ol, ul and "richlist" are indeed still failing.

This is because the new list code happily wraps empty content with a bunch of HTML. It is doing that on a several places and will take some refactoring to fix. I do not have time for this now.

kghbln:
Just added integration tests for plainlist results only

PR #4364, opened 2019-12-08
Restore list format defaulting

JeroenDeDauw:
Fixes SMW 3.0 regression (by restoring 2.5.x behavior)

Reported in https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4363

### DIFF
--- a/src/Query/ResultPrinters/ListResultPrinter.php
+++ b/src/Query/ResultPrinters/ListResultPrinter.php
@@ -55,12 +55,17 @@ class ListResultPrinter extends ResultPrinter {
 	 * @return string
 	 */
 	protected function getResultText( SMWQueryResult $queryResult, $outputMode ) {
-
 		$builder = $this->getBuilder( $queryResult );
 
 		$this->hasTemplates = $this->hasTemplates();
 
-		return $builder->getResultText() . $this->getFurtherResultsText( $queryResult, $outputMode );
+		$result = $builder->getResultText() . $this->getFurtherResultsText( $queryResult, $outputMode );
+
+		if ( $result == '' ) {
+			return $this->params['default'];
+		}
+
+		return $result;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
Tested manually with `{{#ask: [[{{PAGENAME}}]] |?IsNotSet=| mainlabel=- |default=hello }}`. Still needs regression test but I won't have time for that today.

JeroenDeDauw:
As per comments on https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4363, this PR only fixes plainlist. The other list formats are still broken. That ought to not prevent this partial fix from being merged.

kghbln:
no bloody idea why the tests fail

kghbln:
Ok, I finally managed to get the tests working. As a result we need a squash and merge.

src/Query/ResultPrinters/ListResultPrinter.php
```diff
@@ -55,12 +55,17 @@ public function isDeferrable() {
 	 * @return string
 	 */
 	protected function getResultText( SMWQueryResult $queryResult, $outputMode ) {
-
 		$builder = $this->getBuilder( $queryResult );
 
 		$this->hasTemplates = $this->hasTemplates();
 
-		return $builder->getResultText() . $this->getFurtherResultsText( $queryResult, $outputMode );
+		$result = $builder->getResultText() . $this->getFurtherResultsText( $queryResult, $outputMode );
```
s7eph4n: What about just asking for `$queryResult->getCount()` here and not even calling the builder? 
The question is, what is expected for list/ol/ul for getCount === 0. Just an empty string or a wrapped empty string?

JeroenDeDauw: `$queryResult->getCount()` is already compared to zero before this method is invoked. That just checks if there are no subjects though. The defaulting should also happen when there are subjects but no values.

If there are no values the default should be shown, at least if we want to restore the old behavior. Seems sensible enough to me.

PR #4365, opened 2019-12-08
Allow restoring behavior for format=list

JeroenDeDauw:
SMW <3.0 wikis that make extensive use of format=list cannot easily
be upgraded due to the breaking changes SMW 3.0 brings in the form
of added HTML in the list output. In big wikis switching over to
plainlist can be an inpractical amount of work (even with search and replace)
and be very error prone.

This change renames list to richlist and adds list as an alias to richlist.
It thus does not change SMW 3.1 behavior. It however does allow the user
to restore pre-3.0 behavior via the $smwgResultAliases setting:

```php
$smwgResultAliases['richlist'] = [];
$smwgResultAliases['plainlist'][] = ['list'];
```

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -775,7 +775,7 @@ return [
 	'smwgResultFormats' => [
 		'table'      => 'SMW\Query\ResultPrinters\TableResultPrinter',
 		'broadtable' => 'SMW\Query\ResultPrinters\TableResultPrinter',
-		'list'       => 'SMW\Query\ResultPrinters\ListResultPrinter',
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Alright was afraid that some stuff would blow up when renaming the format like this.

Maybe better to leave it be and add a new setting.

kghbln:
> Maybe better to leave it be and add a new setting.

In this case `$smwgListPlain` would probably the the parameter following existing naming scheme.

JeroenDeDauw:
Closed in favor of https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4367

PR #4367, opened 2019-12-09
Add smwgPlainList setting

JeroenDeDauw:
SMW <3.0 wikis that make extensive use of format=list cannot easily
be upgraded due to the breaking changes SMW 3.0 brings in the form
of added HTML in the list output. In big wikis switching over to
plainlist can be an impractical amount of work (even with search and replace)
and be very error prone.

This change allows restoring old behavior.

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -810,6 +810,19 @@ return [
 	],
 	##
 
+	/**
+	 * Affects format=list.
+	 *
+	 * When set to false (the default), format=list will result in lists with HTML markup.
+	 * In this case you can get a plain list via format=plainlist.
+	 *
+	 * To also get plain lists (without HTML markup) when using format=list, set this setting to true.
+	 * In SMW versions older than 3.0 format=list always resulted in a plain list, so this setting allows restoring old behavior.
+	 *
+	 * @since 3.1.2
+	 */
+	'smwgPlainList' => false,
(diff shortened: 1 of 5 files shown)


s7eph4n:
Well, I see the need, but this also incurs debt. Maybe already mark the setting as deprecated with a view to remove it in SMW 4? Considering the release record of SMW this would give users several years to adapt.

JeroenDeDauw:
Why would this need to be deprecated? Why should users be forced to go through a potentially super painful migration that adds no value for them? I'm not sure the behavior change in SMW 3.0 was a good idea. Perhaps adding a new "richlist" format would have been better.

I'm not sure what debt you are talking about. The added complexity is minimal and is significantly less than the _accidental_ complexity in the list related code which itself is a small issue compared to the whole result printer mess (https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4313).

JeroenDeDauw:
So I think this should NOT be deprecated.

kghbln:
We are indeed talking about a mid to long term horizon so we need not do make a final decision on deprecation now. I am currently for / against doing so by 50 %.

kghbln:
@JeroenDeDauw I guess this also affects ol and ul?

JeroenDeDauw:
It does not

PR #4388, opened 2020-01-11
Fix static code analysis in ServicesFactory

JeroenDeDauw:


JeroenDeDauw:
> Return value of SMW\Services\ServicesFactory::newCacheFactory() must be an instance of Onoi\Cache\CacheFactory, instance of SMW\CacheFactory returned

src/Services/ServicesFactory.php
```diff
@@ -156,55 +175,43 @@ public function create( ...$service ) {
 
 	/**
 	 * @since 2.0
-	 *
-	 * @return SerializerFactory
```
mwjames: Why do we remove the comments? Just because typing is added doesn't equally mean that comments can be removed.

src/Services/ServicesFactory.php
```diff
@@ -2,28 +2,47 @@
...
+use SMW\CacheFactory;
+use SMW\Connection\ConnectionManager;
+use SMW\ContentParser;
+use SMW\DataItemFactory;
+use SMW\DataUpdater;
+use SMW\DataValueFactory;
+use SMW\DeferredCallableUpdate;
+use SMW\DeferredTransactionalCallableUpdate;
+use SMW\EntityCache;
+use SMW\EventHandler;
```
mwjames: This sort of change here costs me a lot of grievance (in the past and as of now) as it forces unnecessary work on rebasing PRs.

src/Services/ServicesFactory.php
JeroenDeDauw: Why duplicate what is already in the code in a comment? It makes the code more verbose / less information dense, makes it more costly to modify and adds room for inconsistencies.

mwjames: That is what the comment block is for to inform the user and document generator about what is expected here. Typing is added so the compiler and the PHP engine can verify against a codified expectation but that doesn't nullify the documentation task.

JeroenDeDauw: Are we using a documentation generator that does not use key information added in PHP 7.0? I certainly hope not.

src/Services/ServicesFactory.php
JeroenDeDauw: Merge conflicts are the worst

Maybe I should have disabled the "reorder alphabetically" feature before running the cleanup command for this particular file, or for SMW in general. I'm used to working in projects where this is cleanup a non-event and rarely or never causes merge conflicts.

src/Services/ServicesFactory.php
mwjames: > Are we using a documentation generator that does not use key information added in PHP 7.0? I certainly hope not.

I don't know but https://doc.semantic-mediawiki.org/ runs doxygen 1.8.13 that aside, when you are removing comments like above yet I'm adding comments (independent of whether new methods are typed or not) introduces inconsistent code styling.

Issue #4382, opened 2019-12-28
Performance issues


acnetj:
### Setup and configuration

MediaWiki | 1.34.0
PHP | 7.4.1 (fpm-fcgi)
MySQL | 5.7.25-0ubuntu0.16.04.2
Semantic MediaWiki | 3.1.1

After upgrade I have to save the page twice in order for the edits to show up on query results. If I just save once after edit the query output is still old. Wasn't that way before upgrade.

mwjames:
> MediaWiki | 1.34.0
> PHP | 7.4.1 (fpm-fcgi)

I don't remember that the compatibility matrix mentions anything about
supporting either MW 1.34+ or PHP 7.4+ which means users that deploy
the software on systems that aren't tested (meaning working outside of
the compatibility range) will unlikely be satisfied

There is a reason why there is a compatibility matrix and users should
not expect any support assistance for unsupported environments.

That said, adding MW 1.34 to the CI should be the next step, yet I'm
confident that it will break the SMW test suite and is likely hinting
to an issue that has been introduced with MW 1.34 given that no
fundamental changes have been merged to SMW in the last 4/5 months.

> After upgrade I have to save the page twice in order for the edits to show
> up on query results. If I just save once after edit the query output is
> still old. Wasn't that way before upgrade.

This description is too ambiguous (at least for me) which means I
cannot (and will not) respond to any such issues that don't provide a
clear explanation of how to replicate such issue especially now that
my volunteer time will become more restricted starting with 2020.

Cheers

kghbln:
I have mixed experience and the result is probably dependent on the result format used. On [sandbox](https://sandbox.semantic-mediawiki.org/wiki/Utilisateur:Kghbln/Unknown_count) I have to add the subobject and purge twice to get it to show in the graph (MW 1.34, SMW 3.1, PHP 7.2). On another wiki (MW 1.31, SMW 3.1 and PHP 7.0) the results show immediately for the table format.

kghbln:
Ok, I created some examples on [sandbox](https://sandbox.semantic-mediawiki.org/wiki/Issue/4382) and the behavior is identical for both formats. 1. Add subobject to add data, 2. Purge once to get the results. For me this is kinda expected and I think the [`@deferred` query marker](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Deferred_query_marker) is meant at the way out here.

[...]

PR #4394, opened 2020-01-11
[CI] Add MW 1.34

mwjames:
This PR is made in reference to: https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4382#issuecomment-570825763

- Adds MW 1.34 to the CI as non-voting because it breaks the test suite with 8 errors and 28 failures [0] as predicted in the referenced comment

This PR includes:
- [x] CI build passed

[0] https://travis-ci.org/SemanticMediaWiki/SemanticMediaWiki/jobs/635751414

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ matrix:
       php: 7.1
     - env: DB=mysql; MW=REL1_31; TYPE=benchmark; PHPUNIT=6.5.*
       php: 7.2
+    - env: DB=mysql; MW=REL1_34; PHPUNIT=6.5.*
+      php: 7.2
     - env: DB=sqlite; MW=REL1_31; TYPE=composer; PHPUNIT=5.7.*
       php: 7.1
     - env: DB=mysql; MW=REL1_31; TYPE=relbuild; PHPUNIT=5.7.*
@@ -45,7 +47,7 @@ matrix:
     - env: DB=mysql; MW=REL1_31; TYPE=benchmark; PHPUNIT=6.5.*
     #- env: DB=mysql; MW=REL1_31; SESAME=2.8.7; PHPUNIT=5.7.*
     - env: DB=mysql; MW=REL1_33; ES=5.6.6; PHPUNIT=5.7.*
-
+    - env: DB=mysql; MW=REL1_34; PHPUNIT=6.5.*
 
 # Dec 16, 2015 (GCE switch): Travis support wrote (Tomcat + Java):
 # bug in the JDK: http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7089443.


mwjames:
>  it breaks the test suite with 8 errors and 28 failures [0]

Neither master nor any other release is compatible with MW 1.34, users who try to use that version (or above) do so with no guarantee that the software works as intended.

mwjames:
The most serious issue here are the broken integration tests which indicate (or suggest) that "old" revision data are referenced and used instead.

JeroenDeDauw:
> The most serious issue here are the broken integration tests which indicate (or suggest) that "old" revision data are referenced and used instead.

That would explain why queries are not updating properly on sandbox. @kghbln figured this was caused by 1.34.

mwjames:
> That would explain why queries are not updating properly on sandbox. @kghbln figured this was caused by 1.34.

Yes, and I have a hypothesis of the cause but I will unlikely have time this weekend to verify that.

kghbln:
> figured this was caused by 1.34.

I am nearly a 100 percent sure about this. Currently when you add an annotation it is not saved during the edit process. A reparse is required. Pretty nasty if you ask me. :(

I was not aware that 1.34 is such a "toxic" release. Otherwise I would not have upgraded sandbox in October 2019 but only now.

mwjames:
> I am nearly a 100 percent sure about this. Currently when you add an annotation it is not saved during the edit process. A reparse is required. Pretty nasty if you ask me. :(

Aside from the core issue caused by MW 1.34, I'm planning to add an entity examiner (over the holidays I has some time at hand) that informs the user about issues similar to what we do with Elastic.

image

It will compare the associated revision (which we store in SMW) with the revision of the viewed page content to hint a mismatch but for that to happen it needs some preparation since it will combine current separated indicators (elastic, constraint etc.) into one panel.

kghbln:
> Aside from the core issue caused by MW 1.34, I'm planning to add an entity examiner (over the holidays I has some time at hand) that informs the user about issues similar to what we do with Elastic.

I guess this will  be pretty useful. :)

jaideraf:
> I am nearly a 100 percent sure about this. Currently when you add an annotation it is not saved during the edit process. A reparse is required. Pretty nasty if you ask me. :(

I plan to recreate my whole website and I begun a fresh new instance of MW 1.34 with SMW 3.1.1 (I know it is not endorsed by the [compatibility matrix](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/COMPATIBILITY.md), but I am a SMW tester, so...) and I see the same symptoms you describe. 

Trying to downgrade from MW 1.34 to 1.33 from git/update.php is not possible either due to some database changes, so a hell situation. And I presume more hell situations are coming in the next MW releases due to the change from the traditional PHP parser to the new PHP/Parsoid.

Sadly, I have seeing SMW users here, and on the user mailing list, upgrading their instances to MW 1.34 without seeing the compatibility matrix... maybe a notice on the [README.md](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/README.md) here and on the smw.org could help the awareness. What do you think @kghbln?

kghbln:
Thanks for your feedback.

> Sadly, I have seeing SMW users here, and on the user mailing list, upgrading their instances to MW 1.34 without seeing the compatibility matrix... maybe a notice on the README.md here and on the smw.org could help the awareness.

I have made the compatibility info more intuitive to understand: abf9a8003dac2a8c6f2b92fc6e7b38a952a43302 This should be enough to help the cause.

PR #4408, opened 2020-01-12
[CI] Set minimum PHPUnit to 6.5+

mwjames:

This PR includes:
- [x] CI build passed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,39 +14,39 @@ matrix:
       php: 7.2
     - env: DB=sqlite; MW=REL1_33; PHPUNIT=6.5.*
       php: 7.3
-    - env: DB=sqlite; MW=REL1_31; SITELANG=ja; PHPUNIT=5.7.*
+    - env: DB=sqlite; MW=REL1_31; SITELANG=ja; PHPUNIT=6.5.*
       php: 7.1
-    - env: DB=mysql; MW=REL1_31; BLAZEGRAPH=1.5.2; PHPUNIT=5.7.*
+    - env: DB=mysql; MW=REL1_31; BLAZEGRAPH=1.5.2; PHPUNIT=6.5.*
       php: 7.1
-    - env: DB=mysql; MW=REL1_32; FUSEKI=2.4.0; PHPUNIT=5.7.*
+    - env: DB=mysql; MW=REL1_32; FUSEKI=2.4.0; PHPUNIT=6.5.*
       php: 7.1
-    - env: DB=mysql; MW=REL1_31; VIRTUOSO=6.1; PHPUNIT=5.7.*
+    - env: DB=mysql; MW=REL1_31; VIRTUOSO=6.1; PHPUNIT=6.5.*
       php: 7.1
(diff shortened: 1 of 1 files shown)


mwjames:
In order to prepare for a possible upgrade to PHPUnit 7 or 8 (see [0] for the reasoning), this PR raises the minimum to 6.5+ which allows to start to use some migration path towards the objective.

While there shouldn't be any noticeable impact as of now, when backporting some of the changes to an older release we may hit an issue since the previous release uses 5.7+ (or lower) and can conflict with a test that is included in a backported change hence some additional caution is warranted.

@kghbln FYI

[0] https://phpunit.de/supported-versions.html

kghbln:
> While there shouldn't be any noticeable impact as of now, when backporting some of the changes to an older release we may hit an issue since the previous release uses 5.7+ (or lower) and can conflict with a test that is included in a backported change hence some additional caution is warranted.

@mwjames Why don't we back-port also the changes to the current release branch. It looks to me that 3.1.x would pass on 6.5

@JeroenDeDauw I would greatly appreciate if you could squash and merge or do something else that the merge commit appears coupled with the change commit. Currently is looks like an antipattern to me an it makes it really hard to reconstruct that is happening.

JeroenDeDauw:
squash does not do anything if there is just 1 commit AFAIK. Could do a rebase and merge, in which case there is no merge commit. Would you be more happy with that?

kghbln:
> squash does not do anything if there is just 1 commit AFAIK

I does. Now the have the merge commit at the top and about 30 commits down the change commit. This makes is much worse to get the idea that these are basically the same thing. In between we have jolly good bunch of changes unrelated to this one. Why is it so super painful to do a squash and merge. This worked in the past years for this repo.

JeroenDeDauw:
> Why is it so super painful to do a squash and merge.

Hey, no need to jump to conclusions here. I was not making a deliberate choice, just hitting the green button, as I've been doing 1000s of times over the last 10 years.

kghbln:
> Hey, no need to jump to conclusions here

I am not jumping but grumping here. I think GitHub remembers the last button option set. Anyways, I could do the merges, preferably if also James additionally approves the pull.

JeroenDeDauw:
GH remembers the last merge option used yes, but this is global, not per repository. The other options exist for a reason, and there also is a reason that squash is not the default and is the most recent option added.

mwjames:
> @mwjames Why don't we back-port also the changes to the current release branch. It looks to me that 3.1.x would pass on 6.5

On the matter of squash vs simple merge:

> makes is much worse to get the idea that these are basically the same thing. In between we have jolly good bunch of changes unrelated to this one. Why is it so super painful to do a squash and merge. This worked in the past years for this repo.

@kghbln Is right that for the last couple of years squash has been used as merge option to avoid unnecessary decoupled merge commits.

image

JeroenDeDauw:
Squash and merge destroys author info in git blame though. Isn't rebase the better option if you want to avoid merge commits?

PR #4417, opened 2020-01-12
Use top navigation in JSONView, refs 3673

mwjames:
This PR is made in reference to: #3673

- Moving the interaction buttons into a top navigation

This PR includes:
- [x] CI build passed

## Before

image

## After

image

### DIFF
--- a/res/Resources.php
+++ b/res/Resources.php
@@ -37,7 +37,8 @@ return [
 			'smw/ext.smw.table.css',
 			'smw/ext.smw.tabs.css',
 			'smw/factbox/smw.factbox.css',
-			'smw/smw.indicators.css'
+			'smw/smw.indicators.css',
+			'smw/smw.jsonview.css'
 		],
 		'position' => 'top',
 		'targets' => [ 'mobile', 'desktop' ]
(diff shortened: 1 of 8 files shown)


kghbln:
@mwjames This only seems to work in some places

**works**
- [property groups](https://sandbox.semantic-mediawiki.org/wiki/smw/schema:Group:Exif_special_properties)
- [profiles](https://sandbox.semantic-mediawiki.org/wiki/smw/schema:Search-profile-form-definition)

**does not work**
- [Property constraints](https://sandbox.semantic-mediawiki.org/wiki/smw/schema:Constraint:NonNegativeInteger)
- [Formatters](https://sandbox.semantic-mediawiki.org/wiki/smw/schema:Keyword_link_formatter)

mwjames:
> @mwjames This only seems to work in some places

> does not work    Property constraints

This needs an extra HTML element and when the page is cached (parser cache) this is missing which is why a purge will fix this as in case of [0].

[0] https://sandbox.semantic-mediawiki.org/wiki/smw/schema:Constraint:NonNegativeInteger

kghbln:
Ok, cool. I am pretty sure that I purged the page before writing all of this. Probably I only did an F5 in the browser instead of a Ctrl + F5.

mwjames:
> Probably I only did an F5 in the browser instead of a Ctrl + F5.

Isn't enough as we need to purge the parser cache to create a new HTML output.

PR #4436, opened 2020-01-19
Fix @annotation behavior when used with Page Forms

JeroenDeDauw:
Manually verified and not sure how to write a non-browser
automated test for this.

Ideally this code would be further refactored so it does
not break whenever it is used together with an extension
that adds an action. Might currently also be broken with
some visual editor stuff, did not check.

### DIFF
--- a/src/MediaWiki/Hooks/OutputPageParserOutput.php
+++ b/src/MediaWiki/Hooks/OutputPageParserOutput.php
@@ -95,7 +95,7 @@ class OutputPageParserOutput extends HookHandler {
 
 		$request = $outputPage->getContext()->getRequest();
 
-		if ( in_array( $request->getVal( 'action' ), [ 'delete', 'purge', 'protect', 'unprotect', 'history', 'edit' ] ) ) {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
@mwjames is there is reason to use a blacklist instead of a whitelist for the actions?

mwjames:
> @mwjames is there is reason to use a blacklist instead of a whitelist for the actions?

Those listed are the "known" actions that should be avoided (hence the balacklist) to invoke a postproc via the `OutputPageParserOutput` hook.

mwjames:
> Manually verified and not sure how to write a non-browser
automated test for this.

This is going to be difficult even with the `onOutputPage` option given that the postproc relies on `EditPage::POST_EDIT_COOKIE_KEY_PREFIX` to be set via the `WebRequest`.

JeroenDeDauw:
What about using a whitelist? Or are there unknown actions (or those added by extensions) for which the code should be run?

If whitelist does not work, then perhaps there is another way to make this code more robust? I'm not seeing one but then again I am not very familiar with the code and its context.

PR #4458, opened 2020-01-25
Add `Special:PendingTaskList`

mwjames:

- Extends #3095 so that instead of showing an incomplete task list on each page, a simplified note will be displayed with a link to `Special:PendingTaskList` where users can inform themselves about incomplete (or pending) tasks
- `Special:PendingTaskList` is kept generic to allow to extend the content displayed on this special page for different contexts that informs about pending tasks

This PR includes:
- [x] CI build passed

## Example

image

image


kghbln:
[Documented](https://www.semantic-mediawiki.org/wiki/Help:Special:PendingTaskList)

Issue #4486, opened 2020-01-28
SMW\MediaWiki\PageInfoProvider::getCreationDate() can be called when there is no first revision


hexmode:
### Setup and configuration

- SMW version:  2.5.7 (d528e24)
- MW version:  1.27.3 (73756b7)
- PHP version: 5.6.27
- DB system (MySQL, Blazegraph, etc.) and version: Percona 5.7.23-23-57-log

### Issue
Whether due to my particular config or some other problem in SMW, `SMW\MediaWiki\PageInfoProvider::getCreationDate()`  was called without an initial revision being created for a page.

### Stack trace
Instrumenting the function and adding `debug_print_backtrace()` gave the following:
```
#0 .../extensions/SemanticMediaWiki/src/PropertyAnnotators/PredefinedPropertyAnnotator.php(92): SMW\MediaWiki\PageInfoProvider->getCreationDate()
#1 .../extensions/SemanticMediaWiki/src/PropertyAnnotators/PredefinedPropertyAnnotator.php(69): SMW\PropertyAnnotators\PredefinedPropertyAnnotator->createDataItemByPropertyId(string)
#2 .../extensions/SemanticMediaWiki/src/PropertyAnnotators/PropertyAnnotatorDecorator.php(61): SMW\PropertyAnnotators\PredefinedPropertyAnnotator->addPropertyValues()
#3 .../extensions/SemanticMediaWiki/src/MediaWiki/Hooks/NewRevisionFromEditComplete.php(135): SMW\PropertyAnnotators\PropertyAnnotatorDecorator->addAnnotation()
#4 .../extensions/SemanticMediaWiki/src/MediaWiki/Hooks/NewRevisionFromEditComplete.php(77): SMW\MediaWiki\Hooks\NewRevisionFromEditComplete->doExtendParserOutput(ParserOutput)
#5 .../extensions/SemanticMediaWiki/src/MediaWiki/Hooks/HookRegistry.php(264): SMW\MediaWiki\Hooks\NewRevisionFromEditComplete->process()
#6 [internal function]: SMW\MediaWiki\Hooks\HookRegistry->SMW\MediaWiki\Hooks\{closure}(WikiPage, Revision, boolean, User)
#7 .../includes/Hooks.php(195): call_user_func_array(Closure, array)
#8 .../includes/page/WikiPage.php(1903): Hooks::run(string, array)
[...]

PR #4487, opened 2020-01-28
Add test to demonstrate #4486

hexmode:
This PR is made in reference to: #4486 

This PR includes:
- [ ] CI build passed

### DIFF
--- a/src/MediaWiki/PageInfoProvider.php
+++ b/src/MediaWiki/PageInfoProvider.php
@@ -66,6 +66,10 @@ class PageInfoProvider implements PageInfo {
 	 * @return integer
 	 */
 	public function getCreationDate() {
+		if ( $this->wikiPage->getTitle()->getFirstRevision() === null ) {
+			return wfTimestamp( TS_MW, 0 );
+		}
+
 		return $this->wikiPage->getTitle()->getFirstRevision()->getTimestamp();
 	}
 
--- a/tests/phpunit/src/MediaWiki/PageInfoProviderTest.php
+++ b/tests/phpunit/src/MediaWiki/PageInfoProviderTest.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace SMW\Test;
+
+use MWTimestamp;
+use Revision;
+use SMW\MediaWiki\PageInfoProvider;
+use SMW\PageInfo;
+use User;
+use WikiPage;
+
+/**
+ * @covers \SMW\MediaWiki\PageInfoProvider
+ *
+ *
+ * @group SMW
+ * @group SMWExtension
+ *
+ * @license GNU GPL v3+
+ * @since 3.2.0
+ *
+ * @author MarkAHershberger
+ */
+class PageInfoProviderTest extends SemanticMediaWikiTestCase {
+
+	public function getClass() {
+		return '\SMW\MediaWiki\PageInfoProvider';
+	}
+
+	private function newInstance(
+		WikiPage $wikiPage = null,
+		Revision $revision = null,
+		User $user = null
+	) {
+		if ( $wikiPage === null ) {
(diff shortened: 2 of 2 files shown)


src/MediaWiki/PageInfoProvider.php
```diff
@@ -66,6 +66,10 @@ public function getModificationDate() {
 	 * @return integer
 	 */
 	public function getCreationDate() {
+		if ( $this->wikiPage->getTitle()->getFirstRevision() === null ) {
+			return wfTimestamp( TS_MW, 0 );
+		}
```
JeroenDeDauw: This looks good as a local fix though I'm wondering if it is treating a symptom rather than the root cause. I do not know the code well enough to tell. One for @mwjames

tests/phpunit/src/MediaWiki/PageInfoProviderTest.php
```diff
@@ -0,0 +1,53 @@
...
+	public function testCanConstruct() {
+		$this->assertInstanceOf( $this->getClass(), $this->newInstance() );
+	}
+
+	/**
+	 * @depends testCanConstruct
+	 */
+	public function testCanGetCreationDate() {
+		$this->assertInstanceOf( 'MWTimestamp', $this->newInstance()->getCreationDate() );
+	}
```
JeroenDeDauw: A very high portion of the complexity here is not needed.

You can create an equivalent test by extending only the base PHPUnit testcase class and having a single simple test method like

```php
testCanGetCreationDate() {
    $p = new PageInfoProvider( Title::newFromText( __METHOD__ ), new WikiPage( $title ), null );
    $this->assertInstanceOf( MWTimestamp::class, $p->getCreationDate() );
}
```

But then with proper naming and indent.

hexmode:
re: complexity via @JeroenDeDauw

You are right.  I was just copying the style of other tests and hoping
it would be useful later.

mwjames:
@hexmode I appreciate your effort you showed here to demonstrate the issue you apparently encountered (see my response on the original ticket) but the test in itself doesn't demonstrate the issue because you would need a real `Database` connection since your are working with the `Revision` object.

Instead, you could mock the `Revision`/`Title` object to show that `Title::getFirstRevision` is allowed to return `null`.

A couple of notes:

- `SemanticMediaWikiTestCase` is depreacted so for unit tests we use `PHPUnit_Framework_TestCase` directly 
- For integration tests that require an MW object (Title, Revision, Page) or an integration with the `Database` or MW we rely on `DatabaseTestCase`
- We don't use any of any of the `MediaWiki * TestCase` stuff since that breaks regularly and I cannot waste time adopting tests on each release


[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/README.md#test-cases

mwjames:
> Instead, you could mock the Revision/Title object to show that Title::getFirstRevision is allowed to return null.

Without pulling in an active `Title` or `WikiPage` you could simply do and which gets you "Error: Call to a member function getTimestamp() on null"

```
	public function testWikiPage_TYPE_CREATION_DATE_FirstRevisionIsNull() {

		$title = $this->getMockBuilder( '\Title' )
			->getMock();

		$title->expects( $this->any() )
			->method(  'getFirstRevision' )
			->will( $this->returnValue( null ) );

		$wikiPage = $this->getMockBuilder( '\WikiPage' )
			->getMock();

		$wikiPage->expects( $this->once() )
			->method(  'getTitle' )
			->will( $this->returnValue( $title ) );

		$instance = new PageInfoProvider( $wikiPage );

		$this->assertEquals(
			'something',
			$instance->getCreationDate()
		);
	}
```

kghbln:
Is this one stuck or no longer needed?

PR #4497, opened 2020-02-01
Provide tooltip for the property page search filter

### LINKED ISSUE #4490: Filter function features on property pages are not easy accessible
### Setup and configuration

- SMW version: 3.1.3
- MW version: 1.31.6
- PHP version: 7.2.24-0ubuntu0.18.04.2
- DB system (MySQL, Blazegraph, etc.) and version: 10.1.43-MariaDB-0ubuntu0.18.04.1

### Issue
The filter function on property pages does not work correctly. If you type something in, no results are displayed. Only if you type in an existing value exactly (case sensitive) you get results.

### Steps to reproduce
Go to any property page, for example https://sandbox.semantic-mediawiki.org/wiki/Attribut:Has_colour and try to filter. In this case, you don't get any results when filtering by e.g. "yello", but you do when filtering by "yellow". 
Note that this is case sensitive, "Yellow" won't give any results, too.

kghbln:
This PR is made in reference to: #4490 

- Provides tooltip for the property page search filter as suggested by @mwjames 

This PR includes:
- [x] CI build passed

Fixes #4490

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -944,6 +944,7 @@
 	"smw-parser-recursion-level-exceeded": "The level of $1 recursions was exceeded during a parse process. It is suggested to validated the template structure, or if necessary adjust configuration parameter <code>$maxRecursionDepth</code>.",
 	"smw-property-page-list-count": "Showing $1 {{PLURAL:$1|page|pages}} using this property.",
 	"smw-property-page-list-search-count": "Showing $1 {{PLURAL:$1|page|pages}} using this property with a \"$2\" value match.",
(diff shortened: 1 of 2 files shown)


mwjames:
@kghbln Added [0] with some rudimentary information, surely this needs a bit more "fluff" to be useful.

[0] https://www.semantic-mediawiki.org/wiki/Help:Property_page/Filter

kghbln:
> Added [0] with some rudimentary information, surely this needs a bit more "fluff" to be useful.

Thanks. [Fluffed up!](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AProperty_page%2FFilter&type=revision&diff=74020&oldid=74002}

PR #4499, opened 2020-02-02
Improve filter tooltip, refs 4497

mwjames:
This PR is made in reference to: #4497

This PR includes:
- [x] CI build passed

## Example

image


mwjames:
@kghbln A general note, in case `+` appears next to the number on the tab it indicates that more results are expected to exists when a filter is used.

image

Issue #4501, opened 2020-02-02
Deprecation in implode()


alex-mashin:
PHP 7.4, MW 1.34.0. SMW 3.1.3:
```ErrorException from line 402 of /var/www/wiki/w/extensions/SemanticMediaWiki/src/SQLStore/QueryEngine/QueryEngine.php: PHP Deprecated: implode(): Passing glue string after array is deprecated. Swap the parameters```

JeroenDeDauw:
Thanks for reporting. Fixed by https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4502

PR #4502, opened 2020-02-02
Swap implode arguments

JeroenDeDauw:
Fix https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4501


kghbln:
Officially we do not support PHP 7.4 for SMW 3.1. Not sure whether to back-port or not.

JeroenDeDauw:
Can backport

Should never have been in this order in the first place

mwjames:
> Should never have been in this order in the first place

It what supported in PHP 7.4-.

> Officially we do not support PHP 7.4 for SMW 3.1. Not sure whether to back-port or not.

Can be backported but it doesn't change the fact that PHP 7.4 isn't supported on either 3.1 or master because first of all it isn't in the CI matrix and aside the technicality, it requires more work to get pass all tests.

JeroenDeDauw:
> It what supported in PHP 7.4-.

Many things are still supported in PHP 8 and yet should still not be done.

Issue #4541, opened 2020-02-14
Notices and warnings on PHP 7.4


vapblack:
### Setup and configuration

![image](https://user-images.githubusercontent.com/61056547/74541149-201b8680-4f0f-11ea-8631-2c4d28f2d454.png)

![image](https://user-images.githubusercontent.com/61056547/74541174-29a4ee80-4f0f-11ea-9c4a-57e90260d411.png)

### Issue
Okay so I installed SMW with composer. I added
 "mediawiki/semantic-media-wiki": "~3.1"
and followed the installation instructions at https://www.semantic-mediawiki.org/wiki/Help:Installation/Quick_guide

Now if I go to random pages on my wiki, I get this error at the top and bottom of my pages:

### Stack trace
![image](https://user-images.githubusercontent.com/61056547/74541228-417c7280-4f0f-11ea-8994-bdf844d3b0ff.png)

If I hit refresh (sometime a couple times) the errors go away. On the mage page of the wiki though, the error stays no matter what I do. I tried disabling all extensions. The errors stay. I tried rolling back and doing the install again ( a couple times) and the errors stay.

anyone with any clues on whats going on?

```

Notice: Trying to access array offset on value of type bool in /var/www/worldafropedia/wiki/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Property.php on line 508

Notice: Trying to access array offset on value of type bool in /var/www/worldafropedia/wiki/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Property.php on line 508

Notice: Trying to access array offset on value of type bool in /var/www/worldafropedia/wiki/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Property.php on line 508

Notice: Trying to access array offset on value of type bool in /var/www/worldafropedia/wiki/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Property.php on line 508

Notice: Trying to access array offset on value of type bool in /var/www/worldafropedia/wiki/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Property.php on line 508

Notice: Trying to access array offset on value of type bool in /var/www/worldafropedia/wiki/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Property.php on line 508
```

### Steps to reproduce
If I go to any page while SMW is active, the pages load, but the errors are at the top and/or bottom of the content. If I disable SMW, the errors arent there.

[...]

PR #4542, opened 2020-02-14
Increase type safety in DIProperty

JeroenDeDauw:
Prompted by https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4541

### DIFF
--- a/includes/dataitems/SMW_DI_Property.php
+++ b/includes/dataitems/SMW_DI_Property.php
@@ -52,7 +52,7 @@ class DIProperty extends SMWDataItem {
 	/**
 	 * Either an internal SMW property key (starting with "_") or the DB
 	 * key of a property page in the wiki.
-	 * @var string
+	 * @var string|null
 	 */
 	private $m_key;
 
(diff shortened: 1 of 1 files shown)


mwjames:
Throwing around types isn't going to solve anything here as evident by the broken tests, the implicit behaviour of type casting is part of the historic technical debt in SMW which has to be addressed differently.

The type system isn't the holy grail and should only be used for new classes where the behaviour is tested, adding it in retrospect to "old" classes will most likely fail as shown in the case above. Even if you correct all cases in SMW core, it doesn't mean that is going to be solved in extensions hereby exposing them to unnecessary breakage just to "satisfy" the need for a type.

akuckartz:
> the implicit behaviour of type casting is part of the historic technical debt in SMW which has to be addressed differently.

How?

mwjames:
> How?

There a couple of ways, one is to install PHP 7.4 encounter one issue at a time, identify what is expected, look at what is implicit, and resolve the implicit implication or assumption and make it explicit by either type casting the value directly or find another solution. Anyway, doing this will require diligence, time, and effort just adding a couple of types to the method signature will cause all sorts of trouble.

JeroenDeDauw:
James I knew this would break the tests. It makes the issue more visible. So no this is not proof types are bad.

And yes, obviously types are not a holy grail. Just like tests are not.

> should only be used for new classes

I strongly disagree with this. It goes against both my theoretical understanding and my lived experience over the last 5+ years.

JeroenDeDauw:
On the topic of type casing: if you really want to have this type casting happen you should fix the type docs, because then they are currently _wrong and misleading_. This is one of the great advantages of having the types in the code: you notice when they are wrong.

PR #4544, opened 2020-02-15
Add `MessageLocalizer`, `LocalMessageProvider`

mwjames:

- Adds `MessageLocalizer`
- Adds `LocalMessageProvider` and `runLocalMessageCopy.php`

This PR includes:
- [x] CI build passed

### DIFF
--- a/SemanticMediaWiki.php
+++ b/SemanticMediaWiki.php
@@ -55,6 +55,13 @@ class SemanticMediaWiki {
 			}
 		}
 
+		/**
+		 * @see https://www.mediawiki.org/wiki/Localisation#Localising_namespaces_and_special_page_aliases
+		 */
+		$GLOBALS['wgMessagesDirs']['SemanticMediaWiki'] = __DIR__ . '/i18n';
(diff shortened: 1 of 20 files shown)


mwjames:
@kghbln FYI

`LocalMessageProvider` and `runLocalMessageCopy.php` is a continuation of https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4438#issuecomment-576031115 which provides an interface for local messages.

The background is that maintenance scripts contain mostly `en` text and the `LocalMessageProvider` can be used as bridge to provide translatable text snippets while being independent of the standard MW system. For example, ` i18n/local/maintenance.i18n.json` contains texts that can be used in maintenance scripts as demonstrated in `maintenance/runLocalMessageCopy.php`.

I added the `runLocalMessageCopy.php` (and it notes "This script is not expected to be used by any user or administrator and is only provided for convenience to the Semantic MediaWiki project.") to allow to copy messages from a local file (such as `maintenance.i18n.json` to the canonical `en.json`) and copy translated messages from a MW owned i18n file back to the local file.

Whether we want to make text in scripts using a local message or not is a different discussion, or whether non English text should be shown (based on the `wgLanguageCode` setting) or not is again another discussion. This PR merely opens the possibility to do so in future.

PR #4546, opened 2020-02-15
[ES] Remove redirect from index, refs 4498

mwjames:
This PR is made in reference to: #4498

- Before the #4498, the diff would have marked any `redi` related change as to be deleted, with a `Document` instance created from a `SemanticData` object the check for a redirect (`_REDI`) as to be explicit

This PR includes:
- [x] CI build passed

## Issue 

https://sandbox.semantic-mediawiki.org/wiki/Issue

image


mwjames:
@kghbln FYI

Issue #4308, opened 2019-09-20
PropertyLabelFinder::findCanonicalPropertyLabelById returns bool


JeroenDeDauw:
Users of this code are not expecting that which leads to warnings and likely bugs.

Warnings: https://gist.github.com/mszabo-wikia/bf2745970f308a24c908c1b4734acb37 (most are caused by this issue, some by https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4307) (originally reported in https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4294)

JeroenDeDauw:
@mwjames I expect you have a better idea of how to fix this one than me

JeroenDeDauw:
@mszabo-wikia I take it wikia is working on updating the extensions it uses again? I'm curious what the timeline there is.

JeroenDeDauw:
And what PHP version you (will) use.

mwjames:
> @mwjames I expect you have a better idea of how to fix this one than me

There is nothing wrong with this method as `false` (aka boolean) is
allowed to be returned where it indicates that there is canonical
label for the property available which is not same as an empty label.

This function has been tested this way at least since 2.4, using PHP
7.4 as pretense (or  better its failed test results) to claim
something being wrong here is not a valid argument.

JeroenDeDauw:
There is nothing wrong with the method, but there is something wrong with the combination of the method and its users.

Before the mismatch got silently ignored. To me it looks like there probably is a defect even if we ignore the spamming of the logs.

JeroenDeDauw:
System tests in Maps are failing because of this: https://travis-ci.org/JeroenDeDauw/Maps/jobs/599421578

Relevant code is just creating pages with SMW annotations... https://github.com/JeroenDeDauw/Maps/blob/master/tests/System/SemanticMW/MapQueryTest.php#L61-L91

JeroenDeDauw:
@mwjames your insights into this code would really be appreciated

Issue #4547, opened 2020-02-15
PHP 7.4 compatibillity


akuckartz:
Refs: #4549

This is a feature request.

kghbln:
This will come with SMW 3.2 in May. I guess no need to create a tracking issue for this. @mwjames

mwjames:
Let's keep it as extra ticket since this requires some effort and diligence.

kghbln:
> Let's keep it as extra ticket since this requires some effort and diligence.

Ok, fine with me.

JeroenDeDauw:
@akuckartz "Refs: #4547" is a self reference ;)

kghbln:
> Refs: #4547" is a self reference ;)

Actually this was me. Amended now.

JeroenDeDauw:
Nitpick mode: @kghbln this is either a feature/enhancement or a bug, not a code quality change

JeroenDeDauw:
Is https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4308 the only remaining known issue on PHP 7.4?

kghbln:
> Is #4308 the only remaining known issue on PHP 7.4?

Nobody knows since we currently do not test against PHP 7.4. Current tests already fail with PHP 7.3

mwjames:
> PHP 7.4. Current tests already fail with PHP 7.3

Incorrect, PHP 7.3 + MW master fails while other combinations in regards to PHP 7.3 do pass.

kghbln:
> Incorrect, PHP 7.3 + MW master fails while other combinations in regards to PHP 7.3 do pass.

Ah, indeed. I overlooked the test running against MW 1.33

mwjames:
Anyone could have just added PHP 7.4 as a non voting CI run to see what's missing! Aside from this technical aspect, users who want to see support for PHP 7.4 should have a look at [0] and start contributing to fix the failing tests otherwise support for PHP 7.4 is not guaranteed.

```
Using PHP 7.4.0

Semantic MediaWiki: 3.2.0-alpha, git: 835b2c7
                    SMWSQLStore (mysql:5.6.33-0ubuntu0.14.04.1)

MediaWiki:          1.35.0-alpha, git: 2e3bdb3 (refs/heads/master)
                    Extension vendor autoloader
Site language:      en

Execution time:     2020-02-16 01:21
Debug logs:         Disabled

Xdebug:             Disabled (or not installed)
[...]

PR #4550, opened 2020-02-15
Check for `false`, refs 4308

mwjames:
This PR is made in reference to: #4308, #4547 

- Going to be explicit here, and yes, `PropertyRegistry::getInstance()->findPropertyIdByLabel( $label )` is allowed to return `false`

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/MediaWiki/Page/PropertyPage.php
+++ b/src/MediaWiki/Page/PropertyPage.php
@@ -136,8 +136,12 @@ class PropertyPage extends Page {
 
 		$label = $this->getTitle()->getText();
 
+		if ( ( $key = PropertyRegistry::getInstance()->findPropertyIdByLabel( $label ) ) === false ) {
+			return false;
+		}
+
 		$property = new DIProperty(
-			PropertyRegistry::getInstance()->findPropertyIdByLabel( $label )
+			$key
 		);
 
 		// Ensure to redirect to `Property:Modification date` and not using


kghbln:
@mwjames I guess this may be back-ported?

mwjames:
> @mwjames I guess this may be back-ported?

Yes.

kghbln:
> Yes.

Great, I guess this also fixes #4308?

mwjames:
> Great, I guess this also fixes #4308?

I don't know, it would require me to install PHP 7.4 and have a look at it.

For now this more in line with "There is nothing wrong with this method as `false` (aka boolean) is allowed to be returned where it indicates that there is canonical label for the property available which is not same as an empty label."

kghbln:
> I don't know, it would require me to install PHP 7.4 and have a look at it.

Fair enough. I guess we will find out sooner than later.

src/MediaWiki/Page/PropertyPage.php
```diff
@@ -136,8 +136,12 @@ protected function getRedirectTargetURL() {
 
 		$label = $this->getTitle()->getText();
 
+		if ( ( $key = PropertyRegistry::getInstance()->findPropertyIdByLabel( $label ) ) === false ) {
```
JeroenDeDauw: Having a side effect (variable assignment) inside of an if predicate is unexpected. I recommend against doing this for readability/understandably reasons.

This is also against the MW coding conventions if you care about those.

JeroenDeDauw:
> Great, I guess this also fixes #4308?

I don't think so

kghbln:
Back-ported with 243f0abce4f34e0e04e3dd3dab95430a0f72055b

Issue #4505, opened 2020-02-03
Behaviour of namespaces changed in 3.1, e.g. $smwgNamespacesWithSemanticLinks


krabina:
### Setup and configuration

- SMW version:3.1
- MW version:1.31
- PHP version:7,2

### Issue
The issue showed up when using the CommentStreams extension. What happens is that in a working environment with SMW <3.1 and working CommentStreams, after upgrading to SMW 3.1.x CommentsStremas still worked, but no properties were set.

There are two workarounds for this.
1. set `$smwgNamespacesWithSemanticLinks[844] = true;`
(this step is usually not required, but helps as a workround).
BTW, [documentation](https://www.semantic-mediawiki.org/wiki/Help:$smwgNamespacesWithSemanticLinks) says it should be: `$smwgNamespacesWithSemanticLinks[NS_COMMENTSTREAMS] = true;`
(accoring to [CommentStreams namespaces](https://www.mediawiki.org/wiki/Extension_default_namespaces#CommentStreams))
But this results in PHP Notice: Use of undefined constant NS_COMMENTSTREAMS - assumed 'NS_COMMENTSTREAMS' as described in [here](https://www.semantic-mediawiki.org/wiki/Thread:Help_talk:$smwgNamespacesWithSemanticLinks/Can%27t_activate_smw_on_custom_namespace) (even tough enableSemantics is clearly before this)
2. load the extension CommentStreams **before** SMW

### Steps to reproduce
1. Install CommentStreams in a SMW 3.1.x installation, invoke it after SMW, then no properties are being saved.
2. Move enabling CommentStremas in LocalSettings.php before calling SMW and properties will get saved,

@kghbln indicated, this could be related to #4119

kghbln:
A third alternative will be to add `define( 'NS_COMMENTSTREAMS', 844 );` prior to invoking any extension.

cicalese:
The namespaces are created and $wmwgNamespacesWIthSemanticLinks is set by CommentStreams in an extension registration callback [0].

In SMW 3.0, it did not matter whether CommentStreams was loaded with wfLoadExtension() before or after calling enableSemantics(). In SMW 3.1, CommentStreams must be loaded before enableSemantics() is called. The error only occurs in SMW 3.1 if CommentStreams is loaded after enableSemantics() is called.

[...]

PR #4558, opened 2020-02-15
Add `Settings::override` during `Setup::init`, refs 4505

mwjames:
This PR is made in reference to: #4505

This PR includes:
- [ ] CI build passed

Fixes #

### DIFF
--- a/src/Setup.php
+++ b/src/Setup.php
@@ -98,7 +98,7 @@ final class Setup {
 	 * @since 3.0
 	 */
 	public static function isEnabled() {
-		return defined( 'SMW_VERSION' ) && $GLOBALS['smwgSemanticsEnabled'];
+		return defined( 'SMW_VERSION' ) && defined( 'SMW_EXTENSION_LOADED' );
 	}
 
 	/**
@@ -146,10 +146,6 @@ final class Setup {
 		$this->initMessageCallbackHandler();
 		$this->addDefaultConfigurations( $vars, $rootDir );
 
-		if ( CompatibilityMode::extensionNotEnabled() ) {
-			CompatibilityMode::disableSemantics();
-		}
-
 		$this->registerJobClasses( $vars );
 		$this->registerPermissions( $vars );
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/includes/SettingsTest.php
```diff
@@ -99,6 +99,32 @@ public function testRegisterChangeListener() {
 		$instance->clearChangeListeners();
 	}
 
+	public function testOverride_ListingToOverride() {
+
+		$changeListener = $this->getMockBuilder( '\SMW\Listener\ChangeListener\ChangeListener' )
```
JeroenDeDauw: using `ChangeListener::class` helps static analysis tools find usage of `ChangeListener`

tests/phpunit/includes/SettingsTest.php
```diff
@@ -99,6 +99,32 @@ public function testRegisterChangeListener() {
...
+			->with( $this->equalTo( 'Foo' ) )
+			->will( $this->returnValue( true ) );
+
+		$changeListener->expects( $this->once() )
+			->method( 'setAttrs' )
+			->with( $this->equalTo( [ 'Foo' => 'Foobar' ] ) );
+
+		$changeListener->expects( $this->once() )
+			->method( 'trigger' )
+			->with( $this->equalTo( 'Foo' ) );
```
JeroenDeDauw: Using the mock API like this binds the settings test to ChangeListener in a way not detectable by static analysis. Doing a rename (using a tool) of one of these methods will cause failure here. Maybe that is fair enough inside of a test for ChangeListener or closely related code, but it is strange for a settings test.

mwjames: >  ChangeListener in a way not detectable by static analysis. 

I'm not worried about any static analysis or tools and for this particular case I'm interested in whether the overridden value is send to a change listener in a form I expect it to do.

JeroenDeDauw: My comment is not specific to this particular case either.

My concern is that if you write this kind of tests consistently, you end up with a codebase that is both brittle and hard to refactor.

I've seen much worse offenders than this one in SMW. Every time I see a PHPUnit mock returning another PHPUnit mock I die a bit inside. So. Much. Coupling.

JeroenDeDauw: > Maybe that is fair enough inside of a test for ChangeListener or closely related code, but it is strange for a settings test.

I just had a closer look at this test in my IDE since I did not understand how it worked. Now I see ChangeListener is invoked via `set`. So this is closely related code, making that particular concern not applicable to this test.

includes/Settings.php
```diff
@@ -218,6 +218,23 @@ public static function newFromGlobals() {
 		return self::$instance;
 	}
 
+	/**
+	 * @since 3.2
+	 *
+	 * @param array $vars
+	 */
+	public function override( array $vars ) {
```
JeroenDeDauw: Is this method suppose to be package public? In other words: is it meant to be used by extensions to SMW?

tests/phpunit/includes/SettingsTest.php
mwjames: > Every time I see a PHPUnit mock returning another PHPUnit mock I die a bit inside. So. Much. Coupling.

Yes, I could create a `SpyChangeListener` to only be concerned about the output instead of checking particular methods but then I would have to write another piece of code just to encapsulate the `ChangeListener`. If start doing that I would never finish any code because of encapsulation concerns.

includes/Settings.php
mwjames: > Is this method suppose to be package public? In other words: is it meant to be used by extensions to SMW?

No, it should only be used during `Setup:init`.

JeroenDeDauw: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4560

PR #4575, opened 2020-02-21
Don't call WebRequest::getLimitOffset in 1.35

DannyS712:
See https://phabricator.wikimedia.org/T245779

### DIFF
--- a/includes/specials/SpecialProperties.php
+++ b/includes/specials/SpecialProperties.php
@@ -69,9 +69,9 @@ class SpecialProperties extends SpecialPage {
 		$request = $this->getRequest();
 		if ( method_exists( $request, 'getLimitOffsetForUser' ) ) {
 			// MW 1.35+
-			return = $request->getLimitOffsetForUser( $this->getUser() );
+			return $request->getLimitOffsetForUser( $this->getUser() );
 		} else {
-			return = $request->getLimitOffset();
+			return $request->getLimitOffset();
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


includes/specials/SpecialProperties.php
```diff
@@ -66,7 +66,13 @@ protected function getGroupName() {
...
 
 	private function getLimitOffset() {
-		return $this->getRequest()->getLimitOffset();
+		$request = $this->getRequest();
+		if ( method_exists( $request, 'getLimitOffsetForUser' ) ) {
+			// MW 1.35+
+			return $request->getLimitOffsetForUser( $this->getUser() );
+		} else {
+			return $request->getLimitOffset();
+		}
```
JeroenDeDauw: You can write this without the else part:

```php
		if ( method_exists( $request, 'getLimitOffsetForUser' ) ) {
			// MW 1.35+
			return $request->getLimitOffsetForUser( $this->getUser() );
		}

		return $request->getLimitOffset();
```

DannyS712: True, but I was using the same code block that I was using elsewhere where it wasn't returning, but rather setting a variable, so I kept the else block

JeroenDeDauw: That is part of the reason why using a guard clause (the form in my variation) is nice.

JeroenDeDauw: https://www.entropywins.wtf/blog/2019/01/14/readable-functions-guard-clause/

kghbln:
@DannyS712 Thanks a lot for the pull!

PR #4580, opened 2020-02-22
Restore comments forcefully removed by "Increase type safety in DIProperty"

mwjames:
This PR is made in reference to: #4542

- The next time I see a PR that forcefully removes comments from an existing class I will show no mercy and __revert the PR__ without prio warning. I have been telling people several times that comments remain and are expected at least for public methods independent of whether some fancy PHP 7.1 type is used or not.

This PR includes:
- [ ] CI build passed

Fixes #

### DIFF
--- a/includes/dataitems/SMW_DI_Property.php
+++ b/includes/dataitems/SMW_DI_Property.php
@@ -100,26 +100,48 @@ class DIProperty extends SMWDataItem {
 		$this->m_inverse = $inverse;
 	}
 
-	public function getDIType(): int {
+	/**
+	 * @since 1.6
+	 *
+	 * @return int
+	 */
+	public function getDIType() : int {
 		return SMWDataItem::TYPE_PROPERTY;
 	}
 
-	public function getKey(): ?string {
+	/**
+	 * @since 1.6
+	 *
+	 * @return string|null
+	 */
+	public function getKey() : ?string {
 		return $this->m_key;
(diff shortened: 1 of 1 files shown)


akuckartz:
How are these comments used? Is that documented anywhere?

mwjames:
> How are these comments used? Is that documented anywhere?

Yes, by a developer like myself who wants to know when a public method was added (indicated by `@since`) and as good practice to other developers. If you maintain your own library, you'll can add or remove comments as you like but for SMW we have an established practice which isn't nullified by some suddenly appearing PHP 7.1 type hinting.

Aside from being diligent in documenting an interface, the information is used by/in https://doc.semantic-mediawiki.org/.

akuckartz:
Maybe the information that type comments are needed in addition to type hinting should be added to https://www.semantic-mediawiki.org/wiki/Coding_conventions

JeroenDeDauw:
> Aside from being diligent in documenting an interface, the information is used by/in https://doc.semantic-mediawiki.org/.

Can you show a case where information vanished from this site after it was moved from a doc to the code itself? (So I'm talking about types here, not @since.)

Btw the codebase is full of incorrect docs. They are not reliable. Every time I look at the code I find many issues. This file was particularly atrocious: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4573/files.

Issue #4146, opened 2019-07-13
MW 1.34+, "Move DataAfterContent outside of main content block" break Factbox display


mwjames:
### Setup and configuration

- SMW version: all
- MW version: MW 1.34+

### Issue

- https://github.com/wikimedia/mediawiki-skins-Vector/commit/7d46396dc1773b1584d7ce2c9ec7f042a5a2a7f5
- https://github.com/wikimedia/mediawiki-skins-MonoBook/commit/fa1deee55e2068ed18af7bbd01da1c54bb438eea

The above changes do break the display of the `Factbox` (which has been part of SMW since ~1.0) due to reliance of the `dataAfterContent` section generated by the `SkinAfterContent` hook.

@kghbln @hexmode This cannot be fixed in SMW and is clearly a BC break, this needs to be reverted.

PS: I find a distorting that we have to dicusss such breaks again and again and again, and I am __seriously tired__ of this!

## Before

![image](https://user-images.githubusercontent.com/1245473/61171840-9799c980-a56c-11e9-90d4-66cba2675d7a.png)

## After

![image](https://user-images.githubusercontent.com/1245473/61171839-949ed900-a56c-11e9-8164-148243918907.png)

kghbln:
This is indeed not nice. Can't the [BeforePageDisplay](Manual:Hooks/BeforePageDisplay) be used instead or is this a brute change? Anyways I have the feeling that many more extensions will be affected by these changes.

mwjames:
> Can't the BeforePageDisplay

No.

> Anyways I have the feeling that many more extensions will be affected by these changes.

I don't know about that but this is a typical example of a WMF staff granting +2 on a PR that has not been vetted against extensions that use the `SkinAfterContent` hook with the consequence that things are just broken.

I just happened to came across this by chance given that I used the master vector-skin where the Factbox suddenly appeared to be broken and I had to invested the skin extension rather than looking at chances in MW core.

hexmode:
I've but a comment on the [related bug](https://phabricator.wikimedia.org/T226199).

kghbln:
The related change actually broke a couple of other extensions provided by WMF, the creators and mergers however believe the did a great job doing so. I believe this is the way it is.

PR #4582, opened 2020-02-22
CSS (`smw-factbox-view`, smw.factbox.css), refs 4146

mwjames:
This PR is made in reference to: #4146

- This tries to mend the factbox display on the `.skin-vector` by using some extra CSS rules to make the factbox content to appear as part of the overall `bodyContent` and not as dismembered snippet of the content as shown in the example.

This PR includes:
- [x] CI build passed

## Example

MW 1.34+

image

### DIFF
--- a/res/smw/factbox/smw.factbox.css
+++ b/res/smw/factbox/smw.factbox.css
@@ -196,8 +196,51 @@
 }
 
 /**
+ * #4146
+ *
+ * `#mw-data-after-content` defines when the content is outside of the
+ * `#bodyContent`
+ *
+ * `#bodyContent #mw-data-after-content` negates previous CSS when it is part of
+ * the #bodyContent (being inside)
+ */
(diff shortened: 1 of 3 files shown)


mwjames:
The doesn't fix #4146 , it only adds some CSS rules to make it appear as if the "old" behaviour is retained but the issue remains unresolved.

kghbln:
@mwjames Can this be back-ported? Not sure how this plays with MW 1.31.x to 1.33.x

mwjames:
> @mwjames Can this be back-ported? Not sure how this plays with MW 1.31.x to 1.33.x

I tried to account for this by adding some extra rules [0] but for now we should keep it on master only.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/res/smw/factbox/smw.factbox.css#L198-L227

kghbln:
> I tried to account for this by adding some extra rules [0] but for now we should keep it on master only.

Thanks for the info. Let's keep it as is.

PR #4586, opened 2020-02-22
Add max-width, theme parameter to #info

mwjames:
This PR is made in reference to: https://www.semantic-mediawiki.org/wiki/Thread:Help_talk:Special_characters_in_tooltips/Adding_a_tooltip

- This doesn't answer the question of "what if I would like to add a custom tooltip (i.e. hacking SMW)" (which I don't recommend otherwise any modifications to the smw.tippy.js will be made invalid by the next upgrade) anyway:
  - The `Highlighter` class creates the HTML code in PHP while
  - `smw.tippy.js` is the corresponding tandem on the JS side to create and invoke the `tippy` instance (see atomiks/tippyjs) 
- This PR adds two new parameters to #info:
  - max-width: to define the maximum width with a tooltip, allowing it to be wider than the default to display more information at once
  - theme (square-border, square-border-light) to use a slightly different theme for the tooltip appearance

This PR includes:
- [x] CI build passed

## Example
```
{{#info: message=Integer eu ... |max-width=350 |theme=square-border }}
```

image

image

### DIFF
--- a/includes/Highlighter.php
+++ b/includes/Highlighter.php
@@ -241,10 +241,18 @@ class Highlighter {
 		}
 
 		$language = is_string( $this->language ) ? $this->language : Message::USER_LANGUAGE;
-		$style = [];
+		$attribs = [];
 
 		if ( isset( $this->options['style'] ) ) {
-			$style = [ 'style' => $this->options['style'] ];
+			$attribs['style'] = $this->options['style'];
+		}
+
+		if ( isset( $this->options['maxwidth'] ) && $this->options['maxwidth'] !== '' ) {
+			$attribs['data-maxwidth'] = $this->options['maxwidth'];
+		}
+
+		if ( isset( $this->options['themeclass'] ) && $this->options['themeclass'] !== '' ) {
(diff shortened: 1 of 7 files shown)


mwjames:
@kghbln The added parameters should be enough to cover most use cases that rely on #info while the height remains as-is otherwise this becomes anything but a tooltip.

PS: As an official note, hacking `smw.tippy.js` is not recommended!

kghbln:
Really cool. Thank you!

> otherwise this becomes anything but a tooltip

What about introducing `#supobjects` pages that show instead of the pages they are embedded on. :D

mwjames:
> What about introducing #supobjects pages that show instead of the pages they are embedded on. :D

Not sure I can follow!

kghbln:
> Not sure I can follow!

This was a joke. `#supobject` in contrast to `#subobject`.

mwjames:
> This was a joke. #supobject in contrast to #subobject.

Right. Its quite late for me hence I'm a bit slow.

kghbln:
[Life example](https://sandbox.semantic-mediawiki.org/wiki/Issue/4586) and [documentation]().

PR #4048, opened 2019-05-25
Add new `smwtable-clean` table CSS

mwjames:
This PR is made in reference to: #

This PR addresses or contains:

This PR includes:
- [ ] Tests (unit/integration)
- [ ] CI build passed

### Example

default (`sortable wikitable smwtable`)

image

`sortable smwtable-clean`

image

PR #4587, opened 2020-02-22
Introduce class "smwbroadtable-clean"

kghbln:
This PR is made in reference to: #4048 

- Fixes the `smwtable-clean` class producing a broadtable
- Introduces the `smwbroadtable-clean` class to provide the same styling as the former `smwtable-clean` class

This PR includes:
- [ ] CI build passed

Fixes #

### DIFF
--- a/res/smw/ext.smw.css
+++ b/res/smw/ext.smw.css
@@ -205,6 +205,13 @@ span.smwtlcomment {
 }
 
 .smwtable-clean {
+	max-width: 100%;
+	margin-bottom: 20px;
+	background-color: transparent;
+	border-spacing: 0px;
+}
+
+.smwbroadtable-clean {
 	width: 100%;
 	max-width: 100%;
 	margin-bottom: 20px;
@@ -212,19 +219,25 @@ span.smwtlcomment {
 	border-spacing: 0px;
 }
 
-.smwtable-clean tr {
-border-top: 1px solid #dddddd;
+.smwtable-clean tr,
+.smwbroadtable-clean tr {
+	border-top: 1px solid #dddddd;
 }
 
-.smwtable-clean th {
+.smwtable-clean th,
+.smwbroadtable-clean th {
 	text-align: left;
 }
 
-.smwtable-clean td, .smwtable-clean th {
+.smwtable-clean td,
+.smwbroadtable-clean td,
+.smwtable-clean th,
+.smwbroadtable-clean th {
 	padding: 0;
 }
 
-.smwtable-clean tr > th {
+.smwtable-clean tr > th,
+.smwbroadtable-clean tr > th {
 	padding: 8px !important;
 	line-height: 1.42857143;
 	vertical-align: top;
@@ -235,7 +248,8 @@ border-top: 1px solid #dddddd;
 	vertical-align: middle;
 }
 
-.smwtable-clean tr > td {
+.smwtable-clean tr > td,
+.smwbroadtable-clean tr > td {
 	padding: 8px !important;
 	line-height: 1.42857143;
 	vertical-align: top;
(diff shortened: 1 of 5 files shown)


kghbln:
@mwjames Not sure if this is the correct approach, however it tries to fix unexpected behaviour for the table format. Also not sure how to get special page "Ask" to use this class.

mwjames:
```
.smwtable-clean {
	max-width: 100%;
	margin-bottom: 20px;
	background-color: transparent;
	border-spacing: 0px;
}

.smwbroadtable-clean {
	width: 100%;
```
Is my assumption that the only difference between `smwtable-clean` (now being `max-width: 100%;`) and `smwbroadtable-clean` ( being `width: 100%;`) is the `max-width`?

How does this look different?

mwjames:
>  it tries to fix unexpected behaviour for the table format.

Given this comment, you want that `smwtable-clean` is not a broadtable (== 100% width) by default?

kghbln:
> Given this comment, you want that smwtable-clean is not a broadtable (== 100% width) by default?

Indeed, and this is achieved by the following CSS on most of the wikis I work with:
```css
.smwtable-clean {
	width: auto;
}
```
So I figured not defining width at all would achieve the same.

kghbln:
> Is my assumption that the only difference between smwtable-clean (now being max-width: 100%;) and smwbroadtable-clean ( being width: 100%;) is the max-width?

This is correct.

> How does this look different?

See sandbox.

mwjames:
> So I figured not defining width at all would achieve the same.

Just to confirm, adding `width: auto;` to `.smwtable-clean` will make a query with `|format=table` non-broadtable while using `|format=broadtable` makes it a table with `100%` width, right?

kghbln:
> Just to confirm, adding width: auto; to .smwtable-clean will make a query with |format=table non-broadtable 

yes

> while using |format=broadtable makes it a table with 100% width, right?

here you would have to add `|class=smwbroadtable-clean`. This is why I wrote that I am not sure if this appoarch is correct. This is just an helpless effort to fix the table format I guess.

kghbln:
> here you would have to add |class=smwbroadtable-clean.

This is basically what you also have to do for the table format. So no difference. Open is the question on how to do this for special page "Ask".

mwjames:
> here you would have to add |class=smwbroadtable-clean. This is why I wrote that I am not sure if this appoarch is correct. This is just an helpless effort to fix the table format I guess.

Using your changes from the sandbox should be sufficient without having to introduce a new CSS class.

`|format=table`

https://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial:Ask&x=-5B-5BHas-20number%3A%3A%2B-5D-5D%2F-3FHas-20number&format=table&limit=250&link=all&headers=show&searchlabel=%26hellip%3B%20autres%20r%C3%A9sultats&class=smwtable-clean&sort=&order=asc&offset=0&mainlabel=

`|format=broadtable`

https://sandbox.semantic-mediawiki.org/w/index.php?title=Sp%C3%A9cial:Ask&x=-5B-5BHas-20number%3A%3A%2B-5D-5D%2F-3FHas-20number&format=broadtable&limit=250&link=all&headers=show&searchlabel=%26hellip%3B%20autres%20r%C3%A9sultats&class=smwtable-clean&sort=&order=asc&offset=0&mainlabel=

kghbln:
Ah indeed, so we just need to make the special pages to additionally use the broadtable class.

mwjames:
> Ah indeed, so we just need to make the special pages to additionally use the broadtable class.

Why? If no format is given then `Special:Ask` (same goes for `SpecialConstraintErrorList` etc.) will automatically assume `format=broadtable`.

mwjames:
As I said, your changes should be sufficient for both use cases without having to make any other changes or introduce a new CSS class.

```
.smwtable-clean {
	width: auto;
	max-width: 100%;
	margin-bottom: 20px;
	background-color: transparent;
	border-spacing: 0px;
}
```

kghbln:
> Why? If no format is given then Special:Ask (same goes for SpecialConstraintErrorList etc.) will automatically assume format=broadtable.

Ah, I figured the default is "table". Admittedly I do not use special page "Ask" that often if at all.

kghbln:
Now at https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4589

PR #4589, opened 2020-02-22
Fix "smwtable-clean" class being broad by default

kghbln:
This PR is made in reference to: #4048 

- Fix "smwtable-clean" class being broad by default

This PR includes:
- [ ] CI build passed

Fixes #

### DIFF
--- a/res/smw/ext.smw.css
+++ b/res/smw/ext.smw.css
@@ -205,7 +205,7 @@ span.smwtlcomment {
 }
 
 .smwtable-clean {
-	width: 100%;
+	width: auto;
 	max-width: 100%;
 	margin-bottom: 20px;
 	background-color: transparent;


kghbln:
@mwjames Sometimes solutions are easier than one may expect. Thanks for your guidance.

kghbln:
@mwjames While this works for the special pages including Ask it does not work for format [broadtable](https://sandbox.semantic-mediawiki.org/wiki/Broadtable). This is basically why I wanted to introduce a "smwbroadtable-clean" class.

mwjames:
> This is basically why I wanted to introduce a "smwbroadtable-clean" class. However it is still much better this way around than it was before.

This should work with one CSS class but it requires I slight addition (the `broadtable` is added by `|format=broadtable`):

```
@@ -250,10 +250,14 @@ border-top: 1px solid #dddddd;
 
 .smwtable-clean tr > th.headerSort {
 	padding-right: 21px !important;
 }
 
+.smwtable-clean.broadtable {
+	width: 100%;
+}
```

mwjames:
PS: When doing CSS or documentation changes, use `[skip ci]` as we don't check those via the CI framework.

kghbln:
> This should work with one CSS class but it requires I slight addition (the broadtable is added by |format=broadtable):

It took me a while to figure this out, however came up with the same result. :)

kghbln:
@mwjames Paradise regained. Thanks a lot!

PR #4592, opened 2020-02-23
Add predefined properties group, refs 3749

mwjames:
This PR is made in reference to: #3749

- Adds an import for a group definition of some SMW core predefined properties and as before, if the wiki defines an individual group for a property then this has priority before the pre-deployed group assignment
- Slightly changes the `PROPERTY_GROUP_SCHEMA` schema to make the content more readable

This PR includes:
- [x] CI build passed


kghbln:
[Documented.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AProperty_group&type=revision&diff=74602&oldid=68730)

Issue #4611, opened 2020-02-29
Faceted search


mwjames:
The main task and objective of Semantic MediaWiki is to extend the MediaWiki platform with an ability to store structured information and make them retrievable either as part of a query result or as data for external federation (RDF, JSON, CSV etc,).

Another objective is to provide interfaces to help make stored data queryable and visualisable which is achieved by supporting different technical solutions including embedding results into a wikipage using the `#ask` parser function, building queries with the help of the advanced query builder (also known as `Special:Ask`), or build a bridge to MediaWiki's `Special:Search` using the  "Extended Search" (aka. `SMWSearch`) integration.

## Faceted search

To extend those interfaces and because Semantic MediaWiki is currently missing a vital navigational piece that allows users to easily navigate existing data with a limited preset of information, I have been working for the last week on a prototype to make a "Faceted search" to become a reality with Semantic MediaWiki (the result being shown in the screen dump).

![faceted-search (2)](https://user-images.githubusercontent.com/1245473/75606953-35d1a400-5aea-11ea-891c-b7de1298f7da.gif)

The "Faceted search" requires some technical tweaks but it mostly works out-of-the box by relying on the prefetch mode introduced in 3.1 for query results that minimizes the amount of SQL queries required to find data for query and selected printrequests.

### Motivation

Providing such interface will give users a tool at hand to narrow down possible matches using a faceted search that should not only help advanced users but also be an attractive alternative for novice users that find the `Special:Ask` a bit overwhelming while the `Special:SearchByProperty` is only usable when one knows the property and/or value to match and comes out short when trying to browse and match.

Yes, there is something called `SemanticDrilldown` but this extension is hardly maintained and violates the integration with SMW in part because it uses direct access to SMW tables which is discouraged to any extension that uses SMW as backend. Using raw access to tables makes an assumption about a specific data model that may or may not be supported and in worst cases creates a strain on the wiki due to use of unoptimized SQL queries. Also, due to a possible `smw_terms` migration in future, accessing tables directly will no longer work as it used to be and most likely make  `SemanticDrilldown` incompatible with any future SMW release.

### Notes

I took inspiration for the visual appearance of the search from the http://projectblacklight.org/ project to help devise an interface that has been proven itself in different areas such as [0, 1, 2].

[...]

PR #4613, opened 2020-02-29
Add `smw_countmap` field, refs 4611

mwjames:
This PR is made in reference to: #4611

- Part of the technical tweaks for #4611
- Adds `smw_countmap` as extra field to the `smw_object_aux` (#4227) table which stores a map of counts of properties and categories used per entity
- It allows to prefetch that information for a list of entities without having to lookup each single property table to match the entity which means that for a list of 500 result matches we only require one SQL query to know which properties (and categories) are used on all 500 entities (including they amount of values assigned to each property)
-  Information is stored in a simple blob field type since we don't want query the information in the field (that information is stored separately), we only want to fetch the entire content at once to build an aggregation map
- Changes `smwgUpgradeKey` to `smw:2020-02-29`
- `updateEntityCountMap.php` script is provided to update the field information 

This PR includes:
- [x] CI build passed

## Example

A count map per entity will look similar to something like the following example.

```json
{
	"Has_Url": 1,
	"Has_annotation_uri": 1,
	"Has_area": 1,
	"Has_boolean": 1,
	"Has_date": 1,
	"Has_email": 1,
	"Has_number": 1,
	"Has_page": 3,
	"Has_record": 1,
	"Has_status": 1,
	"Has_telephone_number": 1,
	"Has_temperature": 1,
	"Has_wattage": 3,
	"_ERRC": 1,
	"_INST": {
		"Sample_pages": 1,
		"Lorem_ipsum": 1
	},
	"_MDAT": 1,
	"_SKEY": 1,
	"_SOBJ": 1
}
```


mwjames:
Of the first occasion where the field was added to the table, an incomplete task is registered reminding an administrator to do something.

image
image

kghbln:
Doku:
* [maintenance script "updateEntityCountMap.php"](https://www.semantic-mediawiki.org/wiki/Help:Maintenance_script_updateEntityCountMap.php)
* [field "smw_countmap"](https://www.semantic-mediawiki.org/wiki/Field:smw_countmap)

kghbln:
@mwjames I documented the scrip similarily to the "populateHashField.php" when it comes to them manipulating the ".smw.json" file. However if I look at sandbox's file is see neither of the expected entries. Did the way the completion is being checked change in the meantime?

mwjames:
> Did the way the completion is being checked change in the meantime?

The incomplete task/name are tied to a message key  [0,1] which avoids that the `SetupFile` [2] has to have any specific knowledge about any task. The leftover of line 300 to 315 should be removed from `SetupFile` at once point. Once a task is completed it is removed which signals to the `SetupFile` that there are no incomplete tasks.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/maintenance/populateHashField.php#L85-L92
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/maintenance/updateEntityCountMap.php#L30-L33
[2] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/SetupFile.php#L291-L328

Eventually

kghbln:
Thanks for the info. I adapted the docu.

PR #1896, opened 2016-10-07
Move internationalization from languages to i18n

kghbln:
This PR is made in reference to: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1848#issuecomment-249402052

This PR addresses or contains:
- Moves all internationalization to the i18n directory
- Removes internationalization via php-files except for alias and magic

This PR includes:
- [ ] Tests (unit/integration)
- [x] CI build passed

Follow-up to https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1848

PR #4618, opened 2020-03-01
Remove `smwgContLang`

mwjames:
This PR is made in reference to: #1848, #1896, #1941

- Removes `smwgContLang` which was never a configuration setting but got used as is, we have removed any reference of it in SMW with (#1896, #1941) and now it is time to remove remaining references from the code base
- If for some reason some external user made use of it then the `smwfContLang()` function is provided as convenience 

This PR includes:
- [x] CI build passed

### DIFF
--- a/includes/SMW_PageSchemas.php
+++ b/includes/SMW_PageSchemas.php
@@ -199,7 +199,8 @@ class SMWPageSchemas extends PSExtensionHandler {
 	 * semantic property within the Page Schemas 'editschema' page.
 	 */
 	public static function getFieldEditingHTML( $psTemplateField ) {
-		global $smwgContLang;
+
+		$smwgContLang = smwfContLang();
 
 		$prop_array = [];
 		$hasExistingValues = false;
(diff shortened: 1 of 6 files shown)


mwjames:
@kghbln Could be that PF is relying on that but honestly that isn't any of my concern since SMW isn't using it since at least 2.5 and it should have never been a $GLOBALS in the first place. Anyway, out of compliance I added `smwfContLang()` which they can rely upon if the necessity arises.

With this PR all settings parameters in the `Settings` object are stateless and can now be easily serialized.

kghbln:
How can I find out? When I do a [code search](https://codesearch.wmflabs.org/extensions/?q=smwfContLang&i=nope&files=&repos=) I get no results for any extensions but SMW. In the end I would like to create a somewhat decent error report.

mwjames:
> How can I find out? When I do a code search I get n

You searched for the wrong term, using `smwgContLang` instead of `smwfContLang` because `smwfContLang()` is a function I introduced with this PR.

https://codesearch.wmflabs.org/extensions/?q=smwgContLang&i=nope&files=&repos=

kghbln:
> You searched for the wrong term, using smwgContLang instead of smwfContLang because smwfContLang() is a function I introduced with this PR.

Thanks for the info. Three extensions affected and information provided:
- [Page Forms / T246579](https://phabricator.wikimedia.org/T246579)
- [Semantic Drilldown / T246581](https://phabricator.wikimedia.org/T246581)
- [Semantic Internal Objects / T246582](https://phabricator.wikimedia.org/T246582)

mwjames:
@kghbln FYI https://phabricator.wikimedia.org/T249172 is the consequence of this change because `$smwgContLang` no longer exists as explained above.

		$datatypeLabels = $smwgContLang->getDatatypeLabels();

kghbln:
Thanks for the note. I cross-referenced the two issues.

Issue #4597, opened 2020-02-25
Let SESP to be grouped in Special:Browse via smw/schema:Group:Predefined properties


jaideraf:
### Setup and configuration

- SMW version: 3.2.0-alpha
- MW version: 1.33.2
- PHP version: 7.3.14
- MySQL version: 5.6.46-86.2
- SESP version: 2.1.0

### Enhancement
When aplying "`_CUSER`" (key for `Page creator` special property from SESP - "Criador da página" in Portuguese) in "`adminstrative_group`" section from `smw/schema:Group:Predefined properties` the property does not "move" to `Adminstrative properties` group in Special:Browse.

### Steps to reproduce
See: 
* https://novo.wikincat.org/wiki/Especial:Navegar/:B000001?uselang=en
* https://novo.wikincat.org/w/index.php?title=smw/schema:Group:Predefined_properties&action=edit

mwjames:
> When aplying "_CUSER" (key for Page creator special property from SESP

The actual (internal) key is `___CUSER` as see in [0] which should be used here.

[0] https://github.com/SemanticMediaWiki/SemanticExtraSpecialProperties/blob/master/data/definitions.json#L32-L37

kghbln:
> The actual (internal) key is ___CUSER as seen in [0] which should be used here.

I knew this was already possible but I would not have come up with the internal id. Thanks for you help. Now [documented.](https://www.semantic-mediawiki.org/w/index.php?title=Help%3ASchema%2FType%2FPROPERTY_GROUP_SCHEMA&type=revision&diff=74635&oldid=72777)

jaideraf:
Indeed. Thanks @mwjames for your help, and thanks @kghbln for the ever astonishing SMW docs.
(I was on vacation, that is why did not respond earlier). :heart:

PR #4642, opened 2020-03-08
Add edit protection for imported content, refs 4595

mwjames:
This PR is made in reference to: #4595

- #4595 added support for defining a dedicated import performer (user which cannot be hijacked because it is classified as system user) and this PR will introduce an edit protection for SMW_NS_SCHEMA (for now) on content where the `import_performer` was defined to ensure no other users besides the importer can alter the content (users can delete the page but not edit).
- In general, the importer is only allowed to replace content of imported pages if there was no edit to the content by any other users which is the reason why a mechanisms is needed to lock content so that it can be replaced without fear that some other user modified the content and may loose that information on the next import.
- Adds `smwgImportPerformers` to list those users classified as import performers which should be used in `import_performer` 

This PR includes:
- [x] CI build passed

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -93,6 +93,22 @@ return [
 	##
 
 	###
+	# List of users for import activities
+	#
+	# Users listed here are to be used exclusively for the import task and can
+	# depending on the specific protection level lock certain content from being
+	# altered by any other user.
+	#
+	# The protection is only enabled when a specific import content has defined
+	# the `import_performer` with a listed user.
+	#
+	# @since 3.2
+	# @default []
+	##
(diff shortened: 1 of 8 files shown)


mwjames:
@kghbln A bit of background here; The `FacetedSearch` will introduce a new schema type and requires a default profile which gets deployed (imported) with the software upon its setup. The default profile is a prerequisite but once a user modifies that profile, the importer is unable to replace the content on any other occasion which is why such imported content where `import_performer` has been defined becomes locked. The reason is simple, I have seen users (#4597) to modify imported content whether we tell them not do it or not because it is easier to modify something instead of creating it from scratch.

image

PR #4644, opened 2020-03-14
Check `justinrainbow/json-schema` is available

mwjames:
This PR is made in reference to: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3749#issuecomment-599029003

- Schema validation happens to ensure input data conforms with the requirements set forth so checking that the validator is available is paramount otherwise we produce schemata with incorrect data.

This PR includes:
- [x] CI build passed

## Example

image


kghbln:
This however does not solve the issue that the lib is not obtained via composer in the first place.

mwjames:
> This however does not solve the issue that the lib is not obtained via composer in the first place.

This is correct but it prevents anyone from trying to create or alter schemata without any validation. The whole idea is the "structured" input follows a particular form and only contain data that is permitted through the validation schema.

Adding this as a requirement to the SMW `composer.json` would be the next step.

kghbln:
> Adding this as a requirement to the SMW composer.json would be the next step.

Guess/hope that this is all what it is needed: #4648

PR #4655, opened 2020-03-21
Add namedargs alias to list result format

JeroenDeDauw:
- "namedargs" alias for "named args" to the list format including a test

This PR includes:
- [ ] CI build passed

### DIFF
--- a/src/Query/ResultPrinters/ListResultPrinter.php
+++ b/src/Query/ResultPrinters/ListResultPrinter.php
@@ -163,6 +163,7 @@ class ListResultPrinter extends ResultPrinter {
 				'type' => 'boolean',
 				'message' => 'smw-paramdesc-named_args',
 				'default' => false,
+				'aliases' => [ 'namedargs' ]
 			],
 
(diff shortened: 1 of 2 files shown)


kghbln:
@JeroenDeDauw I added test cases for the alias. Apart from list also "category" and "templatefile" provide "named args". These should be getting the alias, too.

mwjames:
> . Apart from "list" also "category" and "templatefile" provide "named args". These should be getting the alias, too.

This needs to happen otherwise `namedargs` becomes an oddity among result printers. Consistency is the key here.

kghbln:
> This needs to happen otherwise namedargs becomes an oddity among result printers. Consistency is the key here.

Exactly. This is why I mentioned it. In the meantime I created the respective pull requests.

kghbln:
The existing docu about named arguments was an utter mess. I really wonder why nobody justfully complained. Anyhow I created [same docu.](https://www.semantic-mediawiki.org/wiki/Help:Plainlist_format/parameter_named_args)

PR #4656, opened 2020-03-21
Provide method to check record type

mwjames:

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/DataTypeRegistry.php
+++ b/src/DataTypeRegistry.php
@@ -194,6 +194,17 @@ class DataTypeRegistry {
 	}
 
 	/**
+	 * @since 3.2
+	 *
+	 * @param string $typeId
+	 *
+	 * @return boolean
+	 */
+	public function isRecordType( string $typeId ) : bool {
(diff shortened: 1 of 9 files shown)


src/DataTypeRegistry.php
```diff
@@ -193,6 +193,17 @@ public function isRegistered( $typeId ) {
...
 	}
 
+	/**
+	 * @since 3.2
+	 *
+	 * @param string $typeId
+	 *
+	 * @return boolean
+	 */
+	public function isRecordType( string $typeId ) : bool {
```
JeroenDeDauw: <3

src/DataTypeRegistry.php
```diff
@@ -193,6 +193,17 @@ public function isRegistered( $typeId ) {
...
+		return strpos( $typeId, '_rec' ) !== false;
+	}
```
JeroenDeDauw: Hmm... ideally there would be a dedicated type (class/interface) for the data types. Just a semi related idea, not something for this PR

PR #4663, opened 2020-03-22
Add `SpecialFacetedSearch`, refs 4611

mwjames:
This PR is made in reference to: #4611

- General introduction [0]
- `FACETEDSEARCH_PROFILE_SCHEMA` description [1]
- `range_group` and `range_control` specific description [2]

This PR includes:
- [x] CI build passed

Fixes: #4611

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/faceted-search/src/MediaWiki/Specials/FacetedSearch
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/faceted-search/src/Schema/docs/facetedsearch.profile.md
[2]  https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/faceted-search/src/Schema/docs/property.profile.md


mwjames:
The default profile (deployed with SMW) will provide a minimum set of settings to make the faceted search to appear light without any fancy sliders, hierarchy trees, checkboxes, or conditional fields.

image

## Custom profile

A custom profile can alter the appearance by adding a `theme` that controls a CSS set (stored via Common.css etc.) providing a possibility to make the sidebar wider (or generally occupy 100% width) or to modify the color scheme as demonstrated below.

image

There are couple of different value filters available to make the filtering process more adaptable to the context of values so for example, a range filter for numeric values:

image

A range group filter with range groups being maintained on the selected property (as those ranges are context driven by the related property hence they need to be assigned to the property via a `PROPERTY_PROFILE_SCHEMA`)

image

Support for a `hierarchy_tree` to be shown as part of the category and property filter card (as demonstrated above) or instead of the default `list_filter` select a `checkbox_filter`.

image

mwjames:
There is a possibility to enable other result printers such as `map` (as demonstrated below) but that isn't part of the initial deployment and it should be checked whether it makes sense to provide other printers based on use case and merit.

I only have spent 10 minutes on this but https://github.com/JeroenDeDauw/Maps/issues/609 and https://github.com/JeroenDeDauw/Maps/issues/608 are a related outcome of this experiment.

image

akuckartz:
Is the missing documentation the only reason why this is not yet merged?

gesinn-it-gea:
@JeroenDeDauw @kghbln @krabina @akuckartz (and others): shouldn't we make a joint effort to finish the incredible work that @mwjames has done in this PR, #4611 and related commits to get Faceted Search into 4.0.0?

kghbln:
Personally I'd rather get 4.0.0 to the public to support 1.36 and perhaps 1.37 sooner than later. I am afraid that making this faceted search happen will add a delay to the release. I have nothing against doing a 4.1.0 release soon after if someone finishes the coolness of this PR. Anyhow the developers have to do the work.

krabina:
I agree that this PR shoud be finished! Personally, I don't care so much about in between LTS releases. From a dev point of view I can image that it might be easier to make this work in 1.35 and then care for this (and other issues) in 1.36 and 1.37. Otherwise we might keep this lying around way longer, making it harder to get it to work in a later release.

But I agree with Karsten, the devs who will take this on will have the decision.

JeroenDeDauw:
@gesinn-it-gea be my guest. I do not have sufficient incentives to look into this feature and do not expect that to change.

krabina:
@gesinn-it-gea what must be done to accept/test this? How can we help?

PR #4668, opened 2020-03-26
[CI] Fix warnings and notices

kghbln:

- root: deprecated key sudo (The key `sudo` has no effect anymore.) 
- root: key matrix is an alias for jobs, using jobs
- root: missing os, using the default linux

This PR includes:
- [x] CI build passed


kghbln:
Obviously this worked.

PR #4675, opened 2020-03-28
Find and match redirect source for `sequence_map`, refs 4669, 4227

### LINKED ISSUE #4669: sequence_map does not work with redirects
### Setup and configuration

- SMW version: 3.1.0
- MW version: 1.33.0
- PHP version: 7.3.14-1~deb10u1 (apache2handler)
- DB system (MySQL, Blazegraph, etc.) and version: 10.1.44-MariaDB-0+deb9u1

### Issue
The [sequence map feature](https://www.semantic-mediawiki.org/wiki/Help:Sequence_map) introduced in #4227 does not preserve the value order if there are redirects (the value of a property with datatype page redirects to another page).

### Steps to reproduce

1. Create `SMW/Schema:Seqmap` with:

     ```
	{
	    "type": "PROPERTY_PROFILE_SCHEMA",
	    "profile": {
	        "sequence_map": true
	    },
	    "tags": [
	        "property profile"
	    ]
	}
    ```

2. Create `Property:name` with `[[profile schema::seqmap]]`.
3. Run `rebuildData.php`.
4. Create `Foo` with `#redirect [[baz]]`.
5. Create `Test` with `[[name::foo]], [[name::bar]]`.
6. `{{#show: Test |?name}}` yields `Bar, Baz` instead of `Baz, Bar`

mwjames:
This PR is made in reference to: #4669, #4227

- It changes `'smwgUpgradeKey' => 'smw:2020-03-28',` because we need an additional index on the redirect table to make the read more efficient

This PR includes:
- [x] CI build passed

Fixes #4669

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -71,7 +71,7 @@ return [
 	#
 	# @since 3.0
 	##
-	'smwgUpgradeKey' => 'smw:2020-02-29',
+	'smwgUpgradeKey' => 'smw:2020-03-28',
 	##
 
 	###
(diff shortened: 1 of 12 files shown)


kghbln:
@mwjames I'd like to merge this for me to immediately do the script run for sandbox.

kghbln:
Thanks! Just had a look at the script updating the wiki. It should always run update.php every time thus preventing downtime.

PR #4681, opened 2020-04-03
Remove localized extension titles 

kghbln:

- Assure that the fallback in en is always used.

This PR includes:
- [ ] CI build passed

Fixes #


kghbln:
Because Semantic MediaWiki is a _Eigenname_. It should not have been translated in the first place.

akuckartz:
It is a descriptive Eigenname. The translations can still be useful for some communities. Can "smw-title" be renamed to "smw-translated-title" or something similar so that the translations do not get lost?

kghbln:
The translated strings are still available and preserved at translatewiki.net they however will not be synced into the repo. I find it highly irritating to have a different name on Special:Version.

Nikerabbit:
Surely transliterations should be allowed?

PR #4684, opened 2020-04-04
Add `ConfigPreloader`

mwjames:

- `loadDefaultConfigFrom` allows to load some pre-deployed settings that are not used by default and bound to context and can directly be invoked after `enableSemantics`
  - `developer.php` mostly relevant to develop with SMW
  - `media.php` common media specific settings
  - `elastic-fileingest.php`
- `loadConfigFrom` to load a local bound set of configurations 

This PR includes:
- [x] CI build passed

## Example

```
enableSemantics( 'example.org' )->loadDefaultConfigFrom( 'developer.php', 'media.php' );
```
```
enableSemantics( 'example.org' )->loadConfigFrom( __DIR__ . '/locationX/foo.php' );
```

### DIFF
--- a/data/config/developer.php
+++ b/data/config/developer.php
@@ -1,0 +1,62 @@
+<?php
+
+/**
+ * Default settings mostly used in connection with a development environment and
+ * it not meant to be used in production.
+ *
+ * @since 3.2
+ */
+
+error_reporting(E_STRICT|E_ALL);
+ini_set("display_errors", 1);
+
+return [
+
+	/**
+	 *  MediaWiki specific settings to be used within a development environment
+	 */
+	'wgShowExceptionDetails' => true,
+	'wgDevelopmentWarnings' => true,
+	'wgShowSQLErrors' => true,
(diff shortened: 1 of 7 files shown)


mwjames:
@kghbln In most cases we will not change existing default settings but you can create a "profile" that includes a different set of values for settings and even include modifications to things like `$wgDefaultUserOptions['name-of-the-user-preference'] = 1;` so that whatever user scenario (or due to some some commercial interests) requires a different set of defaults can make them available as profile without interfering with the default.

Of course, changing things after `enableSemantics` remains unaltered (meaning is possible) but loading those "profile" can provide means to avoid copying the same settings over and over again.

mwjames:
https://www.semantic-mediawiki.org/wiki/Config_preloading

kghbln:
Also referenced on [enableSemantics](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AEnableSemantics&type=revision&diff=76636&oldid=73294)

Issue #3532, opened 2018-10-26
Date interpretation fails for  "2018-10-11 18:23:59 +0200" 


WolfgangFahl:
## Setup and configuration
see 
- https://sandbox.semantic-mediawiki.org/wiki/Issue3532
- https://www.semantic-mediawiki.org/wiki/User:WolfgangFahl/Workdocumentation_2018-10-26

### Issue
see 
- https://sandbox.semantic-mediawiki.org/wiki/Issue3532
- https://www.semantic-mediawiki.org/wiki/User:WolfgangFahl/Workdocumentation_2018-10-26
![grafik](https://user-images.githubusercontent.com/1336221/47567359-9e28d800-d92e-11e8-9c0f-8593ea744d92.png)

_Unable to interpret the "2018-10-11 18:23:59 +0200" input value as valid date or time component with "Part of the date is out of bounds." being reported_

### Steps to reproduce

```
[[Has date::2018-10-11 18:23:59 +0200]]

{{#ask: [[{{FULLPAGENAME}}]]
|?Has_date
}}
```
similar iso dates do not have this issue e.g.
2018-10-11 18:03:20 +0200
works.

kghbln:
This is indeed a known limitation as [documented](https://www.semantic-mediawiki.org/wiki/Help:Type_Date#Input_examples). Please use a formatter like `ISO` or `F` to get the result in shape.

WolfgangFahl:
I do not understand why this is closed? Is this not a candidate for an enhancement?

kghbln:
> Is this not a candidate for an enhancement?

No.

WolfgangFahl:
We are talking input not output. Your comment is not a work around for the problem ...

PR #4692, opened 2020-04-05
Recognize date with UTC offset, refs 3532

mwjames:
This PR is made in reference to: #3532

- While importing some data I came across the `2002-11-01T00:00:00.000-0800` date format which requires special treatment due to the some UTC offset (see http://www.faqs.org/rfcs/rfc2822.html)
- PR also solves the issue of #3532 with a format like `2018-10-11 18:23:59 +0200`

This PR includes:
- [x] CI build passed


kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help%3AType_Date&type=revision&diff=75494&oldid=73580)

PR #4723, opened 2020-04-13
Add Special:MyLanguage for better user experience

shirayuki:
- Add Special:MyLanguage to MediaWiki.org URLs


kghbln:
@shirayuki Thanks a lot!

PR #4725, opened 2020-04-16
Avoid using deprecated $wgParserConf configuration variable

cscott:
Modern code should use the Parser or ParserFactory service.
See I787f22ea9bf59a049b13631ba6974866a1300988.

### DIFF
--- a/includes/ContentParser.php
+++ b/includes/ContentParser.php
@@ -7,6 +7,7 @@ use ParserOptions;
 use Revision;
 use Title;
 use User;
+use MediaWiki\MediaWikiServices;
 use SMW\MediaWiki\RevisionGuard;
 
 /**
@@ -49,10 +50,6 @@ class ContentParser {
 	private $skipInTextAnnotationParser = false;
 
 	/**
-	 * @note Injecting new Parser() alone will not yield an expected result and
-	 * doing new Parser( $GLOBALS['wgParserConf'] brings no benefits therefore
-	 * we stick to the GLOBAL as fallback if no parser is injected.
-	 *
 	 * @since 1.9
 	 *
 	 * @param Title $title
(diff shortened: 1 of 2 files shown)


kghbln:
Thanks for the patch!

JeroenDeDauw:
This removed compatibility with MW 1.31 (latest LTS).

SMW master supported MW versions here https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/COMPATIBILITY.md

kghbln:
> This removed compatibility with MW 1.31 (latest LTS).

Indeed. This can only be merged after the SMW 3.2.0 release was done which I hope to do around the time MW 1.35 gets released.

cscott:
Sorry -- you can use the technique from
https://github.com/JeroenDeDauw/ParserHooks/pull/24 to maintain backward
compat.
  --scott

>

kghbln:
> Sorry -- you can use the technique from.

Thanks for the info. However I think this can wait her for another eight weeks before can be merged. Since I am not a developer I'd rather focus on documenting stuff at this end.

cscott:
Incidentally, the other SMW patches got backward-compatibility code because they have travis tasks set up for the appropriate older MW versions, and the failing travis tasks let me know what versions my patch needed to support.  I don't think this repo had such automated tests; it would be useful to have them set up in the future so that drive-by compatibility fixes like this one can be tailored appropriately.  When I'm trying to patch tens of different extensions, I don't really have time to figure out the exact backwards-compatibility policy of each one individually.

kghbln:
Actually Semantic MediaWiki is doing intense testing. See [".travis.yml"](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/.travis.yml). I guess this is what you mean. And yes indeed, looking for a compatibility overview somewhere in the repo is not expected if you have this. In the heat of the action. :)

Issue #4724, opened 2020-04-15
Error running update script after installing SemanticMediaWiki


nnaka:
### Setup and configuration

- SMW version: 3.1.5
- MW version: 1.34.0
- PHP version: 7.3
- DB system (MySQL, Blazegraph, etc.) and version: 5.7.25 Percona XtraDB Cluster

### Issue
Installed the SemanticMediaWiki plugin 3.1.5 through composer.
The installation successfully completes through composer, but afterwards when I run the update script (`php .../maintenance/update.php`) to update the DB tables, I see an error:
```
Wikimedia\Rdbms\DBQueryError from line 1603 of <MY_PATH>/mediawiki-1.34.0/includes/libs/rdbms/database/Database.php: A database query error has occurred. Did you forget to run your application's database schema updater after upgrading?
Query: INSERT  INTO `smw_prop_stats` (p_id,usage_count,null_count) VALUES ('1','0','0')
Function: SMW\SQLStore\TableBuilder\Examiner\PredefinedProperties::doUpdate
Error: 1105 Percona-XtraDB-Cluster prohibits use of DML command on a table (<MY_DB_NAME>.smw_prop_stats) without an explicit primary key with pxc_strict_mode = ENFORCING or MASTER (<MY_DB_HOST:MY_DB_PORT>)
```
### Some background
I originally ran into a similar issue when installing MediaWiki itself, but was able to work around it using the command-line installer.  Another solution that was presented at that time is tracked here: https://phabricator.wikimedia.org/T226857.  Due to system restrictions, I am unable to set the `pxc_strict_mode=PERMISSIVE` as some forums seem to suggest.
Also cross posted here: https://www.mediawiki.org/wiki/Topic:Vkjyg9uce01b20p5

mwjames:
> DB system (MySQL, Blazegraph, etc.) and version: 5.7.25 Percona XtraDB Cluster

This isn't the first time someone has trouble with the Percona XtraDB Cluster (see #4507). @hexmode is the point man here.

> Function: SMW\SQLStore\TableBuilder\Examiner\PredefinedProperties::doUpdate
Error: 1105 Percona-XtraDB-Cluster prohibits use of DML command on a table (<MY_DB_NAME>.smw_prop_stats) without an explicit primary key with pxc_strict_mode = ENFORCING or MASTER (<MY_DB_HOST:MY_DB_PORT>)

[...]

Issue #3559, opened 2018-12-23
Add primary key indexes


hexmode:
While trying to import to a [Percona XtraDB Cluster](https://www.percona.com/software/mysql-database/percona-xtradb-cluster), I'm finding tables in various MW extensions that don't have primary key indexes  ([explanation](https://www.percona.com/doc/percona-xtradb-cluster/LATEST/features/pxc-strict-mode.html#tables-without-primary-keys)).

For SMW, I've made the following changes:
``` sql
ALTER TABLE smw_di_blob ADD PRIMARY KEY(p_id,s_id,o_hash);
ALTER TABLE smw_di_bool ADD PRIMARY KEY(p_id,s_id,o_value);
ALTER TABLE smw_di_uri ADD PRIMARY KEY(p_id,s_id,o_serialized);
ALTER TABLE smw_di_coords ADD PRIMARY KEY(p_id,s_id,o_serialized);
ALTER TABLE smw_di_wikipage ADD PRIMARY KEY(p_id,s_id,o_id);
ALTER TABLE smw_di_number ADD PRIMARY KEY(p_id,s_id,o_serialized);
ALTER TABLE smw_fpt_ask ADD PRIMARY KEY(s_id,o_id);
ALTER TABLE smw_fpt_askde ADD PRIMARY KEY(s_id,o_serialized);
ALTER TABLE smw_fpt_askfo ADD PRIMARY KEY(s_id,o_hash);
ALTER TABLE smw_fpt_asksi ADD PRIMARY KEY(s_id,o_serialized);
ALTER TABLE smw_fpt_askst ADD PRIMARY KEY(s_id,o_hash);
ALTER TABLE smw_fpt_cdat ADD PRIMARY KEY(s_id,o_serialized);
ALTER TABLE smw_fpt_conc ADD PRIMARY KEY(s_id);
ALTER TABLE smw_fpt_conv ADD PRIMARY KEY(s_id,o_hash);
ALTER TABLE smw_fpt_dtitle ADD PRIMARY KEY(s_id,o_hash);
ALTER TABLE smw_fpt_impo ADD PRIMARY KEY(s_id,o_hash);
ALTER TABLE smw_fpt_inst ADD PRIMARY KEY(s_id,o_id);
ALTER TABLE smw_fpt_lcode ADD PRIMARY KEY(s_id,o_hash);
ALTER TABLE smw_fpt_ledt ADD PRIMARY KEY(s_id,o_id);
[...]

PR #4732, opened 2020-04-19
Add `db-primary-keys` config profile, refs 4684

### LINKED ISSUE #4507: SMW should implement primary keys by default (at least for InnoDB)
### Setup and configuration

- SMW version: 2.5.7 (d528e24)
- MW version: 1.27.3 (73756b7)
- PHP version: 5.6.27 (apache2handler)
- DB system and version: Percona 5.7.23-23-57-log

### Issue
_After previously filing #3559 and using the back ported fix, I've come across some compelling reasons for using Indexes on InnoDB tables._

In his comments on the above issue, @mwjames writes
>...adding [a PRIMARY KEY] index just for the sake of the PRIMARY seems questionable unless of course there is evidence that the query planner would prefer the PRIMARY over any existing index.
>...adding PRIMARY KEY to all tables arbitrarily is going to be difficult so I'm currently undecided on how to proceed. One possible solution could be for those users that require a more strict key validation to add an extra index definition via a hook (see #3675 (comment)) and gets loaded as soon as SMW related tables are altered (during setup or update).

After reading this, I would agree that if the problem is just my finicky DB, then there is no reason to make a core change in SMW.  I appreciate the method that @mwjames developed to allow me to add primary keys--it works!--but I thought there must be something else going on, so I did quick research and came across [Tuning InnoDB Primary Keys](https://www.percona.com/blog/2018/07/26/tuning-innodb-primary-keys/) that states:
>InnoDB is called an index-organized storage engine. An index-organized storage engine uses the B-Tree of the primary key to stores the data, the table rows. **That means a primary key is mandatory with InnoDB.** If there is no primary key for a table, InnoDB adds a hidden auto-incremented 6 bytes counter to the table and use that hidden counter as the primary key. There are some issues with the InnoDB hidden primary key. You should always define explicit primary keys on your tables. In summary, you access all InnoDB rows by the primary key values.

(Emphasis throughout this issue is mine, not the original.)

Now, that is written by the principal architect at Percona (the makers of the DB that is forcing primary keys on me), so I want a neutral source.

I found something that looks compelling.  The MySQL ref manual, in the section titled "[Clustered and Secondary Indexes](https://dev.mysql.com/doc/refman/8.0/en/innodb-index-types.html)" states:
>Every InnoDB table has a special index called the clustered index where the data for the rows is stored....If the table has no PRIMARY KEY or suitable UNIQUE index, InnoDB internally generates a hidden clustered index named GEN_CLUST_INDEX on a synthetic column containing row ID values. 
>...**All indexes other than the clustered index are known as secondary indexes**. In InnoDB, **each record in a secondary index contains the primary key columns for the row**, as well as the columns specified for the secondary index. InnoDB uses this primary key value to search for the row in the clustered index. 

So, why not just (continue) to rely on InnoDB's hidden clustered index? I think the answer is found in the conclusion of [Scalability issues due to tables without primary keys](https://www.percona.com/blog/2013/10/18/innodb-scalability-issues-tables-without-primary-keys/) (albeit, written in 2013 against MySQL 5.5, [whose documentation describes InnoDB indexes](https://web.archive.org/web/20190808070212/https://dev.mysql.com/doc/refman/5.5/en/innodb-index-types.html) the same as the 8.0 documentation):
>Tables without primary keys cause a wide range of contention because they rely on acquiring dict_sys mutex to generate row-id values. This mutex is used at critical places within InnoDB. **Hence the affect of large amount of INSERTs to tables without primary keys is not only isolated to that table alone but can be seen very widely.** There are a number of times I have seen tables without primary keys being used in many different scenarios that include simple INSERTs to these tables as well as multi-row INSERTs as a result of, for example, INSERT … SELECT into a table that is being temporarily created. **The advice is always to have primary keys present in your tables.** Hopefully I have been able to highlight the true impact non-presence of primary keys can have.

mwjames:
This PR is made in reference to: #4684, #4724 

- PR provides a `db-primary-keys.php` config profile that users can invoke in case they need support for primary keys as in #4724 and as outlined in #4507 we won't have them created by default so for those that need them can do `enableSemantics( 'example.org' )->loadDefaultConfigFrom( 'db-primary-keys.php' );` which should be sufficient 
- Providing the profile doesn't mean SMW is supporting the functionality it just provides the profile for convenience
- Also for convenience, the `ConfigPreloadTableListPrimaryKeysCompleteTest` will check whether all SMW owned tables are listed in `db-primary-keys.php` or not and otherwise fail in case a new table is added without being listed in `db-primary-keys.php`
- PR uses the information from #3559 to match the list of primary keys

This PR includes:
- [x] CI build passed

Fixes #4507 

## Example

Output of `setupStore.php` when the `db-primary-keys.php` is loaded.

```
--- Database setup --------------------------------------------------------

Storage engine:                                                SMWSQLStore3
Database (type/version):                            mysql (10.4.11-MariaDB)

Checking version requirement ...
   ... done.

-------------------------------------------------------- Primary key(s) ---

The following updates adds primary key information for the tables owned by
Semantic MediaWiki.

Checking table definitions ...
   ... run table definition update ...                          39 (tables)
   ... done.

------------------------------------------------------ Core table(s) ------

Checking table smw_object_ids ...
```

### DIFF
--- a/data/config/db-primary-keys.php
+++ b/data/config/db-primary-keys.php
@@ -1,0 +1,152 @@
+<?php
+
+use Onoi\MessageReporter\MessageReporter;
+use SMW\Utils\CliMsgFormatter;
+
+/**
+ * This profile adds primary keys to Semantic MediaWiki owned tables to help DB
+ * environments like `Percona XtraDB Cluster` (#4724, #4507) that require those
+ * keys.
+ *
+ * Beware that while the profile is provided as part of the Semantic MediaWiki
+ * software it does NOT imply that systems like `Percona` are officially
+ * supported as there is no test setup that runs the required test suite.
+ *
+ * @see https://www.semantic-mediawiki.org/wiki/Config_preloading
+ *
+ * @since 3.2
+ */
+
+class ConfigPreloadPrimaryKeyTableMutator {
+
+	// #3559
+	const PRIMARY_KEYS = [
+
+		// Common property value tables
+		'smw_di_blob'     => 's_id,p_id,o_hash',
+		'smw_di_bool'     => 's_id,p_id,o_value',
+		'smw_di_coords'   => 's_id,p_id,o_serialized',
+		'smw_di_number'   => 's_id,p_id,o_serialized',
+		'smw_di_time'     => 's_id,p_id,o_serialized',
+		'smw_di_uri'      => 's_id,p_id,o_serialized',
+		'smw_di_wikipage' => 's_id,p_id,o_id',
+
+		// Fixed property value tables
+		'smw_fpt_ask'     => 's_id,o_id',
+		'smw_fpt_askde'   => 's_id,o_serialized',
+		'smw_fpt_askfo'   => 's_id,o_hash',
+		'smw_fpt_askdu'   => 's_id,o_serialized',
+		'smw_fpt_asksi'   => 's_id,o_serialized',
+		'smw_fpt_askst'   => 's_id,o_hash',
+		'smw_fpt_askpa'   => 's_id,o_hash',
+		'smw_fpt_cdat'    => 's_id,o_serialized',
+		'smw_fpt_conc'    => 's_id',
+		'smw_fpt_conv'    => 's_id,o_hash',
+		'smw_fpt_dtitle'  => 's_id,o_hash',
+		'smw_fpt_impo'    => 's_id,o_hash',
+		'smw_fpt_inst'    => 's_id,o_id',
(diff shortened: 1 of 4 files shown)


kghbln:
Referenced [on `$smwgDefaultStore`](https://www.semantic-mediawiki.org/w/index.php?title=Help%3A%24smwgDefaultStore&type=revision&diff=76633&oldid=74010) and [configuration preloding](https://www.semantic-mediawiki.org/wiki/Help:Configuration_preloading).

Issue #4738, opened 2020-04-23
Wikitext not parsed on "Special:Ask" when used with `intro` parameter


bescheid:
### Setup and configuration

- SMW version: 3.0.0
- MW version: 1.31.1
- PHP version: 7.2.25 (cgi-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: 10.1.40-MariaDB

### Issue
`{{#ask: ... |intro = [[page]] }}` works on the inline query page, but not on Spezial:Semantische_Suche. Here the result is a printout of `<a href="/wiki/page">Page</a>`  

### Steps to reproduce

https://www.univie.ac.at/rel_jap/an/Metalog:Bilder/Skulptur (click on "Weitere Ergebnisse")

kghbln:
@jaideraf I remember we had something like this before but cannot remember exactly. I think it was you who came up with it.

jaideraf:
No, I think you are refering to [this issue](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4467) (my question was about intro not going to the output in the templatefile format).

Actualy I think this is a new  issue, we can reproduce it on sandbox:  
https://sandbox.semantic-mediawiki.org/wiki/Issue/4738

kghbln:
Thanks for answering and creating an example on sandbox! Still I have the feeling we had this before. I think @krabina is our second expert in this field. Senior moments with no end I'm afraid.

mwjames:
> Still I have the feeling we had this before. I 

Question, what integration tests we have with `intro` / `outro`? Any of those test containing a wikitext notation and a test with `Special:Ask` (see the `S` section to test the output for special pages)?

jaideraf:
`s-0011.json` chekcs `intro` / `outro` for `table` format in three cases:
```
#1 (parse [[ ... ]] as links in intro/outro
#2 (parse URL in intro/outro)
#3 (parse template in intro/outro)
```
I expanded this text file to also test the `list`, `plainlist`, `ol` and `ul` formats. Should I send a PR? (I did not provide a fix, I don't know PHP).

bescheid:
Thank you. Actually, I use “plainlist” format combined with “template”.

I think the working solution is introtemplate rather than intro when using wikicode. Still, it is a kind of bug…

Best

Bernhard Scheid

Von: Jaider Andrade Ferreira [mailto:notifications@github.com]
[...]

PR #4749, opened 2020-04-25
Fix wikitext not parsed on "Special:Ask"

jaideraf:
This PR is made in reference to: https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4738 (Wikitext not parsed on "Special:Ask" when used with `intro`  and `outro` parameters)

- Fix per James comment https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4738#issuecomment-619439901

This PR includes:
- [ ] CI build passed

Fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4738

### DIFF
--- a/src/Query/ResultPrinters/ResultPrinter.php
+++ b/src/Query/ResultPrinters/ResultPrinter.php
@@ -382,19 +382,19 @@ abstract class ResultPrinter implements IResultPrinter {
 		// Apply intro parameter
 		if ( ( $this->mIntro ) && ( $results->getCount() > 0 ) ) {
 			if ( $outputmode == SMW_OUTPUT_HTML && $this->isHTML ) {
- 				$result = Message::get( [ 'smw-parse', $this->mIntro ], Message::PARSE ) . $result;
- 			} elseif ( $outputmode !== SMW_OUTPUT_RAW ) {
- 				$result = $this->mIntro . $result;
- 			}
+				$result = Message::get( [ 'smw-parse', $this->mIntro ], Message::PARSE ) . $result;
+			} elseif ( $outputmode !== SMW_OUTPUT_RAW ) {
+				$result = $this->mIntro . $result;
+			}
 		}
 
 		// Apply outro parameter
 		if ( ( $this->mOutro ) && ( $results->getCount() > 0 ) ) {
 			if ( $outputmode == SMW_OUTPUT_HTML && $this->isHTML ) {
- 				$result = $result . Message::get( [ 'smw-parse', $this->mOutro ], Message::PARSE );
- 			} elseif ( $outputmode !== SMW_OUTPUT_RAW ) {
- 				$result = $result . $this->mOutro;
- 			}
+				$result = $result . Message::get( [ 'smw-parse', $this->mOutro ], Message::PARSE );
+			} elseif ( $outputmode !== SMW_OUTPUT_RAW ) {
+				$result = $result . $this->mOutro;
+			}
 		}
 
 		// Preprocess embedded templates if needed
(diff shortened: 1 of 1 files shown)


src/Query/ResultPrinters/ResultPrinter.php
```diff
@@ -381,20 +381,20 @@ protected function handleNonFileResult( $result, QueryResult $results, $outputmo
...
 		// Apply intro parameter
 		if ( ( $this->mIntro ) && ( $results->getCount() > 0 ) ) {
-			if ( $outputmode == SMW_OUTPUT_HTML ) {
-				$result = Message::get( [ 'smw-parse', $this->mIntro ], Message::PARSE ) . $result;
-			} elseif ( $outputmode !== SMW_OUTPUT_RAW ) {
-				$result = $this->mIntro . $result;
-			}
+			if ( $outputmode == SMW_OUTPUT_HTML && $this->isHTML ) {
+ 				$result = Message::get( [ 'smw-parse', $this->mIntro ], Message::PARSE ) . $result;
+ 			} elseif ( $outputmode !== SMW_OUTPUT_RAW ) {
```
mwjames, line 386: We Somehow have some unrelated white space changes here (see left side, indicated by darker green).

mwjames:
While the change and check for `$this->isHTML ` seems simple, it doesn't explain why we suddenly would need this since this part of the code hasn't changed in years. I don't want to introduce a change that may have unanticipated side effects for example when used in combination with HTML code, or some parser functions like `{{#info: ...}}`.

src/Query/ResultPrinters/ResultPrinter.php
jaideraf: Weird, thanks, done.

jaideraf:
> I don't want to introduce a change that may have unanticipated side effects for example when used in combination with HTML code

Fair enough.

mwjames:
I have no problem on merging this change but as I said I need a guarantee that this doesn't suddenly introduce incompatibilities with other printers which means it should be tested more thoroughly hands-on.

jaideraf:
I am without a computer this weekend, so I plan to test it more next week. Though, it would be cool if this change would be applied on sandbox to see if it breaks the current set of examples. Is it possible @kghbln?

mwjames:
> I am without a computer this weekend, so I plan to test it more next week. Though, it would be cool if this change would be applied on sandbox to see if it breaks the current set of examples. Is it possible @kghbln?

I can merge this PR since the change itself is minor but as I said it would be good to confirm afterwards that it doesn't make things worse off.

kghbln:
I was not sticking around yesterday afternoon and evening. Anyhow the change is now on sandbox. @jaideraf Perhaps you could do the field testing now.

jaideraf:
> Anyhow the change is now on sandbox.

Thanks, @kghbln.

> I need a guarantee that this doesn't suddenly introduce incompatibilities with other printers which means it should be tested more thoroughly hands-on.

> as I said it would be good to confirm afterwards that it doesn't make things worse off.

OK. [Done](https://sandbox.semantic-mediawiki.org/wiki/Intro_and_Outro_testing). 

* :relieved: 32 result formats does parse wikitext in `intro` or `outro` as (now) expected in Special:Ask. 
* :relieved: 28 result formats does parse wikitext in `intro` or `outro` as (now) expected in normal pages.
* :expressionless: 7 result formats does **not** parse wikitext in `intro` or `outro` (this is not a regression: they had never parsed wikitext in `intro` or `outro` - confirmed with an earlier version of SMW): `timeseries`, `jqplotchart`, `map`, `valuerank`, `datatables` and `eventcalendar`. 
* :no_mouth: 1 result format (`moderntimeline`) does not show `intro` or `outro` data in either normal pages or Special:Ask (maybe it is not implemented). 
* :alien: 16 result formats were not tested because of several reasons: it is already broken; it is not listed on Special:Ask; it is not expected to have `intro` or `outro` parameters (for example, math formats); missing data to test them; the result format is not activated in the wiki, etc.

kghbln:
@jaideraf WOW Cool Thank you!

mwjames:
@jaideraf Absolute marvelous. This is surely an effort I haven't seen in a long time from someone of the SMW community.

PR #4750, opened 2020-04-26
Improved source code and documentation

Fonata:
This PR makes the code optics a little more consistent:
- Documentation: make it even easier to get PHPUnit running
- Added and improved DocBlocks. This should help IDEs and static code analyzers.
- Removed unnecessary code
- Source code formatting: More consistent space placement
- Add newlines to PHP DocBlocks to separate @annotation types
- Reduce code complexity: no else needed after return
- If the first `__construct` parameter is optional, the second has to be, too
- Removed unneeded ternary operator of the shape `condition ? true : false`
- A few more files now have a new line character at the very end of files. This makes viewing and editing them a little easier.
- If a function parameter can be null, the TypeHint needs to start with `?` (PHP 7.1 feature)
- Use lowercase 'null' like elsewhere

This PR includes:
- No new tests 
- [x] CI build passed


mwjames:
I'm merging this as-is even though I only looked roughly at the change otherwise my I cannot do anything else this weekend. In future, changes to the documentation and removing deprecated functions should be split from each other into separate PRs.

The PR as it is can become stale very quickly because it touches a large class set making reviewing it rather daunting and we should avoid that in future.

kghbln:
@mwjames Is writing what I thought. However a great effort @Fonata and thank you for this.

PR #4479, opened 2020-01-26
Add `SchemaFilter` (`CategoryFilter`, `NamespaceFilter`), refs 4280

mwjames:
This PR is made in reference to: #4280

This PR addresses or contains:

- Ground work for #4280 to allow to define filter conditions (see below) in a schema and case it supports the filter notation
- Implements `CategoryFilter` and `NamespaceFilter`
- `filter.md` contains details about the use those filters 

This PR includes:
- [x] Tests (unit/integration)
- [x] CI build passed

## Example

```
{
	"if": {
		"namespace": "NS_HELP",
		"category": { "allOf" : [ "Foo", "Bar", "Foobar" ] }
	},
	"then": {
		"follow": "step_1"
	}
}
```
```
$compartments = $this->schemaList->newCompartmentIteratorByKey(
	'filter_rules'
);

$schemaFilterFactory = new SchemaFilterFactory();

$categoryFilter = $schemaFilterFactory->newCategoryFilter(
	$categories
);

$namespaceFilter = $schemaFilterFactory->newNamespaceFilter(
	$ns
);

$namespaceFilter->setNodeFilter(
	$categoryFilter
);

$namespaceFilter->filter( $compartments );
```

Issue #4280, opened 2019-09-01
Search suggestion summaries


mwjames:
Allowing users to quickly grasp what an object or subject is about and help them gather information about available facts is an important part of Semantic MediaWiki.

When @s7eph4n first added support (#450) for `Special:Search` we made an important step into the direction of integrating Semantic MediaWiki with the MediaWiki search. This was followed by allowing users to define search forms (#3126) in SMW 3.0 so that they could create use case centered input forms that help narrow the time for users to find particular items with annotated properties.

The next integration will draw the circle even closer by helping user that rely on the completion search to provide suggestion summaries so that information can be displayed even sooner hereby making the search process more focused and efficient.

## Example 

![image](https://user-images.githubusercontent.com/1245473/64070045-835a7a80-cc47-11e9-84cb-339ff6a73c1f.png)

Depending on the context (namespace and category) summaries can be customized (using a new schema type) to retrieve summaries from different properties providing enough flexibility for administrators and curators to define of what should be shown when an item is part of a result set.

![smw-suggestion-summaries](https://user-images.githubusercontent.com/1245473/64070447-86f1ff80-cc4f-11e9-8686-5f2b07d3db40.gif)

## Technical notes

Since MediaWiki doesn't provide any hooks or similar event based system to extend the content of the completion search we rely on the `MutationObserver` to observe changes to the suggestion list and manipulate the content on the fly without having to touch any MediaWiki owned files.

mwjames:
While supporting this function out-of-the-box in SMW core would surely make sense, I decided against it and I will place the code into a separate `CompletionSearchSummaries` extension given that the JS code requires a `MutationObserver` to fetch mutation events on the generated HTML `suggestion` list because there is NO hook or API that would allows to inject context information simply in PHP.

Now, the problem is that in case [0] modifies the `suggestion` list in an incompatible way, the `MutationObserver` might suddenly stop working therefore to lower the risk for a non-functionality feature isolate and deploy it as separate extension.

[0] https://phabricator.wikimedia.org/T244392

kghbln:
Probably "Search Summaries Completion" will be a slightly better name.

PR #4765, opened 2020-05-03
Add `PropertyFilter`, refs 4479

mwjames:
This PR is made in reference to: #4479

- Extends #4479 to provide `property` as conditional filter and is equivalent to what the `category` filter does just that it uses properties as argument to match a rule or if combined with other filters can define more complex constructs such as shown in the example
- `PropertyFilter` only provides the ground work for the filtering, a schema type has to decide whether a `property` filter should be used or not and requires to provide an individual `RuleFinder` (or `RuleMatcher`) as seen in [0]
- Definitions of how conditionals are defined is bound to the schema type including what attributive filter arguments are supported (e.g `allOf`,`oneOf`) as in [1]

This PR includes:
- [x] CI build passed

## Example

```json
{
    "if":{
        "namespace":[
            "NS_MAIN",
            "NS_HELP"
        ],
        "category":{
            "oneOf":[
                "category-6-a",
                "category-6-b"
            ]
        },
        "property":{
            "allOf":[
                "property-6-a",
                "property-6-b"
            ]
        }
    },
    "then":"do_something"
}
```

[0] https://github.com/SemanticMediaWiki/SemanticImageCaption/blob/master/src/RuleFinder.php
[1] https://github.com/SemanticMediaWiki/SemanticImageCaption/blob/master/data/schema/imagecaption-rule-schema.v1.json#L51-L126

### DIFF
--- a/src/Schema/Filters/PropertyFilter.php
+++ b/src/Schema/Filters/PropertyFilter.php
@@ -8,6 +8,7 @@ use SMW\Schema\ChainableFilter;
 use SMW\Schema\CompartmentIterator;
 use SMW\Schema\Compartment;
 use SMW\Schema\Rule;
+use SMW\DIProperty;
 use RuntimeException;
 
 /**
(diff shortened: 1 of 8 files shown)


kghbln:
Also refs #4280

PR #4796, opened 2020-06-18
Escape separator in CSV printer

prudloff-insite:
- When exporting a query result as CSV, the values can contain an unescaped `,` character, so it becomes impossible to properly parse the values from the CSV.

This PR includes:
- [ ] CI build passed

We are stuck on the 2.5 branch for now but I am not sure you still accept PR for this branch.
(If the 3.1 still has the same issue, I think a similar fix could be applied.)

### DIFF
--- a/includes/queryprinters/CsvResultPrinter.php
+++ b/includes/queryprinters/CsvResultPrinter.php
@@ -87,7 +87,7 @@ class CsvResultPrinter extends FileExportPrinter {
 					$growing = array();
 
 					while ( ( $object = $field->getNextDataValue() ) !== false ) {
-						$growing[] = Sanitizer::decodeCharReferences( $object->getWikiValue() );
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Here is the line on master: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Query/ResultPrinters/CsvFileExportPrinter.php#L160

`getWikiValue` got changed to `getShortWikiText`, though that is probably not fixing this issue.

I can't verify this code is correct just by looking at it. Having an unit test for this code would be great. There do not seem to be any yet though.

krabina:
On SMW 4.1.2, I cannot reproduce the problem. See this example
https://test.knowledge.wiki/Test_4796
There are several "," inside values, but all are correctly escaped.

jaideraf:
Since we have not received more responses from the author, I think this pull should closed.

Issue #4727, opened 2020-04-18
Use of the `InternalParseBeforeLinks` hook


mwjames:
Someone pointed me towards [0] "Deprecate InternalParseBeforeLinks hook".

[1] clarifies the task that is expected to be carried out by the hook which has been an integral part of Semantic MediaWiki since the 1.0 [2] (Dec 31, 2007) release. Removing the hook will mean Semantic MediaWiki will no longer work as designed and any wiki that deploys the software will more or less break.

Aside from breaking entire wikis, I'm guessing that around 50% of the tests (~4000 tests) will break as well.

And no, the parser function `#set` is no equivalent replacement for the `[[Foo::Bar]]` syntax.

> TODO: investigate whether SemanticMediaWiki can use alternatives.

I'm ears but tests are to remain untouched and have to work as before otherwise the situation remains unchanged and will mostly likely break wikis in their present form.

[0] https://gerrit.wikimedia.org/r/#/c/mediawiki/core/+/589463/
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/47e1af9e77f59714b9897bbd96954522e6d4e0b4
[2] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/1.0/includes/SMW_GlobalFunctions.php#L141

kghbln:
> And no, the parser function `#set` is no equivalent replacement for the [[Foo::Bar]] syntax.

Indeed is not and there is currently no direct replacement. Not being able to use the in-text annotation syntax `[[Foo::Bar]]` will affect 99,99 % of all wikis using SMW.

Given the insulting and disgraceful comments I personally received from some people regarding the syntax changes concerning the list format rework that came with SMW 3.0 I am not expecting to receive friendlier comments regarding what may lay ahead here in case ...

Personally I think that a) a solution should be found to keep in-text annotations work for existing users and b) an alternative parser function should be provided for starters of new wikis to avoid in-text annotations using the existing syntax.

mwjames:
> an alternative parser function should be provided for starters of new wikis to avoid in-text annotations using the existing syntax.

[...]

PR #4807, opened 2020-06-24
Bump version to 4.x

JeroenDeDauw:
* Update SMW to MW 1.32+
* Clean CI jobs
* Increment major version

We can drop support for further MW versions before releasing. Would wait a bit with the release, even though there is some pressure to have an SMW release that does not cause a trillion warnings on MW 1.35


JeroenDeDauw:
@kghbln so this would bump SMW to 4.0. Maybe we also drop MW 32 and 33?

I rather keep compat back to 1.31 but that seems to be quite a bit of work in case of SMW which I do not see anyone doing at the moment.

jaideraf:
If we will lose the bracket syntax (`[[Foo::Bar]]`) [as noted by James](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4727), it will be also an "opportunity" to make a SMW 4.

JeroenDeDauw:
That would definitely require a major release yes. I'd only drop the brackets if and when we are absolutely forced to, since this is not a good change for our users.

kghbln:
Does this kill the in-text annotation syntax?

JeroenDeDauw:
This does not remove annotation syntax.

There are tests against master which was basically 1.35 when I made this change. We can add stuff later. This change gets rid of the colossal mess with half the tests being optional and failing and the whole thing taking hours to run. I might throw the entire travis setup out in favor of GitHub actions and maybe get something that runs in a reasonable time.

Issue #4674, opened 2020-03-28
Disabling entity issue panel


JeroenDeDauw:
Is there a way to disable the entity issue panel via wiki config?

![image](https://user-images.githubusercontent.com/146040/77826343-16d92880-710f-11ea-92d5-81e53054ef72.png)

Having it show, by default, on pages that do not have semantic annotations on them, does not seem like a good idea to me. For the same reasons listed here: https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4673#issue-589594461. Maybe having it on by default makes sense. Not sure. But on pages without semantic annotations? Uhhh.

User experience: install SMW. New things shows up on page. Complicated technical explanation. Why do I need to deal with this? What does it even mean. I'm not even using SMW yet. Maybe I should give this Cargo thing a try instead?

kghbln:
I agree that we should probably have an configuration parameter which allows to disable this for advanced users. 

However I do not think that it should be disabled by default. Admittedly I myself already had a couple of situation where this helped me identify structural issues fast and not by chance at a later stage or after a couple of people sent me e-mails about an issue. For me as an expert user this is actually an improvement.

> Having it show, by default, on pages that do not have semantic annotations on them, does not seem like a good idea to me.

Do not forget special property annotations so an extra logic would have to be added here which probably complicates things unnecessarily. For all the people who are sitting on SMW and do not want to use it the configuration parameter to enable/disable is good enough I guess.

JeroenDeDauw:
> disable this for advanced users

While I find the new UI element mildly annoying, my concern is about its effect on non-advanced users. Especially in parts of a wiki where SMW is not used, or where it is otherwise not relevant to the user.

That you like having it on by default suggests something about advanced users :) 

> special property annotations

Yeah. But that is technical stuff. This does not help new users that get confused. And it does not matter to admins that get put off from installing SMW because of usability issues.

kghbln:
> While I find the new UI element mildly annoying, my concern is about its effect on non-advanced users. Especially in parts of a wiki where SMW is not used, or where it is otherwise not relevant to the user.

I think that we agree on having a configuration parameter. If everything works the panel does not show up at all. If if does it is about constraint violations and stuff like that. This in turn is useful for novice users.

> Yeah. But that is technical stuff. This does not help new users that get confused. And it does not matter to admins that get put off from installing SMW because of usability issues.

In these cases you would want to use the configuration parameter.

[...]

PR #4842, opened 2020-09-06
Add smwgDetectOutdatedData setting

JeroenDeDauw:
Partially deals with https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4674

This default can be changed when the issues with the current behavior
are fixed.

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -2482,6 +2482,16 @@ return [
 		// [ 'host' => '127.0.0.1', 'port' => 9200, 'scheme' => 'http' ],
 		// [ 'host' => '127.0.0.1', 'port' => 9200, 'scheme' => 'http', 'user' => 'username', 'pass' => 'password!#$?*abc' ]
 		// 'localhost:9200'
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Merging this to retain 3.1 behavior. I don't agree with changing it without first resolving the issues described in the linked issue. And neither does the community, without even knowing about said issues https://twitter.com/SemanticMW/status/1302022612219035648

kghbln:
> Merging this to retain 3.1 behavior.

From the name of the parameter I get that it disables outdated data detection which is the opposite of what 3.1 is doing if set to false.

JeroenDeDauw:
You are right, I got confused here. The behavior that is retained is not showing the issue panel in case the data is outdated. No issue panel in 3.1.

JeroenDeDauw:
Wait no - the revision check was added in 3.2. https://www.semantic-mediawiki.org/wiki/Help:Entity_examinations @kghbln what are you talking about?

PR #4844, opened 2020-09-09
Use Language::formatNumNoSeparators where appropriate

cscott:
Avoids using the deprecated `$noSeparators` parameter to `Language::formatNum` in favor of `Language::formatNumNoSeparators`, which has been around since MW 1.21.


kghbln:
Thanks for the pull!

Issue #4851, opened 2020-09-30
SMW 3.2.0 incorrectly registers available rights


Nikerabbit:
### Setup

- SMW version: 3.2.0
- MW version: 1.36.0-alpha
- PHP version: 7.3.19-1~deb10u1 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: 10.3.23-MariaDB-0+deb10u1

### Issue
First reported in: https://phabricator.wikimedia.org/T264208

Detailed description of the issue and a [stack trace](https://www.semantic-mediawiki.org/wiki/Help:Identifying_bugs) if applicable:

```
https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/GroupPermissions.php#L47

var_dump( $groups ); die();

array(4) {
  ["smwadministrator"]=>
  array(1) {
    ["smw-admin"]=>
    bool(true)
  }
  ["smwcurator"]=>
  array(6) {
    ["smw-patternedit"]=>
    bool(true)
    ["smw-schemaedit"]=>
    bool(true)
    ["smw-pageedit"]=>
    bool(true)
    ["smw-viewjobqueuewatchlist"]=>
    bool(true)
    ["smw-viewentityassociatedrevisionmismatch"]=>
    bool(true)
    ["smw-vieweditpageinfo"]=>
    bool(true)
  }
  ["smweditor"]=>
  array(1) {
    ["smw-vieweditpageinfo"]=>
    bool(true)
  }
  ["user"]=>
  array(1) {
    ["smw-vieweditpageinfo"]=>
    bool(true)
  }
}
```

Steps to reproduce the observation (recommendation is to use the [sandbox](https://sandbox.semantic-mediawiki.org)):
https://sandbox.semantic-mediawiki.org/w/api.php?action=help&recursivesubmodules=1

JeroenDeDauw:
If you know what the desired behavior is, could you briefly describe it?

Nikerabbit:
I was gonna submit patch but it's already 21.37... basically https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/GroupPermissions.php#L46 needs to use `array_keys` or on `$rights`.

PR #4852, opened 2020-10-01
Fix registration of available rights (#4851)

Nikerabbit:


JeroenDeDauw:
Bit disturbing there apparently is no test for this code

Issue #4857, opened 2020-10-08
Make `ask` and `askargs` API return a value for `format=count`


osnard:
Currently when running a query with the Action API modules `ask` or `askargs` no value will be returned.

E.g. `api.php?action=askargs&conditions=Modification%20date::%2B&printouts=Modification%20date&parameters=sort%3DModification%20date|`**`format%3Dcount`**`&format=json` will return 

```
{
	"query": {
		"printrequests": [...],
		"results": [],
		"serializer": "SMW\\Serializers\\QueryResultSerializer",
		"version": 2,
		"meta": {...}
	}
}
```
The query is executed correctly and the `SMWQueryResult` object contains the proper value. it is just not returned. it would be nice to have Action API modules `ask` and `askargs` respect  `format=count`

kghbln:
Reopening this. See also #4890

akuckartz:
Can this be closed? #4890 was closed.

kghbln:
> Can this be closed? #4890 was closed.

No, since this is feature request. #4890 was a bug as a result of an unsuccessful implementation attempt

D-Groenewegen:
**+1** 
I just ran into this. The ask API doesn't support result formats but I would welcome an additional parameter to the ask API, maybe `totalcount`, along with the other data in `meta`. 

If I understand correctly, the only way you can retrieve the total number of results from the API is :
- use the ask API with limit set to infinite (well, 99999999999999999) and read `count` . Cons: painfully slow
- use askarg API with `format=count`: feasible only for rather basic queries like `<some property>::<some value>`
- use MW's parser API on full query syntax with `format=count` : circuitous and potentially slow
- use SMW with ElasticSearch and use its search API : available only to the lucky few

gesinn-it-gea:
Fixed in commit [2be6e6757](https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/2be6e6757) (see also #7006).

A previous attempt (PR #4858, commit c9b7a6ff, reverted after #4890) placed the count check inside `getResult()`, which caused `countValue` to bleed into normal queries and broke existing API clients.

The new fix intercepts in `doFormat()` instead, guarded by `getCountValue() !== null`. Since `setCountValue()` is only called by the query engine when running in `MODE_COUNT`, this guard never fires for normal queries — making the change fully non-breaking.

The API now returns:
```json
{
  "query": {
    "count": 42,
    "meta": { "type": "count" }
[...]

PR #4858, opened 2020-10-08
Make Action API return value for "format=count"

osnard:
See https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4857

### DIFF
--- a/src/MediaWiki/Api/ApiQueryResultFormatter.php
+++ b/src/MediaWiki/Api/ApiQueryResultFormatter.php
@@ -34,6 +34,12 @@ class ApiQueryResultFormatter {
 	protected $isRawMode = false;
 
 	/**
+	 *
+	 * @var SMWQueryResult
+	 */
+	protected $queryResult = null;
+
+	/**
 	 * @since 1.9
 	 *
 	 * @param SMWQueryResult $queryResult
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
@osnard seems like I was to quick here. Your change breaks the tests. Can you fix? Else I will have to revert.

osnard:
@JeroenDeDauw I will have a look. Need to figure out the reason first. The breaking tests do not use `format=count`.

osnard:
I have now committed https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4860 as a follow up to fix the tests.

But to be honest, I'd like to have a recommendation here, regarding the question of how a `format=count` query should be formatted. Currently I am just using `query.results` and put an integer in there. This is not really nice, as usually this fields holds an associative array with page names as keys. All other fields (like `query.meta` and `query.printrequests` ) are just the same.
Having `query.printrequests` with standard "Subject" entry seems to be wrong also.

Should I better introduce a new complete new API output format and change the existing tests accordingly? Also I have no clue why the tests failed in the first place. Non of the test cases involved `format=count`, therefore no `SMWQueryResult` should have the "count value" set.

PR #4860, opened 2020-10-12
Fix AskArgs API UnitTests

osnard:
Re-use standard result format

Follow up for https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4858


osnard:
Those failing tests seem unrelated. @JeroenDeDauw, can you confirm?

PR #4862, opened 2020-10-16
Remove usage of hard-deprecated ParserCache::getETag.

Pchelolo:
See https://gerrit.wikimedia.org/r/c/mediawiki/core/+/634593

ParserCache::getETag is being removed from MW since it's not a ParserCache concern to build tags for the whole page.


Seb35:
This has been fixed independently in 9091adb78697203188f67ec64dd6ac59c229b16e (almost same code, shaped differently, and replacing a second occurence of ParserCache::getETag).

Issue #4868, opened 2020-10-22
Do not write objects into ExtensionData


cicalese:
As part of [ParserCache / RESTBase / Parsoid integration](https://phabricator.wikimedia.org/T250500), the ParserCache serialization format is being changed to JSON. This requires that an objects that are being written to ExtensionData currently be converted to JSON serialization.  It appears that there are places in [ParserData.php](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/ParserData.php) and perhaps elsewhere where this must be fixed.

We plan to submit a patch to address this.

cscott:
There have been some fixes/improvements to JSON serialization since this was filed; though see T313452.

JeroenDeDauw:
https://phabricator.wikimedia.org/T313452

hexmode:
Looks like this was superseded with #5285.  Also, discussed this with Cindy and she said she probably wouldn't use this.

hexmode:
Should have closed as resolved since #5285 has apparently fixed it.

PR #4869, opened 2020-10-22
Make TranslationPropertyAnnotator compatible with new format.

Pchelolo:
We are moving ParserCache from PHP serialization to JSON serialization,
thus instances of classes will no longer be supported in extension
data. https://gerrit.wikimedia.org/r/635901 changes the Translate
extension to write primitives instead of the Title.

See https://phabricator.wikimedia.org/T266268

### DIFF
--- a/src/Property/Annotators/TranslationPropertyAnnotator.php
+++ b/src/Property/Annotators/TranslationPropertyAnnotator.php
@@ -64,11 +64,22 @@ class TranslationPropertyAnnotator extends PropertyAnnotatorDecorator {
 			);
 		}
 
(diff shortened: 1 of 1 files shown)


kghbln:
I believe this refs #4868

Pchelolo:
> Thanks a lot for the pull. I believe this refs #4868

Nope. The ticket is about not WRITING into there, this is about not reading from the data set by other extension

kghbln:
> Nope. The ticket is about not WRITING into there, this is about not reading from the data set by other extension

Thanks for clarifying my mistake!

Issue #4833, opened 2020-08-28
The "Refresh" option in the page menu no longer selectable


kghbln:
### Setup and configuration

- MediaWiki | 1.35.0-rc.2 (0f1a2a2)18:43, 21 August 2020
- PHP | 7.2.24-0ubuntu0.18.04.6 (apache2handler)
- MariaDB | 10.1.44-MariaDB-0ubuntu0.18.04.1
- Semantic MediaWiki | 3.2.0-alpha (1c1c4a2) 19:23, 17 August 2020

### Issue
The "Refresh" option in the page menu allowing to clear the cache of a page is no longer selectable for Vector and Monobook probably others, too.

### Steps to reproduce
[Main Page of s-mw.o](https://sandbox.semantic-mediawiki.org/wiki/Main_Page)

jaideraf:
It coud be related to https://phabricator.wikimedia.org/T262092

H21lab:
Affecting also to Timeless skin. Workaround is to use `?action=purge` parameter in URL

Seb35:
Digging in SMW history, the 'Refresh' option was added [https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2502 on 10 Jun 2017] followed by [https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2631/files a second PR on 19 Aug 2017] to disable this option when JS is not enabled.

For an unknown reason, the JS to un-disable the link is no more loaded in MW 1.35 and/or SMW 3.2 (not sure which part is responsible of this).

kghbln:
> For an unknown reason, the JS to un-disable the link is no more loaded in MW 1.35 and/or SMW 3.2 (not sure which part is responsible of this).

Well, it worked for SMW 3.2 on MW 1.34 so it is safe to say that MW 1.35 causes the code in SMW to no longer be loaded

Seb35:
It appears that the resource 'mediawiki.notify' was deleted because always available (loaded in resource 'mediawiki.base'), see https://phabricator.wikimedia.org/T233676.

For MW 1.35, the patch is simply to remove dependency to 'mediawiki.notify' in `smw/util/ext.smw.util.purge.js` and `res/smw/util/ext.smw.util.postproc.js`

[...]

PR #4881, opened 2020-11-05
Re-enable the "Refresh" menu when JS is enabled

Seb35:
MW 1.35 deleted the resource mediawiki.notify (included in mediawiki.base), resulting in avoiding the undisabling of the "Refresh" menu (removing of the class "is-disabled" when JS is enabled). See  #2631 for the mechanism to enable when JS is loaded.

In order to support MW < 1.35, I declared the dependencies on PHP side to benefit of version_compare, and consequently removed mw.loader.using on JS side.

In JS files, removed one initial tab; the diff rendering seems to be a bit lost.

Issue: #4833

### DIFF
--- a/res/Resources.php
+++ b/res/Resources.php
@@ -273,6 +273,9 @@ return [
 			'smw-purge-update-dependencies'
 		],
 		'position' => 'top',
+		'dependencies' => array_merge( [
+			'mediawiki.api',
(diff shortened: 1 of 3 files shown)


JeroenDeDauw (approved):

kghbln:
Thanks a lot for the pull. This really resolves pain! :)

PR #4898, opened 2020-12-15
Improve special page name

JeroenDeDauw:


kghbln:
@JeroenDeDauw I do not think this is a big leap forward if at all but I do not object.

PR #4927, opened 2021-02-08
Use $1 in the messages on Special:Types

Seb35:
There is a small glitch on [Special:Types with French language](https://sandbox.semantic-mediawiki.org/wiki/Special:Types?uselang=fr) because the message is trimed of the two first characters, which are “” in other languages, but the French starts with «&amp;#​160; so the two first characters are «&amp; and the remaining #​160; is displayed.

Instead of modifying afterwards the message, use the $1 inside the message.

Also, uniformise the style (font-size: 85%) by using it for the fallback case when there is no help message for some specific type.

### DIFF
--- a/includes/specials/SMW_SpecialTypes.php
+++ b/includes/specials/SMW_SpecialTypes.php
@@ -421,22 +421,21 @@ class SMWSpecialTypes extends SpecialPage {
 			$typeValue = TypesValue::newFromTypeId( $typeId );
 			$msgKey = 'smw-type' . str_replace( '_', '-', strtolower( $typeId ) );
 
-			$text = $typeValue->getLongHTMLText( $linker );
-
 			if ( Message::exists( $msgKey ) ) {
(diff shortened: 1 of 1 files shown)


Seb35:
I tested it on a real wiki with MW 1.35.1 and SMW 3.2.2, and it works.

The failing test is about the file i18n/ne.json whose, supposedly, would have different values for the messages smw_kiloseparator and smw_decseparator, but this file has no both such keys. On the other side, all tests on Travis fail on this specific test, so it seems to be quite real -- not a transcient error.

I’m executing locally the tests to hopefully understand the failing test on Travis.

Nikerabbit:
I had a glitch with exports on Monday so I was not able to export a fix for the test failure. I'll try to do one manually now.

Nikerabbit:
Tests on master should work now.

Seb35:
Thanks @Nikerabbit. I amended the commit to relaunch Travis.

kghbln:
Thanks a lot for your efforts! @JeroenDeDauw I guess this is ready to merge?!

Issue #4943, opened 2021-02-23
MW 1.35: s-0001.json - Use of SpecialPageFactory::getPage was deprecated in MediaWiki 1.32.


gesinn-it:
4) SMW\Tests\Integration\JSONScript\JSONScriptTestCaseRunnerTest::testCaseFile with data set "s-0001.json" ('/home/travis/build/SemanticMe...1.json')

Use of SpecialPageFactory::getPage was deprecated in MediaWiki 1.32. [Called from SMW\Tests\Utils\JSONScript\SpecialPageTestCaseProcessor::process in /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/JSONScript/SpecialPageTestCaseProcessor.php at line 88]

https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/jobs/760157304#L2040

Seb35:
See also #4623 and linked issues for previous replacements of this deprecated static method.

Issue #4603, opened 2020-02-29
Use of SpecialPageFactory::getLocalNameFor was deprecated in MediaWiki 1.32. [Called from SMW\DataValues\TypesValue::makeSpecialPageTitleText ...]


mwjames:
### Setup and configuration

```
Using PHP 7.4.0
Semantic MediaWiki: 3.2.0-alpha, git: ad102f9
                    SMWSQLStore (mysql:5.6.33-0ubuntu0.14.04.1)
MediaWiki:          1.35.0-alpha, git: 8ea8171 (refs/heads/master)
                    Extension vendor autoloader
Site language:      en
Execution time:     2020-02-29 01:58
Debug logs:         Disabled
Xdebug:             Disabled (or not installed)
Intl/ICU:           7.4.0 / 52.1
PCRE:               10.33 2019-04-16
```

### Stack trace

```
1) SMW\Test\PropertiesQueryPageTest::testFormatPropertyItemOnUserDefinedProperty
Use of SpecialPageFactory::getLocalNameFor was deprecated in MediaWiki 1.32. [Called from SMW\DataValues\TypesValue::makeSpecialPageTitleText in /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/src/DataValues/TypesValue.php at line 246]
/home/travis/build/SemanticMediaWiki/mw/includes/debug/MWDebug.php:333
/home/travis/build/SemanticMediaWiki/mw/includes/debug/MWDebug.php:265
/home/travis/build/SemanticMediaWiki/mw/includes/GlobalFunctions.php:1035
/home/travis/build/SemanticMediaWiki/mw/includes/specialpage/SpecialPageFactory_deprecated.php:91
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/src/DataValues/TypesValue.php:246
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/src/DataValues/TypesValue.php:170
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/includes/querypages/PropertiesQueryPage.php:202
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/includes/querypages/PropertiesQueryPage.php:146
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/includes/querypages/PropertiesQueryPage.php:103
[...]

PR #4947, opened 2021-02-27
Maintenance: replace deprecated static function

Seb35:
The static calls to methods of SpecialPageFactory are replaced by an instance of the class.

The only other occurence of `SpecialPageFactory::` is in a comment in [tests/phpunit/includes/specials/SpecialsTest.php](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/includes/specials/SpecialsTest.php), but given the class still exists I guess it’s correct.

Kept the old version for MediaWiki 1.31, although I’m not sure if it will be supported by SMW 4.0.

Fixes: #4603
Fixes: #4943

### DIFF
--- a/tests/phpunit/Utils/JSONScript/SpecialPageTestCaseProcessor.php
+++ b/tests/phpunit/Utils/JSONScript/SpecialPageTestCaseProcessor.php
@@ -4,12 +4,12 @@ namespace SMW\Tests\Utils\JSONScript;
 
 use FauxRequest;
 use Language;
+use MediaWiki\MediaWikiServices;
 use OutputPage;
 use RequestContext;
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Only need to support 1.35 and later: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/COMPATIBILITY.md#platform-compatibility-and-release-status

Seb35:
I believe this is ready, I didn’t see there was a syntax error in my previous commits, sorry.
This commit only improve the 1.35 Travis tests where there are now 5 errors instead of 6 (21743 assertions instead of 21739).

Issue #4944, opened 2021-02-23
MW 1.35: Use of Revision::newFromId was deprecated in MediaWiki 1.31.


gesinn-it:
5) SMW\Tests\Integration\MediaWiki\LinksUpdateSQLStoreDBIntegrationTest::testPageCreationAndRevisionHandlingBeforeLinksUpdate

Use of Revision::newFromId was deprecated in MediaWiki 1.31. [Called from SMW\Tests\Integration\MediaWiki\LinksUpdateSQLStoreDBIntegrationTest::retrieveAndLoadData in /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/MediaWiki/LinksUpdateSQLStoreDBIntegrationTest.php at line 190]

https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/jobs/760157304#L2056

Abijeet:
This was fixed in https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/4e0aed2f8084b71442e94c05598c8f71cfb922f0#diff-e522ca6c381e561f70b7c582a2fabea47d77e1b867ead2440bfa90d2045d85d3R191

Issue #4941, opened 2021-02-23
MW 1.35: Use of Revision::getTimestampFromId was deprecated in MediaWiki 1.35.


gesinn-it:
2) SMW\Tests\Integration\InterwikiDBIntegrationTest::testRdfSerializationForInterwikiAnnotation

Use of Revision::getTimestampFromId was deprecated in MediaWiki 1.35. [Called from SMWExportController::printPages in /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/includes/export/SMW_ExportController.php at line 415]

https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/jobs/760157304#L1992

Issue #4714, opened 2020-04-11
Use of Revision::getTimestampFromId was deprecated in MediaWiki 1.35. [Called from SMWExportController::printPages


mwjames:
### Setup and configuration

```
Using PHP 7.4.0

Semantic MediaWiki: 3.2.0-alpha, git: 5a80f51
                    SMWSQLStore (mysql:5.6.33-0ubuntu0.14.04.1)

MediaWiki:          1.35.0-alpha, git: 24b1aae (refs/heads/master)
                    Extension vendor autoloader

Site language:      en
Execution time:     2020-04-11 01:45

Debug logs:         Disabled
Xdebug:             Disabled (or not installed)

Intl/ICU:           7.4.0 / 52.1
PCRE:               10.33 2019-04-16
```

### Issue

```
2) SMW\Tests\Integration\InterwikiDBIntegrationTest::testRdfSerializationForInterwikiAnnotation

Use of Revision::getTimestampFromId was deprecated in MediaWiki 1.35. [Called from SMWExportController::printPages in /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/includes/export/SMW_ExportController.php at line 468]

/home/travis/build/SemanticMediaWiki/mw/includes/debug/MWDebug.php:333
/home/travis/build/SemanticMediaWiki/mw/includes/debug/MWDebug.php:265
/home/travis/build/SemanticMediaWiki/mw/includes/GlobalFunctions.php:1033
/home/travis/build/SemanticMediaWiki/mw/includes/Revision.php:1076
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/includes/export/SMW_ExportController.php:468
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/InterwikiDBIntegrationTest.php:154
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/InterwikiDBIntegrationTest.php:91
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/DatabaseTestCase.php:173
/home/travis/build/SemanticMediaWiki/mw/maintenance/doMaintenance.php:99
```

https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/jobs/673755527

mwjames:
[...]

PR #4949, opened 2021-02-28
Maintenance: replace Revision by RevisionRecord

Seb35:
This is a breaking change to follow Multi-Content Revisions from MediaWiki (i.e. a revision has multiple “slots”, each one is a content).

The previous Revision is replaced by MediaWiki\Revision\RevisionRecord, which is mostly similar (although not a subclass) except there are no static methods (newFromId, getTimestampFromId) and it must be specified in the method getContent the requested slot (MAIN, at least for now).

SMW has an almost-central management of revisions with RevisionGuard.

An upstream MediaWiki hook is replaced: NewRevisionFromEditComplete → RevisionFromEditComplete

A downstream Semantic MediaWiki hook has a modified signature: SMW::RevisionGuard::ChangeRevision( Title, ?RevisionRecord& )

Issue: #4944
Issue: #4941

---

**DRAFT - do not merge!**

> This first draft pull request replaced most of the deprecated class `Revision` and it seems to work with some smoke tests, but it remains details to be checked.

TODO from easiest to most difficult:
- [x] mock RevisionRecord: the two methods `getSize` and `getSha1` must be mocked else the class remains abstract, see Travis
- [ ] mock in ProtectionValidatorTest to fix the warnings?
- [ ] it seems there are issues in deleting the pages from the previous test case in `InterwikiDBIntegrationTest`, fix it
- [ ] `getContentHandler`: check/change the call to this method 
- [ ] `Revision::newFromId`: check/change the replacements - should it use RevisionGuard?
- [ ] should it be kept the `LogicException` added in RevisionGuard, or a fallback to the global `MediaWikiServices::getInstance()->getRevisionLookup()->getRevisionByTitle(…)`?
- [ ] `getTimestampFromId`: check/change the call to this method: the change in this commit is only to avoid the warning, but it should be assert if this use of RevisionStore (the only occurence in SMW) is the right way to do (see also #4714)
- [ ] use of downstream hook `SMW::RevisionGuard::ChangeRevision`: should a new hook be created or do we change signature of this existing one? - I only found SemanticApprovedRevs as public user of this hook in SMW + Gerrit repos
- [ ] keep the two old hooks? (depends on compatibility pre-1.35)

### DIFF
--- a/tests/phpunit/Unit/Serializers/QueryResultSerializerTest.php
+++ b/tests/phpunit/Unit/Serializers/QueryResultSerializerTest.php
@@ -202,7 +202,7 @@ class QueryResultSerializerTest extends \PHPUnit_Framework_TestCase {
 		$provider[] = [
 			[
 				'queryResult' => $this->buildMockQueryResult( $setup )
-				],
+			],
 			[
 				'printrequests' => [
 					[ 'label' => 'Foo-1', 'typeid' => '_num', 'mode' => 2, 'format' => false, 'key' => '', 'redi' => '' ],
@@ -237,7 +237,9 @@ class QueryResultSerializerTest extends \PHPUnit_Framework_TestCase {
 			] );
 
 			$printRequests[] = $printRequest;
-			$getResults[] = \SMW\DIWikiPage::newFromTitle( new \Title( NS_MAIN, $value['printRequest'] ) );
+			# TODO: Should be the following line, but the environment of the test should be adapted
+			# $getResults[] = \SMW\DIWikiPage::newFromTitle( \Title::makeTitle( NS_MAIN, $value['printRequest'] ) );
+			$getResults[] = \SMW\DIWikiPage::newFromTitle( \Title::makeTitle( NS_MAIN, '' ) );
 
 			$dataItem = $this->newMockBuilder()->newObject( 'DataItem', [
 				'getDIType' => DataItem::TYPE_NUMBER,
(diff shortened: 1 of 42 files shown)


Seb35:
I fixed most of the remaining errors in Travis, it remains, for 1.35, 2 failures and 2 warnings for 21782 assertions (there were yesterday 21 errors, 3 failures, 3 warnings for 21392 assertions), and for  1.32-1.34, 1 failure and 2 warnings. I did not really work to solve MW master issues even if I followed the counts (since yesterday: +38 assertions, -6 errors, +3 failures, -1 warning).

I incorporated in this draft PR the other PRs #4995 and #4996 to reduce the number of warnings and be able to see results on MW master, but they should be stripped out before the final merge given they are issues unrelated to RevisionRecord.

tests/phpunit/Unit/Serializers/QueryResultSerializerTest.php
```diff
@@ -237,7 +237,9 @@ private function buildMockQueryResult( $setup ) {
 			] );
 
 			$printRequests[] = $printRequest;
-			$getResults[] = \SMW\DIWikiPage::newFromTitle( new \Title( NS_MAIN, $value['printRequest'] ) );
+			# TODO: Should be the following line, but the environment of the test should be adapted
+			# $getResults[] = \SMW\DIWikiPage::newFromTitle( \Title::makeTitle( NS_MAIN, $value['printRequest'] ) );
+			$getResults[] = \SMW\DIWikiPage::newFromTitle( \Title::makeTitle( NS_MAIN, '' ) );
```
Seb35, line 242: This is an issue in the test: it is tested against an empty title. Previously it was created a Title object with `new Title()`, the constructor of Title is now private, so I used `Title::makeTitle( NS_MAIN, '' )` to continue creating it with the same behaviour, but it is a quickfix and the test should instead be adapted.

JeroenDeDauw:
Can this be closed?

Seb35:
This was fixed by other commits, whose #5046, #5049, #5051.

PR #4952, opened 2021-03-05
CachedFactbox: Avoid master DB connection on read requests

mszabo-wikia:
Currently, CachedFactbox fetches the revision ID of the current page via
RevisionGuard, which always uses the master DB, even on read requests. This
results in warnings from MediaWiki's TransactionProfiler ("Expectation
(masterConns <= 0) by MediaWiki::main not met (actual: 1)"), and can pose an
issue if the master DB is in a remote datacenter and so expensive (latency) to
access, or if it is down for maintenance or otherwise unavailable.

Instead, use OutputPage::getRevisionId to fetch the revision ID here, which
avoids this behavior.

### DIFF
--- a/src/Factbox/CachedFactbox.php
+++ b/src/Factbox/CachedFactbox.php
@@ -10,7 +10,6 @@ use SMW\Parser\InTextAnnotationParser;
 use Title;
 use Psr\Log\LoggerAwareTrait;
 use SMW\Utils\HmacSerializer;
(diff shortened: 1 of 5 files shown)


JeroenDeDauw (approved):

gesinn-it:
This PR is breaking all CI jobs, legacy as well as MW 1.35 and master

mszabo-wikia:
Yea, I still need a moment to update the tests to account for the change.

PR #4957, opened 2021-03-25
Add a protocol to relative urls in FileHandler fixes #4918

### LINKED ISSUE #4918: PHP Warning:  get_headers(): This function may only be used against URLs
### Setup

- SMW version: 3.1.6
- MW version: 1.31.6
- PHP version: 7.0.33-0ubuntu0.16.04.16
- DB system (MySQL, Blazegraph, etc.) and version: 5.7.32-0ubuntu0.16.04.1
- Elasticsearch version: 5.6.16

### Issue

This issue came to light after having found the cause of #4488. In the ElasticSearch file ingestion process, a call is made to the `get_headers()` function in the FileHandler class ([here](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/1f4bbda9bb8f7826ffabf00159cfdc0760043ca3/src/Elastic/Indexer/Attachment/FileHandler.php#L83)) using a URL that is based on `$wgServer`. 

According to [the documentation of $wgServer](https://www.mediawiki.org/wiki/Manual:$wgServer), MediaWiki supports setting it to a protocol-relative URL (e.g., `//www.mediawiki.org`). Doing this however, causes the previously mentioned call to `get_header()` to fail, since it is unable to recognize the given URL as a valid URL:

```
PHP Warning:  get_headers(): This function may only be used against URLs
```

This causes the `get_headers()` function to return `false`, which causes the file ingestion to fail. This should not happen, since a protocol-relative URL should be allowed.

Steps to reproduce the observation:

* Set `$wgServer` to a protocol-relative URL
* Enable `experimental.file.ingest` for `ElasticStore`
* Run `rebuildElasticIndex` maintenance script

octfx:
This PR is made in Reference to #4918 

Tries to add a protocol to a relative url, by checking `wgCanonicalServer` or the request through `RequestContext::getMain`. If both fail `http` is used as a fallback.


kghbln:
@octfx Thanks for providing this patch and please excuse that is has been rotting here since. 

@JeroenDeDauw Could you have a peep at this?

JeroenDeDauw:
I'm afraid to merge this without tests

octfx:
I'll try to add some when my time permits it!

octfx:
> I do not think this would work for two reasons:
> 
> * `file_get_contents` also does not understand protocol-less URIs. The URL also needs to be protocolized before it is passed to that.
> * In the function `protocolizeUrl`, you return the unaltered canonical URL if it has a scheme. This means any path or query that was passed to the function would be lost.
> 
> I believe just removing lines 136 to 145 (the `try-catch` block) would fix the second issue. The first issue may be fixed by assigning the protocolized URL to a variable and passing that variable to both `file_get_contents` as well as `get_headers`.

Oh, the `file_get_contents` part is a good catch, that slipped through!

I've updated the try-catch block to add the canonical server protocol to the input url.

Further I've added some tests, but I couldn't get them to run locally as it always errored out with `Use of undefined constant SMW_PHPUNIT_AUTOLOADER_FILE [...] boostrap.php on line 15` :/

JeroenDeDauw:
@Seb35 does this look good to you?

marijnvanwezel:
@JeroenDeDauw Could you take a look at this again when you have the time? We are facing this issue again.

JeroenDeDauw:
@marijnvanwezel did you test that this patch fixes the issue?

alistair3149:
Thanks for the patch, @octfx, and apologies it sat open this long.

This is now obsolete. The problem in #4918 (`get_headers(): This function may only be used against URLs` when `$wgServer` is protocol-relative) lived in `FileHandler::fetchContentFromURL()`, which the Elastic file ingest used to fetch attachment content over HTTP.

PR #6138 (commit 2a14189dd, merged August 2025) changed `FileIndexer` to read attachment bytes directly from the MediaWiki file backend via `fetchContentFromFile()` instead of fetching over HTTP. As a result `fetchContentFromURL()`, along with its `get_headers()` and `file_get_contents()` calls, is no longer invoked during ingest, so a protocol-relative URL never reaches PHP's URL functions. The `getFullURL()` value is now only stored as inert `file_path` metadata on the Elasticsearch document and is never dereferenced.

Since the code path this PR patches is no longer reached (and the branch has drifted out of mergeability against the refactored `FileHandler`), I'm closing as obsolete in favour of #6138.

Note: `fetchContentFromURL()` itself still exists with the original implementation. It is currently dead code, but if it is ever wired back into a live path it would want a fix along these lines, which can be handled separately.

Issue #4945, opened 2021-02-23
MW 1.35: Use of Parser::__construct was deprecated in MediaWiki 1.34.


gesinn-it:
6) SMW\Tests\Integration\Parser\InTextAnnotationParserTemplateTransclusionTest::testPreprocessTemplateAndParse with data set #0 (0, array(array(true), false, 8, 'hash'), '[[Foo::{{Bam}}]]', '?bar', array('[[:?bar|?bar]]', 1, array('Foo'), array('?bar')))

Use of Parser::__construct was deprecated in MediaWiki 1.34. [Called from SMW\Tests\Integration\Parser\InTextAnnotationParserTemplateTransclusionTest::runTemplateTransclusion in /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Integration/Parser/InTextAnnotationParserTemplateTransclusionTest.php at line 58]

https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/jobs/760157304#L2071

Seb35:
Fixed by e56edac5e, this issue can be closed.

Nikerabbit:
Thanks!

Issue #4940, opened 2021-02-23
MW 1.35: CsvFileExportPrinterTest::testLink - Use of $wgParser was deprecated in MediaWiki 1.32.


gesinn-it:
1) SMW\Tests\Query\ResultPrinters\CsvFileExportPrinterTest::testLink

Use of $wgParser was deprecated in MediaWiki 1.32. [Called from SMW\Query\ResultPrinters\ResultPrinter::handleNonFileResult in /home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/src/Query/ResultPrinters/ResultPrinter.php at line 379]

https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/jobs/760157304#L1973

Seb35:
Fixed by e56edac5e, this issue can be closed.

kghbln:
Thank you!

PR #4996, opened 2021-05-24
Use the services factory to create a parser object

Seb35:
Direct uses of global $wgParser is deprecated in MW 1.35. This removes the three last occurrences of it.

It is backward-compatible thanks to the services factory. It could be deleted the 1.34- compatibility by removing the line in /src/Services/mediawiki.php, but I think it should be done in a commit which clearly removes this compatibility (extension.json, CI).

Fixes: #4940
Fixes: #4945

### DIFF
--- a/src/Parser/RecursiveTextProcessor.php
+++ b/src/Parser/RecursiveTextProcessor.php
@@ -5,6 +5,7 @@ namespace SMW\Parser;
 use Parser;
 use ParserOptions;
 use RuntimeException;
+use SMW\ApplicationFactory;
 use SMW\Localizer;
 use SMW\ParserData;
 use SMW\MediaWiki\Template\TemplateExpander;
@@ -71,7 +72,7 @@ class RecursiveTextProcessor {
 		$this->parser = $parser;
 
 		if ( $this->parser === null ) {
-			$this->parser = $GLOBALS['wgParser'];
+			$this->parser = ApplicationFactory::getInstance()->create( 'Parser' );
 		}
 	}
 
--- a/src/Query/ResultPrinters/FeedExportPrinter.php
+++ b/src/Query/ResultPrinters/FeedExportPrinter.php
@@ -5,6 +5,7 @@ namespace SMW\Query\ResultPrinters;
 use FeedItem;
 use ParserOptions;
 use Sanitizer;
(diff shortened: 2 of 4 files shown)


src/Parser/RecursiveTextProcessor.php
```diff
@@ -71,7 +72,7 @@ public function __construct( ?Parser $parser = null ) {
 		$this->parser = $parser;
 
 		if ( $this->parser === null ) {
-			$this->parser = $GLOBALS['wgParser'];
+			$this->parser = ApplicationFactory::getInstance()->create( 'Parser' );
```
JeroenDeDauw, line 75: This seems like it creates a new parser. If the code that uses this value does mutation, then this change could break things.

Seb35: Indeed, but it can also be expected less side-effects (if any) due to unexpected residual state in the parser between different parses; and it be set a parser if desired in some object RecursiveTextProcessor.

I will test in the following days with `@annotation` queries, which are the more extreme SMW+parser subtleties to my knowledge, but I have no idea how to test more.

Also, there are no differences between the [Travis build of the latest merged commit](https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/builds/771993852) and [this one](https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/builds/772165201), except for MW 1.35 where there are two errors removed in this build.

Nikerabbit: If it is important to use the global parser, you can use `MediaWikiServices::getParser()` (available since 1.29).

Aside, where on earth is this ApplicationFactory class defined? I spend like 15 minutes looking for it, and could not find it.

JeroenDeDauw: Sounds frustrating @Nikerabbit. Do you not have class-based search in your editor?

`class_alias( \SMW\Services\ServicesFactory::class, '\SMW\ApplicationFactory' );` in `Aliases.php`. Probably references in SMW itself should be updated if a class is moved...

Seb35:
Fixed by e56edac5e, which is similar to this PR but uses `MediaWikiServices::getParser()`.

Issue #4997, opened 2021-05-26
Collation issue reported by Travis on triplestores


Seb35:
### Setup

- SMW version: 4.0.0-alpha (uncovered in 02b990260, but could be earlier, still present in current master)
- MW version: 1.32-1.35
- PHP version: 7.1-7.3
- DB system (MySQL, Blazegraph, etc.) and version: multiple triplestores (Fuseki, Sesame, Blazegraph)

### Issue

On Travis builds, since 4 May 2021 when Travis builds were fixed in #4984, there are now permanently a failure in the test f-309.json about the order of results specified by the collation uca-default-u-kn (the ICU default multilingual collation with numeric sorting).

The failure description is below, here specifically extracted from [the first failing Travis build  14722.3 (Fuseki triplestore)](https://travis-ci.org/github/SemanticMediaWiki/SemanticMediaWiki/builds/769420224)
```
1) SMW\Tests\Integration\JSONScript\JSONScriptTestCaseRunnerTest::testCaseFile with data set "f-0309.json" ('/home/travis/build/SemanticMe...9.json')
Failed "#0 (asc sort, expected by DEFAULTSORT: 3 (T1.1), 2 (T1.2), 1 (T1.3), B (T1.10), A (T1.11))"

==== (actual) ====
<div class="mw-parser-output"><div class="smw-columnlist-container" dir="ltr"><div class="smw-column" style="width:100%;" dir="ltr"><div class="smw-column-header">T</div><ul><li>F0309/1 </li><li>F0309/2 </li><li>F0309/3 </li><li>F0309/A </li><li>F0309/B </li></ul></div> 
<br style="clear: both;" /></div>



</div>
==== (StringContains expected) ====
[ <ul><li>F0309/3 </li><li>F0309/2 </li><li>F0309/1 </li><li>F0309/B </li><li>F0309/A </li></ul> ]

Failed asserting that 0 matches expected 1.

/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/StringValidator.php:84
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/Validators/StringValidator.php:33
/home/travis/build/SemanticMediaWiki/mw/extensions/SemanticMediaWiki/tests/phpunit/Utils/JSONScript/QueryTestCaseProcessor.php:246
[...]

PR #4998, opened 2021-05-26
Alternate armouring for the sort key for triplestores

Seb35:
This needs an execution of /maintenance/updateEntityCollation.php for existing installations using a triplestore.

This is a first draft to verify that Travis builds.

Closes: #4997

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,8 +22,8 @@ jobs:
       php: 7.4
     #- env: DB=sqlite; MW=REL1_35; PHPUNIT=8.5.*; SESAME=2.8.7
     #  php: 7.4
-    #- env: DB=sqlite; MW=REL1_35; PHPUNIT=8.5.*; BLAZEGRAPH=1.5.2
-    #  php: 7.4
+    - env: DB=sqlite; MW=REL1_35; PHPUNIT=8.5.*; BLAZEGRAPH=1.5.2
+      php: 7.4
     #- env: DB=sqlite; MW=REL1_35; PHPUNIT=8.5.*; VIRTUOSO=6.1
     #  php: 7.4
     - env: DB=sqlite; MW=REL1_35; PHPUNIT=8.5.*; TYPE=benchmark
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -275,15 +275,7 @@ return [
 	# - SMW_SPARQL_QF_SUBC to resolve subcategories
 	#
 	# - SMW_SPARQL_QF_COLLATION allows to add support for the sorting collation as
-	#   maintained in $smwgEntityCollation. It is not enabled by default as the
-	#   `uca-*` collation generates a UTF-8 string that contains unrecognized
-	#   UTF codepoints that may not be understood by the back-end hence the
-	#   Collator prevents and armors those unrecognized characters by replacing
-	#   them with a ? to avoid a cURL communication failure but of course this
-	#   means that not all elements of the sort string can be transfered to the
-	#   back-end and can therefore cause a sorting distortion for close matches
-	#   as in case of for example "Ennis, Ennis Hill, Ennis Jones, Ennis-Hill,
-	#   Ennis-London"
+	#   maintained in $smwgEntityCollation.
 	#
 	# - SMW_SPARQL_QF_NOCASE to support case insensitive pattern matches
 	#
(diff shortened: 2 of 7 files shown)


Seb35:
The Travis jobs `3` and `9` testing with Fuseki and Blazegraph passes as expected; the `4` testing with Sesame has no more this failure but still faills because of one existing error.

kghbln:
Due to the requirement of executing "updateEntityCollation.php" per https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4997#issuecomment-848533813 we also need an entry in the release notes. I will be great if you could add it to this very pull.

Seb35:
Indeed for the release notes. It has compability implications since it recomputes all values of the `swivt:sort` SPARQL property with a very different scheme compared to current one; at the same time there is [a warning in PHP documentation](https://www.php.net/manual/en/collator.getsortkey.php) that sort keys could change accross ICU library versions (I’m not sure people recompute the sort keys after an upgrade (I just discover this myself), but it could in theory lead to issues).

----

About the code itself: For the encoding of the sort key, I used bin2hex/unpack('H*') because it is easy but it uses 2 times more space compared to original binary string (8 bits used for each 4 bits). Advantages are it is fast and, importantly, it preserves the relative order of bits (so that comparing bit-by-bit in the armoured string is identical to comparing bit-by-bit in the original string).

It could be used other encodings to save space, but it is very important to keep “relative bits order” property, else comparing armoured strings would lead to a different (falsy) collation. In the choices there are [base32hex](https://en.wikipedia.org/wiki/Base32#base32hex) (5 bits converted to 8 bits) or some dialect of base64 (6 bits converted to 8 bits, but probably very nonstandard since classical base64 do not preserve the relative bits order (letters and figures are inverted)). In any case, the implementation should be sufficiently fast ([it was asked on SO for base32](https://stackoverflow.com/questions/16416802/effective-solution-base32-encoding-in-php) with some answers).

Seb35:
I tried various encodings and implementations:
* `unpack( 'H*', $text )` and `bin2hex( $text )`: standard PHP converting a binary string to an hexadecimal string, e.g. `bin2hex( 'a' ) == '61'`
* `base64_encode`: standard PHP converting a binary string to a base64-encoded string, e.g. `base64_encode( 'a' ) == 'YQ=='`
* `convert_uuencode`: standard PHP converting a binary string to [uuencode](https://en.wikipedia.org/wiki/Uuencoding)-encoded string, e.g. ```convert_uuencode( 'a' ) == '!80``\n`\n'``` (where \n is the character 0x0a)
* base32hex: [base32hex](https://en.wikipedia.org/wiki/Base32#base32hex) (defined in [RFC  4648](https://datatracker.ietf.org/doc/html/rfc4648)) with the [implementation here](https://github.com/bbars/utils/blob/master/php-base32-encode-decode/Base32.php), it converts a binary string to base32-encoded string containing figures [0-9a-v], e.g. `Base32hex::encode( 'a' ) == 'c4'`
* base64hex: does not really exist but is almost similar to base64 with the same principles as base32hex: 6 bits converted to 8 bits ASCII letters [0-9A-Za-z{}] without padding at the end, in order to preserve bitwise sort order

The program testing performance of these algorithms is [here](https://github.com/Seb35/bbars-utils/blob/perf-tests/php-base32-encode-decode/base32hex.php) (dirty draft, see mainly the conclusions at the end of the file).

**As a conclusion,** I find base64 with modified figures [0-9A-Za-z{}] (instead of A-Za-z0-9+/]) without end padding gives the better performance/space alternative:
* sort keys are 33% bigger than their original binary form and 20-30% bigger than the current form,
* performance is 3 times better than current implementation.

Seb35:
@mwjames @JeroenDeDauw: this PR is finalised on my side, could you review it?

Summary:
* when the sort of the results is outsourced to the triplestore, I replaced the previous lossy algorithm by a losseless one, based on a modified Base64 with the property that a bitwise comparison in triplestore is equivalent to a comparison with the UCA collation (see [this comment](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4998#issuecomment-855279017) for details),
* consequently, (almost) all tests pass with Blazegraph and Fuseki,
* reverted 6e3b2ea30e4d525641733f17c93b55545d507008,
* skipped two tests for Blazegraph, because the testing mechanics does not see that the underlying SQLStore is sqlite ([this commit works when the SQLStore is mysql](https://app.travis-ci.com/github/SemanticMediaWiki/SemanticMediaWiki/jobs/527454315)).

Issue #4989, opened 2021-05-12
SMW\MediaWiki\Connection\Database::query() always uses the master DB


mszabo-wikia:
### Setup

- SMW version: 3.1
- MW version: 1.33
- PHP version: 7.3.27
- DB system (MySQL, Blazegraph, etc.) and version: MySQL 5.7.25

### Issue
The `SMW\MediaWiki\Connection\Database::query()` method, often used in conjunction with `SMW\MediaWiki\Connection\Query` to issue read (SELECT) queries, always uses the master DB, even when replica DBs are available. Ideally, this method should use replicas when possible to allow splitting the load between multiple instances.

Sample stacktrace of such a query that would be okay to run on a replica but ends up on the master:

```
#0 /extensions/fandom/FandomIncludes/src/Database/SamplingClusterQueryLogger.php(89): Fandom\Includes\Database\SamplingClusterQueryLogger->getContext(array)
	#1 /includes/libs/rdbms/database/Database.php(1361): Fandom\Includes\Database\SamplingClusterQueryLogger->debug(string, array)
	#2 /includes/libs/rdbms/database/Database.php(1237): Wikimedia\Rdbms\Database->attemptQuery(string, string, boolean, string)
	#3 /includes/libs/rdbms/database/DBConnRef.php(53): Wikimedia\Rdbms\Database->query(string, string, integer)
	#4 /includes/libs/rdbms/database/DBConnRef.php(274): Wikimedia\Rdbms\DBConnRef->__call(string, array)
	#5 /extensions/SemanticMediaWiki/src/MediaWiki/Connection/Database.php(364): Wikimedia\Rdbms\DBConnRef->query(string, string, boolean)
	#6 /extensions/SemanticMediaWiki/src/SQLStore/EntityStore/SemanticDataLookup.php(524): SMW\MediaWiki\Connection\Database->query(string, string)
	#7 /extensions/SemanticMediaWiki/src/SQLStore/EntityStore/SemanticDataLookup.php(268): SMW\SQLStore\EntityStore\SemanticDataLookup->fetchFromTable(SMW\MediaWiki\Connection\Query, SMW\SQLStore\PropertyTableDefinition, boolean, SMW\RequestOptions)
[...]

PR #5002, opened 2021-05-31
Use replica DBs for high-volume queries, refs #4989

mszabo-wikia:
Introduce a new method readQuery() on SMW's Database helper class that uses a
replica DB for executing the query, and use it in
SemanticDataLookup::fetchFromTable, which is responsible for a large amount of
queries. This allows for other call sites to be gradually migrated to the new
method without breaking callers that might rely on query() using the primary DB.

### DIFF
--- a/src/MediaWiki/Connection/Database.php
+++ b/src/MediaWiki/Connection/Database.php
@@ -9,6 +9,7 @@ use RuntimeException;
 use SMW\ApplicationFactory;
 use SMW\Connection\ConnRef;
 use UnexpectedValueException;
+use Wikimedia\Rdbms\IDatabase;
 
 /**
  * This adapter class covers MW DB specific operations. Changes to the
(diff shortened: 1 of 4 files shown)


JeroenDeDauw (approved):

src/MediaWiki/Connection/Database.php
```diff
@@ -320,10 +323,52 @@ public function select( $tableName, $fields, $conditions, $fname, array $options
...
 		$this->transactionHandler->muteTransactionProfiler( true );
 
+		$results = $this->executeQuery(
+			$this->connRef->getConnection( 'write' ),
+			$sql,
+			$fname,
+			$ignoreException
+		);
+
+		$this->transactionHandler->muteTransactionProfiler( false );
```
mszabo-wikia, line 335: As a side-effect, this infrastructure can go away as well once we remove the use of master connections from read requests :)

PR #5016, opened 2021-07-05
Replace wgParser

hexmode:
This fixes [a problem with REL1_35](https://app.travis-ci.com/github/SemanticMediaWiki/SemanticMediaWiki/jobs/521821756#L1739) where it complained about wgParser.

### DIFF
--- a/src/MediaWiki/Hooks.php
+++ b/src/MediaWiki/Hooks.php
@@ -301,7 +301,7 @@ class Hooks {
 			'ArticleViewHeader' => [ $this, 'onArticleViewHeader' ],
 			'ContentHandlerForModelID' => [ $this, 'onContentHandlerForModelID' ],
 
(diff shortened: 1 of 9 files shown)


src/MediaWiki/Hooks.php
```diff
@@ -342,6 +341,11 @@ private function registerHandlers() {
 			'AdminLinks' => [ $this, 'onAdminLinks' ],
 			'PageSchemasRegisterHandlers' => [ $this, 'onPageSchemasRegisterHandlers' ]
 		];
+		if ( version_compare( MW_VERSION, '1.35', '<' ) ) {
```
JeroenDeDauw, line 344: Remember that we do not need to support MW older than 1.35 on master

hexmode: thanks, fixing.

PR #5020, opened 2021-07-12
Replace ResourceLoaderTestModules hook with QUnitTestModule

Abijeet:
Use QUnitTestModule in extensions.json if MW > 1.33


JeroenDeDauw:
No need to support MW < 1.35

PR #5022, opened 2021-07-12
Run tests on Ubuntu 20.04 focal

JeroenDeDauw:


JeroenDeDauw:
Apparently this does not allow running with PHP 7.3

Issue #4990, opened 2021-05-14
Exception on rebuildElasticIndex.php


Seb35:
### Setup

- SMW version: 3.1.3
- MW version: 1.33
- PHP version: 7.3.23
- DB version: MariaDB 10.2.36
- Elasticsearch version: 6.5.4

### Issue

When I setup Elasticsearch as query store as documented on [ElasticStore](https://www.semantic-mediawiki.org/wiki/Help:ElasticStore) and execute `extensions/SemanticMediaWiki/maintenance/rebuildElasticIndex.php`, I obtain the following exception:
```
Rebuilding indices ...
   ... creating required indices and aliases ...
   ... updating document                           100% (8471606/8471622)
   ... updating settings and mappings ...
   ... *data* index ... [dd116763e3bb38c2e12ebf64] [no req]   Elasticsearch\Common\Exceptions\NoNodesAvailableException from line 53 of /tmp/mediawiki/vendor/elasticsearch/elasticsearch/src/Elasticsearch/ConnectionPool/StaticNoPingConnectionPool.php: No alive nodes found in your cluster
Backtrace:
#0 /tmp/mediawiki/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Transport.php(78): Elasticsearch\ConnectionPool\StaticNoPingConnectionPool->nextConnection()
#1 /tmp/mediawiki/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Transport.php(96): Elasticsearch\Transport->getConnection()
#2 /tmp/mediawiki/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Namespaces/BooleanRequestWrapper.php(40): Elasticsearch\Transport->performRequest(string, string, array, NULL, array)
#3 /tmp/mediawiki/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Namespaces/IndicesNamespace.php(40): Elasticsearch\Namespaces\BooleanRequestWrapper::performRequest(Elasticsearch\Endpoints\Indices\Exists, Elasticsearch\Transport)
#4 /tmp/mediawiki/extensions/SemanticMediaWiki/src/Elastic/Indexer/Rollover.php(56): Elasticsearch\Namespaces\IndicesNamespace->exists(array)
[...]

PR #5034, opened 2021-08-03
Fix interface to ElasticStore parameters

Seb35:
In SMW\Elastic\Connection\ConnectionProvider::getConnections(), the keys $k in $params [[1]](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Elastic/Connection/ConnectionProvider.php#L71-L87) are used to create method names "set$k" in Elasticsearch\ClientBuilder::fromConfig() [[2]](https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/ClientBuilder.php#L200).
The connection parameters are set with setConnectionParams, this is this fix (see use in Elasticsearch\Connections\Connection, e.g. [[3]](https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Connections/Connection.php#L290)).

This issue can be seen by setting, in `SMW\Elastic\Connection\ConnectionProvider::getConnection()`, the second parameter of `ClientBuilder::fromConfig( $params, $quiet = true )` to `$quiet = false` like in the following diff: an exception is returned.
```diff
@@ -90 +90 @@ class ConnectionProvider implements IConnectionProvider {
-                       $clientBuilder = ClientBuilder::fromConfig( $params, true );
+                       $clientBuilder = ClientBuilder::fromConfig( $params, false );
```

### DIFF
--- a/src/Elastic/Connection/ConnectionProvider.php
+++ b/src/Elastic/Connection/ConnectionProvider.php
@@ -72,14 +72,16 @@ class ConnectionProvider implements IConnectionProvider {
 			'hosts' => $endpoints,
 			'retries' => $this->config->dotGet( 'connection.retries', 1 ),
 
-			'client' => [
+			'connectionParams' => [
+				'client' => [
 
-				// controls the request timeout
-				'timeout' => $this->config->dotGet( 'connection.timeout', 30 ),
+					// controls the request timeout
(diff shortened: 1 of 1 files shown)


Seb35:
I experienced this in https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4990#issuecomment-851589670 when I tried to activate ElasticStore and I had to tune the timeouts.

Issue #4988, opened 2021-05-07
Chained properties not displayed when occuring multiple times


rubenvandeven:
Hi there,
I've an issue whic

rubenvandeven:
This is the debug output of the query. If needed:  
![debug_error](https://user-images.githubusercontent.com/677296/117422284-f59c5b80-af0e-11eb-84b5-daabf24eeadc.png)

smw-user:
See: https://sandbox.semantic-mediawiki.org/wiki/Issue/4988.

edwardsnj:
Does anyone have any insight on this? This is blocker for my SMW site. Is there a way to manipulate the #ask: to eliminate this issue? Based on my exploration, I think the issue occurs when a different primary value has a repeated chained property value that occurs a second (or more) time with some other value in-between. It kinda looks like there is some caching of the chained value being attempted and failing due to the interleaving (different value).

edwardsnj:
I just checked the database query (SQLStore) itself (as shown by the debug option on semantic search), and as far as I can tell, the chained properties logic is implemented at some highly level (as separate/distinct queries?).

smw-user:
I don't want to be impatient but any progress so far?
Since this is a fairly new bug - at least on the wiki where I contribute the bug was introduced with an upgrade from MW 1.34.4 to 1.35.1 and SMW 3.1.6 to 3.2.2 - we have been using queries with property chains there quite frequently. Suddenly we have to deal with missing properties ruining queries and are forced to rewrite queries/templates (calling #show!? in each result of the result format templates instead of having one nice property chain in the query) to provide the same user experience.

gesinn-it:
This indeed looks like a bug. As a workaround you could try to query your subobjects and then query the details using result format "template" (plainlist)

tafinho:
Same here.

Found a workaround by sorting the query by one of the chained property.

Seb35:
The issue comes from #4572: when it is reverted, all results are displayed. The right resolution remains to be find with a middle ground between the old behaviour and the new one without this side effect.

Seb35:
More exactly the only thing to do is to comment, in `src/SQLStore/EntityStore/PrefetchCache.php`, function `prefetch` the instruction `return;` in the middle:
```diff
diff --git src/SQLStore/EntityStore/PrefetchCache.php src/SQLStore/EntityStore/PrefetchCache.php
index d928a2234..e22ae16a4 100644
[...]

PR #5036, opened 2021-08-03
Optimise second-chained properties results, refs 3722

Seb35:
This PR is made in reference to: #3722

* When a chained result is requested, there is already an optimisation for the first-chained property with a bulk request of this property accross all results (#4572); this optimises further the second-chained property (then third, etc) with a bulk request of this property accross all results (results from the first-chained property results).
* solves #4988

The “prefetch cache” operation is now global: all prefetching is done at the first execution of `PrefetchCache::prefetch()` for all lines of the result, instead of one prefetching on each line.

The real cache (PrefetchCache->cache) have keys "PropertyName" or "-PropertyName" and the values are arrays [ "semantic object ID" => [ list of WikiPages ] ] ; so that it is easy to pick values in PrefetchCache::getPropertyValues(). (I say WikiPages here and after, but it could be other data types.)

The second cache (PrefetchCache->lookupCache) is aligned with the first one; its keys are the whole property chain like "Prop1.-Prop2.Prop3…PropN-1" for the N-level chain (so that it is empty for first-level chain) and the values are the WikiPages of the result for this level N; this is mainly used to be able to compute the next level starting from this result.

Earlier versions of this PR failed on tests p-0434.json and p-0467.json due to the vicious repetition of a same property in the chain, which why the whole property chain must be cached somewhere (here in lookupCache) and not only the property.

On a request #ask with 999 results and two 2-levels chain properties (with no common property), it took 18.4 seconds before this patch and 3.7 seconds with this patch (mean over a sample of 5 reloads of the page; it is not impossible other factors influence the performance like MySQL internal cache, but the figures are anyway quite different). This was on SMW 3.2.2 + MW 1.33 + ElasticStore + PHP 7.3.

This PR includes:
- [x] CI build passed

### DIFF
--- a/src/Query/Result/FieldItemFinder.php
+++ b/src/Query/Result/FieldItemFinder.php
@@ -136,7 +136,6 @@ class FieldItemFinder {
 		if ( $this->printRequest->isMode( PrintRequest::PRINT_CATS ) ) {
 			$options = new RequestOptions();
 			$options->isChain = false;
-			$options->isFirstChain = false;
 
 			// Rely on the prefetch
 			self::$catCache = $this->itemFetcher->fetch(
@@ -323,7 +322,6 @@ class FieldItemFinder {
 		}
 
 		$requestOptions->isChain = false;
-		$requestOptions->isFirstChain = false;
 
 		// If it is a chain then try to find a connected DIWikiPage subject that
 		// matches the property on the chained PrintRequest.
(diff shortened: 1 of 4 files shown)


Seb35:
@mwjames: this is the logical continuation of #4572, could you review it?

Seb35:
I will add a test to catch issue 4988, probably by adding a new Ask request in p-0434.json or p-0467.json.

Seb35:
I added the parser test p-0468.json, but I’m not sure about the exact naming logic. I copied q-0601.json and adapted it to trigger issue #4988, which fails as expected before this PR.

gesinn-it-gea:
I did a local test: looks good to me. @JeroenDeDauw can you please have a quick.

jaideraf:
> I did a local test: looks good to me.

@JeroenDeDauw could this be merged? Can we run the current GitHub automated tests here? I don't understant how to do it (or I don't have the permissions to to it here by the UI.

JeroenDeDauw:
I also do not know how to trigger the tests in this case. @Seb35 could do a rebase against master and force-push that, which should do the trick. Or we could do this ourselves and make a new PR.

gesinn-it-gea:
this definitely requires a rebase to get the latest test framework. Then you can locally simply run "make ci"

jaideraf:
> I will rework this PR in medium priority, perhaps in 2-3 months (I’m interested to see the tests passing in the short term).

@Seb35 the tests are passing now (there is only failures, but i think they do not matter to this subject). If you can, take a look at https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5623#issuecomment-2618343570 also, please.

JeroenDeDauw:
Needs rebase

hkwi:
I [cherry-picked](https://github.com/hkwi/SemanticMediaWiki/tree/pr-5036-testcase-rebase) only the test case from PR #5036 onto the current master and ran it locally in Docker. 

```
JSONScript Test Case Runner (SMW\Tests\Integration\JSONScript\JSONScriptTestCaseRunner)
 ✔ Case file with data set "p-0468.json" [2606.15 ms]

Time: 00:03.247, Memory: 151.00 MB

OK (1 test, 2 assertions)
```

So it looks the test case added by PR #5036 already passes on the current master.

jaideraf:
I believe this has been fixed with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5623 (SMW 7.0, currently master), so I am closing this now. Thank you.

PR #5039, opened 2021-08-06
Update tests so they work when used with another extension (e.g. SAR)

hexmode:
Since I was working on updating SAR so that tests would run in a typical way within the MW environment, I discovered and fixed a few things here.


JeroenDeDauw:
What is the deal with the `StoreTest.php` removal?

hexmode:
Consolidating the `SMW\Test` namespace with the `SMW\Tests` resulted in a conflict of the two StoreTest.php instances.

I took the lazy way out and deleted the older test.

Hopefully, I'll be able to reconcile this problem and maintain all the tests tomorrow.

hexmode:
If you can see this, here are the failures: https://github.com/hexmode/SemanticApprovedRevs/runs/3276499618

hexmode:
Yay! the SAR test succeeded as well as these!

PR #5066, opened 2021-09-13
PageDeleter: Pass UserIdentity when calling doDeleteArticleReal 

Abijeet:
This parameter is mandatory for MW 1.37


Abijeet:
This should be merged after #5065

Issue #5055, opened 2021-08-27
ParserCache fails to serialize due to smwdata on MW 1.36


Ostrzyciel:
### Setup

- SMW version: [master](https://github.com/SemanticMediaWiki/SemanticMediaWiki/tree/e20217049bfe5f7119920fdb6e372b3a2bd24c70)
- MW version: 1.36.1 (I know it's unsupported yet)
- PHP version: 7.4.22
- DB system and version: MariaDB 10.6.3, Apache Jena

### Issue

Found this in the error log when saving a page :
![image](https://user-images.githubusercontent.com/56843159/131168385-d1675f2e-9edb-4109-be13-fe72e3828a34.png)
(I am deeply sorry for the awkward formatting, I am still figuring out Logstash, I hope it's readable.)

The reason for this is that MW in 1.36 by default serializes parser cache as JSON, not PHP, which is something that the additional `smwdata` prop apparently does not support. It can be temporarily mitigated by setting [$wgParserCacheUseJson](https://www.mediawiki.org/wiki/Manual:$wgParserCacheUseJson) to false, but that setting is deprecated. I think making `SMW\SemanticData` JSON-serializable will fix that, though I have no idea how hard that might be.

Seb35:
It might be not that hard although it could be some work: the [algorithm to detect serializable objects](https://github.com/wikimedia/mediawiki/blob/04820b38cd9503ab2f6e2eee2750db89d11e948e/includes/json/JsonCodec.php#L164) wants the object implements [MediaWiki\Json\JsonUnserializable](https://github.com/wikimedia/mediawiki/blob/04820b38cd9503ab2f6e2eee2750db89d11e948e/includes/json/JsonUnserializable.php) (the other serializable objects are stdClass, scalars, or arrays of these things).

Seb35:
Well, the interface [MediaWiki\Json\JsonUnserializable](https://github.com/wikimedia/mediawiki/blob/04820b38cd9503ab2f6e2eee2750db89d11e948e/includes/json/JsonUnserializable.php) does not exist in MW 1.35 and is mandatory in 1.36 to avoid this warning, so it introduces a hard break between 1.35 and 1.36: either `SMW\SemanticData` does not implement this interface (compatibility ≤ 1.35 and we have this warning in ≥ 1.36), either it does implement this interface and it breaks in MW ≤ 1.35 because the interface does not exist.

[...]

PR #4874, opened 2020-10-27
Do not write objects into ExtensionData

cicalese:
See #4868.

This is still a work in progress. It needs tests.

PR #5072, opened 2021-09-16
Implements JsonUnserializable on SMW\SemanticData

Seb35:
To avoid the warning “Non-unserializable property set at $.ExtensionData.smwdata” on MediaWiki 1.36.

Closes: #5055

Notes:

1) **Stub interface for 1.35**: I added a stub interface MediaWiki\Json\JsonUnserializable if it does not exist (MW 1.35-) so that classes can implement it unconditionnally, but it is unused in MW 1.35-. It is loaded by Composer’s classmap. Given MediaWiki’s autoloader is loaded before Composer’s autoloader, the real interface from MW 1.36 is found before the stub interface from SMW. The stub interface can be removed when SMW will only support MW 1.36+.

2) **Isomorphic with PHP serialization**: I put in this serialization all properties of objects to stay isomorphic with previous (PHP) serialization. Possibly some properties could be removed if they are not needed in this serialization (currently used for ParserOutput, but I don’t know if it will be used elsewhere in the future).

3) **Issue in MW unserialization of objects**: If you observe the object ParserOutput obtained from cache, you will be surprised that it is the JSON array and not an object SMW\SemanticData. This comes from MediaWiki, precisely I5785596d68e ([link](https://gerrit.wikimedia.org/r/c/mediawiki/core/+/639609)). I put the remark on the Gerrit change, and I will open a Phabricator task.

4) **Alternative PR**: Just when I finished this PR, I remembered there is already a PR from Cindy #4874. It is an alternative to this PR, perhaps/probably in an easier way.

For reviewers:

1) JsonUnserializer->unserializeArray() and JsonUnserializer->unserialize() are not recursive, that’s why there is a loop in SemanticData::newFromJsonArray() for mPropVals. I added a precision in the definition of mPropVals that it is a two-dimensional array.

2) JsonUnserializer->unserialize() does not accept null as parameter, so that are some operators ? :

### DIFF
--- a/includes/SemanticData.php
+++ b/includes/SemanticData.php
@@ -2,6 +2,8 @@
 
 namespace SMW;
 
+use MediaWiki\Json\JsonUnserializable;
+use MediaWiki\Json\JsonUnserializer;
 use SMW\DataModel\SubSemanticData;
 use SMW\DataModel\SequenceMap;
 use SMW\Exception\SubSemanticDataException;
@@ -30,7 +32,7 @@ use SMWDIContainer;
  * @author Markus Krötzsch
  * @author Jeroen De Dauw
  */
-class SemanticData {
+class SemanticData implements JsonUnserializable {
 
 	/**
 	 * Returns the last modified timestamp the data were stored to the Store or
@@ -74,7 +76,7 @@ class SemanticData {
 	 * Array mapping property keys (string) to arrays of SMWDataItem
 	 * objects.
 	 *
-	 * @var SMWDataItem[]
+	 * @var SMWDataItem[][]
 	 */
 	protected $mPropVals = [];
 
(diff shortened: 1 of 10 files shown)


Seb35:
I like the spirit of the PR #4874 of @cicalese to use simple arrays instead of the more complex serialization that I did in this PR, I’m not sure what is the best way. Thoughts?

Seb35:
Side note: I opened the task [T291244](https://phabricator.wikimedia.org/T291244) about unserialization of objects in MW 1.36+ and [a Gerrit change](https://gerrit.wikimedia.org/r/c/mediawiki/core/+/721772/) is proposed.

JeroenDeDauw:
@Seb35 the new code contains a bunch of non-strict checks, ie `if ( $var )` or `$var ? :`. Better to use explicit checks for null, isset or empty string/array, as appropriate. Else the string "0" will end of evaluating to false.

Some tests for the new code also would also increase confidence in it doing what is intended.

JeroenDeDauw:
@Seb35 apologies for the slowest review ever. Your PR is much appreciated!

PR #5075, opened 2021-09-21
PropertyPage: Pass user making the request to fetch time offset

Abijeet:
Not sure if it makes sense to fetch the article owners time offset

Also $this->getPage()->getUser() returns an int, so
hasLocalTimeOffsetPreference would always end up using $wgUser


Abijeet:
Merge after: #5074

PR #5076, opened 2021-09-21
Replace fallbacks for MediaWiki\Permissions\PermissionManager

Abijeet:
Class itself is available since MW 1.33 and methods that we use
are available since 1.34


Abijeet:
Need to fix the test cases.

Abijeet:
Done, ready to be reviewed and merged.

PR #5087, opened 2021-10-04
Always pass user when creating instance of ParserOptions

Abijeet:
Fixes:
```
[2021-10-03T04:12:01.139474+00:00] error.WARNING: [abd1405ad654229e2a932430] /wiki/Special:Ask/-5B-5BSummary:%2B-5D-5D-20-5B-5BStatus::open-5D-5D-20-5B-5BProject::MediaWiki-5D-5D/-3FThread/mainlabel%3D-2D/limit%3D100/offset%3D100/format%3Dol/default%3DNo-20open-20support-20requests.-20This-20is-20good!   ErrorException: ParserOptions being created without a UserIdentity object [Called from SMW\Parser\RecursiveTextProcessor::recursiveTagParse] {"exception":"[object] (ErrorException(code: 0): ParserOptions being created without a UserIdentity object [Called from SMW\\Parser\\RecursiveTextProcessor::recursiveTagParse] at /srv/mediawiki/tags/2021-10-01_09:34:10/extensions/SemanticMediaWiki/src/Parser/RecursiveTextProcessor.php:284)
```

### DIFF
--- a/src/MediaWiki/Template/TemplateExpander.php
+++ b/src/MediaWiki/Template/TemplateExpander.php
@@ -4,6 +4,7 @@ namespace SMW\MediaWiki\Template;
 
 use Parser;
 use ParserOptions;
+use RequestContext;
 use Title;
 use RuntimeException;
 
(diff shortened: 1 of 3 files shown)


src/MediaWiki/Template/TemplateExpander.php
```diff
@@ -70,7 +71,8 @@ public function expand( $template ) {
 		$options = $this->parser->getOptions();
 
 		if ( !$options instanceof ParserOptions ) {
-			$options = new ParserOptions();
+			$user = RequestContext::getMain()->getUser();
+			$options = new ParserOptions( $user );
```
JeroenDeDauw, line 75: Can avoid having this $user var in the whole method scope.

`new ParserOptions( RequestContext::getMain()->getUser() )`

Then again the whole if-body should be in its own `newParserOptions` function.

PR #5128, opened 2021-11-09
COMPOSER_VERSION 2

JeroenDeDauw:


gesinn-it:
+1 since we anyway only work with MW 1.35+ on main branch

gesinn-it (approved):

PR #5132, opened 2021-11-10
Also run tests with MW 37

JeroenDeDauw:


gesinn-it:
There is not yet an official Docker image for 1.37. See https://hub.docker.com/_/mediawiki

JeroenDeDauw:
Well yeah... The way CI was set up for this repo is rather limiting. Look what I have in my recent extensions: https://github.com/ProfessionalWiki/ExternalContent/blob/master/.github/workflows/ci.yml

PR #5136, opened 2021-11-17
Do not rely on MW initialization order when being autoloaded.

Pchelolo:
Currently SMW setup uses both composer and extension.json.
During the composer autoload it executes SemanticMediaWiki.php
to export it's own constants to be used in LocalSettings.php.
Additionally it does self-check to see if it's intalled correctly.

During autoloading you can't rely on ANY part of MediaWiki to be
initialized. In this case it relied on $wgVersion and $wgLanguageCode
to be present. They will not be anymore since core is moving
dependency autoloading earlier.

In general, this pattern will be prohibited in MW 1.38 and will
be broken intentionally - relying on MW to be partially initialized
during class autoloading is a bad, bad, bad idea.

This fix is a stop-gap to avoid breaking translatewiki entirely.

Bug: https://phabricator.wikimedia.org/T295883

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -25,8 +25,8 @@
 		"SemanticMediaWiki::onExtensionFunction"
 	],
 	"ExtensionMessagesFiles": {
-		"SemanticMediaWikiAlias": "/i18n/extra/SemanticMediaWiki.alias.php",
-		"SemanticMediaWikiMagic": "/i18n/extra/SemanticMediaWiki.magic.php"
+		"SemanticMediaWikiAlias": "i18n/extra/SemanticMediaWiki.alias.php",
+		"SemanticMediaWikiMagic": "i18n/extra/SemanticMediaWiki.magic.php"
 	},
 	"QUnitTestModule": {
 		"ext.smw.tests": {
--- a/extension.json
+++ b/extension.json
@@ -24,6 +24,10 @@
 	"ExtensionFunctions": [
 		"SemanticMediaWiki::onExtensionFunction"
 	],
+	"ExtensionMessagesFiles": {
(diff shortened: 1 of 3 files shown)


extension.json
```diff
@@ -24,6 +24,10 @@
 	"ExtensionFunctions": [
 		"SemanticMediaWiki::onExtensionFunction"
 	],
+	"ExtensionMessagesFiles": {
+		"SemanticMediaWikiAlias": "/i18n/extra/SemanticMediaWiki.alias.php",
```
legoktm, line 28: No leading / here

legoktm (changes requested):

JeroenDeDauw:
Perhaps it is best to just remove this whole setup check. Or at least stop running stuff via the composer autoloader.

extension.json
Pchelolo: done

Pchelolo:
> Perhaps it is best to just remove this whole setup check. Or at least stop running stuff via the composer autoloader.

This is a minimal fix to unbreak translatewiki. I believe it will be beneficial to convert the extension to standard extension registration mechanism, but that decision is not up to me.

JeroenDeDauw:
A PR that does that is most welcome! I will definitely review it.

JeroenDeDauw:
I understand that this makes sense to unbreak translatewiki. Not clear it makes sense long term though, which is why I have not merged it yet. If it is broken anyway, then removal (or fix) seems better than putting in some dummy values to make PHP errors go away.

Nikerabbit:
This seems to work in my development environment, though currently setup only as a local hack in non-permanent way. Having it merged (or at least available in a local branch) would help greatly.

PR #5142, opened 2021-11-23
Native loading (WIP)

hexmode:


.github/workflows/main.yml
```diff
@@ -15,9 +15,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - mediawiki_version: 1.35
```
gesinn-it-gea, line 18: Am I right that dropping CI for MW 1.35 is only temp. during this PR is WIP? Otherwise this PR is way too early since MW 1.35 is LTS and needs to be supported until EOL and should be target no. 1 for CI. @JeroenDeDauw @krabina @kghbln any thoughts on this?

kghbln: 4.0.0 should indeed support 1.35 imho.

JeroenDeDauw: Yeah it'd be quite bad to not support MW 35 in at least the next SMW release. Dropping it after that maybe, then we have at least one recent SMW release that supports MW LTS.

hexmode: That was just to get the tests to run.  I need 1.35 support, so I'm not advocating dropping it.

hexmode: brought it back in 7ef61c0

PR #5144, opened 2021-11-24
Minimal pull request to enable native loading and Makefile for local testing

hexmode:
It would be great to use the same Makefile in the future to run the tests on github, but that isn't in this pull request.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -77,8 +77,13 @@
 			"SMW\\Maintenance\\": "maintenance/",
 			"SMW\\Tests\\": "tests/phpunit/"
 		},
-		"files" : [
-			"SemanticMediaWiki.php"
+		"psr-0": {
+			"SemanticMediaWiki": "SemanticMediaWiki.php"
+		},
+		"files": [
+			"src/Aliases.php",
+			"src/Defines.php",
+			"src/GlobalFunctions.php"
 		],
 		"classmap" : [
 			"includes/",
--- a/composer.json
+++ b/composer.json
@@ -94,16 +94,16 @@
 		"process-timeout": 0
 	},
 	"scripts": {
-		"test": "php ../../tests/phpunit/phpunit.php -c phpunit.xml.dist",
-		"phpunit": "php ../../tests/phpunit/phpunit.php -c phpunit.xml.dist",
-		"phpdbg": "phpdbg -qrr ../../tests/phpunit/phpunit.php -c phpunit.xml.dist",
+		"test": "php ${MW_INSTALL_PATH}/tests/phpunit/phpunit.php -c phpunit.xml.dist",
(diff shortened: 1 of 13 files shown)


JeroenDeDauw:
Holy f that is quite the Makefile

hexmode:
Makefiles are just task-oriented shell scripts.  There is a lot of extra stuff in there, so I should trim it to just run the tests.

hexmode:
Ok, this failure looks similar to what I get here.  I guess if it compares well with the most recent failures against master, it should be merged?

hexmode:
Just checked: It doesn't look the same as failures against master.  I'll see what I can do about those.

gesinn-it-gea:
@hexmode I made a PR to Docker Hub so that now MW 1.37 is also available. Feel free to add it to the test matrix.

https://hub.docker.com/_/mediawiki

composer.json
```diff
@@ -94,16 +94,16 @@
 		"process-timeout": 0
 	},
 	"scripts": {
-		"test": "php ../../tests/phpunit/phpunit.php -c phpunit.xml.dist",
-		"phpunit": "php ../../tests/phpunit/phpunit.php -c phpunit.xml.dist",
-		"phpdbg": "phpdbg -qrr ../../tests/phpunit/phpunit.php -c phpunit.xml.dist",
+		"test": "php ${MW_INSTALL_PATH}/tests/phpunit/phpunit.php -c phpunit.xml.dist",
```
hexmode, line 97: @kghbln @JeroenDeDauw Is this ok?  I was having trouble with symlinks and this made it work.

src/Localizer/LocalMessageProvider.php
```diff
@@ -16,38 +16,40 @@ class LocalMessageProvider implements MessageLocalizer {
...
 	 *
 	 * @param string $file
-	 * @param string|null $languageCode
+	 * @param ?string $languageCode
 	 */
 	public function __construct( string $file, ?string $languageCode = null ) {
 		$this->file = $file;
 		$this->languageCode = $languageCode;
-		$this->languageFileDir = $GLOBALS['wgMessagesDirs']['SemanticMediaWiki'];
+		$this->languageFileDir = !is_array( $GLOBALS['wgMessagesDirs']['SemanticMediaWiki'] )
```
hexmode, line 50: The access to $wgMessageDirs is scattered throughout and needs to be fixed.

hexmode:
Yay! Tests pass on 1.35!

hexmode:
> @hexmode I made a PR to Docker Hub so that now MW 1.37 is also available. Feel free to add it to the test matrix.

done.

composer.json
JeroenDeDauw: Seems reasonable enough to me. Have not tried it myself. Nowadays I use make like seen here https://github.com/ProfessionalWiki/ExternalContent/blob/master/Makefile

PR #5147, opened 2021-11-26
Consolidate access to wgMessagesDirs

hexmode:


JeroenDeDauw:
`LocalMessageProvider` does not seem like a good place to put this. Is an entirely different responsibility by the looks of it. Can't say more without further investigation in IDE.

hexmode:
> `LocalMessageProvider` does not seem like a good place to put this. Is an entirely different responsibility by the looks of it. 

Should I just use MW's context object?

JeroenDeDauw:
My comment was not about using static fields and functions VS injecting something. Rather that the new static stuff gets glued onto LocalMessageProvider in this PR, which seems like an odd location. 

(I do not really understand what you have in mind with context object)

hexmode:
> (I do not really understand what you have in mind with context object)

I was thinking about how to get the value of wgMessagesDirs -- but I meant [Config](https://www.mediawiki.org/wiki/Manual:Config.php), not context.

JeroenDeDauw:
A new "class" with the static stuff would be better than sticking it onto `LocalMessageProvider`.

jaideraf:
I don't think this is going to flourish, so I am closing it now. If anyone is still interested the subject, please update de code and send a new pull request.

PR #5150, opened 2021-11-29
Add alias for only class reffed outside of core with SMW\Test

hexmode:
At least, the only one I could find on codesearch.


hexmode:
oops... looks like I need to add something in the autoloader...

hexmode:
or just not have a typo

Issue #5165, opened 2021-12-19
LocalMessageProvider.php:  syntax error, unexpected 'string' (T_STRING), expecting function (T_FUNCTION) or const (T_CONST)


kghbln:
### Setup

MediaWiki | 1.35.5 (56fc6ae)16. Dez. 2021, 17:43
-- | --
PHP | 7.3.31-1~deb10u1 (apache2handler)
MariaDB | 10.3.31-MariaDB-0+deb10u1
Semantic MediaWiki | <s>4.0.0-rc.0</s> 4.0.0-rc1 2c1de56

### Issue

```
[1acd36acdda3667845d6b6dd] /wiki/Speciaal:Beheerdersverwijzingen ParseError from line 19 of /../w/extensions/SemanticMediaWiki/src/Localizer/LocalMessageProvider.php: syntax error, unexpected 'string' (T_STRING), expecting function (T_FUNCTION) or const (T_CONST)
```

### Backtrace
```
#0 [internal function]: AutoLoader::autoload(string)
#1 /../w/extensions/SemanticMediaWiki/src/SetupCheck.php(139): spl_autoload_call(string)
#2 /../w/extensions/SemanticMediaWiki/src/Setup.php(121): SMW\SetupCheck->__construct(array, SMW\SetupFile)
#3 /../w/extensions/SemanticMediaWiki/SemanticMediaWiki.php(76): SMW\Setup->init(array, string)
#4 /../w/includes/Setup.php(806): SemanticMediaWiki::onExtensionFunction()
#5 /../w/includes/WebStart.php(89): require_once(string)
#6 /../w/index.php(44): require(string)
#7 {main}
```

PR #5168, opened 2021-12-19
Fix PHP 7.3 compat in LocalMessageProvider

JeroenDeDauw:
Fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5165

### DIFF
--- a/SemanticMediaWiki.php
+++ b/SemanticMediaWiki.php
@@ -39,6 +39,8 @@ class SemanticMediaWiki {
 			define( 'SMW_EXTENSION_LOADED', true );
 		}
 
+		include_once __DIR__ . '/src/Aliases.php';
+
 		$defaultSettings = include_once __DIR__ . '/DefaultSettings.php';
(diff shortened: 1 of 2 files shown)


SemanticMediaWiki.php
```diff
@@ -39,6 +39,8 @@ public static function initExtension( $credits = [] ) {
 			define( 'SMW_EXTENSION_LOADED', true );
 		}
 
+		include_once __DIR__ . '/src/Aliases.php';
```
hexmode, line 42: What does this have to do with 7.3 compat?

hexmode (commented):
I thought we said 4.x was going to require 7.4?

SemanticMediaWiki.php
JeroenDeDauw: Nothing. Thanks. This is what happens when you have two different clones of SMW open and modifying them at the same time.

PR #5169, opened 2021-12-20
Consolidate loading

hexmode:
* Move SemanticMediaWiki.php and DefaultSettings.php from root
* Consolidate Aliases.php into SemanticMediaWiki.php


JeroenDeDauw:
Looks like you forgot to remove Defines.php

Does this also work when keeping Defines.php and Aliases.php but just loading these later, at the place where they are now inlined?

What made you move SemanticMediaWiki.php and DefaultSettings.php out of the root? I'm especially concerned about DefaultSettings which is referenced from the docs. Also note that `includes/` is deprecated in favor of `src/`.

hexmode:
> Does this also work when keeping Defines.php and Aliases.php but just loading these later, at the place where they are now inlined?

I'm sure it would, but I don't understand why that would be needed.  Having an auto-loaded public static method is better than requiring people who need these definitions and aliases to require a file at a specific location.

> What made you move SemanticMediaWiki.php and DefaultSettings.php out of the root?

Because having SemanticMediaWiki.php in root is confusing for anyone who sees it and expects `wfExtension()` or `require_once` to activate an extension.  I would like to standardize on just using `wfExtension()` and avoid confusion.  A major version release is a good place to do that

> I'm especially concerned about DefaultSettings which is referenced from the docs.

Update the docs?  Again, the major version release is a good time to make these changes.

> Also note that `includes/` is deprecated in favor of `src/`.

I understood `src/` is being used for PSR-4 autoloading, so I used `includes/`.  Ideally, I'd like to get rid of GlobalFunctions.php completely, though.

JeroenDeDauw:
> I'm sure it would, but I don't understand why that would be needed. Having an auto-loaded public static method is better than requiring people who need these definitions and aliases to require a file at a specific location.

One responsibility per file is nice. These should not be included from anywhere else, so the API does not matter. And if it did, you could still have the static methods, and have them do the inclusion of the relevant file.

> Because having SemanticMediaWiki.php in root is confusing for anyone who sees it and expects wfExtension() or require_once to activate an extension

Fair enough

> Update the docs? Again, the major version release is a good time to make these changes.

This is a good way to confuse people for little gain. So keeping DefaultSettings.php where it was is the one change I'd like to see before merging this.

hexmode:
I still want to do a more thorough check for core use of aliased classes.

JeroenDeDauw:
Hmm, I should have been more clear. Not moving DefaultSettings.php is because users can look there for the default values. There is no code, or there should be none, outside of the loading code in SMW that you are modifying, that should load these files. So there is no need for the shims, and they do not help with my concern.

hexmode:
I think I addressed that by updating the docs and links and restoring the file and point the comments in it to the DefaultSettings.php file.

JeroenDeDauw:
Still not happy with the DefaultSettings.php move but whatever :)

JeroenDeDauw:
Bleh now the MW 35 test on master is failing. wtf, it worked here. Maybe unrelated and just bad timing?

hexmode:
Looks like it was caused by composer/composer#10281

JeroenDeDauw:
What part of our stack is using that? Something in SMW? Or is it a more general problem? And is this for Composer 1.x or 2.x or both?

PR #5171, opened 2021-12-21
Replace removed AbstractBlock::getTarget()

ZabeMath:
Reported at https://phabricator.wikimedia.org/T298095

### DIFF
--- a/src/MediaWiki/Hooks/UserChange.php
+++ b/src/MediaWiki/Hooks/UserChange.php
@@ -2,12 +2,12 @@
 
 namespace SMW\MediaWiki\Hooks;
 
+use MediaWiki\User\UserIdentity;
 use SMW\ApplicationFactory;
 use SMW\MediaWiki\Jobs\UpdateJob;
 use SMW\NamespaceExaminer;
 use SMW\MediaWiki\HookListener;
 use Title;
-use User;
 
 /**
  * @see https://www.mediawiki.org/wiki/Manual:Hooks/BlockIpComplete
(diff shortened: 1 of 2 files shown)


Abijeet:
This looks good to me

Abijeet:
Fix was deployed and tested on translatewiki. Works well.

src/MediaWiki/Hooks/UserChange.php
```diff
@@ -55,15 +55,15 @@ public function setOrigin( $origin ) {
...
+	 * @param UserIdentity|string $user
 	 */
 	public function process( $user ) {
 
 		if ( !$this->namespaceExaminer->isSemanticEnabled( NS_USER ) ) {
 			return false;
 		}
 
-		if ( $user instanceof User ) {
+		if ( $user instanceof UserIdentity ) {
```
brightbyte, line 66: This will fail for older versions of MediaWiki that don't have UserIdentity, I think. It probably needs to be
`( $user instanceof User || $user instanceof UserIdentity )`

ZabeMath: UserIdentity is @since 1.31. Is SemanticMediaWiki supposed to support even older versions?

gesinn-it-gea: MW dev-master supports MW 1.35+

JeroenDeDauw: I checked this before merging. Indeed since MW 31. Not an issue

PR #5176, opened 2021-12-24
Replace usage of deprecated wfWikiID() in DefaultSettings

ZabeMath:
The global function wfWikiID() is deprecated since 1.35 and it's usages
should be replaced with WikiMap::getCurrentWikiId().


hexmode (approved):
Thank you!

PR #5180, opened 2021-12-26
Replace usage of deprecated global function wfReadOnly()

ZabeMath:
The global function wfReadOnly() has been deprecated in favor of the
new ReadOnlyMode service. Its usages should be replaced.


JeroenDeDauw:
What is the compatibility implication? ie what is the oldest version of MediaWiki the new code will work with?

ZabeMath:
ReadOnlyMode got introduced in MW 1.29.

PR #5181, opened 2021-12-26
Update broken URL

squirrelslair:
Updated old/broken URL "http://www.foaf-project.org/" to "xmlns.com/foaf/spec/"; broken URL keeps getting tagged by RottenLinks


squirrelslair:
I see a check failed but it doesn't seem to have anything to do with what I changed. What am I missing?

JeroenDeDauw:
Those checks where already failing, don't worry about them.

PR #5186, opened 2021-12-29
Fix missing parenthesis on a method call

Ostrzyciel:
This causes non-existent property PHP warnings.


malberts (approved):
Thanks. I somehow missed that with my recent unit test fixes.

JeroenDeDauw:
@malberts feel free to merge PRs from other people if you feel like you have a handle on them and that they do not require further review

PR #5187, opened 2021-12-29
Remove app factory alias usage

hexmode:
Also fix path

### DIFF
--- a/includes/specials/SMW_SpecialTypes.php
+++ b/includes/specials/SMW_SpecialTypes.php
@@ -1,6 +1,6 @@
 <?php
 
-use SMW\ApplicationFactory;
+use SMW\Services\ServicesFactory as ApplicationFactory;
 use SMW\DataTypeRegistry;
 use SMW\DataValueFactory;
 use SMW\Utils\HtmlTabs;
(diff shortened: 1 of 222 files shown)


includes/specials/SMW_SpecialTypes.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-use SMW\ApplicationFactory;
+use SMW\Services\ServicesFactory as ApplicationFactory;
```
JeroenDeDauw, line 3: I guess you did this because it is easier with standard search and replace. Aliases are not that nice though, just saying for further changes, not an actual issue.

hexmode: It is easy with sed, yes.  And it also brings the alias into the file so that someone who is just looking at the file can see what ApplicationFactory is an alias for.

I did think about the next step and just replacing ApplicationFactory with ServicesFactory but decided to do baby steps instead.

I take it, though, that you would prefer one change instead of two?

If so, I'll do that in the future.

JeroenDeDauw: I think this change was a step in the right direction. No issues. Perfect enemy of the good and all.

PR #5188, opened 2021-12-29
More steps towards less dependence on enableSemantics

hexmode:
Without these changes, special pages are not set up when enableSemantics() is not used.

AFAICT, the only need for enableSemantics here is where some other initialization is done
with it via ConfigPreloader.


JeroenDeDauw:
So you moved registration of this one hook. What about the other hooks there? Do they function or did we just not notice they also do not work when not calling enableSemantics?

hexmode:
> Do they function or did we just not notice they also do not work when not calling enableSemantics?

I don't know what "we" didn't notice, but I didn't notice them.  I have added others as I have seen them.  I'll do a thorough review later today.

Issue #5196, opened 2022-01-09
 Call to a member function newSchemaList() on null 


hexmode:
### Setup

- SMW version: master

### Issue

```
Sub Semantic Data (SMW\Tests\DataModel\SubSemanticData)
 ✔ Can construct [0.03 ms]
 ✔ Add sub semantic data [0.18 ms]
 ✔ Add sub semantic data with mismatched subject throws exception [0.68 ms]
 ✔ Remove sub semantic data [0.05 ms]
 ✘ Remove property [0.43 ms]
   │
   │ Error: Call to a member function newSchemaList() on null
   │
   │ /var/www/html/extensions/SemanticMediaWiki/src/DataModel/SequenceMap.php:50
   │ /var/www/html/extensions/SemanticMediaWiki/includes/SemanticData.php:495
   │ /var/www/html/extensions/SemanticMediaWiki/tests/phpunit/DataModel/SubSemanticDataTest.php:116
   │
```

See [test run](https://github.com/SemanticMediaWiki/SemanticMediaWiki/runs/4750905875#step:10:155)

Seems to be related to the order of the tests run.

In this case, [newSchemaFinder](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/44cc7f39bda4b01f378cb436295f651a2699cdd5/src/Schema/SchemaFactory.php#L156) seems to return null.  WTF?

JeroenDeDauw:
> In this case, newSchemaFinder seems to return null. WTF?

looks impossible

hexmode:
> looks impossible

Agreed. But [it is happening](https://github.com/SemanticMediaWiki/SemanticMediaWiki/runs/4750905875#step:10:155) somehow.

Interestingly, changing the [typehint for newSchemaFinder](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5195/commits/1760dfd3175ee6a0cd4cf76fa5c6a66e2fb786ff) moves the failure [down a few lines](https://github.com/SemanticMediaWiki/SemanticMediaWiki/runs/4753611561#step:10:155) because now [newSchemaList](https://github.com/hexmode/SemanticMediaWiki/blob/1760dfd3175ee6a0cd4cf76fa5c6a66e2fb786ff/src/Schema/SchemaFinder.php#L123) is providing null.

[...]

PR #5198, opened 2022-01-09
Add DataModel tests

hexmode:
Demonstrates bug #5196

### DIFF
--- a/src/Schema/SchemaFactory.php
+++ b/src/Schema/SchemaFactory.php
@@ -148,12 +148,8 @@ class SchemaFactory {
 
 	/**
 	 * @since 3.1
-	 *
-	 * @param Store|null $store
-	 *
-	 * @return SchemaFinder
 	 */
-	public function newSchemaFinder( Store $store = null ) {
+	public function newSchemaFinder( ?Store $store = null ): SchemaFinder {
 
 		$applicationFactory = ApplicationFactory::getInstance();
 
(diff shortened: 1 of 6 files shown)


src/Schema/SchemaFactory.php
```diff
@@ -153,7 +153,7 @@ public function newSchema( $name, $data ) {
 	 *
 	 * @return SchemaFinder
 	 */
-	public function newSchemaFinder( Store $store = null ) {
+	public function newSchemaFinder( Store $store = null ): SchemaFinder {
```
JeroenDeDauw, line 156: In these cases you can remove the @return doc

src/Schema/SchemaFinder.php
```diff
@@ -106,9 +106,9 @@ public function invalidateCache( DIProperty $property, ChangeRecord $changeRecor
 	 *
 	 * @param DataItem $dataItem
 	 *
-	 * @return SchemaList|[]
+	 * @return ?SchemaList
```
JeroenDeDauw, line 109: Is that actually a valid doc? Never seen that. Always like `SchemaList|null`. Or omitted in modern PHP, since you can express it in code.

src/Schema/SchemaFactory.php
hexmode: Thanks, I thought they were a little repetitive.

src/Schema/SchemaFinder.php
hexmode: A quick search of the docs does not show the nullable notation, but I'm sure I've used.

Or maybe I only thought I had because I have been leaving it off where I could.

PR #5207, opened 2022-01-20
Use ElasticFactory object injected into ElasticStore

### LINKED ISSUE #5206: Injected ElasticFactory is ignored in IndexerRecoveryJob
### Setup

- SMW version: 4.0.0
- MW version: 1.35.5
- PHP version: 7.4
- DB system (MySQL, Blazegraph, etc.) and version: ElasticSearch 5.6.16

### Issue

At Wikibase Solutions, we use a custom extension ([WSSemanticParsedText](https://www.mediawiki.org/wiki/Extension:WSSemanticParsedText)) to do preprocessing on the raw text before indexing it into ElasticSearch. This extension implements the `SMWSetupAfterInitializeComplete` hook in order to inject a custom instance of the `ElasticFactory` into the `ElasticStore`. This `ElasticFactory` overrides the normal `newIndexer` method to provide an indexer that does our preprocessing.

The above works great when active replication is in a normal state. However, when active replication fails for some reason and an IndexerRecoveryJob is pushed, the unprocessed version will be stored in the ElasticStore. This occasionally results in our entire ElasticSearch index becoming unprocessed.

After looking at the implementation of the IndexerRecoveryJob, the likely culprit is revealed:

```php
$elasticFactory = $applicationFactory->singleton( 'ElasticFactory' );
```

Instead of using the ElasticFactory from the ElasticStore singleton, as happens during active replication, a singleton for the ElasticFactory is used. Evidently, our injected ElasticFactory is ignored. Ideally, the behaviour of the IndexerRecoveryJob should be identical to that of active replication.

marijnvanwezel:
This pull request closes #5206.

### DIFF
--- a/src/Elastic/ElasticStore.php
+++ b/src/Elastic/ElasticStore.php
@@ -76,6 +76,13 @@ class ElasticStore extends SQLStore {
 	}
 
 	/**
+	 * @return ElasticFactory
+	 */
+	public function getElasticFactory() {
+		return $this->elasticFactory;
+	}
+
+	/**
 	 * @see Store::service
 	 *
 	 * {@inheritDoc}
--- a/src/Elastic/Jobs/IndexerRecoveryJob.php
+++ b/src/Elastic/Jobs/IndexerRecoveryJob.php
@@ -4,15 +4,12 @@ namespace SMW\Elastic\Jobs;
 
 use SMW\Services\ServicesFactory as ApplicationFactory;
 use SMW\DIWikiPage;
-use SMW\Elastic\Connection\Client;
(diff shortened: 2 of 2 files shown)


hexmode:
It would be good to add ElasticSearch to the CI pipeline.  Do you think you could do that in a separate pull request?

marijnvanwezel:
@hexmode I think that would be a good idea, but it is not something I know how to do.

We have tested this pull request locally and it seems to work well. I will mark it as ready to review.

marijnvanwezel:
@JeroenDeDauw @hexmode when do you think you will have the time to look at this? It's not urgent, but it is an issue we are regularly encountering in production.

PR #5234, opened 2022-04-17
i18n: replace Javascript with JavaScript

shirayuki:
Replace Javascript with JavaScript for consistency with other messages


kghbln:
Thanks!

PR #5257, opened 2022-05-26
Replace usage of Sanitizer::removeHTMLtags with removeSomeTags

Abijeet:
Sanitizer::removeHTMLtags was deprecated in MW 1.38

Bug: T305136


Abijeet:
> When was `removeSomeTags` added?

In MW 1.38

PR #5274, opened 2022-07-09
Update to accomodate future changes to JsonCodec

cscott:
JsonCodec will handle recursion itself in the future, so you can directly
include JsonSerializable objects as values in the array returned by
::toJsonArray(), and those objects will already be unserialized values
in the array given to ::newFromJsonArray().

This patch makes the code in SemanticMediaWiki agnostic about whether
unserialization occurs before or after ::newFromJsonArray() occurs,
in order to handle both the old and proposed new behavior of JsonCodec.

### DIFF
--- a/includes/SemanticData.php
+++ b/includes/SemanticData.php
@@ -898,6 +898,8 @@ class SemanticData implements JsonUnserializable {
 	 * @return array
 	 */
 	public function jsonSerialize() {
+		# T312589 explicitly calling jsonSerialize() will be unnecessary
+		# in the future.
 		$json = [
 			'stubObject' => $this->stubObject,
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
This is scary to merge without a test; I'm not sure I would spot a small mistake in there during CR.

jaideraf:
> We're getting log spam due to this in MediaWiki 1.45:

@paladox @alistair3149 did you know if this is already fixed? Can we close it?

paladox:
I think this was done in https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/e36ae539b610fcf0c9584d960f38dac7b64c7d28 ?

PR #5285, opened 2022-07-26
Update to accomodate future changes to JsonCodec

cscott:
JsonCodec will handle recursion itself in the future, so you can directly
include JsonSerializable objects as values in the array returned by
::toJsonArray(), and those objects will already be unserialized values
in the array given to ::newFromJsonArray().

This patch makes the code in SemanticMediaWiki agnostic about whether
unserialization occurs before or after ::newFromJsonArray() occurs,
in order to handle both the old and proposed new behavior of JsonCodec.

Bug: T312589

### DIFF
--- a/includes/SemanticData.php
+++ b/includes/SemanticData.php
@@ -898,16 +898,18 @@ class SemanticData implements JsonUnserializable {
 	 * @return array
 	 */
 	public function jsonSerialize() {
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
@cscott probably better to use the task URL rather than `# T312589` in the comments. Not everyone is a WMF dev, and those comments will get even more obscure if WMF changes issue tracker again.

PR #5288, opened 2022-08-06
Create schema.txt

krabina:
Create a schema.txt file to automatically import Schema.org vocabulary on installation.


data/import/vocabularies/schema.txt
```diff
@@ -0,0 +1,2805 @@
...
+ DigitalDocumentPermissionType|Category
+ DigitalFormat|Category
+ DigitalPlatformEnumeration|Category
+ DisabilitySupport|Category
+ DisagreeAction|Category
+ Discontinued|Category
+ DiscoverAction|Category
+ DiscussionForumPosting|Category
+ DislikeAction|Category
+ Distance|Category
```
JeroenDeDauw, line 1759: There also is a lowecase `distance` earlier on in the text file, of type Text. Is that correct?

JeroenDeDauw:
That is some seriously big txt file :)

data/import/vocabularies/schema.txt
krabina: I guess so. [schema:Distance](https://schema.org/Distance) would be for identifying categories and [schemadistance](https://schema.org/distance) for property usage. I downloaded version 14 as CSV from schema.org

kghbln:
I am unsure if I would like to have about 3000 properties to be imported into my blanc semantic wiki. Sure you can avoid this by changing the default configuration of SMW and preventing this from happening. However, once people realize the properties are already there.

kghbln:
Never mind. After thinking about it I guess the stuff here just gets added to the overview in the MediaWiki namespace. In this case it is ok.

krabina:
Excactly. It creates the file [MediaWiki:Smw_import_schema](https://www.semantic-mediawiki.org/wiki/MediaWiki:Smw_import_schema) that just sits there waiting for you to use the "imported from" property declaration. If you don't use it, nothing changes in your wiki.

krabina:
The rationale of the PR is to make it easier for novice users to re-use vocabularies. Only three old ones are provided, so I did this update.

PR #5293, opened 2022-08-06
Update owl.txt

krabina:
Update of OWL vocabulary

### DIFF
--- a/data/import/vocabularies/owl.txt
+++ b/data/import/vocabularies/owl.txt
@@ -1,4 +1,4 @@
-http://www.w3.org/2002/07/owl#|[http://www.w3.org/2002/07/owl Web Ontology Language (OWL)]
+https://www.w3.org/TR/owl2-syntax/ | [https://www.w3.org/TR/owl2-syntax/ Web Ontology Language (OWL)]
  AllDifferent|Category
  allValuesFrom|Type:Page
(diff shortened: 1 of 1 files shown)


data/import/vocabularies/owl.txt
```diff
@@ -36,7 +36,7 @@ http://www.w3.org/2002/07/owl#|[http://www.w3.org/2002/07/owl Web Ontology Langu
  owl|Type:Page
  priorVersion|Type:Page
  Restriction|Category
- sameAs|Type:Page
+ sameAs|Type:Reference
```
jaideraf, line 39: I don't think `owl:sameAs` should be of type [Reference](https://www.semantic-mediawiki.org/wiki/Help:Type_Reference). Could you explain this change? 

See: https://www.w3.org/TR/2012/REC-owl2-rdf-based-semantics-20121211/
```
owl:sameAs rdf:type rdf:Property .
owl:sameAs rdfs:domain owl:Thing .
owl:sameAs rdfs:range owl:Thing .
```
`rdfs:domain` and `rdfs:range` of `owl:sameAs` is `owl:Thing`, wich, I think, is close to datatype Page.

krabina: You are absolutely right. This was just a mistake, Thank you for pointing it out!

PR #5310, opened 2022-09-14
Remove usage of deprecated PersonalUrl hook

Abijeet:
Instead use the SkinTemplateNavigation::Universal but retain compatibility for MW 1.36 and below by keeping the old code.


Abijeet:
We've also been running this on translatewiki and it works fine. I think we can merge this in.

Issue #4373, opened 2019-12-16
Formatting display of a year-only date as ISO also displays month and day


Dinoguy1000:
### Setup and configuration

- SMW version: 3.0.0
- MW version: 1.31.1
- PHP version: 7.2.13-1+ubuntu16.04.1+deb.sury.org+1 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: MariaDB 10.1.31-MariaDB-1~xenial

### Issue
If you store a date to a date-type property, where the stored date consists only of a year, and then display that date on some page with the `#ISO` syntax to display it as an ISO date, it displays a month and day as well (e.g. storing `2005` and then displaying with `#ISO` results in `2005-01-01` being displayed). This is counter to [Help:Type Date#ISO](https://www.semantic-mediawiki.org/wiki/Help:Type_Date#ISO), which states:
> Display takes into account if parts of a date were omitted. So a date specified as "2008" will also be printed as "2008" everywhere, even though it might be interpreted as "Jan 1 2008 00:00:00" when comparing it to other dates.

For example, [Special:Browse/:Summoned-5FSkull-5F(CM)](https://yugipedia.com/wiki/Special:Browse/:Summoned-5FSkull-5F(CM)) shows `2005` for "Debut date", but [Summoned Skull#In other media](https://yugipedia.com/wiki/Summoned_Skull#In_other_media) (which has as part of the query `?Debut date#ISO`) produces `2005-01-01` for the "Summoned Skull (CM)" row.

Note that while this is being reported in the context of an MW 1.31.1 installation with SMW 3.0.0, it has also been confirmed on MW 1.19.24 running SMW 2.4.6; see [example](https://yugioh.fandom.com/wiki/Special:Ask?q=%5B%5BKonami+official+tournament+2004+participation+card%5D%5D&p=format%3Dbroadtable%2Flink%3Dall%2Fheaders%3Dshow%2Fsearchlabel%3D...-20further-20results%2Fclass%3Dsortable-20wikitable-20smwtable&po=%3FJapanese+release+date%0A%3FJapanese+release+date%23ISO%3DJapanese+release+date+ISO%0A&sort=Japanese+release+date&limit=500&eq=no) (and [Special:Version](https://yugioh.fandom.com/wiki/Special:Version)).

[...]

PR #5312, opened 2022-09-14
Add ISO-P formatter (allowing partial ISO dates)

### LINKED ISSUE #5309: Add formatter that allows partial ISO dates
This issue is a follow-up on #4373 to allow an ISO formatted date for both, complete and partial dates.

Rationale:

- Sometimes, only a year or a year and month are specified for a date, e.g. during planning of events
- At a later point in time, the date is completed with a month and day

Currently, a date with only a year specified will be rendered completed with January first which is wrong for the reader.

2022 should become 2022 instead of 2022-01-01.

A new formatter `ISO-P` (P for "partial") should render partial dates in ISO formatting:

- 2022 → 2022
- April 2022 → 2022-04
- 1 April 2022 → 2022-04-01

gesinn-it-wam:
This PR is made in reference to: #5309

Closes #5309

### DIFF
--- a/src/DataValues/ValueFormatters/TimeValueFormatter.php
+++ b/src/DataValues/ValueFormatters/TimeValueFormatter.php
@@ -77,6 +77,46 @@ class TimeValueFormatter extends DataValueFormatter {
 	 * @return string
 	 */
 	public function getISO8601Date( $mindefault = true ) {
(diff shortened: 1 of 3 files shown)


gesinn-it-gea:
fixes #4373

gesinn-it-gea:
@JeroenDeDauw ready to merge?

PR #5313, opened 2022-09-14
Allow rebuildData.php for smw_ftp_sesp_usereditcntns

alex-mashin:
```
// This next time someone ... I'm going to Alaska
      $field = strpos( $proptable->getName(), 'redi' ) ? [ 's_title', 's_namespace' ] : [ 's_id' ];
```
This fragment returns wrong `$field` when `$proptable->getName() = 'smw_ftp_sesp_usereditcntns'` and crashes `rebuildData.php`, because 'smw_ftp_sesp_usereditcntns' contains `redi`. So, update the check, so that 'redi' is at the end of the table name.


JeroenDeDauw:
WTF is that code and WTF is that comment :D

Thanks for fixing!

PR #5317, opened 2022-10-09
Don’t apply tests to i18n files with isLanguageRedirect

Seb35:
All these tests are currently skipped and will always be skipped, so remove these 12 noisy messages: it will simplify a bit the reading of the CI logs.

Also, use plural for "labels" to be coherent with "aliases".


Seb35:
Well, I tried to name in advance the branch but failed, I close this PR and retry.

PR #5323, opened 2022-10-23
Fix a test error by mocking a MW service

Seb35:
This adapts the test covering the change in d2d941d5e5c and removes the only error in MW 1.38 CI (but there are also a few failures).

I’m not sure what is the best way to mock a MW service, or more exactly to reset the service to its original unmocked value. In the first version of this patch, I reset to the value fixed in [MediaWiki]/includes/ServiceWiring.php but it’s bad since it could evolve in the future (and it would introduce maintenance to keep synchronisation).

It seems MediaWiki has the class [MediaWikiIntegrationTestCase](https://github.com/wikimedia/mediawiki/blob/master/tests/phpunit/MediaWikiIntegrationTestCase.php) for this (it creates a mocked MediaWikiServices, initialises it, defines the mocks, and reset it at the end), but SemanticMediaWiki currently doesn’t use it.

@JeroenDeDauw @gesinn-it-gea @Abijeet
\> Do we use MediaWikiIntegrationTestCase? Specifically when we need to mock a MW service?

### DIFF
--- a/tests/phpunit/includes/ContentParserTest.php
+++ b/tests/phpunit/includes/ContentParserTest.php
@@ -2,6 +2,8 @@
 
 namespace SMW\Tests;
 
+use MediaWiki\MediaWikiServices;
+use MediaWiki\Content\Renderer\ContentRenderer;
 use SMW\ContentParser;
 use SMW\Tests\PHPUnitCompat;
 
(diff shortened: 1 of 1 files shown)


gesinn-it-gea:
ping @gesinn-it-wam

Abijeet:
> Do we use `MediaWikiIntegrationTestCase`? Specifically when we need to mock a MW service?

I think it's fine to use the `MediaWikiIntegrationTestCase` when needed but the approach that you've taken appears to be a used a few times atleast. See: https://codesearch.wmcloud.org/search/?q=redefineService&i=nope&files=&excludeFiles=&repos=

Seb35:
Closed since it was solved by 1ade5a357d00009ff87310aa5e33758fe127be80 and a9b90b782ce0f420e32d9d7cba83bf08b289b4b7.

PR #5339, opened 2022-12-06
Php81compat

HamishSlater:
Refactor 'pass by reference' use


src/Globals.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace SMW;
+
+class Globals {
+
+	/**
+	 * PHP8.1 Compatibility
+	 * @since 4.0
```
JeroenDeDauw, line 9: Not since 4.0. Just drop that tag and perhaps mark the class as @internal

src/GroupPermissions.php
```diff
@@ -44,15 +47,19 @@ public function initPermissions( &$vars ) {
...
 			if ( !isset( $vars['wgGroupPermissions'][$group] ) ) {
-				$vars['wgGroupPermissions'][$group] = [];
+				$newVars['wgGroupPermissions'][$group] = [];
 			}
 
-			$vars['wgGroupPermissions'][$group] = array_merge( $rights, $vars['wgGroupPermissions'][$group] );
+			$newVars['wgGroupPermissions'][$group] = array_merge( $rights, $vars['wgGroupPermissions'][$group] );
 		}
+
+		Globals::replace( $newVars );
```
JeroenDeDauw, line 60: Can't the caller do this side effect? Weird to both return the new value and modify globals. Like this the test also does not check the production behavior.

src/NamespaceManager.php
```diff
@@ -60,7 +60,7 @@ public function init( $vars ) {
 		$newVars = [];
 
 		if ( !$this->isDefinedConstant( 'SMW_NS_PROPERTY' ) ) {
-			Globals::replace( $newVars = $this->initCustomNamespace( $vars ) );
+			Globals::replace( $newVars = $this->initCustomNamespace( $vars )['newVars'] );
```
JeroenDeDauw, line 63: Why is initCustomNamespace returning two variables in an array?

HamishSlater: initCustomNamespace() originally returned $instance.
For testing it also needs to return $newVars.

So a quick solution was to return:
	'instance' => $instance,
	'newVars' => $newVars

I considered setting a class variable '$instance', then call it from 'initCanonicalNamespaces()';
however, both methods are static.

JeroenDeDauw:
Merge conflicts with master

@HamishSlater I'm unfollowing this PR. Please nudge me when it is ready for review

HamishSlater:
> Merge conflicts with master
> 
> @HamishSlater I'm unfollowing this PR. Please nudge me when it is ready for review

@JeroenDeDauw Github states that there are no conflicts.
I have pulled the branch to include the recent commits.
It is awaiting review.

JeroenDeDauw:
image

HamishSlater:
image

JeroenDeDauw:
Wrong upstream?

malberts:
There is some underlying issue with this branch diverging from upstream master which requires a rebase outside of GitHub.

Perhaps it will be easier to recreate the branch with your changes. I tested this in #5353 and it is not complaining about a conflict.

PR #5345, opened 2022-12-08
Add show entity panel preference

malberts:
Refs #4674

I'm not very familiar with SMW architecture, so this might be totally in the wrong place.

In general the idea here is to allow completely disabling the entity examiner status indicator and panel (and the API call itself).

The user preference is `true` by default to maintain the current default behavior. This can be changed in LocalSettings:
```php
$wgDefaultUserOptions['smw-prefs-general-options-show-entity-issue-panel'] = false;
```

Or a user can set it on the Preferences UI:
Screenshot_20221209_002240

Any test changes are still TODO.

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -90,6 +90,9 @@
 		"UnusedProperties": "SMW\\SpecialUnusedProperties",
 		"WantedProperties": "SMW\\SpecialWantedProperties"
 	},
+	"DefaultUserOptions": {
+		"smw-prefs-general-options-show-entity-issue-panel": true
+	},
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
Looks good so far

JeroenDeDauw:
Still WIP?

malberts:
If the naming and everything else is fine, then it's ready.

krabina:
documented here: https://www.semantic-mediawiki.org/wiki/Help:Entity_issue_panel

PR #5348, opened 2022-12-09
Fix matching of pipe character in Ask API

### LINKED ISSUE #5316: Depth parameter fails when using the API
### Setup

- SMW version: 4.0.2
- MW version: 1.37.4
- PHP version: 7.4.30
- DB system (MySQL, Blazegraph, etc.) and version: MySQL 5.7.37

### Issue

When using the [depth](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Hierarchy_depth_restriction) parameter in the ASK api, the request fails with a smw_noclosingbrackets error.

The reason seems to come from function `getAskApiParameters` that splits the query string into parts, which fails to take in account this case:

```
	public function getAskApiParameters() {

		if ( $this->results === null ) {
			$this->results = isset( $this->requestParameters['query'] ) ? preg_split( "/(?<=[^\|])\|(?=[^\|])/", $this->requestParameters['query'] ) : [];
		}

		return $this->results;
	}
```

Which gives, with my request:

```
array(5)
0: "[[Category:Portail"
1: "+depth=0]]"
2: "?=Page"
3: "?A des termes SEO=Description"
4: "limit=5000"
```

If I protect the "depth" parameter with a hack, I get:
```
	public function getAskApiParameters() {

		$this->requestParameters['query'] = str_replace('|+depth=', '¤¤¤¤', $this->requestParameters['query']);

		if ( $this->results === null ) {
			$this->results = isset( $this->requestParameters['query'] ) ? preg_split( "/(?<=[^\|])\|(?=[^\|])/", $this->requestParameters['query'] ) : [];
		}

		foreach ($this->results as $k => $v)
			$this->results[$k] = str_replace('¤¤¤¤', '|+depth=', $v);

		return $this->results;
	}

```

The result is then: 
```
0: "[[Category:Portail|+depth=0]]"
1: "?=Page"
2: "?A des termes SEO=Description"
3: "limit=5000"
```
and the API returns the proper result.

octfx:
Update the regex to only match the pipe character `|` if the immediate next character is not a `+`.

Test Case: `[[Category:Portail|+depth=0]]|?=Page|?A des termes SEO=Description|limit=5000`
Old Regex: `/(?<=[^\|])\|(?=[^\|])/`  
Returns:
```
Array
(
    [0] => [[Category:Portail
    [1] => +depth=0]]
    [2] => ?=Page
    [3] => ?A des termes SEO=Description
    [4] => limit=5000
)
```

New Regex: `/(?<=[^\|])\|(?=[^\|])(?=[^\+])/`  
Returns:  
```
Array
(
    [0] => [[Category:Portail|+depth=0]]
    [1] => ?=Page
    [2] => ?A des termes SEO=Description
    [3] => limit=5000
)
```

Should fix #5316

### DIFF
--- a/src/MediaWiki/Api/ApiRequestParameterFormatter.php
+++ b/src/MediaWiki/Api/ApiRequestParameterFormatter.php
@@ -45,7 +45,7 @@ final class ApiRequestParameterFormatter {
 	public function getAskApiParameters() {
 
 		if ( $this->results === null ) {
-			$this->results = isset( $this->requestParameters['query'] ) ? preg_split( "/(?<=[^\|])\|(?=[^\|])/", $this->requestParameters['query'] ) : [];
+			$this->results = isset( $this->requestParameters['query'] ) ? preg_split( "/(?<=[^\|])\|(?=[^\|])(?=[^\+])/", $this->requestParameters['query'] ) : [];
 		}
 
 		return $this->results;


JeroenDeDauw:
Verified via https://regex101.com/

Issue #5300, opened 2022-08-10
Fix PHP 8.1 compatibility


JeroenDeDauw:
Issue: SMW passes $GLOBALS by reference in its setup code. https://php.watch/versions/8.1/GLOBALS-restrictions

Also reported in
* https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5272
*  https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5265
*  https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5302

osnard:
I have browsed the code and it looks like most places where this issue occurs are related to things that are usually done by `extension.json` (like hook wiring).

I can think of three ways do fix this:
1. Quickfix by directly accessing `$GLOBALS` where it is actually used
2. Quickfix by making the methods return an array and then merging it into `$GLOBALS`
3. Refactor to use `extension.json` mechanisms

There are also some usages that require a different approach (e.g. unsetting a hook handler).

Do you have any preference or do you have a different way in mind?

JeroenDeDauw:
Approach 2 is likely best for testing

osnard:
I have tried approach 1 to identify the places: https://github.com/hallowelt/mediawiki-extensions-SemanticMediaWiki/commit/c90f870d2c93b8f455820c557474eb0511689061

With this change it doesn't break at regular runtime. I have not touched the tests.
With approach 2 I could adapt unit tests better. Approach 3 is probably too heavy.

saerdnaer:
@osnard @JeroenDeDauw Any plans to bring this to master via an PR?

osnard:
Unfortunately I had to stop work on this due to my personal work load. But I think I could move on and provide a PR.

@JeroenDeDauw You said option 2 was preferred. Before I continue with the work, may I ask for your opinion on this again?

As an example, let's use a current implementation like
```php
Setup::initExtension( $GLOBALS ); // $GLOBALS passed by reference
```

Given the information on https://php.watch/versions/8.1/GLOBALS-restrictions, I can probably neither do something like
```php
$GLOBALS = array_merge_recursive( $GLOBALS, Setup::initExtension() ); // `Setup::initExtension` just returning an array
```
nor
```php
$GLOBALS = Setup::initExtension( $GLOBALS ); // $GLOBALS as reference passed by value
```

So I will need to add some "helper logic", that does the "per-key-merging". E.g.
```php
[...]

PR #5354, opened 2022-12-19
Fix 'pass by reference' use

HamishSlater:
Fix PHP 8.1 compatibility #5300


malberts:
The latest test results are looking good!

PR #5359, opened 2022-12-19
Add MW 1.39 to CI

malberts:
Not expecting this to pass, but we might as well start running this.


malberts:
MW 1.39 is running on PHP 8.1 because of https://github.com/wikimedia/mediawiki-docker/pull/117

PR #5360, opened 2022-12-19
WIP Add return types

malberts:


JeroenDeDauw:
Did you use a tool for this? I have not used it myself (at least not recently), but have heard good things about https://getrector.org/

malberts:
That's exactly what I used. But I need to tweak the conversion rules. And perhaps do smaller PRs, because I ran into cases where I had to manually change things after refactoring.

PR #5367, opened 2022-12-22
Add scalar return types

malberts:


includes/datavalues/SMW_DV_Number.php
```diff
@@ -407,8 +407,10 @@ public function getConvertedUnitValues() {
 	 * The returned value is a canonical ID for the main unit.
 	 * Returns the empty string if no unit is given for the value.
 	 * Overwritten by subclasses that support units.
+	 *
+	 * @return bool|string
 	 */
-	public function getUnit(): string {
+	public function getUnit() {
```
malberts, line 413: This is a manual fix where Rector didn't understand that the subclasses return arrays.

includes/datavalues/SMW_DataValue.php
```diff
@@ -968,8 +968,10 @@ abstract protected function loadDataItem( SMWDataItem $dataItem );
 	 * Overwritten by callers to supply an array of parameters that can be used for
 	 * creating servicelinks. The number and content of values in the parameter array
 	 * may vary, depending on the concrete datatype.
+	 *
+	 * @return bool|array
 	 */
-	protected function getServiceLinkParams(): bool {
+	protected function getServiceLinkParams() {
```
malberts, line 974: This is a manual fix where Rector didn't understand that the subclasses return arrays.

malberts:
Eh it messed up some comment indentations. Looks like you need to configure that first.

src/MediaWiki/Page/ListBuilder.php
```diff
@@ -164,7 +164,10 @@ public function getColumnList( array $dataItems, $colsThreshold = 10 ) {
 		return $htmlColumns->getHtml();
 	}
 
-	private function buildList( $dataItems ) {
+	/**
+	 * @return array<int|string, mixed[]>
```
malberts, line 168: This is an example of the automatically added array type. If this kind of documentation is pointless (albeit required for the CS tools) then I can rather remove these now. If somebody looks at this code later and comes up with a better type we can add it then.

src/SQLStore/Lookup/TableStatisticsLookup.php
```diff
@@ -47,7 +47,10 @@ public function get( $key ) {
 		return $this->{$key}( $this->store->getConnection( 'mw.db' ) );
 	}
 
-	private function loadFromDB( $connection ) {
+	/**
+	 * @return array{total_row_count: mixed, last_id: mixed, duplicate_count: int, rows: array{rev_count: mixed, smw_namespace_group_by_count: mixed, smw_iw: array{delete_count: mixed, redirect_count: mixed}, smw_proptable_hash: array{query_match_count: mixed, query_null_count: mixed}}}[]|array{total_row_count: mixed, rows: array{active_links_count: mixed, invalid_links_count: mixed, unassigned_count: mixed}}[]|array{total_row_count: mixed, unique_terms_occurrence_in_percent: int|float, rows: array{blob_field_null_row_count: mixed, terms_occurrence: array{single_occurrence_total_count: mixed, multi_occurrence_total_count: mixed}}}[]|array{query_time: float, snapshot_date: string}[]
```
malberts, line 51: Spam?

src/SQLStore/TableBuilder/MySQLTableBuilder.php
```diff
@@ -143,7 +143,10 @@ protected function doUpdateTable( $tableName, array $attributes = null ) {
 		}
 	}
 
-	private function getCurrentFields( $tableName ) {
+	/**
+	 * @return string[]|string[][]
```
malberts, line 147: This is wrong, but is strictly correct because `$row->Type` could be an array and `str_replace` can run on an array, hence why it thinks you can have 2D array.

src/SQLStore/Lookup/TableStatisticsLookup.php
JeroenDeDauw: Possibly yes

src/MediaWiki/Page/ListBuilder.php
JeroenDeDauw: This one does not seem super helpful. I'm also concerned correctness. If Rector is taking info from existing info in docs, then it might amplify mistakes in those docs. I do not trust type docs in SMW as I've seen them be wrong multiple times in the past.

includes/datavalues/SMW_DataValue.php
```diff
@@ -759,7 +759,7 @@ public function getHash() {
 	 *
 	 * @return boolean
 	 */
-	public function isNumeric() {
+	public function isNumeric(): bool {
```
JeroenDeDauw, line 762: Return type narrowing to base classes or interfaces that are extended in other extensions require a major release. For this particular class, possibly Maps is the only affected extension, with its Geo DataValue.

Changes to ResultPrinter will have a wider impact

JeroenDeDauw: We can always make a pragmatic exception to the rule, but would need to asses the impact of doing so. Or we could bump to 5.x.

Either way, probably good to first do https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5366 before merging potentially disruptive changes.

malberts: What do you suggest? Should we aim for a new major version in the short term, or is adding types like this too much effort? Otherwise I'll rather focus on missing types throwing deprecation notices in PHP 8.1 (e.g. #5370). But I suppose there is still a chance that by adding a missing type there we might still cause narrowing. I'm not familiar with all the downstream extensions, so that makes it a bit difficult to test.

JeroenDeDauw: We should definitely throw out 7.3, maybe also 7.4.

New major version in the short term seems fine to me, but would still do a patch/minor release first.

Issue #5369, opened 2022-12-22
Fix PHP 8.1 deprecations


malberts:

malberts:
I'll log separate tasks/PRs if I run into more deprecations.

PR #5370, opened 2022-12-22
Fix PHP 8.1 deprecations (during test execution)

malberts:
Refs #5369

Based on #5367 in order to see proper CI results. I'll rebase once that is merged.

This PR fixes the deprecations that show up during GH test executions. There are others that show up during maintenance scrips which are still oustanding.

There are 4 instances of this left:
https://github.com/SemanticMediaWiki/SemanticMediaWiki/actions/runs/3757349379/jobs/6384450256#step:11:1164
It is not immediately obvious to me, so I'm leaving that for later.


malberts:
Simpler version: #5373

PR #5124, opened 2021-11-05
Fix: AbstractContent::getParserOutput was deprecated in MW 1.38

Abijeet:
Also fix invalid usage of `$revision->getContentHandler()`

Error stack trace:
```
[2021-11-09T23:55:46.766710+00:00] error.WARNING: [facd9a01bf3e133e44d51dc0] /w/api.php   PHP Deprecated: Use of AbstractContent::getParserOutput was deprecated in MediaWiki 1.38. [Called from SMW\ContentParser::fetchFromContent] {"exception":"[object] (ErrorException(code: 0): PHP Deprecated: Use of AbstractContent::getParserOutput was deprecated in MediaWiki 1.38. [Called from SMW\\ContentParser::fetchFromContent] at /srv/mediawiki/tags/2021-11-04_08:34:43/extensions/SemanticMediaWiki/includes/ContentParser.php:162)
[stacktrace]
#0 [internal function]: MWExceptionHandler::handleError()
#1 /srv/mediawiki/tags/2021-11-04_08:34:43/includes/debug/MWDebug.php(375): trigger_error()
#2 /srv/mediawiki/tags/2021-11-04_08:34:43/includes/debug/MWDebug.php(349): MWDebug::sendRawDeprecated()
#3 /srv/mediawiki/tags/2021-11-04_08:34:43/includes/debug/MWDebug.php(230): MWDebug::deprecatedMsg()
#4 /srv/mediawiki/tags/2021-11-04_08:34:43/includes/GlobalFunctions.php(1005): MWDebug::deprecated()
#5 /srv/mediawiki/tags/2021-11-04_08:34:43/includes/content/AbstractContent.php(534): wfDeprecated()
#6 /srv/mediawiki/tags/2021-11-04_08:34:43/extensions/SemanticMediaWiki/includes/ContentParser.php(162): AbstractContent->getParserOutput()
#7 /srv/mediawiki/tags/2021-11-04_08:34:43/extensions/SemanticMediaWiki/includes/ContentParser.php(129): SMW\\ContentParser->fetchFromContent()
#8 /srv/mediawiki/tags/2021-11-04_08:34:43/extensions/SemanticMediaWiki/src/MediaWiki/Hooks/LinksUpdateConstructed.php(153): SMW\\ContentParser->parse()
[...]

PR #5371, opened 2022-12-22
Fix ContentParser test for MW 1.39

malberts:
This handles the change introduced in #5124 by mocking the `ContentRenderer` service.


malberts:
I'm not sure what's going on with the MW 1.38 test here. It almost looks like the service is not getting reset properly. But the same order (I assume from the result) isn't causing a problem for MW 1.39.

PR #5372, opened 2022-12-22
Fix MW 1.38 test error

malberts:
Follow-up to #5371.

I couldn't figure out how to get `TestEnvironment` to actually restore the original service again. I tried calling `resetServiceForTesting()` in `tearDown()` but that didn't work.

So either that doesn't actually do what I expect, or I failed an IQ test here. Either way, this does fix the MW 1.38 error casued by #5371.

### DIFF
--- a/tests/phpunit/includes/ContentParserTest.php
+++ b/tests/phpunit/includes/ContentParserTest.php
@@ -2,6 +2,8 @@
 
 namespace SMW\Tests;
 
+use MediaWiki\Content\Renderer\ContentRenderer;
+use MediaWiki\MediaWikiServices;
 use SMW\ContentParser;
 use SMW\Tests\PHPUnitCompat;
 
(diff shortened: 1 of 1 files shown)


tests/phpunit/includes/ContentParserTest.php
```diff
@@ -2,6 +2,8 @@
 
 namespace SMW\Tests;
 
+use MediaWiki\Content\Renderer\ContentRenderer;
```
malberts, line 5: Since this was added in MW 1.38, would it be better to not `use` it, and rather have it in the docstring below? Not sure if it makes a difference for < MW 1.38 (at least, tests don't complain).

JeroenDeDauw: `use` statements do not matter. They can reference symbols that do not exist. They only get used for lookup if you actually use the class during code execution. In other words: this is fine.png

PR #5375, opened 2022-12-23
Fix PHP 8.1 null string deprecations

malberts:
Refs #5369 

Various Special spages, when opened without parameters, were passing nulls where strings are expected.

This also fixes the 4 outstanding deprecations from #5373.


src/MediaWiki/Specials/SpecialBrowse.php
```diff
@@ -68,7 +68,7 @@ public function execute( $query ) {
 
 		$dataValue = DataValueFactory::getInstance()->newTypeIDValue(
 			'_wpg',
-			$articletext
+			$articletext ?? false
```
malberts, line 71: This one is `false`, not `''`, because that's `newTypeIDValue`'s default.

PR #5378, opened 2022-12-27
Fix RDF Export null string deprecation

malberts:
Refs #5369 

On PHP 8.1 a deprecation notice is shown when doing an RDF export using a page's "RDF" link: 
Screenshot_20221227_140956
Resulting URL: `/index.php?title=Special:ExportRDF/Foo_Bar&syntax=rdf`

This leads to an invalid XML document:
Screenshot_20221227_141049

Due to the presence of this text:
> <b>Deprecated</b>:  strtotime(): Passing null to parameter 1 ($datetime) of type string is deprecated in <b>/var/www/html/extensions/SemanticMediaWiki/includes/specials/SMW_SpecialOWLExport.php</b> on line <b>185</b><br />

### DIFF
--- a/includes/specials/SMW_SpecialOWLExport.php
+++ b/includes/specials/SMW_SpecialOWLExport.php
@@ -178,7 +178,7 @@ class SMWSpecialOWLExport extends SpecialPage {
 
 		$date = $wgRequest->getText( 'date' );
 		if ( $date === '' ) {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Draft?

malberts:
That's just GH remembering my last choice and me not noticing.

Issue #4950, opened 2021-03-01
smw.entityIdDisposer causes denial of service on job queue


Nikerabbit:
### Setup

- SMW version: 3.2.1
- MW version: 1.35.1
- PHP version: 7.3.27 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: 10.3.27-MariaDB-log

### Issue

smw.entityIdDisposer jobs are very slow and clog the job queue:
```
[nlaxstro@sanat2 workdir]$ php maintenance/showJobs.php --group --wiki=tieteentermipankki.fi
replaceText: 204 queued; 0 claimed (0 active, 0 abandoned); 0 delayed
refreshLinks: 6743 queued; 2297 claimed (0 active, 2297 abandoned); 0 delayed
htmlCacheUpdate: 23 queued; 0 claimed (0 active, 0 abandoned); 0 delayed
recentChangesUpdate: 1 queued; 0 claimed (0 active, 0 abandoned); 0 delayed
smw.entityIdDisposer: 1 queued; 5 claimed (5 active, 0 abandoned); 0 delayed
```

Example:

```
maalis 01 17:11:27 sanat2.csc.fi php[4017]: 2021-03-01 15:11:27 smw.entityIdDisposer SMW\SQLStore\Installer rootJobIsSelf=1 rootJobSignature=f7d4c5a67d33c85004251afffefccf7a1c6b631b rootJobTimestamp=20210301151124 waitOnCommandLine=1 namespace=0 title=SMW\SQLStore\Installer requestId=cbe40d07554587e5de4e5fe2 (id=24673,timestamp=20210301151124) t=3 good
maalis 01 17:11:51 sanat2.csc.fi php[3176]: 2021-03-01 15:11:51 smw.entityIdDisposer SMW\SQLStore\Installer rootJobIsSelf=1 rootJobSignature=f7d4c5a67d33c85004251afffefccf7a1c6b631b rootJobTimestamp=20210301151110 waitOnCommandLine=1 namespace=0 title=SMW\SQLStore\Installer requestId=3bef5bd713b3665d04feaade (id=790935,timestamp=20210301151110) t=118 good
[...]

PR #5383, opened 2022-12-30
add index on "p_id" for DIBooleanHandler

gesinn-it-gea:
fixing #4950

### DIFF
--- a/src/SQLStore/EntityStore/DataItemHandlers/DIBooleanHandler.php
+++ b/src/SQLStore/EntityStore/DataItemHandlers/DIBooleanHandler.php
@@ -44,6 +44,18 @@ class DIBooleanHandler extends DataItemHandler {
 	 *
 	 * {@inheritDoc}
 	 */
+	public function getTableIndexes() {
+		return [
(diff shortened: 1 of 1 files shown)


gesinn-it-gea:
All other handlers do have indizes on p_id. Only DIBooleanHandler did not have one. Tested on my local instance where I had the same performance issues as @Nikerabbit. The fix brought significant performance improvement.

JeroenDeDauw:
Great fix, thanks! Due to [our release policy](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/RELEASE-POLICY.md) we have to bump to 5.0 if we merge this. So let's hold of a little bit an first get another 4.x release out of the door.

JeroenDeDauw:
Actually... if you pull in this code change and do not run the update script, nothing should break, riiiight? The index simply will not be there, just like it currently is not. So this isn't a breaking change and thus does not need to be delayed till 5.0.

Or am I wrong, and does pulling this in without running the update script cause errors or other problems?

gesinn-it-gea:
@JeroenDeDauw it does not break but create the index automatically if required. No update.php is required.

PR #5386, opened 2023-01-08
Fix some PHP 8.2 deprecations

malberts:
First round of fixing a bunch of PHP 8.2 deprecations. These are primarily due to dynamic property creation.

Some of these are `public` scope, which I suppose is bad practice, but this allows the current code to work as-is.

### DIFF
--- a/src/NamespaceManager.php
+++ b/src/NamespaceManager.php
@@ -22,9 +22,9 @@ class NamespaceManager {
 	const DEFAULT_NAMESPACEINDEX = 100;
 
 	/**
-	 * @var LocalLanguage
+	 * @var LocalLanguage|null
 	 */
-	private $LocalLanguage;
+	private $localLanguage;
 
 	/**
 	 * @var string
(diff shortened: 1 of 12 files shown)


src/NamespaceManager.php
```diff
@@ -22,9 +22,9 @@ class NamespaceManager {
 	const DEFAULT_NAMESPACEINDEX = 100;
 
 	/**
-	 * @var LocalLanguage
+	 * @var LocalLanguage|null
 	 */
-	private $LocalLanguage;
+	private $localLanguage;
```
malberts, line 27: There is a case where this is `null` (https://github.com/SemanticMediaWiki/SemanticMediaWiki/actions/runs/3866345626/jobs/6590412554#step:10:26).

This should be replaced with `?LocalLanguage`, but if we're still supporting PHP 7.3 in 4.0.x then this must wait a little bit.

malberts:
For this to run on CI we'll need to make changes to how the test environment is set up. Currently it pulls in the MW docker image which is on PHP 8.1

JeroenDeDauw:
Yeah, I'm not a fan of using the MW images for that exact reason: you cannot vary what is in there. @gesinn-it-gea

PR #5389, opened 2023-01-08
Avoid type error using abs() on string

JeroenDeDauw:


malberts:
The test failure is `0` now being formatted as `0.0e+0`:
https://github.com/SemanticMediaWiki/SemanticMediaWiki/actions/runs/3869360500/jobs/6595407480#step:10:822

Not sure if this is `0` in general, or specific to the temperature unit (`0.0e+0&#160;°C`).

PR #5390, opened 2023-01-09
Refactor EditInfo constructor parameter order

### LINKED ISSUE #5388: Optional parameter $revision declared before required parameter $user
### Setup

- SMW version: master
- MW version: 1.39.1
- PHP version: PHP 8.1.13
- DB system: MariaDB 10.9.4

### Issue

Open a Property page, e.g. `/Property:FooBar`

> <b>Deprecated</b>:  Optional parameter $revision declared before required parameter $user is implicitly treated as a required parameter in <b>/var/www/html/extensions/SemanticMediaWiki/src/MediaWiki/EditInfo.php</b> on line <b>50</b><br />

This change was introduced in #5091.

malberts:
Fixes #5388

This fixes the constructor signature after #5091.

I don't know if this was ever meant to be part of a public API, so I'm not sure if this should be considered a breaking/major change. However, I cannot think of a different way to solve the issue, because there is still some logic where `revision` can be null.

### DIFF
--- a/src/MediaWiki/EditInfo.php
+++ b/src/MediaWiki/EditInfo.php
@@ -44,10 +44,10 @@ class EditInfo {
 	 * @since 1.9
 	 *
 	 * @param WikiPage $page
+	 * @param User $user
 	 * @param ?RevisionRecord $revision
-	 * @param ?User $user
 	 */
-	public function __construct( WikiPage $page, ?RevisionRecord $revision = null, User $user ) {
+	public function __construct( WikiPage $page, User $user, ?RevisionRecord $revision = null ) {
 		$this->page = $page;
 		$this->revision = $revision;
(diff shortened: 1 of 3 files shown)


src/MediaWiki/EditInfo.php
```diff
@@ -44,10 +44,10 @@ class EditInfo {
 	 * @since 1.9
 	 *
 	 * @param WikiPage $page
+	 * @param User $user
```
malberts, line 47: This docblock is redundant. Should we remove the whole thing, or keep only `@since`?

JeroenDeDauw: Both are fine IMO. Definitely good to get rid of the redundant docs. I'd probably kill the whole thing, as these @since is not very useful here

Issue #5004, opened 2021-06-08
Drop SMW_NS_RULE namespace


jaideraf:
`SMW_NS_RULE` namespace [seems to be unused](https://github.com/SemanticMediaWiki/SemanticMediaWiki/search?q=SMW_NS_RULE), I think it was replaced by the awesome `SMW_NS_SCHEMA` in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3431. So, as @mwjames is not actively developing it right now, it shoud be droped in a SMW 4.0. It is currently just an empty namespace displayed in the Special:Search.

kghbln:
[Documented](https://www.semantic-mediawiki.org/w/index.php?title=Help:Namespaces&oldid=90217)

PR #5395, opened 2023-01-11
Droping SMW_NS_RULE namespace

jaideraf:
Fix [#5004](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5004)


JeroenDeDauw:
Yay, but let's do this in 5.0.0 to be fully in line with our versioning policy, even if the feature is unused.

PR #5396, opened 2023-01-13
(URI data value) fixes spaces assumed as underscores also for non-wiki URIs

### LINKED ISSUE #4910: unwanted underscores in url property value
### Setup

- SMW version: 3.1.0
- MW version: 1.34.0
- PHP version: 7.3.25 (apache2handler)
- DB system (MySQL, Blazegraph, etc.) and version: 10.4.17-MariaDB

### Issue

Spaces in urls get  mistakenly underscored by contrast to what stated here
https://www.semantic-mediawiki.org/wiki/Help:Service_links

```
Help:Datatype "URL", Help:Datatype "Annotation URI"
$1: URL-encoded value of the URL
```

by contrast to

```
Help:Datatype "Page"
$1: URL-encoded article name (no namespace), with underscores replaced by spaces
```

Steps to reproduce the observation (recommendation is to use the [sandbox](https://sandbox.semantic-mediawiki.org)):

Create a URL data type and then add a link like the following

http://www.opal.unito.it/psixsite/Miscellanea%20di%20testi%20di%20genere%20diverso/Elenco%20opere/imgMIscE4813.pdf

the link will not work anymore because %20 get replaced with '_' . Here you can find a "working" example 

https://culturaitaliana.org/wiki/Digital_libraries/Universit%C3%A0_degli_Studi_di_Torino,_Biblioteca_storica_di_ateneo_%22Arturo_Graf%22

field "Digital library digital resource example url"

This might be an unwanted effect of the following commit

https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2201

thomas-topway-it:
fixes [issue/4910](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4910) and partial revert of https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2201

fixes underlying problem causing https://github.com/SemanticMediaWiki/SemanticResultFormats/issues/93 (solved in a recent commit of SemanticResultFormats with a cpu-consuming work-around)

As in the title, the commit https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/2201 assumes that all spaces (%20) in the URI data values correspond to underscore symbols, which is semantically true only for WIKI's URIs, either external or internal. (for this reason, legitimate URIs including spaces, not exchangeable with underscores, cease to work once imported in SMW)

This PR tries to address this issue, however the impact of the changes should be evaluated against the methods getShortWikiText, getShortHTMLText, getLongWikiText and getLongHTMLText (includes/datavalues/SMW_DV_URI.php)

Also note that the corresponding  tests have been identified but not yet updated because they imply a little design change however affecting many tests: see for instance the image below which is related to the new output of getShortWikiText (includes/datavalues/SMW_DV_URI.php)

uri-test-problem

(in the new output, the space, if any, is not anymore converted to an underscore, and therefore the caption must be explicitly declared in the wikitext, otherwise the url gets broken in two parts, one of them considered the caption -- this is correct, however other reviewers might evaluate if the new output is at the same time appropriate for the different uses of the method getShortWikiText)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved URI processing to preserve spaces and correctly encode components for various link types.
  - Enhanced error handling during URI creation for more reliable operation.

- **New Features**
  - Introduced support for retaining original link captions, ensuring more accurate and clear display in appropriate modes.


thomas-topway-it:
Ideally, before approving the Pull request, a key SMW developer is expected to verify that the new output produced by the changes of the `SMWURIValue -> getShortWikiText` method (as mentioned above) is acceptable and consistent with all use-cases throughout SMW of the same method.
(the new output assumes that the caption cannot anymore be used to denote a wikitext url, since spaces aren't anymore converted into underscores in the caption itself, and a wiki url distinguishes the url from the caption precisely by a space: for this reason, both the url and the caption have to be explicitly declared also in the `SMWURIValue -> getShortWikiText` method)

thomas-topway-it:
please also check the thread here https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4910

alistair3149:
Superseded by #7121 (#7105).
The discussion for a cleaner design is at #7123

PR #5397, opened 2023-01-17
Add fallback for use of deprecated TransactionProfiler::setSilenced()

Abijeet:
See: https://gerrit.wikimedia.org/r/c/mediawiki/core/+/839552

Tests have been removed and some backward compatibility for MW < 1.28 is dropped

TransactionProfiler::silenceForScope was added in
https://gerrit.wikimedia.org/r/c/mediawiki/core/+/691979 (MW 1.37)


Abijeet:
@JeroenDeDauw I was able to test this patch on translatewiki.net

Issue #5400, opened 2023-01-19
Control display of namespace in printouts


krabina:
I am very surprised that this is not yet possible in SMW. The default link behaviour for pages is to hide the namespace in links. This is true for the mainlabel (the page names) as well as for datatype page.

I think there are two places where this has to be done:
1. For the main result column it would be good the either extend the [limit parameter](https://www.semantic-mediawiki.org/wiki/Help:Inline_queries) or introduce a new parameter. So in addition to `link=all, subject, or none`, we could have additionalle, allns, subjectns. So to extend the values to: `link=all, allns, subject, subjectns, none`
2. For individual printouts of datatype "page", there are already a lot of [formatting options](https://www.semantic-mediawiki.org/wiki/Help:Displaying_information), so we could add something comparable to `?Has URL#nowiki`, e. g. in the form of `?Has Page#namespace`

Any other suggestions?

krabina:
An example can be found here: https://sandbox.semantic-mediawiki.net/wiki/Issue/5400
The rationale behind it is that it might be very important to see that the page "**French accents**" is actually the page "**Discussion:French_accents**", which is especially important id such a query results both pages that then cannot be distinguished from each other (other than with hovering over the links).

While I can see that often you don't want to display the namespaces (which should stay the default behaviour), I would like to see more flexibility having the above mentioned options.

bertrandgorge:
I've never had the issue, but I use the template format extensively so I guess I'm not really concerned. In any case, I believe it would be better to have a formating option for that (option 2).

For option 1, you mean link, not limit ? I'm not sure it's the right place for doing that since you might want some columns with that behaviour but not some others ?

thomas-topway-it:
> https://github.com/Knowledge-Wiki/SemanticMediaWiki/commit/6f1cbce25c7537ddd18ea3538881de50e9c5f783

perhaps it could be used as default ? i.e. the prefix is always shown for namespaces other than main

bertrandgorge:
>  prefix is always shown for namespaces other than main

Hu, I would rather not ! For categories for a start, that would really not be that nice.

krabina:
Yes, I meant limit, not link, sorry. So it seems there is an easy fix for this. Would this be a problem for you @bertrandgorge in the template? I guess not, because you can always use magic words in template. And for large amounts of data, putting everything into templates for formatting slows things down more. So I think it is good to see the namespaces in links by default.

[...]

PR #5408, opened 2023-01-27
Control display of namespace in printouts #5400

thomas-topway-it:
fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5400

Adds askquery parameter `prefix` with values `'all', 'subject', 'none', 'auto' `

* The PR does not affect the default behavior of askquery results
* The new parameter does not break the abstraction of the SMWDataValue class and sub-classes and as such does not change the interface of existing methods with additional parameters (only the class `ValueTextsBuilder` is created with a constructor taking the SMWQueryResult.) Instead, the existing SMWDataValue's `getShortText` and `getLongText` methods are concurrently used based on the new parameter
* The new parameter also introduces an option "auto", checking whether the class  `SMW\Query\Language\Description`, used the create the query, is an instance of `NamespaceDescription`, which ensures that namespaces of the query results aren't mixed.
* The PR does not make use of global parameters or global variables in order to work


src/Query/ResultPrinters/ListResultPrinter/ValueTextsBuilder.php
```diff
@@ -21,6 +23,15 @@ class ValueTextsBuilder {
 
 	private $linker;
 
+	private $mixedResults;
```
JeroenDeDauw, line 26: Missing type

src/Query/ResultPrinters/ListResultPrinter/ValueTextsBuilder.php
```diff
@@ -21,6 +23,15 @@ class ValueTextsBuilder {
...
+
+	/**
+	 * @param SMWQueryResult $queryResult
+	 */
+	public function __construct( QueryResult $queryResult ) {
+		$this->mixedResults = !( $queryResult->getQuery()->getDescription() instanceof NamespaceDescription );
+	}
```
JeroenDeDauw, line 33: If you only need the boolean, then don't inject the super complex QueryResult. That makes testing excessively difficult.

src/Query/ResultPrinters/TableResultPrinter.php
```diff
@@ -347,6 +350,34 @@ protected function getCellContent( array $dataValues, $outputMode, $isSubject )
 		return $html;
 	}
 
+	/**
+	 * @param @param SMWDataValue $value
+	 * @param int $outputMode
+	 * @param Linker|null|bool $linker
+	 * @return string
+	 */
+	protected function getDataValueText( $value, $outputMode, $isSubject ) {
```
JeroenDeDauw, line 359: Please us PHP types, not doc types. Yes, SMW mostly has types in docs, but that is because that code was written in the PHP 5.x days.

src/Query/ResultPrinters/TableResultPrinter.php
```diff
@@ -347,6 +350,34 @@ protected function getCellContent( array $dataValues, $outputMode, $isSubject )
...
+		} elseif ( $prefix === 'auto' && $isSubject ) {
+			$text = $this->mixedResults ? $value->getLongText( $outputMode, $linker ) : 
+				$value->getShortText( $outputMode, $linker );
+
+		} else {
+			$text = $value->getShortText( $outputMode, $linker );
+		}
+
+		return $text;
+	}
```
JeroenDeDauw, line 378: https://www.entropywins.wtf/blog/2019/01/14/readable-functions-guard-clause/

src/Query/ResultPrinters/TableResultPrinter.php
JeroenDeDauw: Btw, any reason for this to be protected? Typically something should be private unless a collaborator needs it to be public. Protected is rarely the right choice.

src/Query/ResultPrinters/TableResultPrinter.php
```diff
@@ -347,6 +350,34 @@ protected function getCellContent( array $dataValues, $outputMode, $isSubject )
...
+
+		} elseif ( $prefix === 'auto' && $isSubject ) {
+			$text = $this->mixedResults ? $value->getLongText( $outputMode, $linker ) : 
+				$value->getShortText( $outputMode, $linker );
+
+		} else {
+			$text = $value->getShortText( $outputMode, $linker );
+		}
+
+		return $text;
```
JeroenDeDauw, line 377: This looks very similar to what you added to ValueTextsBuilder, and like the type of duplication that is actually bad by leading to inconsistencies and shotgun surgery.

src/Query/Processor/DefaultParamDefinition.php
```diff
@@ -89,6 +89,11 @@ public static function buildParamDefinitions( $vars, $context = null, ResultPrin
 			'values' => [ 'all', 'subject', 'none' ],
 		];
 
+		$params['prefix'] = [
+			'default' => 'none',
+			'values' => [ 'all', 'subject', 'none', 'auto' ],
+		];
```
JeroenDeDauw, line 95: What is the rationale for adding the new parameter to all result formats? As far as I can tell this PR only adds handling of the parameter to a few formats. (I did not check how ValueTextsBuilder is used so might be wrong here.) Having the parameter in all formats but ignored in most seems odd from a product perspective.

krabina: From a user perspective, we want these options in as many formats as possible. Maps is the only one I am aware of that already has this. If it means touching every format to bring this feature to is, we will try this, because I think it is important to be able to control the namespace behaviour. 
Thanks for reviewing, btw!

JeroenDeDauw: Yeah, but do users want the parameter to be there even if it is not actually taken into account?

krabina: No. But would not doing it this way then require changes to SMW core again if we implement it in any other result format? Then it might be better having this in and just adding the availability to the documentation of those formats where it has been implemented.

JeroenDeDauw: Each result format defines what parameters it supports. There where no changes in SMW core needed for Maps to support an extra parameter.

thomas-topway-it: for now the new parameter is only used for "table" and "list" formats. However I see, is it sufficient to declare it using the method `getParamDefinitions` in the specific printer, right ? 
Then, @krabina we could do the same declaration as we add the new parameter to other formats

JeroenDeDauw: > However I see, is it sufficient to declare it using the method getParamDefinitions in the specific printer, right ?

Yes

thomas-topway-it:
I did an update on the basis of the suggestions replacing several methods with a trait. I think I have also used the "one level of abstraction" principle, now I will add notes to specific comments

src/Query/ResultPrinters/ListResultPrinter/ValueTextsBuilder.php
thomas-topway-it: done within the trait

src/Query/ResultPrinters/ListResultPrinter/ValueTextsBuilder.php
thomas-topway-it: this is the only point for which I have added a comment with further explanation within the trait src/Query/ResultPrinters/PrefixParameterProcessor

src/Query/ResultPrinters/TableResultPrinter.php
thomas-topway-it: this method is now removed (I think I adopted the same access declaration from the caller method)

src/Query/ResultPrinters/TableResultPrinter.php
thomas-topway-it: done (in the same trait)

src/Query/ResultPrinters/TableResultPrinter.php
thomas-topway-it: fixed using the trait

src/Query/ResultPrinters/PrefixParameterProcessor.php
```diff
@@ -0,0 +1,40 @@
...
+		// homogeneous or mixed prefixes
+		return !( $query->getDescription() instanceof NamespaceDescription );
+	}
+
+	public function useLongText( bool $isSubject ) : bool {
+		$prefix = $this->prefix;
+
+		if ( $prefix === 'all'
+			|| ( $prefix === 'subject' && $isSubject )
+			|| ( $prefix === 'auto' && $isSubject && $this->mixedResults ) ) {
```
JeroenDeDauw, line 33: This class has various design issues.

Code will blow up is setQuery was not called first.

setQuery does work that is not always needed, since mixedResults is only read in some cases

Actually it is not a class, it is a trait, which have been dubbed "compiler-assisted copy-paste"

ValueTextsBuilder is still requiring a complex object in its constructor. Instead you could inject the boolean, which seems to be all it needs, or wrap figuring out the boolean into a class and then inject said class. You could turn this trait into a class.

thomas-topway-it: redesigned using a class. Regarding the boolean, I'm not sure that a boolean is all we need. There could be other ways in the method PrefixParameterProcessor->getMixedResults to determine whether the query is expected to return mixed namespaces. For now I've found one, but there could be more I'm not aware of. (for this reason this method at least by principle needs to evaluate the query itself.)

JeroenDeDauw:
Needs rebase

krabina:
Closing because of https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5465

PR #5409, opened 2023-01-28
Move Factbox text into a service

### LINKED ISSUE #5387: Creation of dynamic property OutputPage::$mSMWFactboxText is deprecated
### Setup

- SMW version: master
- MW version: 1.39.1
- PHP version: PHP 8.2.1
- DB system (MySQL, Blazegraph, etc.) and version: MariaDB

### Issue

Open a Property page, e.g. `/Property:FooBar`

Deprecation notice:
> <b>Deprecated</b>:  Creation of dynamic property OutputPage::$mSMWFactboxText is deprecated in <b>/var/www/html/extensions/SemanticMediaWiki/src/Factbox/CachedFactbox.php</b> on line <b>171</b><br />

----

I haven't checked if there is an easy solution yet, but this seems to be a bad violation if we're modifying MW objects.

malberts:
Fixes #5387

This takes a very literal approach to replacing the dynamic property with a service. 

Is this the right approach, or is there a better way to do this? The FactboxText class is basically a glorified global string or singleton. I'm not sure if there is a similar example in SMW which I can follow.

Is it necessary to inject this new class into the affected classes? Or is this fine, since they already use ApplicationFactory internally?

### DIFF
--- a/src/MediaWiki/Hooks/OutputPageParserOutput.php
+++ b/src/MediaWiki/Hooks/OutputPageParserOutput.php
@@ -131,13 +131,13 @@ class OutputPageParserOutput implements HookListener {
 
 	protected function addFactbox( OutputPage $outputPage, ParserOutput $parserOutput ) {
 
-		$applicationFactory = ApplicationFactory::getInstance();
-
 		$request = $outputPage->getContext()->getRequest();
 
 		if ( $this->factboxText->hasText() && $request->getCheck( 'wpPreview' ) ) {
 			return '';
 		}
+
+		$applicationFactory = ApplicationFactory::getInstance();
 
 		$cachedFactbox = $applicationFactory->singleton( 'FactboxFactory' )->newCachedFactbox();
 
--- a/src/MediaWiki/Hooks/OutputPageParserOutput.php
+++ b/src/MediaWiki/Hooks/OutputPageParserOutput.php
@@ -4,6 +4,7 @@ namespace SMW\MediaWiki\Hooks;
 
 use OutputPage;
 use ParserOutput;
+use SMW\Factbox\FactboxText;
 use SMW\Services\ServicesFactory as ApplicationFactory;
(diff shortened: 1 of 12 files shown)


src/MediaWiki/Hooks/OutputPageParserOutput.php
```diff
@@ -146,7 +148,7 @@ protected function addFactbox( OutputPage $outputPage, ParserOutput $parserOutpu
 		// Due to how MW started to move the `mw-data-after-content` out of the
 		// `bodyContent` we need a way to distinguish content from a top level
 		// to apply additional CSS rules
-		if ( isset( $outputPage->mSMWFactboxText ) && $outputPage->mSMWFactboxText !== '' ) {
+		if ( $factboxText->hasText() && $factboxText->getText() !== '' ) {
```
JeroenDeDauw, line 151: Suggests a new method

JeroenDeDauw:
> Is it necessary to inject this new class into the affected classes? Or is this fine, since they already use ApplicationFactory internally?

Injecting is cleaner, and I would go with that unless it opens a big can of worms. If the class is already being constructed via ApplicationFactory, then adding a new dependency to its constructor should be easy.

> Is this the right approach, or is there a better way to do this? The FactboxText class is basically a glorified global string or singleton. I'm not sure if there is a similar example in SMW which I can follow.

Nothing comes to mind immediately, though I have not given it a lot of thought.

tests/phpunit/Factbox/FactboxTextTest.php
```diff
@@ -0,0 +1,92 @@
...
+ * @group semantic-mediawiki
+ *
+ * @license GNU GPL v2+
+ * @since 4.1.1
+ *
+ * @author Morne Alberts
+ */
+class FactboxTextTest extends \PHPUnit_Framework_TestCase {
+
+	use PHPUnitCompat;
```
JeroenDeDauw, line 19: Used?

PR #5411, opened 2023-01-29
Fix 4 failing tests on MW 1.38+

Seb35:
MediaWiki 1.38 deprecated Sanitizer::removeHTMLtags and introduced Sanitizer::removeSomeTags, but it is not a strict equivalent function: HTML codes (&#dd;) and their equivalent in UTF-8 are sometimes exchanged.

It was introduced in SMW in 2df4093a4d04761f4bb283938623a2b7616800ca (4.0.2+).

The 4 times, I duplicated the tests with excluding `skip-on` cases:
* Test a-0001.json `#8`: The labels for the properties inside the `<a>` have `&#160;` instead of a plain no-break space; I moved the pre-1.38 file to Fixtures/res.a-0001.8-pre-mw-1.38.txt, and Fixtures/res.a-0001.8.txt is a copy with 3 occurences of `\u00a0` remplaced by `&#160;`
* Test s-0013.json `#0`: similar, but directly in TestCases/s-0013.json
* Test f-0201.json `#5`: the contrary: it was `&#10003;` and `&#10005;` before and it is now `✓` and `✕`
* Test s-0014.json `#6` and `#7`: it was `&` before and it is now `&amp;`


Seb35:
There shouldn’t be other tests to be fixed with this method -- I launched the tests with and without Sanitizer::removeHTMLtags and only these 4 tests are changing.

With the last version 087e3f2, we had:
`Tests: 2198, Assertions: 10188, Failures: 7, Skipped: 38.`

With this patch, we have:
`Tests: 2198, Assertions: 10196, Failures: 3, Skipped: 38.`

JeroenDeDauw (approved):

PR #5415, opened 2023-01-31
Add refreshData() return type

### LINKED ISSUE #5325: method signature mismatch
I'm posing this as a question because, even though this looks obvious, I want to get input from someone else because of how long this has been here.

This is a fairly trivial Q for @JeroenDeDauw:

I was looking at refreshData() today and noticed that the [`@return` type](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/f056d4798afa0e06688e3c3e9068d823eeb6aee8/src/Store.php#L458) does not match [what is actually being returned](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/91546f4c02af0409f97bacecd142a6170d78cde2/src/SQLStore/SQLStore.php#L467).  This appears to be because of [a change](https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/afc1a7ad2107b799a5b869c520d6d53d95694a52) that mwjames made to address design issues.

I realize this isn't a major issue, but it messes with my language server.  Here is an example with VSCode:
[Screencast from 2022-10-27 22-44-41.webm](https://user-images.githubusercontent.com/43581/198478303-16247c49-b580-47d6-ba77-5fb760a4a9f3.webm)

Fixing the type in `@return` is a no-brainer, right?

malberts:
Fixes #5325 

This seems to be the type that is currently being used.

Without knowing anything the class does: it seems a little strange that SPARQLStore is relying on a SQLStore Rebuilder, but maybe that's just the way it is. Perhaps a new interface (or moving it) might be more correct, but I'm not going to look at that right now since there is only one implementation.

### DIFF
--- a/src/Maintenance/DataRebuilder.php
+++ b/src/Maintenance/DataRebuilder.php
@@ -10,6 +10,7 @@ use SMW\Services\ServicesFactory as ApplicationFactory;
 use SMW\DIWikiPage;
 use SMW\MediaWiki\TitleFactory;
 use SMW\Maintenance\DataRebuilder\OutdatedDisposer;
+use SMW\SQLStore\Rebuilder\Rebuilder;
 use SMW\Utils\CliMsgFormatter;
 use SMW\Options;
 use SMW\Store;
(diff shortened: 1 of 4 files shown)


kghbln:
SPARQLStore is, unlike ElasticStore, not a standalone backend for SMW. I am trying to remember precisely however SPARQLStore requires SQLStore for some mapping functionality.

PR #5431, opened 2023-02-11
Fix TypeErrors as noted by on Page Schemas Talk

willcohen:
As noted by
https://www.mediawiki.org/wiki/Extension_talk:Page_Schemas#c-Yaron_Koren-20220728130400-Sneeze5555-20220728075000, trying to create a schema using Page Schemas fails with a TypeError. This change fixes those errors.


willcohen:
This is minor but might be worth making it into a point bug fix release soon. Page Schemas are totally nonfunctional without this fix, at least with current PHP, MW, and SMW, which makes for a very confusing onboarding experience for anyone trying to get their heads around this on a fresh MediaWiki!

willcohen:
Thanks!

PR #5437, opened 2023-02-12
Remove line breaks from the beginning of messages

amire80:
translatewiki cannot conveniently be used to see and insert this linebreak.

However, this doesn't seem to be so necessary anyway, because in the code, each of these messages appears to be used in a separate block element.


JeroenDeDauw:
Can someone verify that this does not cause a regression in the layout?

jaideraf:
I believe there is no harm because those messages are in a bulleted list so,  the line break is done anyway. If you are curious where it is shown: https://sandbox.semantic-mediawiki.net/w/index.php?title=Special:SemanticMediaWiki&action=stats%2Ftable
(it needs permission to see them).

PR #5443, opened 2023-02-18
Fix dynamic property creation

malberts:
Missed one in #5413


malberts:
CI failure is because of a recent TranslateWiki commit that breaks some tests: https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/520e131822e1aa3f1c20c9e051030acc5c12ef7b

PR #5444, opened 2023-02-18
Replace "hereby" with "therefore" in messages

amire80:
If I understand these messages correctly, I think that the word "therefore" is more appropriate grammatically in English.

If I don't understand correctly, feel free to discard it.


hexmode (approved):
Thanks for these.  Your assessment looks like it is correct.

amire80:
> Thanks for these. Your assessment looks like it is correct.

Thanks :)

i18n/en.json
```diff
@@ -836,7 +836,7 @@
...
 	"smw-datavalue-constraint-violation-single-value": "The \"[[Property:$1|$1]]\" property has a <code>single_value</code> constraint and value \"$2\" is violating that requirement.",
 	"smw-constraint-violation-uniqueness": "A <code>unique_value_constraint</code> constraint is assigned to the \"[[Property:$1|$1]]\" property which only permits unique value assignments and the ''$2'' value annotation was already found to be annotated in the \"$3\" subject.",
-	"smw-constraint-violation-uniqueness-isknown": "A <code>unique_value_constraint</code> constraint is assigned to the \"[[Property:$1|$1]]\" property hereby only permits unique value annotations, ''$2'' already contains an annotated value with \"$3\" violating the uniqueness constraint for the current subject.",
+	"smw-constraint-violation-uniqueness-isknown": "A <code>unique_value_constraint</code> constraint is assigned to the \"[[Property:$1|$1]]\" property, therefore unique value annotations are permitted. ''$2'' already contains an annotated value with \"$3\", violating the uniqueness constraint for the current subject.",
```
malberts, line 839: I think the word "only" should be put back here, otherwise the sentence no longer explicitly excludes non-unique values.

amire80: My bad. Sending a fix.

amire80: #5445

PR #5448, opened 2023-02-20
Correct a message about import performer

amire80:
It's unclear to what does "means" refer. This patch clarifies it.


amire80:
Thanks for all the reviews of my tiny patches, @malberts ! It means a lot <3

This is probably the last one (or close to that).

PR #5463, opened 2023-03-10
Use ParserOutput::getCategoryNames() instead of ::getCategories()

cscott:
ParserOutput::getCategories() returns a reference to an internal
array and we'd like to discourage its use.

ParserOutput::getCategoryNames() was added in MW 1.38 so minimum
required MW version was bumped accordingly.

## Summary by CodeRabbit

- **Documentation**
	- Updated the required MediaWiki version from 1.35 to 1.38 to ensure compatibility with newer features and security updates.


Nikerabbit:
This has actually been resolved already in master with BC code (obvious after resolving conflicts).

PR #5465, opened 2023-03-16
Control display of namespace in printouts #5400 (rebase)

thomas-topway-it:
fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5400

Adds askquery parameter `prefix` with values `'all', 'subject', 'none', 'auto' `

* The PR does not affect the default behavior of askquery results
* The new parameter does not break the abstraction of the SMWDataValue class and sub-classes and as such does not change the interface of existing methods with additional parameters (only the class `ValueTextsBuilder` is created with a constructor taking the SMWQueryResult.) Instead, the existing SMWDataValue's `getShortText` and `getLongText` methods are concurrently used based on the new parameter
* The new parameter also introduces an option "auto", checking whether the class  `SMW\Query\Language\Description`, used the create the query, is an instance of `NamespaceDescription`, which ensures that namespaces of the query results aren't mixed.
* The PR does not make use of global parameters or global variables in order to work


thomas-topway-it:
@see https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5408 (rebased)

thomas-topway-it:
I have resync, is it enough ?

https://github.com/Knowledge-Wiki/SemanticMediaWiki/tree/Control-display-of-namespace-in-printouts-%235400-(b)

PR #5466, opened 2023-03-17
Fix sort order on Special:PropertyLabelSimilarity

mmeller-wikia:
Fix regression from https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5464 that changed sorting order.


malberts:
I guess the tests are not running into this.

PR #5473, opened 2023-03-23
Update semanticdata.access.md

YOUR1:
Added SubSemanticData example


Toppole6:
> Added SubSemanticData example

krabina (commented):
Since this is documentation, this should be ready to be integrated. What do you think @JeroenDeDauw @gesinn-it-gea ?

PR #5478, opened 2023-04-07
Fix "use of ParserOutput::addModules with non-array argument was deprecated in MediaWiki 1.38"

D-Groenewegen:


Seb35:
This PR is no more useful:
* the occurences of `ParserOutput::addModules()` were modified in 0a233c2ccb and 058592215,
* and it is not mandatory to use arrays for `OutputPage::addModules()`, it may be though (if it is a string it is converted to an array, [see here](https://github.com/wikimedia/mediawiki/blob/b85439f1a03f53bb0c17914aa5a0fb5b45ca69ca/includes/Output/OutputPage.php#L700-L702))

PR #5490, opened 2023-05-26
DuplicateFinder: Avoid primary DB query on pageview

mszabo-wikia:
As far as I can tell, this only gets called via
DeclarationExaminer::check() when a Property: page is viewed. So, switch it to use a replica DB instead to avoid generating undue load on the primary.


mszabo-wikia:
`\SMW\SQLStore\EntityStore\PropertySubjectsLookup::fetchFromTable` can probably use the same treatment as well. That one is used more extensively (via `getPropertySubjects`) but ultimately all from places where the current process shouldn't need to read data it has just written, so a replica should be fine.

r0bj (approved):

PR #5491, opened 2023-05-30
Use PagerNavigationBuilder to build prev/next pager links

Abijeet:
PagerNaviagationBuilder was introduced in MW 1.39

PrevNextNavigationRenderer was deprecated in MW 1.39

Bug: https://phabricator.wikimedia.org/T332687

### DIFF
--- a/src/MediaWiki/MessageBuilder.php
+++ b/src/MediaWiki/MessageBuilder.php
@@ -4,8 +4,10 @@ namespace SMW\MediaWiki;
 
 use IContextSource;
 use Language;
+use MediaWiki\Navigation\PagerNavigationBuilder;
 use MediaWiki\Navigation\PrevNextNavigationRenderer;
 use Message;
+use RequestContext;
 use RuntimeException;
 use Title;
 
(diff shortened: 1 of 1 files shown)


Abijeet:
The code has been picked up from: https://gerrit.wikimedia.org/r/c/mediawiki/core/+/831370/6/includes/Navigation/PrevNextNavigationRenderer.php#65

src/MediaWiki/MessageBuilder.php
```diff
@@ -98,11 +100,31 @@ public function listToCommaSeparatedText( array $list ) {
...
+
+			if ( $offset > 0 ) {
+				$navBuilder->setPrevLinkQuery( [ 'offset' => (string)max( $offset - $limit, 0 ) ] );
+			}
+
+			if ( !$isAtTheEnd ) {
+				$navBuilder->setNextLinkQuery( [ 'offset' => (string)( $offset + $limit ) ] );
+			}
+
+			return $navBuilder->getHtml();
```
JeroenDeDauw, line 123: I think it is much cleaner if you put the body of the if statement into its own private method.

Or, alternatively, perhaps the if-else is not needed, and the method body can be replaced by the if body. Would need to check that this works down to MW 35 though, so is more work, but ends up with simpler code.

JeroenDeDauw: I still recommend you extract into a private method.

Issue #4971, opened 2021-04-12
Querying for "imported from" property produces weird printout


krabina:
### Setup

MediaWiki | 1.31.14 (f85ff27)22:33, 8 April 2021
-- | --
PHP | 7.2.24-0ubuntu0.18.04.7 (apache2handler)
MariaDB | 10.1.47-MariaDB-0ubuntu0.18.04.1
ICU | 60.2
LuaSandbox | 3.0.1
Lua | 5.1.5

### Issue

When using the most simple query for the property "imported from" it results in strange printouts

```
{{#ask: [[Imported from::+]]
|?Imported from
|format=plainlist
}}
```

result in something like: `<a target="_blank" rel="noreferrer noopener" class="external text" href="http://xmlns.com/foaf/0.1/">foaf</a>`

Steps to reproduce the observation (recommendation is to use the [sandbox](https://sandbox.semantic-mediawiki.org)):
https://www.semantic-mediawiki.org/wiki/User:Krabina_Bernhard/ontology

Querying for this property is important in order to document your usage of external vocabularies.

krabina:
https://sandbox.semantic-mediawiki.org/wiki/Issue/4971

kghbln:
I beautified the output a bit to make it even easier to see what is happening. Yeah, this indeed needs some work.

krabina:
The problem is still there in MW 1.35 and PHP 7.3, btw.

thomas-topway-it:
see here for a possible solution
https://github.com/Knowledge-Wiki/SemanticMediaWiki/commit/3fe85b6b2a082e1877e28e2b65f7deb431d2a610

there are 2 issues in the same time, there was also the caption of the term in the parenthesis that was undefined (so I've created a new function, 'getDeclarativeName')

krabina:
Closed via #5606

PR #5492, opened 2023-05-30
query with "imported from" property

thomas-topway-it:
fixes issue https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4971

- fixes mixed wikitext/html in Imported from printout caption
- fixes empty declarativeName (i.e. vocabulary name used as caption of the imported property between parenthesis) (see here https://www.semantic-mediawiki.org/wiki/User:Krabina_Bernhard/ontology)


gesinn-it-gea:
@thomas-topway-it if you are still working on PRs, please set them to "draft". Otherwise we get spammed by commit notifications. This way it is also clear for the reviewers, that they should not start with the review yet.

thomas-topway-it:
@gesinn-it-gea ah, right, I hoped nobody else was subscribed to the pull request. In fact I had to do a lot of tests on the remote branch itself since there was a discrepancy with my local tests. They fail because `RevisionGuard` in the file `JSONScriptServicesTestCaseRunner.php` is null and on the other side it cannot be a mockup  like the following
```
		$revisionGuard = $this->getMockBuilder( '\SMW\MediaWiki\RevisionGuard' )
			->getMock();
```

 otherwise other tests not related to the current edits will fail. I have now verified that setting a revisionGuard with the following 
`
$mediaWikiNsContentReader->setRevisionGuard( SMW\Services\ServicesFactory::getInstance()->singleton( 'RevisionGuard' ) );
`
seems to set a null revisionGuard. So following I have to check this point.

thomas-topway-it:
@gesinn-it-gea it is now ready for review. I have added a test, based on a simple query like the following

```
{{#ask: [[Imported from::+]]
|?Imported from
|format=plainlist 
}}   
```

the test fails with the original class and succeeds with the updated class. (the original class returns a result like the following, https://www.semantic-mediawiki.org/wiki/User:Krabina_Bernhard/ontology, with escaped html and without captions)

thomas-topway-it:
@JeroenDeDauw can this be finally merged ?

YOUR1:
Seems good to me.

krabina:
Reviewd by @YOUR1 (thank you!). Can we merge, @gesinn-it-gea @JeroenDeDauw ?

krabina:
Closiing this because of the rebased https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5606

Issue #5342, opened 2022-12-06
Property links don't work in de-formal language


krabina:
### Setup

Software | Version
-- | --
MediaWiki | 1.35.8
PHP | 7.4.3 (fpm-fcgi)
MariaDB | 10.3.37-MariaDB-0ubuntu0.20.04.1
ICU | 66.1
Elasticsearch | 6.8.23
Semantic MediaWiki | 4.0.2

### Issue

The [@@@-Syntax to create property links](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1855) does work when the property _Has property description_ is set so some description:
` [[Has property description::Testdescription@de]]` 

It is working if the user has "de" as their language set. However, If a user sets "de-formal" the mouse-over-effect is not working. 

Trying to set ` [[Has property description::Testdescription@de-formal]]` fails with the massage that is not a valid language code.

So I guess this must be some kind of bug. If I cannot set de-formal as language value in monolinual text properties than somehow _de-formal_ (and _de-AT_, _de-CH_) should fallback to _de_.

See also https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1865 and https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/3562

krabina:
Thank you, it is working for me!

krabina:
Unfortunately, this does not solve the issue for me. You can see it here:
https://test.knowledge.wiki/Adolf_Loos
the "Objektbezug" property should be called "Siehe auch": https://test.knowledge.wiki/Attribut:Objektbezug
![grafik](https://github.com/SemanticMediaWiki/SemanticMediaWiki/assets/4318745/ff34166a-9c85-4c8b-be08-736b1319effa)
When I switch from de-forma. to de (as user), the correct name is shown:
![grafik](https://github.com/SemanticMediaWiki/SemanticMediaWiki/assets/4318745/63f90b9a-ae32-43f2-9c97-ea39df3e96d4)
This happens only for non-logged-in users, btw.

thomas-topway-it:
@JeroenDeDauw I'm waiting for some feedback, here https://test.knowledge.wiki/index.php?title=Attribut:Objektbezug&action=edit  I don't see de-format specified

krabina:
That is right. I was not expecting having to set a property description to de-formal, I was expecting it to fall back to de if the wiki is set to de-formal.
[...]

PR #5493, opened 2023-05-31
Fixes Property links don't work in de-formal language

thomas-topway-it:
fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5342

However before pulling the edits, it should be further tested and the following issue should be addressed:

- to what extent the tag de-x-formal is used in MW and SMW and why de-formal isn't completely replaced with de-x-formal in i18n files ? (or on the contrary de-formal inserted in the language-subtag-registry ? https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

otherwise, if we keep the mapping, we have to ensure that every time the language code is handled, it expects either 'de-formal' or de-x-formal' (the present pull request do it within the scope of SMW, but not of MW)

Also refer to the following:

- https://schneegans.de/lv/?tags=de-x-formal&Format=text
- https://doc.wikimedia.org/mediawiki-core/1.33.0/php/LanguageCode_8php_source.html
- https://phabricator.wikimedia.org/T125073
- https://datatracker.ietf.org/doc/html/rfc5646 (Private use subtag)


thomas-topway-it:
@gesinn-it-gea and @kghbln do you want to take a look ?

src/DataValues/MonolingualTextValue.php
```diff
@@ -277,7 +289,12 @@ public function getTextValueByLanguageCode( $languageCode ) {
 			return null;
 		}
 
-		if ( $list['_LCODE'] !== Localizer::asBCP47FormattedLanguageCode( $languageCode ) ) {
+		$mappedLanguageCode = !array_key_exists( $list['_LCODE'], $this->nonstandardLanguageCodeMapping ) ? $list['_LCODE']
```
Nikerabbit, line 292: More idiomatic way to write this is `$mappedLanguageCode = $this->nonstandardLanguageCodeMapping[$list['_LCODE']] ?? $list['_LCODE']`

src/DataValues/LanguageCodeValue.php
```diff
@@ -49,15 +64,22 @@ protected function parseUserValue( $userValue ) {
 
 		// Checks whether the language tag is valid in MediaWiki for when
 		// it is not executed in a query context
-		if ( !$this->getOption( self::OPT_QUERY_CONTEXT ) && !Localizer::isKnownLanguageTag( $languageCode ) ) {
+		
+		// ensure non-standard language codes are mapped to
+		// their canonical form (e.g. de-x-formal to de-formal)
+
+		$mappedLanguageCode = !in_array( $languageCode, $this->nonstandardLanguageCodeMapping ) ? $languageCode
```
Nikerabbit, line 71: `in_array` isn't needed. `array_search( $languageCode, $this->nonstandardLanguageCodeMapping ) ?: $languageCode` does the same thing.

Nikerabbit: Yes bcp47 should be fine to use when formatting language codes for display. The list of deprecated language codes is changing over time, so it's possible the behavior is not the same across all MediaWiki versions.

jaideraf:
Due to https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5530 merge, this one can be closed, right?

JeroenDeDauw:
> Due to #5530 merge, this one can be closed, right?

Dunno, @krabina reported the underlying issue (https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5342) was not fixed. So perhaps we should revert #5530.

PR #5498, opened 2023-06-23
fix SPARQL "data" protocol (GSP)

hkwi:
This patch fixes SMW to send turtle content directly as SPARQL 1.1 GSP POST.
https://www.w3.org/TR/sparql11-http-rdf-update/#http-post

The problem was that internal curl_handler reused the curl options of previous call, and CURLOPT_INFILE did not set the HTTP content.

There was a comment about 4store in GenericRepositoryConnector, but FourstoreRepositoryConnector overrides this part.
4store POST /data/ endpoint only supports content-type of form-urlencoded. https://github.com/4store/4store/blob/fe348295e0e20e1b2192dd08e3ef380d9750ff77/src/http/httpd.c#L1397

In SPARQL 1.1 Graph Store HTTP Protocol (GSP) spec, POST with RDF graph content is the first class, and multipart/form-data may be used as well.

By the way, sending in multipart/form-data is performed by CURLOPT_POSTFIELDS with array of CURLFile instances.


paladox:
This was changed in https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/2eefb599808542a4573db95bc1096ffa19c0f242

Replaces Onoi\HttpRequest with MediaWiki HttpRequestFactory.

Is this change still needed? If so it will need merge conflicts being fixed and to adapt.

hkwi:
Not verified, but that replacement code looks good.

Issue #5510, opened 2023-07-20
update.php must be run twice PHP 8.1 MAINTENANCE_MODE


HamishSlater:
### Setup

- SMW version: 4.1.1 & master
- MW version: 1.39.4
- PHP version: 8.1.20
- DB system and version: MariaDB 10.11.4

### Issue

`PHP 8.1`
After running `update.php`, the system enters `MAINTENANCE_MODE`.
The 'progress bar' reads 'Adding supplement jobs ...'

Note: Doesn't happen with PHP 8.0.

### Workaround

Run `update.php` twice.

osnard:
Can confirm. This is what is looks like:
```
> # First run
> php maintenance/update.php --quick
...
Done in 3.5 s.
> cat extensions/SemanticMediaWiki/.smw.json 
{
    "...": {
        "upgrade_key": "...",
        "maintenance_mode": false
    }
}
> # Next run
> php maintenance/update.php --quick
...
Done in 3.5 s.
> cat extensions/SemanticMediaWiki/.smw.json 
{
    "...": {
        "upgrade_key": "...",
        "maintenance_mode": {
            "supplement-jobs": 80
        }
    }
}
```

osnard:
Looks like is is again the "`$GLOBALS` must not be passed by reference" issue here: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/4.1.1/src/SetupFile.php#L359-L366

If one replaces this with
```
		foreach ( $args as $key => $value ) {
			// NULL means that the key key is removed
			if ( $value === null ) {
				unset( $GLOBALS[self::SMW_JSON][$id][$key] );
				unset( $vars[self::SMW_JSON][$id][$key] );
			} else {
				$vars[self::SMW_JSON][$id][$key] = $value;
				$GLOBALS[self::SMW_JSON][$id][$key] = $value;
			}
		}
```
the content of the `.smw.json` is stable after each run:
```
{
    "...": {
        "upgrade_key": "...",
        "maintenance_mode": false,
        "latest_version": "4.1.1",
        "incomplete_tasks": [],
        "last_optimization_run": "2023-07-20 11:00"
    }
}
```

HamishSlater:
Yup, fixed with #5511; thanks!

PR #5511, opened 2023-07-20
Fix 'pass by reference use'

HamishSlater:
Fixes update.php must be run twice #5510

`SetupFile::remove` assigned `$vars = $GLOBALS` before calling `write( [ $key => null ], $vars )`

`PHP 8.1` passes `$GLOBALS` by value, so set set `$GLOBALS[$key]` explicitly in `write()`


JeroenDeDauw:
What about the unset calls further down in the method?

Issue #5528, opened 2023-09-05
PHP 8.1 Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in includes/dataitems/SMW_DI_Blob.php on line 26


jaideraf:
### Setup

- SMW version: 4.1.2
- MW version: 1.39.4
- PHP version: 8.1.23 
- DB system (MySQL, Blazegraph, etc.) and version: 11.1.2-MariaDB

### Issue

When running `maintenance/updateEntityCollation.php` I get

```
Deprecated: trim(): Passing null to parameter #1 ($string) of type string is deprecated in 
/w/extensions/SemanticMediaWiki/includes/dataitems/SMW_DI_Blob.php on line 26
```
Here is the spot: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/b955904ef964f7e8b14671f5eb707ef9d6c64da9/includes/dataitems/SMW_DI_Blob.php#L26C1-L26C37

PR #5529, opened 2023-09-06
Avoid null error

JeroenDeDauw:
Fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5528


jaideraf:
Thank you both 👍🏼

Issue #5527, opened 2023-09-02
Imported_from seems to break ask-queries with SPARQL-backend


simontaurus:
### Setup

- SMW version: 4.1.1 (also tested: 3.2.3)
- MW version: 1.39.3 (also tested: 1.35.7)
- PHP version: 7.4.33 (also tested: 8.1)
- DB system (MySQL, Blazegraph, etc.) and version: MySQL 8.0.17, Blazegraph 2.1.5

### Issue

`[[Imported from::...]]` statements seem to break ask-queries when a SPARQL-backend is used.
I asume this is due to the change from the local IRI (/id/<title>) to the imported one (e. g. foaf:Person) as primary identifier.

Steps to reproduce the observation:

1. Create a page with an `[[Imported from::foaf:Person]]` statement, e. g. [ImportedFromTestPage](https://demo.open-semantic-lab.org/wiki/Category:ImportedFromTestPage)
1. Query for that page with `[[Imported_from::foaf:Person]]`, e. g. [https://demo.open-semantic-lab.org/w/api.php?action=ask&query=[[Imported_from::foaf:Person]]](https://demo.open-semantic-lab.org/w/api.php?action=ask&query=[[Imported_from::foaf:Person]])
1. Result is empty while in the [sandbox instance](https://sandbox.semantic-mediawiki.net/wiki/Cat%C3%A9gorie:ImportedFromTestPage) it works as expected

simontaurus:
Update: 

The test query above works after purging the page and the iri replacement within the SPARQL query seems fine, so `category:ImportedFromTestPage` gets replaced with `foaf:Person`

But when I set the displaytitle to ImportedFromTestPageDisplayTitle with `{{DISPLAYTITLE: ImportedFromTestPageDisplayTitle}}`,  the query https://demo.open-semantic-lab.org/w/api.php?action=ask&query=[[:Category:ImportedFromTestPage]]|?Imported_from returns
```json
        "results": {
            "Category:ImportedFromTestPageDisplayTitle": {
                "printouts": {
                    "Imported from": []
                },
                "fulltext": "Category:ImportedFromTestPageDisplayTitle",
                "fullurl": "https://demo.open-semantic-lab.org/wiki/Category:ImportedFromTestPageDisplayTitle",
                "namespace": 14,
                "exists": "",
                "displaytitle": ""
            }
        },
```
[...]

PR #5531, opened 2023-09-17
fix: wrong dbkey lookup for imported vocab

simontaurus:
Refs #5527

Please let me know how to address the actual lookup beside the early return if a valid dataItem exists

### DIFF
--- a/src/Exporter/DataItemMatchFinder.php
+++ b/src/Exporter/DataItemMatchFinder.php
@@ -57,6 +57,12 @@ class DataItemMatchFinder {
 			return $dataItem;
 		}
 
+		$dataItem = $expElement->getDataItem();
+		if ( $dataItem !== null && $dataItem instanceof DIWikiPage ) {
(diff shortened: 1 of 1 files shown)


src/Exporter/DataItemMatchFinder.php
```diff
@@ -57,6 +57,12 @@ public function matchExpElement( ExpElement $expElement ) {
 			return $dataItem;
 		}
 
+		$dataItem = $expElement->getDataItem();
+		if ($dataItem && $dataItem instanceof DIWikiPage) {
```
JeroenDeDauw, line 61: Need spaces to match code style and an explicit check against null or false or whatever the value might be:

`if ( $dataItem !== null && $dataItem instanceof DIWikiPage ) {`

simontaurus: thanks, updated accordingly

PR #5532, opened 2023-09-17
Update Simplified Chinese translations

LittlePaw365:
add new translations and modify existing translations for Simplified Chinese

### DIFF
--- a/i18n/extra/SemanticMediaWiki.alias.php
+++ b/i18n/extra/SemanticMediaWiki.alias.php
@@ -814,12 +814,17 @@ $specialPageAliases['zh-hans'] = [
 	'PageProperty' => [ '页面属性' ],
 	'Properties' => [ '属性' ],
 	'Concepts' => [ '概念' ],
-	'SMWAdmin' => [ 'SMW管理' ],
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Need sanity check from someone that understands Chinese

LittlePaw365:
Explanation of some major changes

1. Removed '型' in the names of types: ‘型’ is a short form of ‘类型’, which means ‘type’. It is a common translation pattern in Chinese in some old-fashioned textbooks, however, if ‘型’ is not removed, it is redundant in the translations of system messages like ‘Properties of type […]’, and does not make sense in others like ‘[…] is a type and predefined property …’
2. Changed ‘具有' to '拥有': In Chinese, '具有' usually means that something inherently possesses a certain quality, and often takes abstract things as its object, while '拥有' is a more universal word that can take both concrete and abstract things as its object, and is more suitable for translating phrases like 'Has fields', 'Has query', 'Has area'.

LittlePaw365:
@JeroenDeDauw sorry to bother you. I used an email address that I shouldn't have used in this pull request, so I closed it and I will open a new one. I'm sorry for causing these troubles.

PR #5534, opened 2023-09-27
Elastic/Indexer/Bulk::$response can also be a string

wgevaert:
That is, if there was an error.

But you should not get a useless TypeError then, you should get a useful error.

### DIFF
--- a/src/Elastic/Indexer/Bulk.php
+++ b/src/Elastic/Indexer/Bulk.php
@@ -35,7 +35,7 @@ class Bulk implements JsonSerializable {
 	private $head = [];
 
 	/**
-	 * @var array
+	 * @var array|string
 	 */
 	private $response = [];
 
@@ -133,9 +133,9 @@ class Bulk implements JsonSerializable {
 	/**
 	 * @since 3.2
 	 *
-	 * @return array
+	 * @return array|string
 	 */
-	public function getResponse() : array {
+	public function getResponse() : array|string {
 		return $this->response;
 	}
 
--- a/src/Elastic/Indexer/Bulk.php
+++ b/src/Elastic/Indexer/Bulk.php
@@ -135,7 +135,7 @@ class Bulk implements JsonSerializable {
 	 *
 	 * @return array|string
 	 */
-	public function getResponse() : array|string {
+	public function getResponse() {
 		return $this->response;
 	}
 


JeroenDeDauw:
This can only go into SMW 5.x if you use PHP union types, which got introduced in PHP 8.0. See https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/COMPATIBILITY.md

I suppose this code is not tested at all, since I doubt the MW 35 tests run with PHP 8.0. Or do they? @gesinn-it-wam @gesinn-it-gea

wgevaert:
TL;DR: I will just remove the type hint, such that it can be merged sooner.

> [...] I doubt the MW 35 tests run with PHP 8.0. Or do they?

It does run on php8.0, see [here](https://github.com/SemanticMediaWiki/SemanticMediaWiki/actions/runs/6322832533/job/17169185781?pr=5534#step:2:137). I believe all docker containers for MW 1.35.6+ use php8.
Note that php7.4 has been EOL for almost a year now. It should not be used (Even though [wikipedia still does](https://en.wikipedia.org/wiki/Special:Version)).

> This can only go into SMW 5.x if you use PHP union types, which got introduced in PHP 8.0. See https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/COMPATIBILITY.md

I see. I will then just remove the type hint, such that it can be merged sooner.

The problem I get without this change is the following during `php extensions/SemanticMediaWiki/maintenance/rebuildElasticIndex.php`:
```
TypeError: SMW\Elastic\Indexer\Bulk::getResponse(): Return value must be of type array, string returned
Backtrace:
from /var/www/html/extensions/SemanticMediaWiki/src/Elastic/Indexer/Bulk.php(139)
#0 /var/www/html/extensions/SemanticMediaWiki/src/Elastic/Indexer/Indexer.php(349): SMW\Elastic\Indexer\Bulk->getResponse()
#1 /var/www/html/extensions/SemanticMediaWiki/src/Elastic/Indexer/Rebuilder/Rebuilder.php(317): SMW\Elastic\Indexer\Indexer->indexDocument()
#2 /var/www/html/extensions/SemanticMediaWiki/maintenance/rebuildElasticIndex.php(481): SMW\Elastic\Indexer\Rebuilder\Rebuilder->rebuild()
#3 /var/www/html/extensions/SemanticMediaWiki/maintenance/rebuildElasticIndex.php(419): SMW\Maintenance\rebuildElasticIndex->rebuildFromRow()
#4 /var/www/html/extensions/SemanticMediaWiki/maintenance/rebuildElasticIndex.php(186): SMW\Maintenance\rebuildElasticIndex->rebuild()
#5 /var/www/html/maintenance/includes/MaintenanceRunner.php(309): SMW\Maintenance\rebuildElasticIndex->execute()
#6 /var/www/html/maintenance/doMaintenance.php(85): MediaWiki\Maintenance\MaintenanceRunner->run()
#7 /var/www/html/extensions/SemanticMediaWiki/maintenance/rebuildElasticIndex.php(552): require_once(string)
#8 {main}
```
After which I am left with a dangling lock on my ES database which I have to manually remove (And I don't feel very confident about manually removing locks). When I do not manually remove the lock, the job queue fills with upwards of 50k of `SMW\Elastic\Jobs\IndexerRecoveryJob` type jobs (That keep re-queueing themselves).

Issue #5474, opened 2023-03-23
Special:SearchByProperty does not work with reference datatype


krabina:
### Issue

The [reference datatype](https://www.semantic-mediawiki.org/wiki/Help:Type_Reference) works well in inline queries, but the special page SeachByProperty does not show any results if this datatype is used.

See this example on sandbox:
https://sandbox.semantic-mediawiki.net/wiki/Berlin
In the factbox you can see the magnifying glass icon next to the value. If you click on it, you are directed to the Special:SeachByProperty page. This page does not find any results. 
https://sandbox.semantic-mediawiki.net/w/index.php?title=Sp%C3%A9cial%3ARecherche+par+propri%C3%A9t%C3%A9&property=Has+population&value=3520061

I suppose this bug has existed for a very long time, probably since the introduction of the reference datatype.

PR #5550, opened 2023-11-03
fixes Special:SearchByProperty with reference datatype

thomas-topway-it:
- fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5474
by adding a `destructureDIContainer` method that retrieves the correct DIProperty and dataItem for the [reference](https://www.semantic-mediawiki.org/wiki/Help:Type_Reference) datatype (the first property of the multiValue dataItem) and passing them to `Store` -> `getPropertySubjects`


src/MediaWiki/Specials/SearchByProperty/QueryResultLookup.php
```diff
@@ -207,14 +211,42 @@ private function doQueryForNonValue( PageRequestOptions $pageRequestOptions, Req
 			$requestOptions
 		);
 	}
+	
+	private function destructureDIContainer( $DIProperty, $dataItem, $pageRequestOptions ) {
```
JeroenDeDauw, line 215: Missing types

src/MediaWiki/Specials/SearchByProperty/QueryResultLookup.php
```diff
@@ -207,14 +211,42 @@ private function doQueryForNonValue( PageRequestOptions $pageRequestOptions, Req
...
+		$multiValue = DataValueFactory::getInstance()->newDataValueByItem(
+			$dataItem,
+			$DIProperty
+		);
+		$properties = $multiValue->getPropertyDataItems( $property );
+
+		// Property for the reference value
+		$DIProperty = $properties[0];
```
JeroenDeDauw, line 223: Reusing a variable name like this isn't great

src/MediaWiki/Specials/SearchByProperty/QueryResultLookup.php
thomas-topway-it: types added

src/MediaWiki/Specials/SearchByProperty/QueryResultLookup.php
thomas-topway-it: I've renamed the variable and read the blog page. In this case I wasn't able to find a better way. I've added a comment with some explanation. An alternate way would be to create the correct dataitem manually using `DataValueFactory` however the class `PageRequestOptions` does some additional operation like unescaping the data value. I'm not sure if that can be avoided in this particular context.

Issue #5536, opened 2023-10-05
Special:PageProperty does not work 


krabina:
### Setup

- SMW version: 4.1
- MW version: 1.35/1.39

### Issue
The special page "SpecialPageProperty" does not let you enter a page name and does not retieve results
* https://www.semantic-mediawiki.org/wiki/Special:PageProperty
* https://sandbox.semantic-mediawiki.net/wiki/Sp%C3%A9cial:Propri%C3%A9t%C3%A9s_de_la_page

krabina:
see https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3088

PR #5557, opened 2023-11-13
Fix special page property issue 5536

thomas-topway-it:
fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5536

(the input "From page" in the special page `Special:PageProperty` was not working, due to a change in the class name and the module `ext.smw.autocomplete.page` (from `ext.smw.autocomplete.article` ) entitled to handle the input.

Also, is there a reason to not list this special page, `Special:PageProperty` , in `Special:SpecialPages` , which actually could be used independently ?


jaideraf:
Couldn't this be merged?

Issue #5547, opened 2023-10-28
Error 1091: Can't DROP 'PRIMARY'


krabina:
### Setup

- SMW version: 4.1.2
- MW version: 1.39
- PHP version:7.4
- DB system (MySQL, Blazegraph, etc.) and version: MySQL

### Issue

On a wiki that was already upgraded to 1.39 I changed an extension via composer.local.json. Had to delete the vendor directory due to the composer issue described [here ](https://github.com/SemanticMediaWiki/SemanticMetaTags/issues/76)

On update.php (setupStore.php) the following error shows

```
... changing type of field smw_id from 'INT UNSIGNED NOT NULL AUTO_INCREMENT' to 'INT(11) UNSIGNED NOT NULL KEY AUTO_INCREMENT' ... Wikimedia\Rdbms\DBQueryError from line 1618 of /includes/libs/rdbms/database/Database.php: Error 1091: Can't DROP 'PRIMARY'; check that column/key exists
Function: SMW\SQLStore\TableBuilder\MySQLTableBuilder::doUpdateFieldType
Query: ALTER TABLE `smw_object_ids` DROP PRIMARY KEY

#0 /includes/libs/rdbms/database/Database.php(1602): Wikimedia\Rdbms\Database->getQueryException()
#1 /includes/libs/rdbms/database/Database.php(1576): Wikimedia\Rdbms\Database->getQueryExceptionAndLog()
#2 /includes/libs/rdbms/database/Database.php(952): Wikimedia\Rdbms\Database->reportQueryError()
#3 /includes/libs/rdbms/database/DBConnRef.php(103): Wikimedia\Rdbms\Database->query()
#4 /includes/libs/rdbms/database/DBConnRef.php(293): Wikimedia\Rdbms\DBConnRef->__call()
#5 /extensions/SemanticMediaWiki/src/SQLStore/TableBuilder/MySQLTableBuilder.php(231): Wikimedia\Rdbms\DBConnRef->query()
#6 /extensions/SemanticMediaWiki/src/SQLStore/TableBuilder/MySQLTableBuilder.php(201): SMW\SQLStore\TableBuilder\MySQLTableBuilder->doUpdateFieldType()
#7 /extensions/SemanticMediaWiki/src/SQLStore/TableBuilder/MySQLTableBuilder.php(131): SMW\SQLStore\TableBuilder\MySQLTableBuilder->doUpdateField()
[...]

PR #5561, opened 2023-11-26
Properly truncate sortkeys in TableFieldUpdater.php

Nikerabbit:
The old code had two issues: it used mb_substr which counts characters and not bytes, and it was called before asking the collator to create the sortkey, which may be longer than the input.


krabina:
Thank you! Any chance this might be related to #5547?

Nikerabbit:
I don't see connection between these two.

Issue #5555, opened 2023-11-11
API call to `run-entity-examiner` throws exception


octfx:
### Setup

- SMW version: 4.1.2 (master branch)
- MW version: 1.39.5
- PHP version: 8.1
- DB system (MySQL, Blazegraph, etc.) and version: MariaDb

### Issue

Page views try to run the `run-entity-examiner` task. This call started failing with the following exception approx. 2 months ago (?):

```
Exception caught: Invalid language code "{{uselang}}"

MWException at /var/www/html/includes/language/LanguageFactory.php(158)from /var/www/html/includes/language/LanguageFactory.php(158)
#0 /var/www/html/includes/language/LanguageFactory.php(144): MediaWiki\Languages\LanguageFactory->newFromCode(string)
#1 /var/www/html/includes/libs/MapCacheLRU.php(274): MediaWiki\Languages\LanguageFactory->MediaWiki\Languages\{closure}()
#2 /var/www/html/includes/language/LanguageFactory.php(145): MapCacheLRU->getWithSetCallback(string, Closure)
#3 /var/www/html/includes/language/LanguageFactory.php(128): MediaWiki\Languages\LanguageFactory->getRawLanguage(string)
#4 /var/www/html/includes/language/Message.php(840): MediaWiki\Languages\LanguageFactory->getLanguage(string)
#5 /var/www/html/extensions/SemanticMediaWiki/src/Setup.php(243): Message->inLanguage(string)
#6 [internal function]: SMW\Setup->SMW\{closure}(array, string)
#7 /var/www/html/extensions/SemanticMediaWiki/src/Localizer/Message.php(233): call_user_func_array(Closure, array)
#8 /var/www/html/extensions/SemanticMediaWiki/src/Localizer/MessageLocalizerTrait.php(44): SMW\Localizer\Message::get(array, integer, string)
#9 /var/www/html/extensions/SemanticMediaWiki/src/Indicator/EntityExaminerIndicators/EntityExaminerDeferrableCompositeIndicatorProvider.php(229): SMW\Indicator\EntityExaminerIndicators\EntityExaminerDeferrableCompositeIndicatorProvider->msg(string, integer, string)
[...]

PR #5564, opened 2023-11-27
Sanitise language code in Api::execute

paladox:
#5555


Nikerabbit:
This just fixes the symptom, not the cause, right? Why is the front-end passing invalid language code?

paladox:
> This just fixes the symptom, not the cause, right? Why is the front-end passing invalid language code?

Yeh.

Nikerabbit (approved):

Nikerabbit:
This fix makes sense in addition to the root cause fix in 28e94fe.

PR #5571, opened 2023-12-18
Fix "PHP Deprecated: Creation of dynamic property SMW\\Localizer\\LocalLanguage\\LocalLanguage::$months is deprecated"

paladox:


src/Localizer/LocalLanguage/LocalLanguage.php
```diff
@@ -49,6 +49,11 @@ class LocalLanguage {
 	 */
 	private $monthMap = [];
 
+	/**
+	 * @var array
+	 */
+	private $months = [];
```
JeroenDeDauw, line 55: FYI you can use `private array $months = []`

PR #5573, opened 2023-12-26
Fix PHP deprecation warning in SemanticDataLookup.php

Nikerabbit:
```
PHP Deprecated: str_replace(): Passing null to parameter #3 ($subject) of type array|string
is deprecated at SemanticDataLookup.php:733

str_replace('#', '|', NULL)
```

### DIFF
--- a/src/SQLStore/EntityStore/SemanticDataLookup.php
+++ b/src/SQLStore/EntityStore/SemanticDataLookup.php
@@ -729,7 +729,7 @@ class SemanticDataLookup {
 
 		// Avoid issues with `$row->$sortField` containing other `#` as for
 		// example in case of a subobject name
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
@gesinn-it-gea tests are all broken for some reason

JeroenDeDauw (approved):

Dinoguy1000:
There appears to be a typo in the added check: `softField` instead of `sortField` (i.e. `f` instead of `r`).

Nikerabbit:
Thanks for the fix. Who should merge?

Nikerabbit:
And it doesn't work :/. This is what I get for not properly testing my patches. Will submit a follow-up to this.

Issue #5585, opened 2024-01-26
Semantic Data Sorting Issue with MediaWiki 1.38+ and DisplayTitle Extension


YvarRavy:
### Setup

- SMW version: 4.1.2
- MW version: 1.38 and higher
- PHP version: 7.4
- DB system (MySQL, Blazegraph, etc.) and version: MariaDB-10.3.38

### Issue

**Description:**
According to [SMW DisplayTitle of Page](https://www.semantic-mediawiki.org/wiki/Help:Special_property_Display_title_of#Features_and_limitations), "Unless no other [variable DEFAULTSORTKEY](https://www.mediawiki.org/wiki/Help:Magic_words#Defaultsort) is annotated, the DISPLAYTITLE value will be used as SORTKEY and replaces the content of the database sortfield." However, starting from MW version 1.38 and higher, this behavior is not working correctly. Instead, the PageTitle is used as the default Sortkey. So that causes the sorting to be messed up

**Steps to reproduce:**
1. Ensure that the DisplayTitle extension is installed and configured correctly with the following settings:
   ```php
   $wgAllowDisplayTitle = true;
   $wgRestrictDisplayTitle = false
   ```

2. create the following pages:

    Category:Test

    Page1:
   ```
   {{DISPLAYTITLE:BTest}}
   [[Category:Test]]
   ```
   Page2:
   ```
   {{DISPLAYTITLE:ATest}}
   [[Category:Test]]
   ```
   Page with Ask Query:
   ```
   {{#ask: [[Category:Test]]
    |format=broadtable
    |sort=
    |order=asc
    |propsep=, 
    |valuesep=, 
    |sep=, 
   }}
   ```

PR #5586, opened 2024-01-26
When using |sort= the default sorting value isn't being applied

YvarRavy:
Issue described in: #5585


YOUR1:
@JeroenDeDauw care to take a look, since this one is quite nasty and a hotfix would be good...

JeroenDeDauw:
@YOUR1 did you test the patch? If so, on what MediaWiki version(s)?

@YvarRavy could you rebase this against master? All the tests are failing, hopefully due to unrelated reasons and fixed by rebase

mszabo-wikia (commented):
I think the bug might actually be in `DisplayTitlePropertyAnnotator.php`. It checks whether the provided `defaultsort` is an empty string, and only then applies the displaytitle as the sort key. However, on newer MW, if `defaultsort` is absent then it will be `null`, not an empty string. Perhaps we just need a corresponding null check in `ParserAfterTidy.php`?

mszabo-wikia:
https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5592 seems to confirm that assumption, can you check if that works for you?

YvarRavy:
@mszabo-wikia let me check. i didnt see your comment untill now. give me a moment tho i'm having some trouble with the tests at the moment

mszabo-wikia:
Thanks for the update. I should clarify—I meant that it should be possible to fix the already existing logic that controls displaytitle->defaultsort behavior by adding such a null check there, instead of adding it to this patch. https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5592 does so.

YvarRavy:
> Thanks for the update. I should clarify—I meant that it should be possible to fix the already existing logic that controls displaytitle->defaultsort behavior by adding such a null check there, instead of adding it to this patch. #5592 does so.

well i haven't added your patch to this. i'm pretty sure the test still would've failed even with your patch. since the _DTITLE was "overwriting" the DEFAULTSORT. so i think its still a good idea to check if DEFAULTSORT was set or not

mszabo-wikia:
Yup—but in the end, adding all this logic to ParserData should not be necessary, because DisplayTitlePropertyAnnotator already tries to do this, it just fails on MW 1.38+ because it gets `null` instead of an empty string when a defaultsort is absent, so it does not fall back to the displaytitle in that case. That's why I would prefer https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5592, as it should fix the existing logic instead of duplicating it in a different place.

YvarRavy:
> Yup—but in the end, adding all this logic to ParserData should not be necessary, because DisplayTitlePropertyAnnotator already tries to do this, it just fails on MW 1.38+ because it gets `null` instead of an empty string when a defaultsort is absent, so it does not fall back to the displaytitle in that case. That's why I would prefer #5592, as it should fix the existing logic instead of duplicating it in a different place.

i agree with what you are saying. but the makeSortKey function in SQLStoreUpdate.php uses the last item in the PropertieValues, when we don't have this check it will always use _DTITLE even when DEFAULTSORT is defined and in my opinion the it shouldn't use that value and should just work in de old way without DTITLE. So while it doesn't have any effect on the sorting itself i would assume that we want to keep the Data in the DB the correct value. 
example:
with check
afbeelding
without check
afbeelding

mszabo-wikia:
Hmm, what were your steps to reproduce that? With https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5592 checked out, I created two pages:
* `PageWithDisplayTitle` with content `{{DISPLAYTITLE:Test}}`
* `PageWithDTAndSort` with content `{{DISPLAYTITLE:Test2}}{{DEFAULTSORT:Test3}}`

This appears to behave correctly—the DT is used as the sort key for the first page, while the defaultsort is preferred for the second:
```
*************************** 49. row ***************************
            smw_id: 516
     smw_namespace: 0
         smw_title: PageWithDisplayTitle
            smw_iw: 
     smw_subobject: 
       smw_sortkey: PageWithDisplayTitle
          smw_sort: PageWithDisplayTitle
smw_proptable_hash: a:2:{s:12:"smw_fpt_inst";s:32:"29d19f562c8e15897d3766b4d8202e90";s:12:"smw_fpt_mdat";s:32:"1094a71722d1dec42302eb335921d3a1";}
          smw_hash: 74e3e38016a0e5f0f2d068f76dd34423cf3e4c12
           smw_rev: 108
       smw_touched: 20240130112926
*************************** 50. row ***************************
            smw_id: 517
     smw_namespace: 0
         smw_title: PageWithDTAndSort
            smw_iw: 
     smw_subobject: 
       smw_sortkey: Test3
          smw_sort: Test3
smw_proptable_hash: a:2:{s:14:"smw_fpt_dtitle";s:32:"de11bda771555192f50700b59e71a457";s:12:"smw_fpt_mdat";s:32:"b2d5f7596dbcde882a40c2ee6144e32c";}
          smw_hash: 7be2f450f7c4b2fb01dc282614a3cf53b413df5d
           smw_rev: 109
       smw_touched: 20240130113024
```

YvarRavy:
i haven't touched #5592
i made one page with the content of: 
```
{{DISPLAYTITLE:UsingDTitle}}
{{DEFAULTSORT:UsingDefaultSortKey}}
[[Category:Test]]
```

with the added if statement checking if the defaultsortkey exists
the smw_sort will be "UsingDefaultSortKey"
without this check
the smw_sort will be "UsingDTitle"
and since the defaultsortkey is defined it should put that one in the DB instead of de DTITLE

mszabo-wikia:
Sorry, I will try to rephrase:

SMW currently has logic, in DisplayTitlePropertyAnnotator, to derive the sortkey from either the defaultsort (if set), or else the displaytitle (if set), or else the page title. This logic breaks on MW 1.38+ because 42b895844b697c0f0636911e1551d4ff6404bcf4 did not account for the fact that the defaultsort may be null when retrieved with the new `getPageProperty` API.

Your patch, as-is, adds a workaround for that problem, effectively duplicating the existing logic in DisplayTitlePropertyAnnotator. So now we have two pieces of code attempting to solve the same problem: the partially working old approach in DisplayTitlePropertyAnnotator and the new code added in this patch to ParserData.

What I am proposing in #5592 instead is to fix the existing logic instead of duplicating it in a second place.

Hope this makes sense!

jaideraf:
@JeroenDeDauw I think https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5592 should be merged and @YvarRavy should install the master version to see if it is what he expects.

YvarRavy:
i understand what you are saying @mszabo-wikia but the problem that i have doesn't seem to be fixed by what your fix. when i create the pages as mentioned in my issue. the same behaviour still exists.

mszabo-wikia:
Hmm, I just created a page with the content you gave:
```
{{DISPLAYTITLE:UsingDTitle}}
{{DEFAULTSORT:UsingDefaultSortKey}}
[[Category:Test]]
```
and got:
```
*************************** 51. row ***************************
            smw_id: 518
     smw_namespace: 0
         smw_title: TestSortFix
            smw_iw: 
     smw_subobject: 
       smw_sortkey: UsingDefaultSortKey
          smw_sort: UsingDefaultSortKey
smw_proptable_hash: a:3:{s:12:"smw_fpt_inst";s:32:"dc7398e67a77fc9012131acfd03694e9";s:14:"smw_fpt_dtitle";s:32:"acdbd936a01f6df43a7a960bec4b7b02";s:12:"smw_fpt_mdat";s:32:"4a67d7ac973167c5ca3031c07a2436a2";}
          smw_hash: 11359ebc07758fcfbd3c4ad70d4943402b95f3d1
           smw_rev: 113
       smw_touched: 20240130121742
```

Is the patch checked out correctly?

mszabo-wikia:
> > Hmm, I just created a page with the content you gave:
> > ```
> > {{DISPLAYTITLE:UsingDTitle}}
> > {{DEFAULTSORT:UsingDefaultSortKey}}
> > [[Category:Test]]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and got:
> > ```
> > *************************** 51. row ***************************
> >             smw_id: 518
> >      smw_namespace: 0
> >          smw_title: TestSortFix
> >             smw_iw: 
> >      smw_subobject: 
> >        smw_sortkey: UsingDefaultSortKey
> >           smw_sort: UsingDefaultSortKey
> > smw_proptable_hash: a:3:{s:12:"smw_fpt_inst";s:32:"dc7398e67a77fc9012131acfd03694e9";s:14:"smw_fpt_dtitle";s:32:"acdbd936a01f6df43a7a960bec4b7b02";s:12:"smw_fpt_mdat";s:32:"4a67d7ac973167c5ca3031c07a2436a2";}
> >           smw_hash: 11359ebc07758fcfbd3c4ad70d4943402b95f3d1
> >            smw_rev: 113
> >        smw_touched: 20240130121742
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Is the patch checked out correctly?
> 
> the data you provide right now is correct. but when you dont use DEFAULTSORT, you will see that the smw_sort will be the pagetitle but in 1.37 and lower it will use the displaytitle which should be the correct behaviour according to https://www.semantic-mediawiki.org/wiki/Help:Special_property_Display_title_of#Features_and_limitations

I see, thanks.

I created a page with the content:
```
{{DISPLAYTITLE:UsingDTitle}}
[[Category:Test]]
```
accordingly, but it still seems to work fine:
```
*************************** 53. row ***************************
            smw_id: 520
     smw_namespace: 0
         smw_title: TestSortFix2
            smw_iw: 
     smw_subobject: 
       smw_sortkey: UsingDTitle
          smw_sort: UsingDTitle
smw_proptable_hash: a:3:{s:12:"smw_fpt_inst";s:32:"78c3880dc683023faf6327816876f8ff";s:14:"smw_fpt_dtitle";s:32:"da5f9daa3c6b985b9302e5a51275d763";s:12:"smw_fpt_mdat";s:32:"d85ee8a4d994060d926e769b929c1e5f";}
          smw_hash: 06a8d4e24541b9d9cf05cef6c5ebdd777ea48533
           smw_rev: 115
       smw_touched: 20240130122953
53 rows in set (0.01 sec)
```

The DT is picked up, as expected.

YvarRavy:
what mw version are you using?
i just checked out your patch again. and did the issue way of creating all the pages and stuff and im still getting the smw_sort pagetitle  and the content of the page is

`{{DISPLAYTITLE:ATest}}
[[Category:Test]]
`
afbeelding

mszabo-wikia:
This is on a 1.42 checkout, I also just switched back to SMW master temporarily to confirm that the original bug is still present there. It also seems to work well on a 1.39 checkout.

Is `$wgRestrictDisplayTitle = false;` set in your setup?

YvarRavy:
yes $wgRestrictDisplayTitle = false on my wiki

YvarRavy:
@mszabo-wikia i'm sorry this took so long. i dont know what happend. i swear i checked out the patch correctly. but i might've not been paying attention. your patch fixes the problem

mszabo-wikia:
@YvarRavy no problem, thank you for taking the time to verify this and for filing this bug report in the first place :)

JeroenDeDauw:
If I understand correctly this PR is now obsolete

mszabo-wikia:
Yeah, https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5592 should resolve the sorting issue.

YvarRavy:
i'm closing this pull request

PR #5587, opened 2024-01-26
Fix assorted PHP deprecation notices

mszabo-wikia:
* setSortKey() may end up receiving nulls in some cases, so account for that to avoid warnings about nulls being passed to string functions on PHP 8.1 and newer. Do the same in Deserializer.
* Avoid using 'self' references in ApiRequestParameterFormatter callables.
* Annotate ChunkedIterator methods with #[ReturnTypeWillChange] to avoid method signature mismatch warnings until PHP 7.4 support can be dropped.
* Declare a missing property in QuerySegment.
* Account for annotation data being false (instead of null) in RecursiveTextProcessor, as array autovivification no longer works for `false` on newer PHP.

### DIFF
--- a/src/Parser/RecursiveTextProcessor.php
+++ b/src/Parser/RecursiveTextProcessor.php
@@ -136,11 +136,7 @@ class RecursiveTextProcessor {
 		}
 
 		$parserOutput = $this->parser->getOutput();
-		$track = $parserOutput->getExtensionData( ParserData::ANNOTATION_BLOCK );
-
-		if ( $track === null ) {
-			$track = [];
-		}
+		$track = $parserOutput->getExtensionData( ParserData::ANNOTATION_BLOCK ) ?: [];
 
 		// Track each embedded #ask process to ensure to remove
 		// blocks on the correct recursive iteration (e.g Page A containing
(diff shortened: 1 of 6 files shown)


src/Parser/RecursiveTextProcessor.php
```diff
@@ -136,11 +136,7 @@ public function transcludeAnnotation( $transcludeAnnotation ) {
 		}
 
 		$parserOutput = $this->parser->getOutput();
-		$track = $parserOutput->getExtensionData( ParserData::ANNOTATION_BLOCK );
-
-		if ( $track === null ) {
-			$track = [];
-		}
+		$track = $parserOutput->getExtensionData( ParserData::ANNOTATION_BLOCK ) ?: [];
```
JeroenDeDauw, line 139: This should be `??`

JeroenDeDauw: At least to retain the exact same behavior as the old code

mszabo-wikia: Unfortunately that's not sufficient because this property may be set to `false` explicitly—some other places e.g. L313 already check for that. On PHP < 8.1, this causes the var to be silently converted into an array when L145 accesses a property on it, which triggers a warning on PHP >= 8.1. So changing it to `?:` just makes that behavior explicit.

Issue #5199, opened 2022-01-09
TypeError: Argument 2 passed to SMW\Services\ServicesFactory::newParserData() must be an instance of ParserOutput, bool given


hexmode:

Seb35:
I investigated this issue two days ago during a few hours, but did not find for now how to solve it. Below are some of my findings.

First, to reproduce it, I used the Docker used in the CI and launched this test alone with MW 1.36 and SMW master (4.0.2+) (`composer run-script -- phpunit --filter FileUploadIntegration`) but it worked fine. I was able to reproduce it, outside of Docker (but it should work inside Docker also), with MW 1.36 + SMW master with `composer run-script -- phpunit --filter LinksUpdateTest`. This class LinksUpdateTest has 4 test functions, and the second one fails with this exact message. It can be remarked that it fails on the first test function in the CI. These specific conditions for reproduction show that it is **probably due to some interactions between different tests**.

My preliminary conclusions are that it is probably related to RevisionGuard and/or to the fact the MediaWiki Parser is no more a global object but available through a MW service:
1. About the RevisionGuard, an hypothesis is that if a page is saved with a given title, MW does not save again with a different content
2. About the Parser, an hypothesis is that there are different unaligned objects of type Parser and/or ParserOutput, or the Parser cannot access the revision content – this is not very precise, I have difficulties to track the various objects; for instance there are two PageCreator classes, one in src/MediaWiki/PageCreator.php (\SMW\MediaWiki\PageCreator) and one in tests/phpunit/Utils/PageCreator.php (\SMW\Tests\Utils\PageCreator), I guess the first one is used but I’m not completery sure.

Seb35:
I dived more deeply in this issue. After spending some hours in comparing stacktraces in 1.35 and 1.36 on the test LinksUpdateTest, I changed strategy and bisected the MW version in branch 1.36 on this same test LinksUpdateTest.

It appears the test LinksUpdateTest worked before [`fd71596f97e`](https://github.com/wikimedia/mediawiki/commit/fd71596f97e7ae97f4075747aea529750a302913) and does not more work with this commit. More exactly, it can be made working with last version in branch REL1_36 with this patch:
```diff
diff --git includes/user/ActorStore.php includes/user/ActorStore.php
@@ -47 +47 @@ class ActorStore implements UserIdentityLookup, ActorNormalization {
-       private const LOCAL_CACHE_SIZE = 5;
[...]

PR #5589, opened 2024-01-28
Stabilize tests on MW 1.39

mszabo-wikia:
Most of the error spam in the 1.39 test run is ultimately due to services not being reset between tests, which (a) can cause errors in itself due to stale data as seen in #5199 and co. and (b) can cause resulting error states to propagate to other tests. Previous attempts at fixing this took the approach of selectively resetting services that were found to be problematic, such as ActorStore. However, this is not in itself sufficient: not only is the task of identifying services not safe for reuse between tests quite Sisyphean, resetting them won't reset services or static singletons that injected them as a dependency.

So, reset MW services, SMW services and assorted static singletons before each integration test. This conveniently allows for reenabling several tests that had been previously disabled due to running into exactly the kind of problems described above. Since resetting the MW service container also resets hook handlers, add appropriate hook registration logic to tests that previously implicitly relied on hook registrations "leaking" over from another test (mostly Query-related ones). As we're here, declare a bunch of missing properties in test classes that pollute the output on PHP > 8.2.

On MW 1.39 (and possibly older), the test output appears to be buffered by default, which is inconvenient because one cannot see the progress until the whole suite has concluded. Set stderr="true" in the config to rectify that (matching what core tests already do), and update the environment reporter to write to stderr as well.

This change does not yet make the suite green on MW 1.39 - there are various failures that need to be investigated. However, it should fix most issues that are purely test setup-related.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -14,6 +14,7 @@
          beStrictAboutTestsThatDoNotTestAnything="true"
          printerClass="SMW\Tests\PHPUnitResultPrinter"
          printerFile="tests/phpunit/PHPUnitResultPrinter.php"
+         stderr="true"
          verbose="true">
     <listeners>
         <listener file="tests/phpunit/ExecutionTimeTestListener.php" class="SMW\Tests\ExecutionTimeTestListener">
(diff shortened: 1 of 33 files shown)


JeroenDeDauw:
This is great! Unfortunately the tests are now failing on the MW versions they worked for before.

mszabo-wikia:
@JeroenDeDauw Yup, I still need to make some changes to ensure the state is in order for older versions.

mszabo-wikia:
Looks better now. Some of the remaining failures or newer versions seem to be actually not MW version but PHP version related, as a local MW 1.35 checkout produced some failures for me when run with PHP 8.0. Still, this is a step forward IMO.

PR #5594, opened 2024-01-31
Fix P0110 on PHP 8.1 and newer

mszabo-wikia:
On PHP 8.1 and newer, the default flags of htmlspecialchars() and related functions were changed so that they no longer leave single quotes alone, causing escaped single quotes to show up in the title attribute of the error tested by P0110. The single quotes are not very useful in a title attribute context, so remove them instead.


jaideraf:
Thanks @mszabo-wikia for all this work! ❤️ It is wonderful that all tests are passing again. 🟢🟢🟢🟢🟢

PR #5451, opened 2023-02-27
Make SemanticMediaWiki compatible with new ElasticSearch

marijnvanwezel:
This pull request makes Semantic MediaWiki compatible with new ElasticSearch (8.x tested). It is currently still a draft, since I have not yet updated the tests. It is working in my personal testing environment.

There are a number of major changes in the ElasticSearch API, as well as some breaking changes in the ElasticSearch PHP package, that this pull request addresses. Most notably:

* Mapping types have been removed (https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html);
* Elasticsearch PHP methods now return `Promise` objects, which need to be explicitly converted to a type;
* The `cat` API no longer returns a JSON by default;
* The associative array syntax for hosts is no longer supported (https://github.com/elastic/elasticsearch-php/issues/1214). This required the addition of a configuration variable, `$wgElasticsearchCredentials`;
* `doc_values` is no longer supported on properties of type `text`.

PR #5596, opened 2024-02-02
Run the unit tests again

mszabo-wikia:
In #5039, the unit tests were moved from tests/phpunit/Unit to directly under tests/phpunit, but the PHPUnit configuration was not updated to match. Due to this, 5k+ unit tests have not been running for ~3 years.

So, update the PHPUnit configuration to run these tests again and make fixes as needed. Due to the size of the changes, where nontrivial changes to non-test code would have been required to fix the tests, I opted to skip the test instead with a "SUT needs refactoring" message so that the necessary changes could be made later without complicating the review.

Many tests emit deprecation warnings on newer PHP. I only fixed missing property declarations in test code in this patch; the rest will need investigation to determine whether they are the result of incomplete or invalid stubbings or actual issues in the code.


src/Constraint/Constraints/ShapeConstraint.php
```diff
@@ -181,7 +181,7 @@ private function hasMinLength( $minLength, $property ) {
 				continue;
 			}
 
-			if ( mb_strlen( $dataItem->getString() < $minLength ) ) {
```
mszabo-wikia, line 184: 1/3 of trivial non-test code changes in this patch.

src/Localizer/LocalLanguage/LocalLanguage.php
```diff
@@ -474,7 +474,7 @@ public function getPreferredDateFormatByPrecision( $precision = null ) {
 		);
 
 		foreach ( $dateOutputFormats as $key => $format ) {
-			if ( @constant( $key ) === $precision ) {
```
mszabo-wikia, line 477: 2/3 of trivial non-test code changes in this patch.

src/Property/Annotators/EditProtectedPropertyAnnotator.php
```diff
@@ -64,11 +64,6 @@ public function addTopIndicatorTo( ParserOutput $parserOutput ) {
 			return false;
 		}
 
-		// FIXME 3.0; Only MW 1.25+ (ParserOutput::setIndicator)
-		if ( !method_exists( $parserOutput, 'setIndicator' ) ) {
-			return false;
-		}
```
mszabo-wikia, line 70: 3/3 of trivial non-test code changes in this patch.

tests/phpunit/MediaWiki/Hooks/SpecialPageListTest.php
mszabo-wikia, line 1: This hook handler no longer exists.

tests/phpunit/ParserFunctions/SubobjectParserFunctionTest.php
```diff
@@ -627,7 +627,7 @@ public function tokuFixedParameterProvider() {
 				],
 				'propertyValues' => [
 					'Foo Bar',
-					'1001'
```
mszabo-wikia, line 630: Updated expectations here and in annotator tests to account for a25c0263b376abe6858eaee627a18b8c189c3e58.

tests/phpunit/Integration/SpecialsTest.php
```diff
@@ -38,7 +39,7 @@
  * @group SMWExtension
  * @group medium
  */
-class SpecialsTest extends SemanticMediaWikiTestCase {
```
mszabo-wikia, line 41: This test accesses the database and the service container, so it's an integration test - moving it accordingly.

mszabo-wikia:
So for now I decided to run the unit tests separately in CI, as some of them seem to affect global state, causing test failures in other suites down the road. Also excluded it from < 1.38 as there are some additional failures there that need investigation.

mszabo-wikia:
After some changes, the unit tests should now pass on all versions (when run separately)

src/MediaWiki/PageInfoProvider.php
```diff
@@ -189,7 +189,7 @@ public static function isProtected( Title $title, string $action = '' ) {
 		}
 
 		// MW < 1.37
-		return $title->isProtected();
```
mszabo-wikia, line 192: 3/4 of trivial non-test code changes in this patch.

tests/phpunit/Constraint/Constraints/ShapeConstraintTest.php
```diff
@@ -69,10 +70,10 @@ public function testCheckConstraint_shape_constraint_missing_property() {
 			->with( $this->equalTo( 'Foo' ) )
 			->will( $this->returnValue( false ) );
 
-		$dataValue = $this->getMockBuilder( '\SMWDataValue' )
-			->disableOriginalConstructor()
-			->setMethods( [ 'addError', 'getCallable' ] )
-			->getMockForAbstractClass();
+		$dataValue = $this->createMock( SMWDataValue::class );
+		$dataValue->expects( $this->any() )
```
Nikerabbit, line 74: Isn't `->expects( $this->any() )` the default behaviour?

tests/phpunit/Elastic/Indexer/IndexerTest.php
```diff
@@ -81,7 +82,6 @@ public function testCreate() {
 
 		$expected = [
 			'index' => '_index_abc',
-			'type' => 'data',
```
Nikerabbit, line 84: This change is not obvious to me.

Nikerabbit:
Very nice!

tests/phpunit/Elastic/Indexer/IndexerTest.php
mszabo-wikia: This is to account for https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5451/commits/f020de1b52ff7ff833fdf0394a3ae81ee40139e2 which removed these types.

tests/phpunit/Constraint/Constraints/ShapeConstraintTest.php
mszabo-wikia: Yeah, calling `->method()` (via a magic method) without setting up expects would work too.

PR #5603, opened 2024-02-08
fix: user readQuery in Query class when performing reads

harnash:
Query class is a helper class which is used to build SELECT statements. There is no reason to request 'write' connection when performing read only queries.
This should help in situations where primary DB is not available.


harnash:
🤣
I was afraid that it will break tests. I will fix them tomorrow.

harnash:
I believe tests should now pass. They do on my local setup.

PR #5604, opened 2024-02-08
Use RIGOR_QUICK for ProtectionValidator checks

mszabo-wikia:
The main purpose of this check is to check whether an user is blocked from creating or editing a page by page protection settings. Since this only hinges on the presence of said protection settings and the user's set of rights, and this check is meant to control the rendering of UI elements on read requests, RIGOR_QUICK is more appropriate for this than RIGOR_SECURE which would trigger a primary database connection.


JeroenDeDauw:
I didn't even know this difference existed. Would be cool if someone that does can review. @reedy @Nikerabbit ?

PR #5606, opened 2024-02-14
Imported from property

thomas-topway-it:
fixes issue https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4971

- fixes mixed wikitext/html in Imported from printout caption
- fixes empty declarativeName (i.e. vocabulary name used as caption of the imported property between parenthesis) (see here https://www.semantic-mediawiki.org/wiki/User:Krabina_Bernhard/ontology)

supersedes https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5492

added test `q-1301.json`

the test fails with the original class and succeeds with the updated class. (the original class returns a result like the following, https://www.semantic-mediawiki.org/wiki/User:Krabina_Bernhard/ontology, with escaped html and without captions)

### DIFF
--- a/src/DataValues/ImportValue.php
+++ b/src/DataValues/ImportValue.php
@@ -73,9 +73,8 @@ class ImportValue extends DataValue {
 	 * @var string
 	 */
 	private $declarativeName = '';
-	
-	/** @var array */
-	private $declarativeNames = [];
+
+	private array $declarativeNames = [];
 
 	private MediaWikiNsContentReader $mediaWikiNsContentReader;
 
--- a/src/DataValues/ImportValue.php
+++ b/src/DataValues/ImportValue.php
@@ -6,8 +6,8 @@ use SMW\Message;
 use SMWDataItem as DataItem;
 use SMWDataValue as DataValue;
 use SMWDIBlob as DIBlob;
-use SMW\MediaWiki\MediaWikiNsContentReader as MediaWikiNsContentReader;
-use SMW\Services\ServicesFactory as ApplicationFactory;
+use SMW\MediaWiki\MediaWikiNsContentReader;
+use SMW\Services\ServicesFactory;
 
 /**
  * This datavalue implements datavalues used by special property '_IMPO' used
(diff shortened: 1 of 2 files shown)


src/DataValues/ImportValue.php
```diff
@@ -70,12 +72,19 @@ class ImportValue extends DataValue {
 	 * @var string
 	 */
 	private $declarativeName = '';
+	
+	/** @var array */
+	private $declarativeNames = [];
+
+	/** @var mediaWikiNsContentReader */
+	private $mediaWikiNsContentReader;
```
JeroenDeDauw, line 80: Type is wrong capitalization. Also should use PHP type, not a doc

`private ClassName $foobar;`

src/DataValues/ImportValue.php
```diff
@@ -142,13 +151,45 @@ protected function loadDataItem( DataItem $dataItem ) {
...
 		}
 
 		return true;
 	}
 
+	/**
+	 * @param string $namespace
+	 * @return string
+	 */
+	private function getDeclarativeName( $namespace ) {
```
JeroenDeDauw, line 165: Please use PHP types, not docs

src/DataValues/ImportValue.php
```diff
@@ -6,6 +6,8 @@
 use SMWDataItem as DataItem;
 use SMWDataValue as DataValue;
 use SMWDIBlob as DIBlob;
+use SMW\MediaWiki\MediaWikiNsContentReader as MediaWikiNsContentReader;
```
JeroenDeDauw, line 9: The `as MediaWikiNsContentReader` isn't needed

src/DataValues/ImportValue.php
```diff
@@ -6,6 +6,8 @@
...
+use SMW\Services\ServicesFactory as ApplicationFactory;
```
JeroenDeDauw, line 10: Why import it as a different name? That makes the code harder to understand

JeroenDeDauw:
This is too complicated for me to be confident it works without manual test. @krabina can you verify the relevant SMW features still work?

thomas-topway-it:
done, an easy way to test it would be the following

```
{{#ask: [[Imported from::+]]
|?Imported from
|format=plainlist 
}}
```

krabina:
It does work: https://test.knowledge.wiki/Ontologie
on the same page the issue was there before I chedked out the PR. Here it is in an older version with the issue: https://www.geschichtewiki.wien.gv.at/Wien_Geschichte_Wiki/Ontologie#Kategorie:Attribute_.28Properties.29

PR #5607, opened 2024-02-16
Improve query log output

JeroenDeDauw:


thomas-topway-it:
using `Html::element` looks indeed a good idea

Issue #5634, opened 2024-05-04
MediaWiki 1.43 DBAL compatibility


osnard:
### Setup

- SMW version: master branch at #f6b3855f
- MW version: 1.43.0-wmf.2 (pre-release)
- PHP version: 8.1.20 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: 10.11.4-MariaDB-1:10.11.4+maria~ubu2204

### Issue

During the Wikimedia Hackathon 2024, I tried to run the latest version of Semantic MediaWiki on a pre-release version of MediaWiki 1.43 (upcomming LTS). The goal was to identify issues and maybe provide patches right away. Unfortunately, the available time will not allow to provide a fix, therefore I document the issue here. 

First: Installation (running `maintenance/run.php update`) and activation of the extension ran smoothly, No immediate error.

But when trying to access `Special:Browse` the API call with the parameters

```
action: smwbrowse
format: json
browse: subject
params: {"subject":"Main_Page","ns":0,"iw":"","subobject":"","options":{"dir":null,"lang":"en","group":null,"printable":null,"offset":null,"including":false,"showInverse":false,"showAll":true,"showGroup":true,"showSort":false,"api":true,"valuelistlimit.out":"30","valuelistlimit.in":"20"},"type":"html"}
```

failed with the error
```
Exception caught: Table name component contains unexpected quote or dot character

"Wikimedia\\Rdbms\\DBLanguageError at /includes/libs/rdbms/platform/SQLPlatform.php(1051)
from /includes/libs/rdbms/platform/SQLPlatform.php(1051)
#0 /includes/libs/rdbms/platform/SQLPlatform.php(991): Wikimedia\\Rdbms\\Platform\\SQLPlatform->extractTableNameComponents(string)
#1 /includes/libs/rdbms/platform/SQLPlatform.php(957): Wikimedia\\Rdbms\\Platform\\SQLPlatform->qualifiedTableComponents(string)
#2 /includes/libs/rdbms/platform/SQLPlatform.php(936): Wikimedia\\Rdbms\\Platform\\SQLPlatform->tableName(string)
#3 /includes/libs/rdbms/platform/SQLPlatform.php(837): Wikimedia\\Rdbms\\Platform\\SQLPlatform->tableNameWithAlias(string, string)
[...]

PR #5636, opened 2024-05-05
Use MW43 DBAL

osnard:


krabina:
Great, thank you. This is a first attempt to get SMW working with 1.43 LTS

gesinn-it-gea:
I'm not sure if changing all the individual calls to getConnection is the right approach. After the change, the DB is accessed directly without SMW's Store being involved. Wouldn't it be better to change the access logic somewhere in SMW's Store abstraction layer?

class SQLStore extends Store:
```
	public function getConnection( $type = 'mw.db' ) {
		return parent::getConnection( $type );
	}
```

abstract class Store implements QueryEngine
```
	public function getConnection( $type ) {
		if ( $this->connectionManager === null ) {
			$this->connectionManager = ApplicationFactory::getInstance()->getConnectionManager();
		}

		return $this->connectionManager->getConnection( $type );
	}
```

class ConnectionManager
```
	public function getConnection( $id = null ) {
		$id = strtolower( $id );

		if ( self::$isConnectable === null ) {
			self::$isConnectable = $this->isConnectable();
		}

		if ( isset( self::$connectionProviders[$id] ) ) {
			return self::$connectionProviders[$id]->getConnection();
		}

		throw new RuntimeException( "{$id} is not registered as connection provider" );
	}
```

class ConnectionProvider implements IConnectionProvider ...

osnard:
Closing this in favor of https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5679

My change was only a quick test / draft. The other one is much more mature already. Doesn't make sense to keep this.

squirrelslair:
It appears that this was closed in favour of [#5679](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5679), but that one was closed mentioning this one. 
I tried running composer and update after these were closed and was still getting the error. So, is this fixed?

jaideraf:
@squirrelslair did you follow these steps? https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5634#issuecomment-2344458646

There is not a release yet asserting support for MediaWiki above 1.41. See: 
https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/docs/COMPATIBILITY.md

PR #5660, opened 2024-07-05
Test with more MediaWiki versions

JeroenDeDauw:
## Summary by CodeRabbit

- **Chores**
  - Added support for newer versions of MediaWiki (1.40, 1.41, 1.42) with MySQL as the database type in the workflow configuration.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,6 +30,16 @@ jobs:
           - mediawiki_version: 1.39
             database_type: mysql
             experimental: true
+          - mediawiki_version: 1.40
+            database_type: mysql
+            experimental: true
+          - mediawiki_version: 1.41
+            database_type: mysql
+            experimental: true
+          - mediawiki_version: 1.42
+            database_type: mysql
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Seems like the tests do not run at all due to infra issues. For 1.40 there is a PHPUnit compat issue, and for 41 and 42 there is an issue with the MediaWiki test runner. Can't fix this easily myself since SMW uses a different test setup than I am used to. @gesinn-it-gea

gesinn-it-gea:
@JeroenDeDauw if you like, we can take care to get CI running with more flexible matrix. This would afterwards look like for SRF (https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/master/.github/workflows/ci.yml)

JeroenDeDauw:
@gesinn-it-gea that sounds nice. Would it fix the test runner compatibility issues?

gesinn-it-gea:
@JeroenDeDauw ... I think we can close this PR

PR #5674, opened 2024-07-19
Issue 5667 - fix sniff PSR12.Functions.ReturnTypeDeclaration

gesinn-it-ilm:
Clear and fix sniff **PSR12.Functions.ReturnTypeDeclaration** reported by phpcs and improve code quality.


JeroenDeDauw:
If you are being paid per line changed: congrats on becoming a trillionaire

gesinn-it-ilm:
We have "fix" command we are using to clear specific sniff automatically. Otherwise, it takes time to fix something like this manually. :)
Some of those sniffs we will need to fix manually, because of the workflow and logic itself, but for the sniffs like this, we can use automatic approach.

Issue #5682, opened 2024-07-22
Special:FacetedSearch has an issue


YOUR1:
### Setup

- SMW version: 4.2.0
- MW version: 1.39
- PHP version: 7.4

### Issue

See issue [here](https://sandbox.semantic-mediawiki.net/wiki/Sp%C3%A9cial:FacetedSearch?q=%5B%5BCategory%3AYouri%5D%5D&utf8=%E2%9C%93&limit=50&offset=0&csum=711364995&pv%5BHas+text%5D%5B%5D=&format=table&order=asc&size=50)

PR #5683, opened 2024-07-22
Improve Special:FacetedSearch output

YOUR1:
See https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5682


JeroenDeDauw (approved):
@YOUR1 can you double check that my small tweak is OK?

YOUR1:
Checked. Is okay!

PR #5691, opened 2024-07-26
Use IDBAccessObject::READ_LATEST on 1.42 onwards

redbluegreenhat:
Fixes page protection on 1.42

Miraheze bug: https://issue-tracker.miraheze.org/T12379

## Summary by CodeRabbit

- **New Features**
	- Enhanced compatibility for MediaWiki versions prior to 1.42, improving the handling of updates.
- **Bug Fixes**
	- Adjusted logic in the update process to ensure it operates correctly across different software versions.

### DIFF
--- a/src/MediaWiki/RevisionGuard.php
+++ b/src/MediaWiki/RevisionGuard.php
@@ -52,8 +52,12 @@ class RevisionGuard {
 	public function isSkippableUpdate( Title $title, &$latestRevID = null ) {
 		// MW 1.34+
 		// https://github.com/wikimedia/mediawiki/commit/b65e77a385c7423ce03a4d21c141d96c28291a60
-		if ( defined( 'Title::READ_LATEST' ) && Title::GAID_FOR_UPDATE == 512 ) {
+		if ( defined( 'Title::READ_LATEST' ) ) {
 			$flag = Title::READ_LATEST;
+		// MW 1.42+
+		// https://github.com/wikimedia/mediawiki/commit/5ca00251c607d2fd11e2df66336fcbb9925eea91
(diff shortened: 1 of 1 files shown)


redbluegreenhat:
The force pushes are due to writing incorrectly the logic for version_compare and noticing I had to remove `&& Title::GAID_FOR_UPDATE == 512` from the first if block, otherwise it would still throw an exception.

src/MediaWiki/RevisionGuard.php
```diff
@@ -52,8 +52,12 @@ public function __construct( RevisionLookup $revisionLookup ) {
 	public function isSkippableUpdate( Title $title, &$latestRevID = null ) {
 		// MW 1.34+
 		// https://github.com/wikimedia/mediawiki/commit/b65e77a385c7423ce03a4d21c141d96c28291a60
-		if ( defined( 'Title::READ_LATEST' ) && Title::GAID_FOR_UPDATE == 512 ) {
+		if ( defined( 'Title::READ_LATEST' ) ) {
 			$flag = Title::READ_LATEST;
+		// MW 1.42+
+		// https://github.com/wikimedia/mediawiki/commit/5ca00251c607d2fd11e2df66336fcbb9925eea91
+		} else if ( version_compare( MW_VERSION, '1.42' , '>=' ) ) {
+			$flag = IDBAccessObject::READ_LATEST;
```
JeroenDeDauw, line 60: Can't you use `if ( defined( 'IDBAccessObject::READ_LATEST' ) )`? Seems after than checking the version, and is consistent with the other code. Support for MW older than 35 can also be dropped, so we only need one check, not 2.

PR #5697, opened 2024-08-01
Issue 5695 - Call to undefined method BacklinkCache::clear()

gesinn-it-ilm:
Issue with unit tests for **MW 1.40** has been fixed. 
According to [BacklinkCache: Remove unused internal clear() method](https://github.com/wikimedia/mediawiki/commit/2249870a0f3725c70a8cba3b9eb0f16350b68245#diff-9f2668e3e3ce05624e6d6165f532aca730c6c70fb9c81f973e6f661259b88d72), we have excluded method call for 1.40+ and now CI is working fine. 
Those issues are also cleared for higher MW version - **1.41 and 1.42**, but there are still some errors needed to be fixed. 
(another issue opened: [issue-5698](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5698))

### DIFF
--- a/.env-40
+++ b/.env-40
@@ -1,5 +1,0 @@
-# docker images
-MW_VERSION?=1.40
-PHP_VERSION?=8.1
-DB_TYPE?=mysql
-DB_IMAGE?="mariadb:11.2"
--- a/.env-40
+++ b/.env-40
@@ -1,0 +1,5 @@
+# docker images
+MW_VERSION?=1.40
+PHP_VERSION?=8.1
+DB_TYPE?=mysql
+DB_IMAGE?="mariadb:11.2"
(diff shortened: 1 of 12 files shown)


.env-40
```diff
@@ -0,0 +1,5 @@
+# docker images
+MW_VERSION?=1.40
+PHP_VERSION?=8.1
+DB_TYPE?=mysql
+DB_IMAGE?="mariadb:11.2"
```
JeroenDeDauw, line 5: What is using these files? And do they have to be in the project root?

alexander-gesinn: These are placeholders that can be renamed / symlinked to .env and are then picked up by the "make ci" call. This makes it easier during development to switch between different "targets". But technically, not required. Maybe defining those different targets inside the Makefile would be smarter

JeroenDeDauw (approved):

PR #5700, opened 2024-08-02
Remove unused scrutinizer file

kghbln:


kghbln:
My impression is that this repo will not be piped through scrutinizer again in the future

gesinn-it-gea:
... yes, we can run it through codecov but this requires adding a token. Please contact @gesinn-it-ilm if you need help with that (gesinn.it team does not have permissions to change those settings)

PR #5703, opened 2024-08-08
Issue 5702 - clear deprecation warnings for PHP 8.1 and MW 1.39+ - null to param

gesinn-it-ilm:
**Null to param deprecation warnings** checked and fixed for MW 1.39+ and PHP 8.1.
There are still some other deprecation warnings that need to be checked how it can be updated not to change or break the workflow. (_maybe to open another issue and check_)


jaideraf:
@JeroenDeDauw could you please review this? I am not a developer, but the changes seem ok to me.

PR #5704, opened 2024-08-09
Use table prefix in bootstrap, ignore key check during tests

gesinn-it-ilm:
Set $smwgIgnoreUpgradeKeyCheck to true to ignore verification that a correct upgrade path was selected. 
Define test table prefix used in tests.


JeroenDeDauw:
@gesinn-it-ilm  What is this file for? https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/__setup_extension__

Issue #5706, opened 2024-08-13
Adding smwLikelyOutdatedDependencies to Title object is deprecated


alex-mashin:
SMW 4.2.0, MW 1.39.8, PHP 8.3.10.

Deprecated: Creation of dynamic property `Title::$smwLikelyOutdatedDependencies` is deprecated in `.../includes/debug/DeprecationHelper.php`.

Caused by the following in `SemanticMediaWiki/src/DependencyValidator.php`:
```php
	/**
	 * Signal to the `OutputPageParserOutput` hook to we want a possible purge
	 * action.
	 *
	 * @since 3.1
	 *
	 * @param Title $title
	 */
	public function markTitle( Title $title ) {
		$title->smwLikelyOutdatedDependencies = true;
	}

	/**
	 * @since 3.1
	 *
	 * @param Title $title
	 *
	 * @return boolean
	 */
	public static function hasLikelyOutdatedDependencies( Title $title ) {
		return isset( $title->smwLikelyOutdatedDependencies ) && $title->smwLikelyOutdatedDependencies;
	}
```

Obviously, something less hacky is needed.

PR #5707, opened 2024-08-13
Fix deprecation of dynamic properties

alex-mashin:
Fixes: #5706.

## Summary by CodeRabbit

- **New Features**
	- Improved dependency tracking by centralizing management within the `DependencyValidator` class.
	- Enhanced performance with the use of a static array for marking outdated dependencies.

- **Bug Fixes**
	- Addressed issues related to direct manipulation of dependency properties, leading to more reliable tests and improved state management.

- **Tests**
	- Updated test logic to utilize the new dependency validation method, ensuring tests accurately reflect the revised functionality.

### DIFF
--- a/src/DependencyValidator.php
+++ b/src/DependencyValidator.php
@@ -46,6 +46,11 @@ class DependencyValidator {
 	private $eTag;
 
 	/**
+	 * @var array $titles Title IDs marked as having outdated dependencies.
+	 */
+	private static $titles = [];
+
+	/**
 	 * @since 3.1
 	 *
 	 * @param NamespaceExaminer $namespaceExaminer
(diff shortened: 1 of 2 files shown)


src/DependencyValidator.php
```diff
@@ -45,6 +45,11 @@ class DependencyValidator {
 	 */
 	private $eTag;
 
+	/**
+	 * @var array $titles Title IDs marked as having outdated dependencies.
+	 */
+	private static $titles = [];
```
JeroenDeDauw, line 51: Why static?

alex-mashin:
Because, previously, the `Title` object was marked as having outdated dependencies, and that applied to any instance of `DependencyValidator` request-wide that could exist. If `$titles` were not static, any outdated dependency mark would have the scope of only one `DependencyValidator` instance.

Although, I am not sure that `DependencyValidator` is not effectively a singleton, thus making the difference between static and non-static irrelevant.

Issue #5708, opened 2024-08-15
Deferred updates not running in background due to namespacing


dctucker:
### Setup

- SMW version: 4.2.0 ([sheep-y fork](https://github.com/Sheep-y/SemanticMediaWiki-MW14))
- MW version: 1.4.2
- PHP version: 8.3.10
- DB system (MySQL, Blazegraph, etc.) and version: MySQL 10.5.23-MariaDB-0+deb11u1

### Issue

Detailed description of the issue and a [stack trace](https://www.semantic-mediawiki.org/wiki/Help:Identifying_bugs) if applicable:

```
[DeferredUpdates] DeferredUpdates::run: started MediaWiki\Deferred\AutoCommitUpdate_MediaWiki\Deferred\LinksUpdate\LinksUpdate::doUpdate #1785
[exception] [c694115d35df408a715aa2a4] /index.php?title=Poison_Spray&action=submit
TypeError: SMW\MediaWiki\Hooks\LinksUpdateComplete::process(): Argument #1 ($linksUpdate) must be of type LinksUpdate, MediaWiki\Deferred\LinksUpdate\LinksUpdate given, called in /home/wiki/mediawiki-1.42.1/extensions/SemanticMediaWiki/src/MediaWiki/Hooks.php on line 839
#0 /home/wiki/mediawiki-1.42.1/extensions/SemanticMediaWiki/src/MediaWiki/Hooks.php(839): SMW\MediaWiki\Hooks\LinksUpdateComplete->process()
#1 /home/wiki/mediawiki-1.42.1/includes/HookContainer/HookContainer.php(159): SMW\MediaWiki\Hooks->onLinksUpdateComplete()
#2 /home/wiki/mediawiki-1.42.1/includes/HookContainer/HookRunner.php(2346): MediaWiki\HookContainer\HookContainer->run()
#3 /home/wiki/mediawiki-1.42.1/includes/deferred/LinksUpdate/LinksUpdate.php(194): MediaWiki\HookContainer\HookRunner->onLinksUpdateComplete()
#4 /home/wiki/mediawiki-1.42.1/includes/deferred/AutoCommitUpdate.php(47): MediaWiki\Deferred\LinksUpdate\LinksUpdate->MediaWiki\Deferred\LinksUpdate\{closure}()
[...]

PR #5715, opened 2024-09-02
Convert SQL select() to use only table names and join conditions

Seb35:
MediaWiki 1.42 constrains not to use raw SQL in table names. Almost all calls were already fine except the one ine QueryEngine where JOINed tables are used: these are converted into two parallel variables fromTables and joinConditions compatible with MediaWiki’s select().

Also,
* RedirectStore is converted to a simple JOIN
* TraversalPropertyLookup is converted to a JOIN + Wikimedia\Rdbms\Subquery, as there is no SMW-native equivalent for a subquery.

Issue: #5634

----

**Warning:** These should be merged in separate PRs since they are about different topics:
* the change in `src/MediaWiki/Hooks/LinksUpdateComplete.php`
* the change in `includes/ContentParser.php`
They should be removed before final merge, but included here to have better results in CI.

----

## Summary by CodeRabbit

- **New Features**
	- Enhanced SQL query construction with improved structures for joins and selections.
	- Introduced new properties for managing tables and join conditions in query segments.

- **Bug Fixes**
	- Improved robustness of query processing, especially for nested queries and complex joins.

- **Documentation**
	- Added clarifications and examples for new properties to enhance developer understanding.

- **Refactor**
	- Streamlined SQL query logic for better readability and maintainability across multiple components.

### DIFF
--- a/includes/ContentParser.php
+++ b/includes/ContentParser.php
@@ -157,12 +157,17 @@ class ContentParser {
 		// Avoid "The content model 'xyz' is not registered on this wiki."
 		try {
 			$services = MediaWikiServices::getInstance();
+			// MW 1.38+
 			if ( method_exists( $services, 'getContentRenderer' ) ) {
+				// MW 1.42+
+				if ( version_compare( MW_VERSION, '1.42', '<' ) ) {
+					$revision = $revision->getId();
+				}
 				$contentRenderer = $services->getContentRenderer();
 				$this->parserOutput = $contentRenderer->getParserOutput(
 					$content,
 					$this->getTitle(),
-					$revision->getId()
+					$revision
 				);
 			} else {
 				$this->parserOutput = $content->getParserOutput(
--- a/src/MediaWiki/Hooks/LinksUpdateComplete.php
+++ b/src/MediaWiki/Hooks/LinksUpdateComplete.php
@@ -2,7 +2,6 @@
 
 namespace SMW\MediaWiki\Hooks;
 
-use LinksUpdate;
 use SMW\Services\ServicesFactory as ApplicationFactory;
 use SMW\MediaWiki\RevisionGuardAwareTrait;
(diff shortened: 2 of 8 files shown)


Seb35:
I see there are multiple errors in the CI, I will review them in the next days, but there is the following unrelated thing to make MW 1.42 compatible:
* #5708 : convert `use LinksUpdate;` to `use MediaWiki\Deferred\LinksUpdate\LinksUpdate;` without this the semantic values are not saved

JeroenDeDauw:
You definitely broke something. On master the 35, 39 and 40 tests are all passing

image

Are your changes meant to drop support for some MediaWiki versions? (I think it's fine to drop support for MediaWiki older than 39)

Seb35:
> Are your changes meant to drop support for some MediaWiki versions? (I think it's fine to drop support for MediaWiki older than 39)

It was not intended, I will search why.

Seb35:
Note in my final change: in disjunctions and when $wgDBprefix is not an empty string, temporary tables are created with a name `'t1'`, `'t2'` etc, which means the tables `prefix_t1` etc are created. But in MediaWiki’s `select()`, when the table names are `[ 't0' => 't0', 't1' => 't1' ]` and the joinConditions is `[ 't1' => [ 'JOIN', 't0.smw_id=t1.smw_id' ] ]` then MediaWiki converts this to SQL `SELECT * FROM prefix_t0 JOIN prefix_t1 ON t0.smw_id=t1.smw_id` but `t1` does not exist, only `prefix_t1` but it is not convenient to use the prefix in the join condition.

It may be a bug/edge case in MediaWiki (a better behaviour would be to write `JOIN prefix_t1 t1` so that we can always use `t1` in the join condition). I used a workaround in `QuerySegmentListProcessor::table()` by using the prefixed table name, so that it is different than the alias. It is a bit suboptimal though.

PR #5733, opened 2024-10-07
Fix DependencyLinksValidator for MW 1.42.x

Ferdi265:
MW 1.42.x apparently only allows table name aliases to be specified as `"alias" => "table"` pairs.

This PR fixes a crash with MW 1.42.x or later when setting ` $smwgEnabledQueryDependencyLinksStore = true;`. See also [this issuecomment](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5634#issuecomment-2396556344) in #5634.

This is a followup to #5715.

Note that this is my first PR to a MediaWiki extension and my first PHP change in a long time, even though it's quite minimal.

## Summary by CodeRabbit

- **Chores**
	- Updated SQL query syntax for improved clarity in table name references.

### DIFF
--- a/src/SQLStore/QueryDependency/DependencyLinksValidator.php
+++ b/src/SQLStore/QueryDependency/DependencyLinksValidator.php
@@ -116,10 +116,8 @@ class DependencyLinksValidator {
 		// INNER JOIN smw_object_ids AS p ON ((s_id=p.smw_id))
 		// INNER JOIN smw_object_ids AS v ON ((o_id=v.smw_id))
 		// WHERE p.smw_hash = 'xxx' AND (p.smw_iw!=':smw') AND (p.smw_iw!=':smw-delete')
-		$id_table = $connection->tableName( SQLStore::ID_TABLE );
-
 		$rows = $connection->select(
-			[ $proptables[$tableid]->getName(), $id_table . ' AS p', $id_table . ' AS v' ],
+			[ $proptables[$tableid]->getName(), "p" => SQLStore::ID_TABLE, "v" => SQLStore::ID_TABLE ],
 			[
 				'v.smw_id', 'v.smw_subobject', 'v.smw_touched'
 			],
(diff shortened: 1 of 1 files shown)


jaideraf:
@Seb35, as you have had more experience with that subject, I ask you: does this change seem ok to you?

JeroenDeDauw (approved):

src/SQLStore/QueryDependency/DependencyLinksValidator.php
```diff
@@ -119,7 +119,7 @@ public function hasArchaicDependencies( DIWikiPage $subject ) {
 		$id_table = $connection->tableName( SQLStore::ID_TABLE );
 
 		$rows = $connection->select(
```
Seb35, line 121: I cannot point to the right lines 136-137 since they are not edited, but in the last parameter (join_conditions) it would be better to use only as key the table name or the alias (see [MW doc](https://github.com/wikimedia/mediawiki/blob/a11f20989b829f4ede6c731e6126eea3defdbba6/includes/libs/rdbms/database/IReadableDatabase.php#L415-L428)). Given the same table is used multiple times here, I guess the alias is better here to distinguish the two same tables.

Consequently, the variable `$id_table` just above should be deleted.

src/SQLStore/QueryDependency/DependencyLinksValidator.php
```diff
@@ -169,7 +169,7 @@ public function hasArchaicDependencies( DIWikiPage $subject ) {
 		$links_table = $connection->tableName( SQLStore::QUERY_LINKS_TABLE );
 
 		$row = $connection->selectRow(
```
Seb35, line 171: Same comment for the join_conditions line 183 and the now-unused variable `$links_table`.

Seb35 (changes requested):
Thanks for this PR! Indeed I missed this occurence in #5715.

I think it is better to update accordingly the parameter $join_conditions of select() by removing the 'table AS alias'. Else it is fine.

src/SQLStore/QueryDependency/DependencyLinksValidator.php
Ferdi265: Oh, so just using `"p" => [ "INNER JOIN" ... ]` on its own in the join condition is fine if I understand this right? I was wondering about whether that can be cleaned up as well, will try that tomorrow.

Ferdi265: implemented as you suggested.

Seb35 (approved):
Great! I just tested the last version 1a518e333 and it works well :+1:

BTW I was wondering why no tests caught this, I found the test p-0906.json testing this but it is not executed in the "unit tests" stage, so it is one failure fixed in advance when the full tests will run^^

PR #5741, opened 2024-10-25
Do not skip data update on null edit

### LINKED ISSUE #5740: Manual NULL edits don't work
### Setup

- SMW version: 4.2.0
- MW version: 1.39.10 (2967077) 02:44, 8 October 2024
- PHP version: 	8.1.30 (fpm-fcgi)
- DB system and version: 10.3.39-MariaDB-log

### Issue

SMW defines the concept of a "Null Edit" at: https://www.semantic-mediawiki.org/wiki/Null_edit 

The ability to locally and manually perform null edits on pages with semantic data is a core part of our test and checkout workflow.

We recently upgraded to the above setup from our old system running SMW version: 3.2.3, MW version: 1.34.4, PHP version: 7.2.33, DB system: MySQL 5.5.68

After our update, manual null edits (adding some whitespace to the end of a page and saving) no longer worked.

The page would "show" the updated data, but the "Browse Properties" page would not have the updated SMW data and queries of the pages would not return the updated data.  

Steps to reproduce the observation:

**Please see the set-up at:** 
https://sandbox.semantic-mediawiki.net/wiki/Null_Edit_Test_-_Page_B

yaronkoren:
Simplify DataUpdater::isSkippable() to not return true if the current revision being generated is also the latest revision, i.e. if it's a null edit - because sometimes you do want SMW data to be refreshed on a null edit, like if any of the property values are dynamic.

Fixes #5740 .

### DIFF
--- a/src/DataUpdater.php
+++ b/src/DataUpdater.php
@@ -166,17 +166,7 @@ class DataUpdater {
 	 * @return boolean
 	 */
 	public function isSkippable( Title $title, ?int &$latestRevID = null ) {
-		if ( $this->revisionGuard->isSkippableUpdate( $title, $latestRevID ) ) {
-			return true;
-		}
-
-		$associatedRev = $this->store->getObjectIds()->findAssociatedRev(
-			$title->getDBKey(),
-			$title->getNamespace(),
-			$title->getInterwiki()
-		);
-
-		return $associatedRev == $latestRevID;
+		return $this->revisionGuard->isSkippableUpdate( $title, $latestRevID );
 	}
 
 	/**


yaronkoren:
@JeroenDeDauw - what do you think about this?

JeroenDeDauw:
It looks plausible. Would be nice if someone can verify the linked issue is fixed by this PR.

JeroenDeDauw:
This PR is breaking the tests. The 41 and 42 tests are broken on master, but the rest should pass:
image

gesinn-it-gea:
What's your use case for the null edits? Updating annotations based on query results?

gesinn-it-gea:
we should be careful to not get into regressions. See also #2549 and mwjames' comments.

yaronkoren:
Yes, that's the use case, as noted in #5740. (To be clear, it's a NASA issue, I'm just a consultant on this one.)

amcgillivray-nasa:
> What's your use case for the null edits? Updating annotations based on query results?

That is one of our use cases. The other is that we are running a number of custom scripts which rely on conditional categorization to perform an action. After the action is performed, the page which triggered the action is null-edited so that the category then updates.

yaronkoren:
By the way, as far as I know, this PR no longer causes any additional validation errors.

gesinn-it-gea:
In the meanwhile, this PR has some conflicts.

Some thoughts on this change:

**Risk of Regressions:**

* If there were scenarios where the "associated revision" logic was crucial, this change could introduce regressions.
* Example: A case where `$this->revisionGuard->isSkippableUpdate` returns false, but the "associated revision" matches $latestRevID, which previously would have returned true, will now return false. 

**Additional Questions to Evaluate Regressions**
- Does `$this->revisionGuard->isSkippableUpdate` handle repeated updates fully?
  - If not, there could be scenarios where the same update is unnecessarily processed multiple times.
- Are there cases where `findAssociatedRev `identified relevant revisions that `$this->revisionGuard` cannot detect?
  - The removed logic might have been a fallback for edge cases not covered by `$this->revisionGuard`.
- Why was `findAssociatedRev` included originally?
  - Was it compensating for limitations in `$this->revisionGuard` or addressing specific needs tied to SMW or MW's behavior with JobSpecification updates?
- Have the underlying assumptions changed?
  - If the environment (e.g., `$this->revisionGuard`, SMW's internal mechanisms) has evolved, the removal might be valid—but this needs confirmation.
  
 @JeroenDeDauw what do you think?

revansx:
Some context, Yaron developed this patch to fix an issue with data updates not working on page saves. Until his patch gets implemented we have to maintain our SMW instance with this as a local hack.

jaideraf:
@revansx It would be nice if you could answer about the [Annotation query marker](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Annotation_query_marker) on the issue page https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5740#issuecomment-2444934366

revansx:
Ok. I will

jaideraf:
For the record, I do not approve this change due to the reasons [appointed](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5741#issuecomment-2571150677) by @gesinn-it-gea. It seems to me that [Annotation query marker](https://www.semantic-mediawiki.org/wiki/Help:Ask_syntax/Annotation_query_marker) was created precisely to address this use case.

Issue #5692, opened 2024-07-28
SchemaContent overrides fillParserOutput


keepersdungeon:
### Setup

- SMW version: 4.2.0
- MW version: 1.41
- PHP version: 8.2

### Issue
Hello I'm getting this error when I access any of the 'Administration' 'Classification' or 'Content' properties in the 'Browse Properties'

```
PHP Deprecated:  SMW\MediaWiki\Content\SchemaContent overrides fillParserOutput which was deprecated in MediaWiki 1.38. [Called from MediaWiki\Content\Renderer\ContentRenderer::getParserOutput
PHP Deprecated:  Use of AbstractContent::getParserOutput was deprecated in MediaWiki 1.38. [Called from ContentHandler::callDeprecatedContentGPO 
```
Other properties link seems to be working fine

krabina:
I am seeing the same in MW 1.39 when on a schema page, e. g. SMW/Schema:Keyword link formatter

jaideraf:
It also shows up in the page "smw/schema:Group:Schema_properties":

MW version: 1.39.10
SMW version: 5.0.0-alpha (b045a2d)
PHP: 7.4.33

paladox:
I think this was fixed with https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/21b4a5a84f82eb7b5f09d92c4ee9b1adb109030a and https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5750

PR #5750, opened 2024-11-14
Fix schema content deprications

YvarRavy:
described in: #5692

i've basically moved the logic from Content class to the ContentHandler. also i've changed a few variables from private to public so the ContentHandler can call them. i hope this is okay. this seemed to be the quickest way to fix the deprecation warning.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling and processing logic for schema content rendering.
	- Improved visibility of certain properties and methods for greater accessibility.

- **Bug Fixes**
	- Removed deprecated method for populating parser output, streamlining content handling.

- **Tests**
	- Removed outdated test methods related to the deprecated `fillParserOutput` functionality.


YvarRavy:
i forgot to change the tests, my bad. one problem i can move the tests from schemaContent to schemaContentHandler but since the fillParserOuput is protected in that class i can't call it. i could use the fillParserOutputInternal but i don't know if that is smart

gesinn-it-gea:
Regarding testing etc. I think rewriting the logic to use setter/getter would be a better approach

YvarRavy:
> Regarding testing etc. I think rewriting the logic to use setter/getter would be a better approach

Yeah i was planning todo that after seeing the response from the bot. But that doesn't fix the problem regarding the test case or do i not fully understand what you said?

gesinn-it-gea:
... just wanted to give my 5 cents to why using getters / setters is better in the context of testing:

Using setters and getters instead of public variables provides encapsulation, allowing you to add validation, logging, or debugging checks directly within these methods. This control ensures that data changes adhere to specific rules, which is particularly useful in testing scenarios. With public variables, you lose this ability, risking inconsistent or invalid states that can lead to hard-to-diagnose bugs. Setters/getters also make future modifications easier (e.g., adding validation) without changing the class interface, keeping tests stable.

This should to my opinion be considered when code and tests are changed.

JeroenDeDauw:
Agreed. Seems like you are only reading the property, so a public getter is better than making the property public, since a getter does not allow for mutation.

YvarRavy:
> Agreed. Seems like you are only reading the property, so a public getter is better than making the property public, since a getter does not allow for mutation.

do you have an idea on what todo with the unitTests? as said before i can move them but the fillParserOutput is protected inside the ContentHandler

YvarRavy:
@JeroenDeDauw i'm not sure what todo with the unit tests, since the fillParserOutput is protected. can i just remove the unit tets?

src/MediaWiki/Content/SchemaContent.php
```diff
@@ -268,6 +268,24 @@ public function initServices() {
 		}
 	}
 
+	/**
+	 * Gets the content formatter.
+	 *
+	 * @return SchemaContentFormatter|null The content formatter instance or null if not set.
+	 */
+	public function getContentFormatter() {
```
JeroenDeDauw, line 276: These comments do not add much value. You could just do 

```php
public function getContentFormatter(): SchemaContentFormatter|null {
```

tests/phpunit/MediaWiki/Content/SchemaContentTest.php
```diff
@@ -245,109 +245,4 @@ public function testPreSaveTransform() {
...
-
-		$instance = new SchemaContent(
-			json_encode( [ 'Foo' => 42 ] )
-		);
-
-		$generateHtml = true;
-		$revId = 42;
-
-		$instance->fillParserOutput( $title, $revId, $parserOptions, $generateHtml, $parserOutput );
-	}
```
JeroenDeDauw, line 351: Throwing out these poorly designed and brittle tests is reasonable enough. At the same time it's a pity to not have any good tests for the modified code.

Issue #5743, opened 2024-10-30
SMW should not load unnecessary ResourceLoader modules 


alistair3149:
### Setup

- SMW version: `5.0.0-alpha`
- MW version: `1.39.10`
- PHP version: `8.1.30`
- DB system (MySQL, Blazegraph, etc.) and version: `11.4.3-MariaDB-ubu2404`

### Issue

SMW is loading unnecessary ResourceLoader modules even when the page do not contain any elements that requires them.
These unused styles and scripts are render-blocking and significantly affect the load time of the page.

- ✅ `ext.smw.style` - load on all pages [(source)](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/b045a2db20bfe4d303c5fbe045b4645e63c11b5b/src/MediaWiki/Hooks/BeforePageDisplay.php#L71) 
    - Partially addressed with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5752, we should break down the styles
    - Moved away from base64 images with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5761
- ✅ `ext.smw.tooltip.styles` - load on all pages [(source)](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/b045a2db20bfe4d303c5fbe045b4645e63c11b5b/src/MediaWiki/Hooks/BeforePageDisplay.php#L72)
    - Fixed with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5752
- ✅ `ext.smw.special.styles` - load on all pages under `Special` namespace  [(source)](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/b045a2db20bfe4d303c5fbe045b4645e63c11b5b/src/MediaWiki/Hooks/BeforePageDisplay.php#L79)
    - Fixed with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5752 and https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5753
- ✅ `ext.smw.skin.css` - load on all skins, it should use skinStyles instead.
    - Fixed with https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5755
- ✅ `ext.libs.tippy` - load on all pages, this is especially bad because it is an external library
[...]

Issue #5546, opened 2023-10-26
SMW should use Vue.js and Codex


krabina:
VueJS is the Wikimedia Foundation's official choice for adoption as the JavaScript framework for use within MediaWiki, available since MW 1.35.
Codex is the new design system for MediaWiki and will replace OOUI. Therfore we should talk about using it in SMW and related extensions.

- https://www.mediawiki.org/wiki/Vue.js
- https://www.mediawiki.org/wiki/Codex

Comments are welcome

Nikerabbit:
At least for Codex to be usable, it would require waiting for a long time or reducing the number of supported MediaWiki versions. Or bundling Codex itself.

Vue might be more practical, but even though it's supported since 1.35, it's not the same version across all those releases.

On the other hand, I don't feel like the UI of SMW has big problems usability- or tech-wise. I'd prefer fixing scalability issues and addressing deprecation warnings.

krabina:
I agree that there are more important issues, just wanted to open a thread about this as well since it might affect SMW at some point.

alistair3149:
We have to figure out how to handle the compatibility issues with different MediaWiki version.

Like OOUI, each version of MediaWiki comes with a different version of Codex.
We probably can't use Codex fully in PHP at least until we move minimum MediaWiki version to 1.43 since CSS-only components and PHP support are not in 1.39. The most we can do is like #5786, we start converting existing UI to use Codex design tokens so that the UI ais at least more visually consistent with Codex.

krabina:
Thank you for looking into this!

PR #5754, opened 2024-11-15
5.0.0 release

JeroenDeDauw:
Looking at the changes since 4.2.0, they all seem to be compatibility related. 

Let me know if I missed something that should be mentioned in the release notes. New features, bug fixes, or something else.

## Summary by CodeRabbit

- **New Features**
	- Semantic MediaWiki version 5.0.0 released with improved compatibility for MediaWiki 1.42 and 1.43, and PHP 8.1 and above.
	- Added support for SQL databases including MySQL, SQLite, and PostgreSQL.
	- Introduced a feature to disable the RDF link in the head element.
	- New sections in release notes detailing highlights, user interface changes, and performance improvements.

- **Bug Fixes**
	- Addressed issues with the SMWSearch feature and importing vocabulary.

- **Documentation**
	- Updated compatibility details in the `COMPATIBILITY.md` document.
	- Release notes for version 5.0.0 revised for clarity and accuracy, including upgrade process details.

PR #5758, opened 2024-11-16
Replace smw-callout with MW core message boxes in PHP

alistair3149:
MW core already comes with a message box system that is simliar to SMW callout. Switching to the core implementation will reduce the maintenance needed, keep the styles aligned with core, and reduce unnessecary code.

The JS usage of smw-callout will be replaced in a follow up PR.

Related: #5743, #5546

## Release Notes

- **New Features**
  - Enhanced error message presentation across various components using standardized HTML output methods.
  
- **Bug Fixes**
  - Corrected parameter type annotations for methods to improve clarity and documentation.

- **Tests**
  - Introduced new methods to standardize error message generation in test cases, improving consistency and maintainability.
  - Updated expected output in multiple test cases to reflect changes in error message formatting.

### DIFF
--- a/src/Property/DeclarationExaminerMsgBuilder.php
+++ b/src/Property/DeclarationExaminerMsgBuilder.php
@@ -87,7 +87,7 @@ class DeclarationExaminerMsgBuilder {
 		}
 
 		$class = "plainlinks $msgKey";
-		switch( $type ) {
+		switch ( $type ) {
 			case 'plain':
 				return Html::rawElement( 'div', [ 'class' => $class ], $msg );
 			case 'error':
--- a/src/Property/DeclarationExaminerMsgBuilder.php
+++ b/src/Property/DeclarationExaminerMsgBuilder.php
@@ -86,18 +86,19 @@ class DeclarationExaminerMsgBuilder {
 			return;
 		}
 
-		if ( $type !== 'plain' ) {
-			$class = "smw-callout smw-callout-$type";
+		$class = "plainlinks $msgKey";
+		switch( $type ) {
+			case 'plain':
+				return Html::rawElement( 'div', [ 'class' => $class ], $msg );
+			case 'error':
+				return Html::errorBox( $msg, '', $class );
+			case 'info':
+				return Html::noticeBox( $msg, $class );
+			case 'warning':
+				return Html::warningBox( $msg, $class );
+			default:
+				return '';
 		}
-
-		return Html::rawElement(
-			'div',
-			[
-				'id' => "$msgKey",
-				'class' => "plainlinks $msgKey $class"
-			],
-			$msg
-		);
 	}
 
 	private function msg( $msg, $type = Message::PARSE, $lang = Message::USER_LANGUAGE ) {
--- a/tests/phpunit/Integration/JSONScript/TestCases/s-0018.json
+++ b/tests/phpunit/Integration/JSONScript/TestCases/s-0018.json
@@ -23,7 +23,7 @@
 			},
 			"assert-output": {
 				"to-contain": [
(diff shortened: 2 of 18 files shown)


tests/phpunit/Integration/JSONScript/TestCases/s-0018.json
```diff
@@ -23,7 +23,7 @@
 			},
 			"assert-output": {
 				"to-contain": [
-					"<div id=\"ask-status\" class=\"smw-ask-status plainlinks\"><noscript><div class=\"smw-callout smw-callout-error\">"
+					"smw-error-noscript"
```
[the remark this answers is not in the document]
alistair3149: It is intended. The test is looking for the `smw-error-noscript` class to test if the output is correct.

src/Property/DeclarationExaminerMsgBuilder.php
```diff
@@ -86,18 +86,19 @@ private function makeHTML( $type, $message ) {
 			return;
 		}
 
-		if ( $type !== 'plain' ) {
-			$class = "smw-callout smw-callout-$type";
+		$class = "plainlinks $msgKey";
+		switch( $type ) {
+			case 'plain':
+				return Html::rawElement( 'div', [ 'class' => $class ], $msg );
```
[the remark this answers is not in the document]
alistair3149: I am not sure whether `$msg` contains HTML, so someone else more familiar with it should take a look.

tests/phpunit/MediaWiki/Specials/Ask/ErrorWidgetTest.php
```diff
@@ -57,6 +58,18 @@ public function testGetFormattedQueryErrorElement( $errors, $expected ) {
...
+	 * Return an error box using core MW method
+	 * This is required because the output is different depending on the MW version
+	 *
+	 * @param string $message
+	 * @return $string HTML of the error message box
+	 * @since 5.0.0
+	 */
+	private function getErrorMessageHTML( $message ) {
+		return Html::errorBox( $message, '', 'smw-error-result-error' );
+	}
```
JeroenDeDauw, line 71: Computing the expected value like this is somewhat dodgy. Normally it is good to avoid mirroring the production code in the test.

JeroenDeDauw (commented):
It's be good to mention that some CSS classes changed in the release notes (added in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5754)

tests/phpunit/MediaWiki/Specials/Ask/ErrorWidgetTest.php
alistair3149: In that case do we want to hardcode the HTML for the error box in the test? We'll need to do multiple versions because it is using Codex in the newer MW version, and Codex components are not stable

JeroenDeDauw: Well, nvm then. This undermines the value of the tests, but they still seem to provide something.

alistair3149: I am not super familiar with PHP tests so I am not sure if it is possible or the right way to do it.
We could validate the HTML/content that we feed into the error box, since the `Html::errorBox` method only provides the wrapper. Would that be a more sensible way to do it?

kghbln:
@alistair3149 What do the boxes look like now? I have about 50 wikis where I reused the "smw-callout" class for regular templates. This included semantic-mediawiki.org, sandbox.semantic-mediawiki.org, etc.

alistair3149:
@kghbln it depends on your MediaWiki version and the skin that you're using. By default it looks something like this: https://doc.wikimedia.org/codex/main/components/demos/message.html

PR #5775, opened 2024-11-20
Drop jQuery modules from ext.smw.api

alistair3149:
- `ext.jquery.jStorage` - that was replaced by `mediawiki.storage`
- `ext.jquery.md5` - it is wasteful to use a whole MD5 library to generate an unique ID for one line of code, we can use cyrb53 instead.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new hashing method in the API for generating hashes from strings.
- **Bug Fixes**
	- Updated the storage mechanism from jQuery's storage to MediaWiki's storage for improved compatibility.
- **Deprecations**
	- Removed outdated jQuery plugins related to local storage and MD5 hashing, streamlining the codebase.


jaideraf:
`Error: Unknown module: ext.jquery.jStorage`

It was required here, see the console:
https://wikincat.org/wiki/Wikincat:Home

JeroenDeDauw:
That is presumably coming from https://github.com/search?q=repo%3ASemanticMediaWiki%2FSemanticResultFormats%20ext.jquery.jStorage&type=code

@alistair3149 I've been merging those changes so far since we are making a 5.0 release anyway. But do keep in mind that changing CSS classes and resource loader module names are all breaking changes, and that in some cases we have to think about compatibility.

JeroenDeDauw:
> As for compatibility, I'll comment a list of breaking changes in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5754 to make the transition smoother. Would that be a suitable approach?

image

alistair3149:
> `Error: Unknown module: ext.jquery.jStorage`
> 
> It was required here, see the console: https://wikincat.org/wiki/Wikincat:Home

Follow up PR in SRF here: https://github.com/SemanticMediaWiki/SemanticResultFormats/pull/862

jaideraf:
Confirmed. The error `Unknown module: ext.jquery.jStorage` is gone. 👍🏼

PR #5776, opened 2024-11-20
Allow RDF link in the head element to be disabled

alistair3149:
Used to avoid bots scraping ExportRDF by not advertising on every page because the page is not a cheap call.

The idea is based on the implementation from the WeirdGloop: https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/c7ea14bb20f5d5cfe8dc9134f4f5336e6d2109c3

TODO: When this is merged, update https://www.semantic-mediawiki.org/wiki/Help:Configuration#Export_settings

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration setting to control the visibility of the RDF export link in page headers.
  
- **Bug Fixes**
	- Enhanced logic to ensure the export link is only added when the new setting is enabled.

- **Chores**
	- Minor formatting adjustments for improved readability in the code.

### DIFF
--- a/includes/DefaultSettings.php
+++ b/includes/DefaultSettings.php
@@ -2510,7 +2510,16 @@ return ( function () {
 		 *
 		 * @since 3.2
 		 */
-		'smwgDetectOutdatedData' => false
+		'smwgDetectOutdatedData' => false,
+
+		/**
+		 * Whether Special:ExportRDF <link> element is added to the <head> element
+		 * Used to avoid bots scraping ExportRDF by not advertising on every page
+		 * Because the page is not a cheap call
+		 *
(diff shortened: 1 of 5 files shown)


jaideraf:
> TODO: When this is merged, update https://www.semantic-mediawiki.org/wiki/Help:Configuration#Export_settings

Done: https://www.semantic-mediawiki.org/wiki/Help:$smwgEnableExportRDFLink

krabina:
Question: maybe if it is disabled on every source page and you are providing RDF dumps, maybe there should be a way to point crawlers to the RDF dump on every page instead of Special:ExportRDF

PR #5797, opened 2024-12-01
Remove redundant methods and values

alistair3149:


src/Localizer/Message.php
```diff
@@ -150,10 +150,8 @@ public static function encode( $message, $type = null ) {
 	 * @since 2.5
 	 *
 	 * @param string $messageId
-	 *
-	 * @return boolean
 	 */
-	public static function exists( $message ) {
+	public static function exists( $message ): bool {
```
JeroenDeDauw, line 154: Any reason why you put a PHP return type but did not put the parameter type?

src/MediaWiki/Page/ConceptPage.php
```diff
@@ -22,36 +19,19 @@
...
-
-	/**
-	 * @var Store
-	 */
-	private $store;
-
-	/**
-	 * @var DataValue
-	 */
-	private $propertyValue;
```
JeroenDeDauw, line 38: Random unused fields :thinking:

JeroenDeDauw (approved):

src/MediaWiki/Page/ConceptPage.php
JeroenDeDauw: (I verified they are unused)

PR #5806, opened 2024-12-11
Fix "Deprecated: Return type of x should either be compatible with Iterator::x"

paladox:
I have made the types match https://www.php.net/manual/en/class.iterator.php

We require php 7.4+ so there is no incompatibility.

### DIFF
--- a/src/Iterators/CsvFileIterator.php
+++ b/src/Iterators/CsvFileIterator.php
@@ -119,7 +119,7 @@ class CsvFileIterator implements Iterator, Countable {
 	 *
 	 * {@inheritDoc}
 	 */
-	public function rewind() {
+	public function rewind(): rewind {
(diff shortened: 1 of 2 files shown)


paladox:
@JeroenDeDauw if you prefer, I can use "#[\ReturnTypeWillChange]". Which would be a noop and fix the warning.

JeroenDeDauw:
If the tests are indeed running with PHP 7.4, then that is an issue that should be fixed, though presumably in a separate PR.

paladox:
@JeroenDeDauw this is ready to be merged.

JeroenDeDauw:
The PHP CI issue is now tracked by
* https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5813

PR #5809, opened 2024-12-11
Replace removed "LanguageEn::getConverter()" method

paladox:
We require MW 1.39+ so this does not break any support.

This method were removed in https://gerrit.wikimedia.org/r/c/mediawiki/core/+/936685


paladox:
@coderabbitai full review

PR #5816, opened 2024-12-11
Fix "Creation of dynamic property SMW\Tests\ParserFunctionFactoryTest::$testEnvironment is deprecated"

paladox:
Also fixes "Creation of dynamic property SMW\Tests\ParserFunctionFactoryTest::$parserData is deprecated".


tests/phpunit/ParserFunctionFactoryTest.php
```diff
@@ -19,6 +19,10 @@ class ParserFunctionFactoryTest extends \PHPUnit_Framework_TestCase {
 
 	use PHPUnitCompat;
 
+	private $testEnvironment;
+
+	private $parserData;
```
JeroenDeDauw, line 24: You can use property types

PR #5825, opened 2024-12-11
Allow phpcodesniffer-composer-installer

alistair3149:
- [Allow phpcodesniffer-composer-installer](https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/cceb93b789d7857950cdc9ab2d4dbe26ce07d690), which should gets Tesa CI working again

## Summary by CodeRabbit

- **New Features**
	- Enhanced autoloading for test classes with the addition of the `SMW\\Tests\\` namespace.
	- Improved plugin management by allowing the `dealerdirect/phpcodesniffer-composer-installer`.
  
- **Bug Fixes**
	- Adjusted classmap entries for better organization of test-related files.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -83,6 +83,7 @@
 		"psr-4": {
 			"SMW\\": "src/",
 			"SMW\\Maintenance\\": "maintenance/",
+			"SMW\\Tests\\": "tests/phpunit/",
 			"Onoi\\Tesa\\": "src/Tesa/src/"
 		},
 		"psr-0": {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw (changes requested):
I suspect that some code from `tests/phpunit/includes/` is used by extensions to SMW like Semantic Result Formats, in their own test suites. Placing it in autoload-dev makes it unavailable there.

alistair3149:
> I suspect that some code from `tests/phpunit/includes/` is used by extensions to SMW like Semantic Result Formats, in their own test suites. Placing it in autoload-dev makes it unavailable there.

Yes you are right, it seems to be [that way](https://codesearch.wmcloud.org/search/?q=SMW%5C%5CTest&files=&excludeFiles=&repos=). I will revert that commit then.

JeroenDeDauw (approved):

PR #5847, opened 2024-12-14
Replace PHPUnit classes with their namespaced equivilent

paladox:
Under MW 1.43+ it uses namespaced phpunit classes now.

This change is a NOOP for MW 1.39-1.42.


paladox:
@gesinn-it-gea this should get phpunit to run.

PR #5849, opened 2024-12-14
QueryPage: Fix "contains 1 abstract method"

paladox:
In MW 1.43, getQueryInfo was made abstract so we have to declare it. Before MW 1.43 it returned null so we return null here.

Also fix getQueryInfo  return type in PropertiesQueryPage.php. It has to match the one in QueryPage.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method to retrieve query-related information.

### DIFF
--- a/includes/querypages/QueryPage.php
+++ b/includes/querypages/QueryPage.php
@@ -247,4 +247,11 @@ abstract class QueryPage extends \QueryPage {
 
 		return $num;
 	}
+
+	/**
+	 * @return array|bool
+	 */
+	public function getQueryInfo() {
+		return null;
+	}
 }
--- a/includes/querypages/QueryPage.php
+++ b/includes/querypages/QueryPage.php
@@ -249,7 +249,7 @@ abstract class QueryPage extends \QueryPage {
 	}
 
 	/**
-	 * @return array|bool
+	 * @return array|null
 	 */
 	public function getQueryInfo() {
 		return null;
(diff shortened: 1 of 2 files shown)


includes/querypages/QueryPage.php
```diff
@@ -247,4 +247,11 @@ function doQuery( $offset = false, $limit = false, $property = false ) {
 
 		return $num;
 	}
+
+	/**
+	 * @return array|bool
+	 */
+	public function getQueryInfo() {
+		return null;
```
JeroenDeDauw, line 255: Does not match the type doc. If the real type is array|bool, then you can do

```php
public function getQueryInfo(): ?array {
```

paladox: Thanks, done.

JeroenDeDauw: Why not the PHP type instead of the comment?

paladox: Oh, I didn’t realise you wanted that. Done now. Thanks!

includes/querypages/QueryPage.php
```diff
@@ -247,4 +247,11 @@ function doQuery( $offset = false, $limit = false, $property = false ) {
 
 		return $num;
 	}
+
+	/**
+	 * @return array|null
+	 */
+	public function getQueryInfo(): ?array {
```
JeroenDeDauw, line 254: Once the type is in PHP, the comment becomes noise, so is better removed

PR #5850, opened 2024-12-14
PropertiesQueryPage: Fix getQueryInfo value to be null

paladox:
This matches before it was made abstract. It uses isset checks and returns true even if you use false.

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated the output of the query method to enhance result interpretation, allowing for a null return value.

### DIFF
--- a/includes/querypages/PropertiesQueryPage.php
+++ b/includes/querypages/PropertiesQueryPage.php
@@ -273,7 +273,7 @@ class PropertiesQueryPage extends QueryPage {
 	}
 
 	/**
-	 * @return array|bool
+	 * @return array|null
 	 */
 	public function getQueryInfo() {
 		return null;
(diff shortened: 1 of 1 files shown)


includes/querypages/PropertiesQueryPage.php
```diff
@@ -276,7 +276,7 @@ function getResults( $requestOptions ) {
 	 * @return array|bool
 	 */
 	public function getQueryInfo() {
-		return false;
+		return null;
```
JeroenDeDauw, line 279: This does not match with the above type doc

paladox: Fixed, thanks, done.

paladox:
I’ve had to put this into https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5849 as it requires fixing at the same time to match QueryPage.

PR #5855, opened 2024-12-15
Replace 'PHPUnit_Framework_MockObject_Builder_InvocationMocker' with 'PHPUnit\Framework\MockObject\Builder'

paladox:


paladox:
Note that withConsecutive within InvocationMocker is removed in PHPUnit 10 so an alternative will have to be found to support version 10.

PR #5856, opened 2024-12-15
Replace usage of 'PHPUnit_Framework_MockObject_MockObject'

paladox:


paladox:
Note that MockObject is removed in PHPUnit 10 so a alternative will have to be found to support version 10.

PR #5859, opened 2024-12-15
JsonUnserializable: Fix checking if interface exists

paladox:
Because we namespace under MediaWiki\Json, we have to either act like the class is under it or we have to use \<name>.

This should fix:

```
Warning: Cannot declare interface MediaWiki\Json\JsonUnserializable, because the name is already in use in /var/www/html/includes/json/JsonDeserializable.php on line 51
```

## Summary by CodeRabbit

- **Bug Fixes**
	- Simplified the check for the existence of the `JsonUnserializable` interface, enhancing flexibility.

### DIFF
--- a/includes/compat/JsonUnserializable.php
+++ b/includes/compat/JsonUnserializable.php
@@ -2,7 +2,7 @@
 
 namespace MediaWiki\Json;
 
-if( !interface_exists( 'MediaWiki\Json\JsonUnserializable' ) ) {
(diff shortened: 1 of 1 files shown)


paladox:
```
> namespace MediaWiki\Json;

> var_dump(!interface_exists( 'JsonUnserializable' ));
bool(true)

> var_dump(!interface_exists( 'MediaWiki\Json\JsonUnserializable' ));
bool(false)
```

PR #5860, opened 2024-12-15
Remove JsonUnserializable

paladox:
We support MW 1.39+ only now. We can remove this.

## Summary by CodeRabbit

- **Chores**
	- Removed the `JsonUnserializable` interface to streamline compatibility with MediaWiki versions 1.36 and above.


paladox:
Getting:

```
Warning: Cannot declare interface MediaWiki\Json\JsonUnserializable, because the name is already in use in /var/www/html/includes/json/JsonDeserializable.php on line 51
```

under MW 1.43.

This change should fix it I think?

PR #5871, opened 2024-12-19
JSONScriptTestCaseRunner: Remove unneeded definition of testDatabaseTableBuilder var

paladox:
It's already defined in SMWIntegrationTestCase

## Summary by CodeRabbit

- **Bug Fixes**
	- Removed deprecated property to streamline test case execution.
  
- **New Features**
	- Introduced a new preferred method for page creation, enhancing flexibility in testing.

### DIFF
--- a/tests/phpunit/JSONScriptTestCaseRunner.php
+++ b/tests/phpunit/JSONScriptTestCaseRunner.php
@@ -7,7 +7,6 @@ use SMW\Localizer\Localizer;
 use SMW\Tests\Utils\JSONScript\JsonTestCaseContentHandler;
 use SMW\Tests\Utils\JSONScript\JsonTestCaseFileHandler;
 use SMW\Tests\Utils\UtilityFactory;
-use SMW\Tests\Utils\Connection\TestDatabaseTableBuilder;
 
 /**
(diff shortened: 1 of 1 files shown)


paladox:
Not sure if this will fix https://github.com/SemanticMediaWiki/SemanticDrilldown/pull/85. But this is already defined in the class it's extending. We don't want to overwrite it by defining it in JSONScriptTestCaseRunner as well.

paladox:
Ahhh

doing something like:

```
<?php

class test {
    protected $test = 'test';
}

class test2 extends test {
    protected $test;

    public function getProtectedProperty(){
        return $this->test;
    }

}

$testCase = new test2();
echo $testCase->getProtectedProperty();
```

shows that the protected var from the first class gets overwritten. So instead of 'test' nothing returns.

gesinn-it-ilm:
@paladox good point. Yes, that var is unneeded there. Not sure if this update will affect the MW 1.42 for SemanticDrilldown in a proper way, but let's see.

gesinn-it-ilm (approved):

paladox:
Nope that didn't fix it for Drilldown :(

Issue #5880, opened 2024-12-22
MW 1.40 - fix failing Unit test


gesinn-it-ilm:
After logs in CI for MW `1.40` were checked, there is one error in **Unit** tests, but still it looks like CI for 1.40 is green, or passing. 

The error can be checked here: 
 [error_MW_1.40_ci](https://github.com/SemanticMediaWiki/SemanticMediaWiki/actions/runs/12446828850/job/34749648550#step:5:2757)

Unit tests result:
[unit_test_result_1.40](https://github.com/SemanticMediaWiki/SemanticMediaWiki/actions/runs/12446828850/job/34749648550#step:5:11758)

gesinn-it-ilm:
Failing test is in `ScopeMemoryLimiterTest.php`. 
The error indicates that the test fails because class `ScopeMemoryLimiter` attempted to set the memory limit to lower value then the current memory usage is. 
This error is fixed by increasing the memory limit to be the same or greater than the current memory usage.
CI for `1.40` now works without issues. 
The PR will be opened.

PR #5882, opened 2024-12-22
Issue 5880 - fix failing Unit test CI MW 1.40

gesinn-it-ilm:
This PR is related to the issue #5880.

This PR contains:

- increase the memory limit in `ScopeMemoryLimiterTest.php`
- the error occurs because memory attempted to be set to the value which is lower than the current memory usage

## Summary by CodeRabbit

- **Tests**
	- Improved clarity with added comments in the `ScopeMemoryLimiterTest`.
	- Adjusted memory limit calculations for more accurate testing.
	- Added new assertions to verify callable execution with expected memory limits.


tests/phpunit/Elastic/Indexer/Attachment/ScopeMemoryLimiterTest.php
```diff
@@ -89,7 +89,7 @@ public function testExecute() {
 		$converter = new ScopeMemoryLimiter();
 
 		if ( $memoryLimitBefore === "-1" ) {
-			$memoryLimitBefore = memory_get_usage() + $converter->toInt( '10M' );
+			$memoryLimitBefore = memory_get_usage() + $converter->toInt( '20M' );
```
gesinn-it-gea, line 92: A fixed value like '20M' doesn’t consider the current memory usage or the system's overall memory limit. If the current memory usage already exceeds the fixed value, the test will fail. For more complex or memory-intensive tests, 20 MB might be insufficient, while in some cases, it's unnecessarily large.

@JeroenDeDauw what do you think? I'd suggest to follow a more dynamic solution that should:

- Take the current memory usage into account.
- Ensure that the new memory limit is always above the current memory usage.
- Provide a configurable buffer to account for additional memory needs.

gesinn-it-ilm: @gesinn-it-gea, @JeroenDeDauw maybe this approach or something like this:

```
public function testExecute() {
		// Store the original memory limit
		$originalMemoryLimitBefore = ini_get( 'memory_limit' );
		$converter = new ScopeMemoryLimiter();
	
		// Get current memory usage
		$currentMemoryUsage = memory_get_usage();
	
		// Set a configurable buffer (e.g., 20MB) to ensure the new memory limit is above current usage
		$buffer = $converter->toInt( '20M' );
		
		// Calculate new memory limit (current usage + buffer)
		$memoryLimitBefore = $currentMemoryUsage + $buffer;
	
		// If the original memory limit is set to -1 (unlimited), set it to the calculated value
		if ( $originalMemoryLimitBefore === "-1" ) {
			ini_set( 'memory_limit', $memoryLimitBefore );
		}
	
		// Mock a callable to test the execution
		$this->testCaller = $this->getMockBuilder( '\stdClass' )
			->setMethods( ['calledFromCallable'] )
			->getMock();
	
		$this->testCaller->expects( $this->once() )
			->method( 'calledFromCallable' );
	
		// Calculate the final target memory limit (including a small increment)
		$memoryLimit = $memoryLimitBefore + $converter->toInt( '1M' );
	
		// Create an instance of ScopeMemoryLimiter with the new memory limit
		$instance = new ScopeMemoryLimiter( $memoryLimit );
	
		$instance->execute( [ $this, 'runCallable' ] );
	
		$this->assertEquals(
			$memoryLimit,
			$this->memoryLimitFromCallable,
			"Limit we expected got set."
		);
	
		$this->assertEquals(
			$memoryLimitBefore,
			$instance->getMemoryLimit(),
			"Limit was reset successfully."
		);
	
		ini_set( 'memory_limit', $originalMemoryLimitBefore );
	}
```

paladox:
Looks like this is needed to remove "MediaWiki.PHPUnit.SetMethods.SetMethods" from .phpcs.xml. Otherwise tests fail (took me a bit to track this down, was wondering why it was failing).

paladox:
This needs rebasing, it'll fail phpcs so needs rebasing and resolving.

paladox:
You probably want to rebase since master has some new phpcs checks and I think this would cause failure as phpcs needs fixing for this pull.

paladox:
> @paladox now there are more errors for MW 1.40 after some of the sniffs are fixed. [errors_MW_1.40](https://github.com/SemanticMediaWiki/SemanticMediaWiki/actions/runs/12505439707/job/34888808744#step:5:11812)
> 
> The error linked to this PR is fixed, but 4 remain still there.

I guess it could either be removed or changed to be fixed. I'm not sure it ever worked as it's a void?

gesinn-it-ilm:
@paladox not sure as well. There are a lot of tests, possible that some of those are not up to date anymore, or should be removed.

gesinn-it-ilm:
This PR is covered as a part of PR #5896. 
Will be closed.

PR #5886, opened 2024-12-24
Fix "TypeError: SearchEngineFactory::getSearchEngineClass(): Argument #1 ($dbProvider) must be of type x"

paladox:
## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for connection parameters in the search engine classes.
	- Increased flexibility in connection type acceptance for search engine instantiation.

- **Bug Fixes**
	- Improved type validation to prevent runtime errors related to connection types.


YOUR1:
The code seems good to me, can you coverage with tests as well?

paladox:
@gesinn-it-gea / @gesinn-it-ilm are you able to look into this please? Specifically $connection = $applicationFactory->getConnectionManager()->getConnection( DB_REPLICA ); which is broken because it returns the wrong type. (Would need to be versioned to use it only on like MW 1.39 and a few others (I forgot which MW release changed the type required).

paladox:
Unless we can just not use getConnectionManager and instead use MW cores functions?

PR #5891, opened 2024-12-25
Fix phpunit phpcs partially

paladox:


JeroenDeDauw:
C'mon @paladox. Those are rookie numbers. Only 4 digits?

image

paladox:
> C'mon @paladox. Those are rookie numbers. Only 4 digits?
> 
> image

I don't think I understand what you mean?

PR #5903, opened 2024-12-31
Also test with PHP 8.0

JeroenDeDauw:
## Summary by CodeRabbit

- **Chores**
	- Updated GitHub Actions workflow configuration to adjust PHP version for MediaWiki 1.39 from 8.1 to 8.0
	- Maintained existing CI/CD pipeline structure and test configurations

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
       matrix:
         include:
           - mediawiki_version: '1.39'
-            php_version: 8.1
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
@gesinn-it-ilm @gesinn-it-gea not being able to test with one of the versions at the boundary of our supported range is problematic. It's easy to accidentally introduce incompatibilities and break support for versions you declare you support (PHP 8.0 in this case).

PR #5908, opened 2025-01-01
Fix skipping tests on MW 1.41+

paladox:
It was skipping it for MW 1.39 as well:

> MediaWiki 1.39.11 version is not supported (Check failing assertions for MW 1.41+.)

### DIFF
--- a/tests/phpunit/Integration/JSONScript/TestCases/s-0009.json
+++ b/tests/phpunit/Integration/JSONScript/TestCases/s-0009.json
@@ -61,7 +61,7 @@
 	},
 	"meta": {
 		"skip-on": {
-			"mw-1.40>": "Check connection provider, should be IProviderConnection for MW 1.41+."
+			">mw-1.40": "Check connection provider, should be IProviderConnection for MW 1.41+."
 		},
 		"version": "2",
 		"is-incomplete": false,
--- a/tests/phpunit/Integration/JSONScript/TestCases/s-0009.json
+++ b/tests/phpunit/Integration/JSONScript/TestCases/s-0009.json
@@ -16,6 +16,9 @@
 			"type": "parser",
 			"about": "#0",
 			"subject": "Example/S0009/1",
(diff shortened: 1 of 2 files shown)


paladox:
@coderabbitai review

paladox:
@gesinn-it-gea this seems to not skip on mw 1.41+. I tried to figure out but couldn’t figure it out.

this mw 1.41 tests are broken again I think

paladox:
Any idea how to fix it @gesinn-it-gea ?

gesinn-it-ilm:
@paladox yes, the CI is broken again for MW 1.41. I will take a look.

gesinn-it-ilm:
@paladox, this will work:

```
"meta": {
	"skip-on": {
		"mw-1.41.x": "Check connection provider, should be IProviderConnection for MW 1.41+."
		},
	"version": "2",
	"is-incomplete": false,
	"debug": false
}
```

Using this skip-on part in `meta` section.

gesinn-it-ilm:
@gesinn-it-gea, @paladox according to the docs I found here [design.md](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Integration/JSONScript/docs/design.md#skipping-a-test-or-mark-as-incomplete), looks like the skip-on should use this approach:

`"mediawiki": [ ">1.30.x", "MediaWiki changed ..." ],`

For some reason in this test case `s-0009.json` it won't work. Not sure why, but only way to skip it is by using current approach commited this morning as a part of PR #5918.

paladox:
@gesinn-it-ilm the revert can be merged for now.

gesinn-it-ilm:
@gesinn-it-gea @paladox I have checked the skip-on, the issue with this specific test `s-0009.json` is there because the test fails before it could trigger the skip-on. 
So, until we fix the test, it should remain like this. After that, we can update skip-on part all across the SMW.

PR #5909, opened 2025-01-01
Fix division by zero in CliMsgFormatter#progressCompact

paladox:
## Summary by CodeRabbit

- **Refactor**
	- Improved progress calculation logic in the CLI message formatter.
	- Centralized progress percentage calculation with a new private method.
	- Enhanced modularity of progress-related methods.


paladox:
```
5) SMW\Tests\Integration\Maintenance\UpdateEntityCountMapTest::testRun
DivisionByZeroError: Division by zero
/var/www/html/extensions/SemanticMediaWiki/src/Utils/CliMsgFormatter.php:121
/var/www/html/extensions/SemanticMediaWiki/maintenance/updateEntityCountMap.php:262
/var/www/html/extensions/SemanticMediaWiki/maintenance/updateEntityCountMap.php:144
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/Utils/Runners/MaintenanceRunner.php:108
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/Integration/Maintenance/UpdateEntityCountMapTest.php:47
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/SMWIntegrationTestCase.php:205
```

src/Utils/CliMsgFormatter.php
```diff
@@ -118,7 +118,7 @@ public function progressCompact( int $i, int $total, ?int $current = null, ?int
 			$last = $total;
 		}
 
-		$progress = min( 100, round( ( $i / $total ) * 100 ) );
+		$progress = min( 100, round( ( $total ? $i / $total : 1 ) * 100 ) );
```
Nikerabbit, line 121: Don't put a ternary in middle of the code. It makes it hard to understand, especially without parenthesis to clarify the processing order. It would be better to use if-statement or something like ` $progress = !$total ? 100 : min( 100, round( ( $i / $total ) * 100 ) );`

paladox: Thanks, done.

src/Utils/CliMsgFormatter.php
```diff
@@ -118,7 +118,7 @@ public function progressCompact( int $i, int $total, ?int $current = null, ?int
 			$last = $total;
 		}
 
-		$progress = min( 100, round( ( $i / $total ) * 100 ) );
+		$progress = !$total ? 100 : min( 100, round( ( $i / $total ) * 100 ) );
```
JeroenDeDauw, line 121: I see there is a copy of this logic in the `progress` method. It'd be nice to unify that by extracting a caclulateProgress method used in both places. That new method could use a guard clause instead of this new trinry, and it could use an explicit check for $total being 0, rather than casting the ineger to boolean.

https://www.entropywins.wtf/blog/2019/01/14/readable-functions-guard-clause/
https://dev.to/jeroendedauw/readable-comparisons-2f4o

paladox: Thanks, done.

paladox:
@coderabbitai review

paladox:
@coderabbitai review

src/Utils/CliMsgFormatter.php
```diff
@@ -95,7 +95,26 @@ public function yellow( string $text ): string {
...
 	public function progress( int $i, int $total ): string {
-		return min( 100, round( ( $i / $total ) * 100 ) ) . " %";
+		// @codeCoverageIgnoreStart
+		return $this->calculateProgress( $i, $total, true );
+		// @codeCoverageIgnoreEnd
+	}
+
+	/**
+	 * @codeCoverageIgnore
+	 */
```
JeroenDeDauw, line 105: What is the deal with those ignore tags? I don't see why this code should be ignored. If you added them because the coverage bot gets mad otherwise, we should change the coverage bot

src/Utils/CliMsgFormatter.php
```diff
@@ -95,7 +95,26 @@ public function yellow( string $text ): string {
...
+	private function calculateProgress( int $i, int $total, bool $percentage ) {
+		$value = 100;
+
+		if ( $i > 0 && $total > 0 ) {
+			$value = min( 100, round( ( $i / $total ) * 100 ) );
+		}
+
+		if ( $percentage ) {
+			$value .= " %";
```
JeroenDeDauw, line 114: Did you see this part? I don't think the way you did the extraction makes sense and is correct.

Better if `calculateProgress` returns `int` (it can have this PHP return type) and that `progress` appends the ` %`. `progressCompact` expects an int.

src/Utils/CliMsgFormatter.php
paladox: We can't test this function as it's private.

src/Utils/CliMsgFormatter.php
paladox: Thanks, done.

src/Utils/CliMsgFormatter.php
```diff
@@ -95,7 +95,22 @@ public function yellow( string $text ): string {
 	 * @return string
 	 */
 	public function progress( int $i, int $total ): string {
-		return min( 100, round( ( $i / $total ) * 100 ) ) . " %";
+		// @codeCoverageIgnoreStart
+		return $this->calculateProgress( $i, $total ) . " %";
+		// @codeCoverageIgnoreEnd
+	}
```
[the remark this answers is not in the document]
paladox: @coderabbitai fixed

paladox:
@coderabbitai review

paladox:
@coderabbitai review

gesinn-it-gea:
@JeroenDeDauw can you please merge if you think it's okay as you were more involved in reviewing this PR

PR #5912, opened 2025-01-01
p-0414: Remove check for hhvm

paladox:
HHVM isn't supported by MW anymore and in addition HHVM doesn't support php anymore.


gesinn-it-gea:
@paladox can you please have a look why the test is failing

paladox:
> @paladox can you please have a look why the test is failing

Oh, looks like https://bugs.php.net/bug.php?id=76822. I guess we have to skip on php 7.2 or higher?

paladox:
Done in https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/5913 instead.

PR #5925, opened 2025-01-03
Add @codeCoverageIgnore* to maintenance scripts

paladox:
Code coverage fails on this so we use @codeCoverageIgnoreStart/End for these parts. MediaWiki does this to its maintenance scripts as well.

### DIFF
--- a/maintenance/disposeOutdatedEntities.php
+++ b/maintenance/disposeOutdatedEntities.php
@@ -13,11 +13,13 @@ use Title;
 /**
  * Load the required class
  */
+// @codeCoverageIgnoreStart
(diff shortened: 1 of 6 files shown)


gesinn-it-gea:
@paladox thanks for this PR! Just out of curiosity: What exactly was the failure that occurred?

paladox:
> @paladox thanks for this PR! Just out of curiosity: What exactly was the failure that occurred?

I just followed https://github.com/wikimedia/mediawiki/blob/master/maintenance/addChangeTag.php#L68

I didn’t look much at the codecov report although I have looked at it but not in much great detail.

PR #5927, opened 2025-01-03
BrowseByProperty: Fix 'limit' api type

paladox:
Use 'limit' over 'string' as it's failing under MW 1.43:

> MediaWiki\Api\ApiUsageException: Parameter "limit" accepts only a string value, but got 50


paladox:
See https://github.com/wikimedia/mediawiki-extensions-Cargo/blame/734204e398e483b899ab1df4c3e95ad99a6b92d4/includes/api/CargoQueryAPI.php#L66C5-L66C29

gesinn-it-ilm (approved):

Issue #5957, opened 2025-01-09
Replace deprecated wfGetDB()


freephile:
### Setup

- SMW version:5.0.0-alpha (dev-master)
- MW version:1.43.0
- PHP version:8.1.31 (fpm-fcgi)
- DB system (MariaDB) version: 10.3.39-MariaDB-log

### wfGetDB() is deprecated (since 1.39)
> [!IMPORTANT]
> [See T273239](https://phabricator.wikimedia.org/T273239)

**The only affected file in SMW proper is includes/export/SMW_ExportController.php**

A global search in my setup finds these occurrences in SMW and related extensions:

7 results - 5 files

/opt/htdocs/mediawiki/extensions/Maps/tests/Integration/Semantic/AreaDescriptionTest.php:
53: $area->getSQLCondition( 'geo_table', [ 'id_field', 'lat_field', 'long_field' ], wfGetDB( DB_MASTER ) )
65: $area->getSQLCondition( 'geo_table', [ 'id_field', 'lat_field', 'long_field' ], wfGetDB( DB_MASTER ) )

/opt/htdocs/mediawiki/extensions/SemanticExtraSpecialProperties/src/AppFactory.php:
75: $this->connection = wfGetDB( DB_REPLICA );

/opt/htdocs/mediawiki/extensions/SemanticExtraSpecialProperties/src/PropertyAnnotators/ApprovedDatePropertyAnnotator.php:
66: $dbr = wfGetDB( DB_REPLICA );

/opt/htdocs/mediawiki/extensions/SemanticMediaWiki/docs/releasenotes/RELEASE-NOTES-1.9.2.md:
32: * #236, #265 Removed wfGetDB from SMWSQLStore3Writers and extended test coverage

/opt/htdocs/mediawiki/extensions/SemanticMediaWiki/includes/export/SMW_ExportController.php:
486: $db = wfGetDB( DB_REPLICA );
553: $db = wfGetDB( DB_REPLICA );

## Solution
We can do a straight replacement of 
`$db = wfGetDB( DB_REPLICA );`
with
`$dbr = MediaWikiServices::getInstance()->getConnectionProvider()->getReplicaDatabase();`
(this assumes a read handle)
and updating occurrences of the handle from '$db' to '$dbr' as is the suggested practice.

### Optionally rewrite queries
Queries *should* work as-is, but could be rewritten in the fluent style where they aren't already.

### Optionally create a mockable version
[...]

PR #5958, opened 2025-01-09
pull req to Replace deprecated wfGetDB

freephile:
Use an instance of MediaWikiServices ConnectionProvider ReplicaDatabase (for reading)
Fixes Issue #5957

## Summary by CodeRabbit

- **Refactor**
	- Enhanced database connection handling for improved compatibility with MediaWiki versions.
	- Updated methods to utilize a new approach for retrieving database connections consistently. 
	- Introduced a new method for obtaining a database handle.

### DIFF
--- a/includes/export/SMW_ExportController.php
+++ b/includes/export/SMW_ExportController.php
@@ -488,7 +488,7 @@ class SMWExportController {
 	 *
 	 * @return \Wikimedia\Rdbms\IDatabase|\Wikimedia\Rdbms\IReadableDatabase
 	 */
-	public static function getDBHandle () {
+	public static function getDBHandle() {
 		if ( version_compare( MW_VERSION, '1.42', '>=' ) ) {
 			return MediaWikiServices::getInstance()->getConnectionProvider()->getReplicaDatabase();
 		} else {
--- a/includes/export/SMW_ExportController.php
+++ b/includes/export/SMW_ExportController.php
@@ -479,11 +479,29 @@ class SMWExportController {
 	}
 
 	/**
+	 * Get a handle for performing database read operations.
+	 *
+	 * This is pretty much wfGetDB() in disguise with support for MW 1.39+
+	 * _without_ triggering WMF CI warnings/errors.
+	 *
+	 * @see https://phabricator.wikimedia.org/T273239
+	 *
+	 * @return \Wikimedia\Rdbms\IDatabase|\Wikimedia\Rdbms\IReadableDatabase
+	 */
+	public static function getDBHandle () {
+		if ( version_compare( MW_VERSION, '1.42', '>=' ) ) {
+			return MediaWikiServices::getInstance()->getConnectionProvider()->getReplicaDatabase();
+		} else {
+			return MediaWikiServices::getInstance()->getDBLoadBalancer()->getConnection( DB_REPLICA );
+		}
+	}
+
+	/**
 	 * @since 2.0 made protected; use printAllToFile or printAllToOutput
 	 */
 	protected function printAll( $ns_restriction, $delay, $delayeach ) {
 		$linkCache = MediaWikiServices::getInstance()->getLinkCache();
-		$dbr = MediaWikiServices::getInstance()->getConnectionProvider()->getReplicaDatabase();
+		$dbr = self::getDBHandle();
 
 		$this->delay_flush = 10;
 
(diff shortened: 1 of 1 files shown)


freephile:
Checks failed because `Error: Call to undefined method MediaWiki\MediaWikiServices::getConnectionProvider()`

getConnectionProvider() is only since 1.42

I'm not too familiar with the development of MediaWikiServices so I'm looking at it now to understand the proper backwards compatible (to 1.39) methods.

paladox:
> Checks failed because `Error: Call to undefined method MediaWiki\MediaWikiServices::getConnectionProvider()`
> 
> getConnectionProvider() is only since 1.42
> 
> I'm not too familiar with the development of MediaWikiServices so I'm looking at it now to understand the proper backwards compatible (to 1.39) methods.

Just use a version check and apply the change only on MW 1.42+ and fallback to wfGetDB in the else.

```
if ( version_compare( MW_VERSION, '1.42', '>=' ) ) {
...
} else {
...
}
```

Nikerabbit:
> > Checks failed because `Error: Call to undefined method MediaWiki\MediaWikiServices::getConnectionProvider()`
> > getConnectionProvider() is only since 1.42
> > I'm not too familiar with the development of MediaWikiServices so I'm looking at it now to understand the proper backwards compatible (to 1.39) methods.
> 
> Just use a version check and apply the change only on MW 1.42+ and fallback to wfGetDB in the else.
> 
> ```
> if ( version_compare( MW_VERSION, '1.42', '>=' ) ) {
> ...
> } else {
> ...
> }
> ```

Interface `IConnectionProvider` is present since 1.40 and before `getConnectionProvider` was introduced the correct way to get it is to use `\MediaWiki\MediaWikiServices::getDBLoadBalancerFactory()`. (Which is still instructed in https://gerrit.wikimedia.org/g/mediawiki/core/+/46c01b4712e63de5a60bf8ae9e253c39ba726c56/includes/libs/rdbms/lbfactory/IConnectionProvider.php#29 even though it's outdated now).

freephile:
On my local instance, I tried
`composer phpunit -- --filter RdfFileResourceTest`
and got 
`OK (2 tests, 3 assertions)`

(plus a lot of *unrelated* Deprecation warnings and PHP Stack traces coming from other extensions)

But this only shows that it can work on 1.43.

Should this exporter be using the SMW Services::ServicesFactory?

Should it be using the 'store' abstraction?
`$connection = $this->store->getConnection( 'mw.db' );`

Or should we try to avoid the SMW Services by directly using MediaWikiServices with something like
```php
	/**
	 * Get a handle for performing database operations.
	 *
	 * This is pretty much wfGetDB() in disguise with support for MW 1.39+
	 * _without_ triggering WMF CI warnings/errors.
	 *
	 * @see https://phabricator.wikimedia.org/T273239
	 *
	 * @return \Wikimedia\Rdbms\IDatabase|\Wikimedia\Rdbms\IReadableDatabase
	 */
	public static function getDBHandle() {
		$services = MediaWikiServices::getInstance();
		if ( method_exists( $services, 'getConnectionProvider' ) ) {
			return $services->getConnectionProvider()->getReplicaDatabase();
		} else {
			return $services->getDBLoadBalancer()->getConnection( DB_REPLICA );
		}
	}
```

freephile:
> > > Checks failed because `Error: Call to undefined method MediaWiki\MediaWikiServices::getConnectionProvider()`
> > > getConnectionProvider() is only since 1.42
> > > I'm not too familiar with the development of MediaWikiServices so I'm looking at it now to understand the proper backwards compatible (to 1.39) methods.
> > 
> > 
> > Just use a version check and apply the change only on MW 1.42+ and fallback to wfGetDB in the else.
> > ```
> > if ( version_compare( MW_VERSION, '1.42', '>=' ) ) {
> > ...
> > } else {
> > ...
> > }
> > ```
> 
> Interface `IConnectionProvider` is present since 1.40 and before `getConnectionProvider` was introduced the correct way to get it is to use `\MediaWiki\MediaWikiServices::getDBLoadBalancerFactory()`. (Which is still instructed in https://gerrit.wikimedia.org/g/mediawiki/core/+/46c01b4712e63de5a60bf8ae9e253c39ba726c56/includes/libs/rdbms/lbfactory/IConnectionProvider.php#29 even though it's outdated now).

Thanks Niklas, I didn't see your reply until now (after posting multiple options/questions). The version compare seems the simplest/straightforward approach.

paladox (approved):

jaideraf:
@freephile, thanks, it would be nice if you could create a PR to the [SemanticExtraSpecialProperties](https://github.com/SemanticMediaWiki/SemanticExtraSpecialProperties), too.

PR #5976, opened 2025-01-12
Fixes "InvalidArgumentException: No matchable hierarchy type, expected a property or category entity."

paladox:
This was due to services caching before we set / change ContentServices, and thus when using Title::newFromText it was using a stale language object deep down. Specifically in MediaWikiTitleCodec.

We fixed this by resetting TitleParser and _MediaWikiTitleCodec services.

## Summary by CodeRabbit

- **Tests**
	- Enhanced test environment setup by adding service reset mechanisms for MediaWiki services during testing.


gesinn-it-gea:
@paladox awesome! Good catch!

PR #6028, opened 2025-02-15
SearchResultSet: Remove termMatches

paladox:
We now support the minimum MW version that has that method.


paladox:
Done https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/6035 instead

PR #6035, opened 2025-02-15
SearchResultSet: Remove unnecessary check

paladox:
We support MW 1.39+

## Summary by CodeRabbit

- **Refactor**
  - Simplified the process that handles search queries to deliver a consistent experience. This update streamlines the internal logic behind processing search tokens, ensuring that users receive reliable search results with improved efficiency.

### DIFF
--- a/src/MediaWiki/Search/SearchResultSet.php
+++ b/src/MediaWiki/Search/SearchResultSet.php
@@ -177,12 +177,7 @@ class SearchResultSet extends \SearchResultSet {
 	 * @return string[]
 	 */
 	public function termMatches() {
-		$tokens = $this->getTokens()
-		if ( $tokens !== [] ) {
-			return $tokens;
-		}
-
-		return [];
+		return $this->getTokens();
 	}
 
 	private function getTokens() {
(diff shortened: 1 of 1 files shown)


src/MediaWiki/Search/SearchResultSet.php
```diff
@@ -174,24 +174,10 @@ public function getTotalHits() {
...
-		if ( ( $tokens = $this->getTokens() ) !== [] ) {
-			return $tokens;
-		}
-
-		if ( method_exists( '\SearchHighlighter', 'highlightNone' ) ) {
-			return [];
-		}
-
-		// Will cause the highlighter to match every line start, thus returning the first few lines of found pages.
-		return [ '^' ];
```
alistair3149, line 194: If I understand it correctly, when `$this->getTokens()` returns `[]`, this function should return `[ '^' ]`.

paladox: I'm not sure because you then have the if method exist check above which will return true in MW 1.39+.

alistair3149: In that case should all calls to this function just point to `getTokens` directly?

paladox: I'm not sure what you mean? This function is extending the MW class one. Removing this function causes the tests to fail and I'm not sure how to get it to not use this one.

alistair3149: I missed that it is extending the MW class so you're right!

PR #6045, opened 2025-02-18
CustomForm: Fix "Trying to access array offset on value of type null "

### LINKED ISSUE #5644: Notice: Trying to access array offset on value of type null in CustomForm.php on line 131
### Setup

- MediaWiki | 1.39.7 (64425c7)16:12, 15 May 2024
- PHP | 7.4.33 (apache2handler)
- MariaDB | 10.5.23-MariaDB-0+deb11u1
- Elasticsearch | 7.10.2
- [Semantic MediaWiki](https://www.semantic-mediawiki.org/)	4.1.3

### Issue
Happening on https://sandbox.semantic-mediawiki.net/

When trying to search for something on "Special:Search" I get:

```
 Notice: Trying to access array offset on value of type null in /.../w/extensions/SemanticMediaWiki/src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php on line 131
```

paladox:
This happens because ->getArray() can return null if it can't find the key. We fix this by instead changing the return default for getArray to an array. We then change the isset check to use new php syntax and also return a null, instead of using $vals[0]. We already check for that so this was a duplicate and was error prone.

Fixes #5644

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved form input handling to ensure that when no values are found, an empty array is returned, enhancing robustness in form processing.
	- Simplified logic for value assignment in forms, improving clarity and reducing potential errors in data handling.

### DIFF
--- a/src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
+++ b/src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
@@ -66,7 +66,7 @@ class CustomForm {
 	/**
 	 * @since 3.0
 	 *
-	 * @return
+	 * @return array
 	 */
 	public function getParameters() {
 		return $this->parameters;
@@ -122,10 +122,10 @@ class CustomForm {
 
 			// Find request related value for the active form
 			if ( $this->isActiveForm ) {
-				$vals = $this->request->getArray( $name );
+				$vals = $this->request->getArray( $name, [] );
 
 				$i = $this->fieldCounter[$name];
-				$value = isset( $vals[$i] ) ? $vals[$i] : $vals;
+				$value = $vals[$i] ?? null;
 				$this->parameters[$name] = $value;
 			}
 
--- a/src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
+++ b/src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
@@ -125,7 +125,7 @@ class CustomForm {
 				$vals = $this->request->getArray( $name );
 
 				$i = $this->fieldCounter[$name];
-				$value = isset( $vals[$i] ) ? $vals[$i] : $vals[0];
+				$value = isset( $vals[$i] ) ? $vals[$i] : $vals;
 				$this->parameters[$name] = $value;
 			}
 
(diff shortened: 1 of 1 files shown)


src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
```diff
@@ -122,10 +122,10 @@ public function makeFields( $definition ) {
 
 			// Find request related value for the active form
 			if ( $this->isActiveForm ) {
-				$vals = $this->request->getArray( $name );
+				$vals = $this->request->getArray( $name, [] );
 
 				$i = $this->fieldCounter[$name];
-				$value = isset( $vals[$i] ) ? $vals[$i] : $vals[0];
+				$value = $vals[$i] ?? null;
```
paladox, line 128: I'm not sure if this should be '' instead of null.

src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
```diff
@@ -66,7 +66,7 @@ public function __construct( WebRequest $request ) {
 	/**
 	 * @since 3.0
 	 *
-	 * @return
+	 * @return array
 	 */
 	public function getParameters() {
```
alistair3149, line 71: We can put a type declaration here and drop the `@return` doc

paladox: do you mean like `: array {`?

src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
alistair3149: Should be just `''` because it was the previous value on line 106

src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
paladox: Thanks, done.

src/MediaWiki/Search/ProfileForm/Forms/CustomForm.php
paladox: Thanks, done.

Issue #5872, opened 2024-12-19
Use of MediaWiki\Parser\Parser::getOutput before initialization was deprecated in MediaWiki 1.42


paladox:
> Deprecated: Use of MediaWiki\Parser\Parser::getOutput before initialization was deprecated in MediaWiki 1.42. [Called from SMW\Parser\RecursiveTextProcessor::getParserOutputSafe in /var/www/html/extensions/SemanticResultFormats/extensions/SemanticMediaWiki/src/Parser/RecursiveTextProcessor.php at line 298] in /var/www/html/includes/debug/MWDebug.php on line 379

paladox:
> So, this issue should belong to the SemanticResultFormats repository, right?

I don’t think so? Not sure tho, it’s coming from SMW.

jaideraf:
Oh, this extract mislead me:

`/var/www/html/extensions/SemanticResultFormats/`

YOUR1:
I think this should be a part of the 5.0.0 milestone, since it's visible for every user.

YOUR1:
This notice also displayes when the user searches for something on Special:FacettedSearch. 

```
Deprecated: Use of MediaWiki\Parser\Parser::getOutput before initialization was deprecated in MediaWiki 1.42. [Called from SMW\Parser\RecursiveTextProcessor::getParserOutputSafe in /var/platform/mediawiki/1.43/core/20250127133002/extensions/SemanticMediaWiki/src/Parser/RecursiveTextProcessor.php at line 298] in /var/platform/mediawiki/1.43/core/20250127133002/includes/debug/MWDebug.php on line 385
```

jaideraf:
I couldn't reproduce here: https://wikincat.org/w/index.php?title=Especial:FacetedSearch&q=%5B%5BCategory%3AObra%5D%5D&profile=work

Sophivorus:
This deprecation notice is currently our top log spam

YOUR1:
> I couldn't reproduce here: https://wikincat.org/w/index.php?title=Especial:FacetedSearch&q=%5B%5BCategory%3AObra%5D%5D&profile=work

Does it have error reporting on?

> This deprecation notice is currently our top log spam, so we would appreciate a fix before the 5.0 milestone.
> 
> ![Image](https://github.com/user-attachments/assets/c08092ba-b402-4234-8ddb-1198d912ab6b)

Same here.

YOUR1:
[...]

PR #6047, opened 2025-02-19
RecursiveTextProcessor: getOutput before initialization was deprecated in MediaWiki 1.42

YOUR1:
Fixed https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5872

### DIFF
--- a/src/Parser/RecursiveTextProcessor.php
+++ b/src/Parser/RecursiveTextProcessor.php
@@ -295,6 +295,11 @@ class RecursiveTextProcessor {
 
 	private function getParserOutputSafe(): ?ParserOutput {
 		try {
+			// The getOutput() method is deprecated in MW 1.42 before initializing
+			// So we need to check if the parser was initialized.
+			if ( version_compare( MW_VERSION, '1.42', '>=' ) && $this->parser->getOptions() === null ) {
+				$this->parser->resetOutput();
+			}
 			return $this->parser->getOutput();
 		} catch ( Error $e ) {
 			return null;


YOUR1:
I think the errors of the pipelines can be ignored, since unit tests are not working at all -- and the codecov is a false positive 😂

src/Parser/RecursiveTextProcessor.php
```diff
@@ -295,6 +295,11 @@ private function pruneCategory( &$text ) {
 
 	private function getParserOutputSafe(): ?ParserOutput {
 		try {
+			// The getOutput() method is deprecated in MW 1.42 before initializing
+			// So we need to check if the parser was initialized.
+			if ( version_compare( MW_VERSION, '1.42', '>=' ) && $this->parser->getOptions() === null ) {
+				$this->parser->resetOutput();
```
paladox, line 301: Does this work for you? When I tried this for another extension, investigating a problem, it failed on $this->mOptions I think? As it wasn't defined.

paladox: Oh we’re checking if getOptions is null and if it is we are calling resetOutput. That will fail because of https://github.com/wikimedia/mediawiki/blob/b3f3e47f61ad544241319cb292b27e8acfcec13c/includes/parser/Parser.php#L648.

this should be reverted I think?

paladox: @alistair3149

alistair3149: That should be caught by the try block and return null. Is there an alternative way to initialize parser output?

paladox: I’m not aware of any alternative.

YOUR1: I don't understand why this is reverted, the fix was nested in a try/catch, so the original functionality (what it did before) stayed the same. 

The class is called with a parser 99th of the time, and if it doesn't, the parser is initialized with `MediaWikiServices::getInstance()->getParser()` which is causing the notice to show up (hence the resetOutput fix).

YOUR1: For future reference, could we please wait a bit before reverting changes? I’d appreciate a chance to respond or discuss any concerns first. Thank you for understanding.

paladox: You were checking if getOptions is null and if it was, calling resetOutput . Which would fail on this https://github.com/wikimedia/mediawiki/blob/b3f3e47f61ad544241319cb292b27e8acfcec13c/includes/parser/Parser.php#L648 line?

I realise now that the catch would make sure it wouldn’t fatal so sorry.

YOUR1: Yeah your right about that. The fix is slightly different indeed, the function should return false if the options we're not set, since its an invalid state. I'll update the code (and comments) accordingly.

PR #6056, opened 2025-02-21
SMWIntegrationTestCase: Don't use getTestUser

paladox:
We don't use the prefix that it's expecting in [0]. We change the prefix with https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/20b77db03cf37abc1b3fb59cc2adeabf60dd0190/tests/phpunit/TestEnvironment.php#L158.

[0] https://github.com/wikimedia/mediawiki/blob/master/tests/phpunit/includes/TestUser.php#L30

## Summary by CodeRabbit

- **Tests**
  - Enhanced the test setup by refining how the primary user context is retrieved during execution, ensuring more consistent and reliable test outcomes.


paladox:
But we also have:

```
4) SMW\Scribunto\Integration\JSONScript\SemanticScribuntoJsonTestCaseScriptRunnerTest::testCaseFile with data set "ask-002.json" ('/var/www/html/extensions/Sema...2.json')
RuntimeException: Tracking is already enabled

/var/www/html/includes/libs/rdbms/ChangedTablesTracker.php:29
/var/www/html/tests/phpunit/MediaWikiIntegrationTestCase.php:694
/var/www/html/tests/phpunit/MediaWikiIntegrationTestCase.php:624
/var/www/html/extensions/SemanticMediaWiki/tests/phpunit/SMWIntegrationTestCase.php:203
```

hmm

paladox:
@gesinn-it-gea / @gesinn-it-ilm something for you to look into? Since SMWIntegrationTestCase was added that uses MediaWikiIntegrationTestCase.php. I don't know how you would fix this or whether the fact we change db prefixes is what breaks this.

PR #6065, opened 2025-02-27
Clean JsonSchemaValidator

JeroenDeDauw:
Replaces usage of deprecated `check` method

Adds installability with justinrainbow/json-schema 6.x, so closes https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/6064

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
 		"onoi/callback-container": "~2.0",
 		"symfony/css-selector": "^5|^4|^3.3",
 		"seld/jsonlint": "^1.7",
-		"justinrainbow/json-schema": "~5.2",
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
> PHPUnit\Framework\MockObject\IncompatibleReturnValueException: Method validate may not return value of type boolean, its declared return type is "void"

alistair3149:
> PHPUnit\Framework\MockObject\IncompatibleReturnValueException: Method validate may not return value of type boolean, its declared return type is "void"

It is caused by the tests here: https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/tests/phpunit/Schema/SchemaValidatorTest.php

JeroenDeDauw:
If someone wants to continue with this PR, feel free

PR #6133, opened 2025-05-15
Sanitize query param in HtmlBuilder.php

YOUR1:
Should be backported in SMW4.x as well. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of special characters in search queries to prevent display issues in the search form.

### DIFF
--- a/src/MediaWiki/Specials/FacetedSearch/HtmlBuilder.php
+++ b/src/MediaWiki/Specials/FacetedSearch/HtmlBuilder.php
@@ -186,7 +186,7 @@ class HtmlBuilder {
 			[
 				'action' => $title->getLocalUrl(),
 				'method' => 'get',
-				'q' => $urlArgs->get( 'q', '' ),
+				'q' => htmlspecialchars( $urlArgs->get( 'q', '' ) ),
 				'csum' => crc32( $urlArgs->get( 'q', '' ) ),
 				'limit' => $params['limit'],
 				'offset' => $params['offset'],


src/MediaWiki/Specials/FacetedSearch/HtmlBuilder.php
```diff
@@ -186,7 +186,7 @@ public function buildHTML( Title $title, UrlArgs $urlArgs ): string {
 			[
 				'action' => $title->getLocalUrl(),
 				'method' => 'get',
-				'q' => $urlArgs->get( 'q', '' ),
+				'q' => htmlspecialchars( $urlArgs->get( 'q', '' ) ),
```
JeroenDeDauw, line 189: I'll merge this as a hotfix, but after some inspection of the code, there seem to be further holes in the cheese, even if they do not go all the way through.

General best practice is to escape at the place of output. That is in the template, not here. I see other parameters are also passed to the template without escaping. Probably we are just lucky these are not user-controlled, get cast, or otherwise are made "safe" somehow.

Just a few lines above I see 

```php
$hidden .= '<input name="' . "cstate[$key]" . '" type="hidden" value="' . $value . '">';
```

no escaping either :(

Btw, the mapping in `$templateEngine->bulkLoad()` in `SpecialFAcetedSearch` makes it difficult to find template code.

JeroenDeDauw: I've since realized that q is already escaped in the template, but not sufficiently so, due to the context. Using `htmlspecialchars` here seems like a good fix.

YOUR1: When we make use of the 'TemplateParser' of MediaWiki, Variables retrieved like `{{{var}}}` will NEVER be sanitized. When using `{{var}}` it will. James has a different way of loading the templates tho, so it requires a bit refactoring.

JeroenDeDauw: The triple-brace syntax is not used in the template, but because you added escaping, I first incorrectly assumed nothing was being escaped in the template. Still, various improvements can be made in this PHP file wrt escaping.

PR #6138, opened 2025-05-27
Use file backend for ingest

marijnvanwezel:
Previously, `file_get_contents` was used to fetch the binary content of a file in the file ingest pipeline. This had several problems: (1) it doesn't work when `img_auth.php` is used (2) it doesn't work with certain wikis where SSL is not configured (3) it doesn't work for certain settings of `$wgServer` (#4918).

This pull request fixes that by using MediaWiki functions to fetch the file content for us, which handle these scenarios appropriately.

### DIFF
--- a/src/Elastic/Indexer/Attachment/FileHandler.php
+++ b/src/Elastic/Indexer/Attachment/FileHandler.php
@@ -2,9 +2,9 @@
 
 namespace SMW\Elastic\Indexer\Attachment;
 
-use MediaWiki\Title\Title;
 use File;
 use FileBackend;
+use MediaWiki\Title\Title;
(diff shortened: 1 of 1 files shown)


marijnvanwezel:
This fixes an issue we had with file ingestion for a long time. It would be awesome if this could be merged before the release of 6.0.0! Let me know if anything needs changing.

marijnvanwezel:
@alistair3149 Done.

alistair3149:
Thanks, the change looks good to me. Would you be able to sort the class import please?
https://github.com/SemanticMediaWiki/SemanticMediaWiki/actions/runs/16751486208/job/47422642185?pr=6138#step:4:1586

PR #6143, opened 2025-06-10
Preserve full title in SMW\Utils\Pager::filter() form

### LINKED ISSUE #6142: Filtering form on property pages truncates title at slash
### Setup

- SMW version: master
- MW version: 1.43
- PHP version: 8.4
- DB system (MySQL, Blazegraph, etc.) and version: MySQL

### Issue

On property pages with slashes in the title (e.g. `Property:Task/Desc`), the filter form incorrectly sets `title=Property:Task` due to the use of `strtok( ..., '/' )` in `SMW\Utils\Pager::filter()`.

Steps to reproduce:

Create a property with a title containing a slash (e.g., `Property:Task/Desc`).
Visit the property's page.
Observe that the filter form’s title input has the value `Property:Task`, not `Property:Task/Desc`.

HamishSlater:
Avoid truncating titles at slash (e.g. Property:Task/Desc → Property:Task) by removing strtok()

Fixes: #6142


src/Utils/Pager.php
```diff
@@ -53,7 +53,7 @@ public static function filter( Title $title, $limit = 0, $offset = 0, $filter =
 			],
 			Html::hidden(
 			'title',
-			strtok( $title->getPrefixedText(), '/' )
+			$title->getPrefixedText()
```
JeroenDeDauw, line 66: Presumably the `strtok` was added for some reason, so it seems somewhat likely some other desired behavior is broken now.

HamishSlater: Perhaps @mwjames can shed some light on this.

JeroenDeDauw: They are no longer around

alistair3149: Looked into this. You're right that the `strtok` was there for a reason, but it was the wrong reason for this form.

The idiom originated in `HtmlFormBuilder` back in 2014 (027fc993d), where the form title is a Special page title such as `Special:SearchByProperty/Prop/Value`. There the slash-delimited tail is a subpage-encoded parameter, and stripping to the base special page (so the reload re-posts cleanly with the params in hidden fields) is correct.

It was then copied verbatim into the property/concept filter form in #2785 (716e88f19), where the title is a content page and a slash is part of the actual page name. SMW resolves the entity via `DIProperty::newFromUserLabel( $title->getText() )`, so `Property:Task/Desc` is a distinct property literally named `Task/Desc`, not a subpage of `Task`. The Property and Concept namespaces also have subpages disabled. So on this form the truncation just points the reload at the wrong page, which is exactly #6142.

Removing it here is correct and regression-free: on slash-free titles `strtok` is a no-op, and `Html::hidden()` escapes the value so a slash in the GET query is fine. The near-identical call in `HtmlFormRenderer.php` (line 394) is context-appropriate and should be left as is, since its callers are Special pages that genuinely carry subpage args.

alistair3149:
Thanks for the fix! I opened #7013 to add a regression test and changlog entry, since the PR is created from a forked `master` branch.

Issue #6144, opened 2025-06-10
feat (api/smwbrowse): add support for inverse properties


gesinn-it-gea:
In addition to the regular properties, `smwbrowse` (subject) should also return 'inverse properties'. The change should not break existing API implementations. The change should also include tests.

Current data structure:

```
{
    "query": {
        "subject": "Project_A1#0##",
        "data": [
            {
                "property": "BelongsToProjectArea",
                "dataitem": [
                    {
                        "type": 9,
                        "item": "A#0##"
                    }
                ]
            },
...
```

Proposal:
In addition to "property" add "inverse property" to the `data` array.

PR #6145, opened 2025-06-11
Issue 6144 - add support for inverse properties in smwbrowse (subject)

gesinn-it-ilm:
This PR is related to the issue #6144.

This PR contains:

- Update the API serialization logic to support inverse (incoming) properties
- Add serialization of inverse properties in `SemanticDataSerializer.php`
- Add new key-value par in $data object in serialization, key `direction` with values `direct` or `inverse`
- Maintained backward compatibility with existing direct property logic
- Added new JSONScript test cases to cover inverse property serialization

### Checklist:
- [x] Code compiles and passes all tests
- [x] Inverse properties appear correctly in serialized output
- [x] JSONScript tests cover direct and inverse cases
- [x] No breaking changes to existing API consumers


src/MediaWiki/Api/Browse/SubjectLookup.php
```diff
@@ -129,13 +132,64 @@ private function doSerialize( $params ) {
...
+			// Transform the entries so that the output clearly distinguishes inverse properties
+			foreach ( $incomingDataSerialized['data'] as $entry ) {
+				if ( isset( $entry['property'] ) ) {
+					$incomingData[] = [
+						'inverse property' => $entry['property'],
+						'dataitem' => $entry['dataitem'] ?? [],
+					];
+				}
+			}
+		}
```
JeroenDeDauw: Please compute $incomingData in a dedicated function

src/MediaWiki/Api/Browse/SubjectLookup.php
```diff
@@ -129,13 +132,64 @@ private function doSerialize( $params ) {
...
+			// Serialize the inverse SemanticData separately to extract its entries
+			$incomingDataSerialized = $serializer->serialize( $semanticDataIncoming );
+
+			// Transform the entries so that the output clearly distinguishes inverse properties
+			foreach ( $incomingDataSerialized['data'] as $entry ) {
+				if ( isset( $entry['property'] ) ) {
+					$incomingData[] = [
+						'inverse property' => $entry['property'],
+						'dataitem' => $entry['dataitem'] ?? [],
+					];
```
JeroenDeDauw: So now the API result will contain entries without a `property` key? That can break existing consumers.

src/MediaWiki/Api/Browse/SubjectLookup.php
```diff
@@ -129,13 +132,64 @@ private function doSerialize( $params ) {
...
+		}
+
+		// Merge the direct properties into the main SemanticData object for standard serialization
+		$semanticData->importDataFrom( $semanticDataDirect );
+
+		// Serialize the full direct data
+		$semanticData = $serializer->serialize( $semanticData );
+
+		// Append the previously processed inverse properties with custom "inverse property" label
+		$semanticData['data'] = array_merge( $semanticData['data'], $incomingData );
```
JeroenDeDauw: It's awkward design to have some of the serialization happen in SemanticDataSerializer and some in this SubjectLookup.

src/MediaWiki/Api/Browse/SubjectLookup.php
gesinn-it-ilm: No, it will still contain the result with `property` key. Only for the inverse property, the key will be like `inverse property`.
So, if subject for example Page A has both direct and inverse props, the API result will have them inside `data` as `property` key for direct and `inverse property` for inverse. In this way we can tell the difference between what is direct and what inverse or incoming property for the given subject.

src/MediaWiki/Api/Browse/SubjectLookup.php
gesinn-it-ilm: @JeroenDeDauw should we then move that to SemanticDataSerializer as well?

src/MediaWiki/Api/Browse/SubjectLookup.php
JeroenDeDauw: So if you have a script that loops through the results and assumes the property key is there, it will blow up after your change is merged, right?

gesinn-it-ilm: Yes, that's technically correct if a script assumes that every data entry has a `property` key without checking for it.
However, subjects with no properties at all already led to an empty data array, or array containing no 'property' key.
What changed now is that we return `inverse property` entries instead of omitting them completely. This is an extension of what data is available.

To be defensive, we can:
- use a unified key like `property` and add a `direction` key (direct or inverse)

gesinn-it-ilm: This is an example JSON response for example subject which has both properties and inverse properties. 
You can check here how the response looks like:

[example.txt](https://github.com/user-attachments/files/20704080/example.txt)

JeroenDeDauw: > However, subjects with no properties at all already led to an empty data array, or array containing no 'property' key.

That does not matter. If the array is empty, iteration goes fine. The issue is with removing an existing key from the objects that are the elements in this array.

JeroenDeDauw: > use a unified key like property and add a direction key (direct or inverse)

That prevents existing scripts from blowing up. But it will still change their behavior by including inverse properties, which might not be what existing consumers want.

gesinn-it-gea: Okay, let's add a 'direction' key and 'property' as before. I'm not sure if this would have the same impact as before, as consumers would now get additional properties that they would treat as direct properties unless they look at the 'direction'. I personally would have implemented it the same way as @gesinn-it-ilm did.

src/MediaWiki/Api/Browse/SubjectLookup.php
gesinn-it-ilm: This has been removed because SubjectLookup.php has been updated and serialization moved to SemanticDataSerializer.

src/MediaWiki/Api/Browse/SubjectLookup.php
gesinn-it-ilm: Complete serialization has been moved to `SemanticDataSerializer.php`.

src/MediaWiki/Api/Browse/SubjectLookup.php
gesinn-it-ilm: @JeroenDeDauw direction flag has been added, key `direction` with values `direct` or `inverse`, depends of property. 
Key `property` is as before. 
Serialization has been moved to `SemanticDataSerializer.php`.

```
"data": [
          {
              "property": "Links_To",
              "dataitem": [
                  {
                      "type": 9,
                      "item": "Root_Link#0##"
                  }
              ],
              "direction": "direct"
          },
          {
              "property": "Links_To",
              "dataitem": [
                  {
                      "type": 9,
                      "item": "Root_Link_1#0##"
                  }
              ],
              "direction": "inverse"
          }
      ]
```

src/Serializers/SemanticDataSerializer.php
```diff
@@ -19,12 +20,22 @@
...
 	 */
-	public function serialize( $semanticData ) {
+	public function serialize( $semanticData, $includeInverse = false ) {
 		if ( !$semanticData instanceof SemanticData ) {
 			throw new OutOfBoundsException( 'Object is not supported' );
 		}
 
-		return $this->doSerialize( $semanticData ) + [ 'serializer' => __CLASS__, 'version' => 2 ];
+		$data = $this->doSerialize( $semanticData, $includeInverse );
+
```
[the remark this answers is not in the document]
gesinn-it-ilm: Unnecessary parameter has been removed from the method call.

kghbln:
It will be cool to get the fix for https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/6097 in 5.1.0

Issue #6150, opened 2025-06-26
Issue with DB queries from SMW: RuntimeException: Identifier must not contain quote, dot or null characters


it-spiderman:
### Setup

- SMW version: 5.0.2
- MW version: 1.43.1
- PHP version: 8.4.5

### Issue

In many places, SMW calls MW DBAL with table names already processed by  $db->tableName(). This will quote the table name. When this quoted name is passed to the platform, it will try to quote it again. It fails when it tries to compare aliases (instead of simple $table === $alias as it was before, now it will call table name on alias again, which unquotes it)

Example exception this causes:

wiki/Special:Statistics RuntimeException: Identifier must not contain quote, dot or null characters: got '`smw_object_ids`'
#0 /app/bluespice/w/includes/libs/rdbms/platform/SQLPlatform.php(1044): Wikimedia\Rdbms\Platform\SQLPlatform->addIdentifierQuotes()
#1 /app/bluespice/w/includes/libs/rdbms/platform/SQLPlatform.php(924): Wikimedia\Rdbms\Platform\SQLPlatform->tableNameWithAlias()
#2 /app/bluespice/w/includes/libs/rdbms/platform/SQLPlatform.php(711): Wikimedia\Rdbms\Platform\SQLPlatform->tableNamesWithIndexClauseOrJOIN()
#3 /app/bluespice/w/includes/libs/rdbms/database/Database.php(3305): Wikimedia\Rdbms\Platform\SQLPlatform->selectSQLText()
#4 /app/bluespice/w/includes/libs/rdbms/database/DatabaseMySQL.php(644): Wikimedia\Rdbms\Database->selectSQLText()
#5 /app/bluespice/w/includes/libs/rdbms/database/Database.php(1337): Wikimedia\Rdbms\DatabaseMySQL->selectSQLText()
#6 /app/bluespice/w/includes/libs/rdbms/database/DBConnRef.php(127): Wikimedia\Rdbms\Database->select()
#7 /app/bluespice/w/includes/libs/rdbms/database/DBConnRef.php(351): Wikimedia\Rdbms\DBConnRef->__call()
#8 /app/bluespice/w/extensions/SemanticMediaWiki/src/MediaWiki/Connection/Database.php(250): Wikimedia\Rdbms\DBConnRef->select()
[...]

PR #6151, opened 2025-06-26
Remove redundant call of $db->tableName()

huaj1ng:
as MW DBAL resolves names already

#6150

### DIFF
--- a/includes/export/SMW_ExportController.php
+++ b/includes/export/SMW_ExportController.php
@@ -589,8 +589,7 @@ class SMWExportController {
 				$query .= 'page_namespace = ' . $dbr->addQuotes( $ns );
 			}
 		}
-		$res = $dbr->select( $dbr->tableName( 'page' ),
-							'page_id,page_title,page_namespace', $query,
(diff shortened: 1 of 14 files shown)


JeroenDeDauw:
> There was 1 error:

> 1) SMW\Tests\SQLStore\Lookup\DisplayTitleLookupTest::testPrefetchFromList
> foreach() argument must be of type array|object, null given

> /var/www/html/extensions/SemanticMediaWiki/src/SQLStore/Lookup/DisplayTitleLookup.php:82
> /var/www/html/extensions/SemanticMediaWiki/tests/phpunit/SQLStore/Lookup/DisplayTitleLookupTest.php:125

JeroenDeDauw:
Change itself looks good, so perhaps it's a poorly written test

huaj1ng:
thank you for the review, though I realized that there might be some confusion here:
apparently I was confusing a method "->tableName()" with another "->tablename()", where only one of these cause the problem. I will need to investigate closer to know whether the original way of test makes sense.

PR #6154, opened 2025-07-05
Ukrainian translation for extras

METROKOP228:


METROKOP228:
Why is this pull request not getting merged? Is there any problems or should I just wait more?

PR #6160, opened 2025-07-22
Add experimental CI for MediaWiki 1.44

paladox:


alistair3149:
On a somewhat related matter, we should probably test for PHP 8.3, 8.4, and MariaDB 11.8 (current LTS).
Perhaps we can cover more ground by having MW 1.43 uses PHP 8.3, and having MW 1.44 uses PHP 8.4 and MariaDB 11.8.
We already have image for PHP 8.3 [upstream](https://hub.docker.com/r/gesinn/mediawiki-ci), just needing the 8.4 image.

@gesinn-it-ilm what do you think?

PR #6166, opened 2025-07-24
Bump minimum MW requirement to MW 1.43+

paladox:
@alistair3149 has told me the minimum is 1.43 now or to target that. So lets bump the requirements.


paladox:
Actually for now, we need CI to work on MW 1.43 I think? As code coverage is run on MW 1.40.

PR #6170, opened 2025-07-24
Require MediaWiki 1.43

alistair3149:
Bump SMW to 6.0.0-beta and raise minimum MediaWiki to 1.43

With the breaking changes in MW class imports and database access methods,
it is more maintainable to bump the requirements to 1.43.

5.1.0 release has been tagged for maintaining 1.39 support.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,46 +16,22 @@ jobs:
     strategy:
       matrix:
         include:
-          - mediawiki_version: '1.39'
-            php_version: 8.1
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -16,42 +16,12 @@ jobs:
     strategy:
```
paladox, line 16: May want to run MW 1.43 twice, once with code coverage on and it off (so normal tests run).

paladox (approved):

PR #6177, opened 2025-07-26
Experiment: DO NOT MERGE

paladox:


tests/phpunit/SMWIntegrationTestCase.php
```diff
@@ -161,6 +162,7 @@ protected function tearDown(): void {
 
 	public function run( ?TestResult $result = null ): TestResult {
```
paladox, line 163: @coderabbitai can you update this for MediaWiki 1.43? Specifically all the changes since MediaWiki 1.41? See https://github.com/wikimedia/mediawiki/commit/406b844a427aa2306bf715dea19645b9c6012c34

paladox: can you show me the changes

paladox: But we need the stuff done in the run(). We need to remove run but the stuff in there needs to be moved to elsewhere.  See https://github.com/wikimedia/mediawiki/commit/406b844a427aa2306bf715dea19645b9c6012c34

Issue #6201, opened 2025-08-13
Fix PHP Error "TypeError: array_slice(): Argument #3 ($length) must be of type ?int, string given"


kghbln:
### Setup
- [MediaWiki](https://www.mediawiki.org/)	[1.43.3](https://www.mediawiki.org/wiki/MediaWiki_1.43) [(6748784)](https://gerrit.wikimedia.org/g/mediawiki/core.git/+/67487848a3e40429f6f5e06549a6562f08cfdf68) 12:08, 12 August 2025
- [PHP](https://php.net/)	8.2.28 (apache2handler)
- [ICU](https://icu.unicode.org/)	72.1
- [MariaDB](https://mariadb.org/)	10.11.11-MariaDB-0+deb12u1
- [Elasticsearch](https://www.elastic.co/elasticsearch)	7.10.2
- [Semantic MediaWiki](https://www.semantic-mediawiki.org/)	6.0.0

### Issue

When running rebuildData.php I get the following error:

```
   ... updating 2570      [T: Catégorie:C0615]
[fb828b0a3b51789d24eee832] [no req]   TypeError: array_slice(): Argument #3 ($length) must be of type ?int, string given
```

### Backtrace

```
from /../w/extensions/SemanticMediaWiki/src/Elastic/QueryEngine/ConditionBuilder.php(379)
#0 /../w/extensions/SemanticMediaWiki/src/Elastic/QueryEngine/ConditionBuilder.php(379): array_slice()
#1 /../w/extensions/SemanticMediaWiki/src/Elastic/QueryEngine/DescriptionInterpreters/ClassDescriptionInterpreter.php(63): SMW\Elastic\QueryEngine\ConditionBuilder->findHierarchyMembers()
#2 /../w/extensions/SemanticMediaWiki/src/Elastic/QueryEngine/ConditionBuilder.php(415): SMW\Elastic\QueryEngine\DescriptionInterpreters\ClassDescriptionInterpreter->interpretDescription()
#3 /../w/extensions/SemanticMediaWiki/src/Elastic/QueryEngine/ConditionBuilder.php(336): SMW\Elastic\QueryEngine\ConditionBuilder->interpretDescription()
#4 /../w/extensions/SemanticMediaWiki/src/Elastic/QueryEngine/QueryEngine.php(143): SMW\Elastic\QueryEngine\ConditionBuilder->makeFromDescription()
[...]

PR #6213, opened 2025-08-19
Fix cast for $depth in LegacyParser->finishLinkDescription

paladox:
setHierarchyDepth needs to be a integer not string.


paladox:
I'm not sure if this fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/6201. But this is a needed fix regardless.

PR #6231, opened 2025-08-31
Create a extended DateTime class to add property

paladox:
We work around php not allowing dynamic properties anymore unless class allows it specifically with `#[\AllowDynamicProperties]`.


paladox:
Bump :)

PR #6236, opened 2025-09-04
Remove SearchEngine::replacePrefixes

hartman:
This has been deprecated since MW 1.32 and is not actually doing anything any longer


hartman:
https://gerrit.wikimedia.org/r/c/mediawiki/core/+/1184916

JeroenDeDauw:
This change makes the tests fail

Issue #6223, opened 2025-08-26
Named args and template do not work under certain conditions


manuel-hegner:
### Setup

- SMW version: 6.0.1
- MW version: 1.43.3
- PHP version:8.1.33 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: MariaDB

### Issue
```
$smwgParserFeatures = SMW_PARSER_STRICT | SMW_PARSER_INL_ERROR | SMW_PARSER_HID_CATS | SMW_PARSER_UNSTRIP | SMW_PARSER_LINV;
```

I have a a text property called `Description` storing this:
```
a[[Cyclops|cyclopean]]b<mytag>test test</mytag>c
```
Now I use an `#ask` with `named args` and a `template`

I would expect the template to be called with a parameter called with the content of the property. This is what happens if I remove either the tag or the link from the content.
Instead the template is called with `Description=a[[Cyclops` and `1=cyclopean]]b<mytag>test test</mytag>c`

alex-mashin:
Could not reproduce; but I use MW 1.39.

What happens, if you insert the following in `extensions/SemanticMediaWiki/src/MediaWiki/Renderer/WikitextTemplateRenderer.php`, in `WikitextTemplateRenderer::packFieldsForTemplate()`, immediately after the `foreach ( $this->fields as $key => $value ) {` line?

```php
$value = str_replace( '|', '{{!}}', $value);
```

manuel-hegner:
I am using `{{#set:`.
Tried everything with `SMW_PARSER_LINV` disabled, but it behaves the same.

I think the problem is in https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/8afcbae2cadd42831110a6515e574d48fcd462dc/src/MediaWiki/StripMarkerDecoder.php#L90

This class checks for strip markers and only if it finds any it runs its decoding. This explains why the problem only happens if a tag is actually present.

The method unstrip replaces some characters, like `[`, which then causes the pipe to be parsed in the wrong way. 

Removing the `str_replace` in this methods fixes my problem, as does adding `|` and `{{!}}` in this location. I am not sure what those replacements are meant to do.

manuel-hegner:
Looking more into the unstripping I really don't get how it is supposed to work. I think the issue is that if the code would just simply unstrip the wikitext then any `<nowiki>` content would be parsed again as wikitext when the content is queried.

[...]

PR #6239, opened 2025-09-09
Fix for values with tags rendered by templates with named parameters

manuel-hegner:
A fix for values with tags rendered by templates with named parameters in `{{#ask:}}` under specific circumstances.

This fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/6223


manuel-hegner:
I am not very familiar with testing in php and would be happy for some help with this.

PR #6246, opened 2025-09-27
issue-5421 SMW should not add dynamic property to MediaWiki's WikiFilePage

emwiemaikel:
dynamic property was only used in one place (PageInfoProvider). moved it from File to the InfoProvider

### DIFF
--- a/src/MediaWiki/MwCollaboratorFactory.php
+++ b/src/MediaWiki/MwCollaboratorFactory.php
@@ -168,9 +168,10 @@ class MwCollaboratorFactory {
 	public function newPageInfoProvider(
 		WikiPage $wikiPage,
 		?RevisionRecord $revision = null,
-		?User $user = null
+		?User $user = null,
+		?bool $fileReUploadStatus = true
 	) {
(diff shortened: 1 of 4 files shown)


src/MediaWiki/MwCollaboratorFactory.php
```diff
@@ -168,9 +168,10 @@ public function newConnectionProvider( $provider = null ) {
 	public function newPageInfoProvider(
 		WikiPage $wikiPage,
 		?RevisionRecord $revision = null,
-		?User $user = null
+		?User $user = null,
+		?bool $fileReUploadStatus = true
```
JeroenDeDauw, line 172: This new argument seems to be unused. 

Any reason to not just

```php
$pageInfoProvider = new PageInfoProvider( $wikiPage, $revision, $user, true );
```

paladox:
This was done in https://github.com/SemanticMediaWiki/SemanticMediaWiki/commit/d0cc4298fc40f493bafca57f58aa91e62be1730e

Issue #5790, opened 2024-11-26
SMW should use Mustache instead of TemplateEngine


alistair3149:
Mustache have been available in MediaWiki core since MW 1.25: https://www.mediawiki.org/wiki/Manual:HTML_templates

In SMW, the usage of the [`TemplateEngine`](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/2ca976f3bc2d58c64351d863611e3d2c98177d09/src/Utils/TemplateEngine.php#L14) class is almost identical to the Mustache `TemplateParser` class in core. We should replace them with Mustache because:
- Improve maintainability by dropping the `TemplateEngine` methods, no need to maintain a separate HTML template implementation
- Improve performance since TemplateParser templates are cached automatically

gesinn-it-gea:
we did this for SD already: https://github.com/SemanticMediaWiki/SemanticDrilldown/tree/master/includes/Specials/BrowseData/templates

PR #6257, opened 2025-10-30
Use MW TemplateParser instead of TemplateEngine for FacetedSearch

alistair3149:
#5790

This is the minimal amount of changes required to switch from TemplateEngine to TemplateParser.

Did some manual testing on the `Special:FacetedSearch` page, but it will be great if someone who are more familiar with the UI to take a second look to see if there are any regression.


YOUR1:
I'll check it on monday.

jaideraf:
I have tested and it seems ok to me. Although, I have found a bug not related to this change (https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/6260)

jaideraf (approved):

PR #6287, opened 2026-01-08
remove bullet-point styling from FacetedSearch list

D-Groenewegen:
Just a patch to remove unnecessary bullet points from facets in Special:FacetedSearch in favour of a plainer, space-saving style.


D-Groenewegen:
That''s a good idea. 

Screenshot 2026-01-08 at 11 11 26

Screenshot 2026-01-08 at 11 12 17

I could have given an example with longer strings as values but no need to belabour the point I guess

PR #6301, opened 2026-02-03
Improve reload notice wording for users

gesinn-it-gea:
Rephrased system message to be less technical and to mention that the reload may take a moment and can repeat.


kghbln (commented):
Only the source language is being changed in the repo. Thus, the change in de.json should be removed. It will be overwritten by what comes from translatewiki.

kghbln:
> I'm not sure if what comes from Translatewiki is 'good' then. Will it be overwritten if we set it here?

We have been using translatewiki.net for ages now. Yes, everything except the source language, English, and message documentation will be overwritten by the next sync from there. I suggest trusting the translators if you think that the source is good. There is also the message documentation (qqq) to provide context if your source is not easily comprehensible.

PR #6405, opened 2026-03-15
Fix "MediaWiki.Usage.AssignmentInReturn.AssignmentInReturn"

paladox:


alistair3149 (changes requested):
A few issues with transformations that change behavior:

1. **`src/DataValues/Number/UnitConverter.php`** — `$this->errors[] = 'smw_nounitsdeclared'` evaluates to the string `'smw_nounitsdeclared'`, not an array. The PR returns `[ 'smw_nounitsdeclared' ]` (an array), which changes the return type. Should be `return 'smw_nounitsdeclared';` (or just `return;` since callers don't use the return value).

2. **`src/SQLStore/Rebuilder/Rebuilder.php`** — `$title->getPrefixedDBKey()` is called twice (once for the push, once for the return). Should use a temporary variable:
   ```php
   $entry = [ 'skipped' => $title->getPrefixedDBKey() ];
   $this->dispatchedEntities[] = $entry;
   return $entry;
   ```

3. **`includes/Highlighter.php`** — The PHPDoc comment documenting the `$content` array structure (`$content['caption']`, `$content['context']`) was removed. This is useful documentation unrelated to the PHPCS fix and should be kept.

PR #6406, opened 2026-03-15
Fix param/var types in Lru

paladox:


paladox:
> The change looks good to me.
> 
> Might be out of scope but we might as well just use native types here instead of phpdocs. We can also use constructor property promotion.

That can be done separately I think.

PR #6412, opened 2026-03-15
Timer: Fix property types

paladox:


src/Utils/Timer.php
```diff
@@ -34,6 +34,7 @@ class Timer {
 	 * @return string|bool
 	 */
 	public static function getTimestamp( $outputType = TS_UNIX, $ts = 0 ) {
+		/* @phan-suppress-next-line PhanUndeclaredFunction Needs fixing to detect MW core functions */
```
alistair3149: I tested Phan locally and did not get the same issue. The import in the config looks right to me too.

paladox: Oh how did you do it? I just cloned this repo and ran composer phan. I did not put it in the mediawiki directory.

alistair3149: You need to have it in MediaWiki directory because it has to [import the config](https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/8410be5e5ab9e6844cf157f30f96eec81d97ff85/.phan/config.php#L2) from MW.

paladox: Oh! Thanks.

PR #6512, opened 2026-03-22
UnusedPropertiesQueryPage: Remove codeCoverageIgnore

paladox:


tests/phpunit/Unit/QueryPages/UnusedPropertiesQueryPageTest.php
```diff
@@ -129,4 +157,71 @@ public function testFormatPropertyItemOnPredefinedProperty() {
...
+		);
+
+		$requestOptions = new RequestOptions();
+		$results = $instance->getResults( $requestOptions );
+
+		$this->assertIsArray( $results );
+		$this->assertEmpty( $results );
+	}
+
+	public function testFormatResultWithInvalidTypeThrowsException() {
```
alistair3149: It covers the same code path as the existing `testInvalidResultThrowsException` — both pass an invalid input to `formatResult()` and expect `PropertyNotFoundException`. Either remove the new one, or consolidate them with a `@dataProvider` if you want to cover multiple invalid types (`null`, `[]`, etc.).

paladox: Removed, thanks.

tests/phpunit/Unit/QueryPages/UnusedPropertiesQueryPageTest.php
```diff
@@ -129,4 +157,175 @@ public function testFormatPropertyItemOnPredefinedProperty() {
...
+		$instance->getResults( $requestOptions );
+
+		$cacheInfo = $instance->getCacheInfo();
+
+		$this->assertIsString( $cacheInfo );
+		$this->assertNotEmpty( $cacheInfo );
+		$this->assertStringContainsString( 'smw-sp-properties-cache-info', $cacheInfo );
+	}
+
+	public function testGetPageHeader() {
```
alistair3149, line 276: testGetPageHeader and testGetPageHeaderContainsSearchForm are essentially testing the same method with the same setup, just asserting different substrings. They could be a single test.

paladox: Thanks, done.

Issue #6520, opened 2026-03-23
Empty HTML nodes provoke PHP 8.1 compatibility issues


freephile:
### Setup
 
| Product	| Version  |
| ---           |   ---    |
| SMW           | 6.0.1    |
| MediaWiki	| 1.43.6 (1e92393) 2026-03-16T17:30:14 |
| PHP	        | 8.1.34 (fpm-fcgi) |
| ICU	        | 74.2     |
| MariaDB	| 10.3.39-MariaDB-log |
| Elasticsearch	| 7.10.2   |
| Lua	        | 5.1.5    |
| Pygments	| 2.2.0    |

### Issue
If you have 'empty' self-closing HTML elements like `<section begin=wvs2/>` in your content and call MediaWiki's `maintenance/refreshLinks.php` you will get a ton of the "familiar" PHP8.1 deprecation warnings **Passing null to parameter #1 ($string) of type string is deprecated** as MW core uses functions like `strlen()`, `strspn()`, `strcspn()`, `strrev()` in `includes/parser/Preprocessor_Hash.php`. But I don't think it's MW Core's fault (although they could perhaps use [a patch to `preprocessToObj()`](https://github.com/wikimedia/mediawiki/blob/master/includes/Parser/Preprocessor_Hash.php#L116) to do the same thing by null-coalescing `$text` to an empty string). The problem comes from the SMW SectionTag.php `parse()` method when it receives a null value and passes it on.

Steps to reproduce the observation (recommendation is to use the [sandbox](https://sandbox.semantic-mediawiki.org)):

Create content with a self-closing HTML **section** tag like `<section begin=wvs2/>` and then run `maintenance/refreshLinks.php`

I can't run maintenance scripts on the sandbox, but the content page on my wiki that triggered this problem is my [ISBN 'Book Sources'](https://en.wikipedia.org/wiki/Help:ISBN) project namespace page defining ISBN lookups at https://wiki.freephile.org/wiki/index.php?title=Freephile_Wiki:Book_sources&action=edit and having `<section begin=wvs2/>` tags.

<details><summary>Important Details of attached Stack trace</summary>
<p>
SMW\ParserFunctions\SectionTag->parse($input = NULL, $args = ['begin' => 'wvs2']) /opt/htdocs/mediawiki/extensions/SemanticMediaWiki/src/ParserFunctions/SectionTag.php:54
</p>
</details> 

[...]

PR #6521, opened 2026-03-23
Fix SectionTag.php for empty tags

freephile:
Use nullable type for $input on the parse() method so it is clear that we accept either string or null. Then use coalescing operator in the recursiveTagParse call so that we don't get PHP 8.1 deprecation warnings for passing null through to downstream parser functions.

Fixes Issue #6520

### DIFF
--- a/src/ParserFunctions/SectionTag.php
+++ b/src/ParserFunctions/SectionTag.php
@@ -69,7 +69,7 @@ class SectionTag {
 			}
 		}
 
-		if ( $title !== null && $title->getNamespace() === SMW_NS_PROPERTY ) {
+		if ( $title->getNamespace() === SMW_NS_PROPERTY ) {
 			$attributes['class'] = ( isset( $attributes['class'] ) ? $attributes['class'] . ' ' : '' ) . "smw-property-specification";
 		}
 
--- a/tests/phpunit/Unit/ParserFunctions/SectionTagTest.php
+++ b/tests/phpunit/Unit/ParserFunctions/SectionTagTest.php
@@ -211,30 +211,7 @@ class SectionTagTest extends TestCase {
 		// appended with a leading space to form a valid class list.
 		$result = $instance->parse( 'Foo', [ 'class' => 'myclass' ] );
 
-		$this->assertStringContainsString( 'myclass', $result );
-		$this->assertStringContainsString( 'smw-property-specification', $result );
-	}
-
-	public function testParse_WithNullTitle() {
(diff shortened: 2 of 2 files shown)


tests/phpunit/Unit/ParserFunctions/SectionTagTest.php
```diff
@@ -112,4 +112,129 @@ public function testParse_PropertyNamespace() {
...
+			$this->frame
+		);
+
+		// When a class arg is already present, the property spec class is
+		// appended with a leading space to form a valid class list.
+		$result = $instance->parse( 'Foo', [ 'class' => 'myclass' ] );
+
+		$this->assertStringContainsString( 'myclass', $result );
+		$this->assertStringContainsString( 'smw-property-specification', $result );
+	}
```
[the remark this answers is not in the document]
freephile: suggested fix (which totally makes sense) was implemented in the next commit

src/ParserFunctions/SectionTag.php
```diff
@@ -70,13 +70,13 @@ public function parse( $input, array $args ) {
 		}
 
 		if ( $title !== null && $title->getNamespace() === SMW_NS_PROPERTY ) {
```
alistair3149: Unrelated but should be cleaned up: `$title = $this->parser->getTitle();` is not nullable. The null guard is not needed.

tests/phpunit/Unit/ParserFunctions/SectionTagTest.php
```diff
@@ -112,4 +112,129 @@ public function testParse_PropertyNamespace() {
...
+
+	public function testParse_WithNullTitle() {
```
alistair3149: Title is not nullable, so you can drop the test, and that should fix the CI.

tests/phpunit/Unit/ParserFunctions/SectionTagTest.php
```diff
@@ -112,4 +112,129 @@ public function testParse_PropertyNamespace() {
...
+		$instance = new SectionTag(
+			$this->parser,
+			$this->frame
+		);
+
+		// When a class arg is already present, the property spec class is
+		// appended with a leading space to form a valid class list.
+		$result = $instance->parse( 'Foo', [ 'class' => 'myclass' ] );
+
+		$this->assertStringContainsString( 'myclass', $result );
```
alistair3149: `class="myclass smw-property-specification` is a stronger assertion.

PR #6528, opened 2026-03-23
Add optimizeStore.php maintenance script

### LINKED ISSUE #4696: Add "optimizeStore.php" maintenance script
### Issue
Currently system administratores do `php setupStore.php --quiet --skip-import` on a regular basis via Cron to optimize the existing tables SMW is using.

"setupStore.php" however also hooks into "update.php" and goes through a setup routine every time you do the maintenance as described above.

I think it will be better to decouple optimization of the database from setting it up. Moreover the new script could have a `--with-maintenance-log` flag to keep track of execution and runtime which would reveal important statistical information from the perspective of a system administrator.

Note this is not about removing the `--skip-optimize` flag and related functionality from "setupStore.php".

vm-pranavan:
Fixes #4696


alistair3149:
Superseded by #7011

PR #6569, opened 2026-04-02
Drop buildResult return type

alistair3149:
Datatable in SemanticResultFormats can return array. We can only type this when SRF is patched.

SRF: https://github.com/SemanticMediaWiki/SemanticResultFormats/blob/7517c40d0cccbe8b5c7aa788e6a83d66f6965edf/formats/datatables/DataTables.php#L421

Tagging @thomas-topway-it since they are the author of the Datatable format


thomas-topway-it:
@alistair3149 noted, thanks

PR #6574, opened 2026-04-04
fix src/Query phan errors

paladox:


src/Query/QueryProcessor.php
```diff
@@ -311,7 +311,7 @@ public static function getResultFromQuery( Query $query, array $params, $outputM
 			// process is picked-up by the `deferred.js` loader that will
 			// initiate an API request to finalize the query after MW has build
 			// the page.
-			if ( $printer->isDeferrable() !== $printer::DEFERRED_DATA ) {
+			if ( $printer->isDeferrable() ) {
 				return Deferred::buildHTML( $query );
 			}
```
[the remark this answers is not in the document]
paladox: isDeferrable is a bool though. Doesn't return a string.

src/Query/Language/ThingDescription.php
```diff
@@ -17,7 +17,7 @@
 class ThingDescription extends Description {
 
 	public function getQueryString( $asValue = false ): string {
-		return $asValue ? ( isset( $this->isNegation ) ? '!' : '' ) . '+' : '';
+		return $asValue ? ( $this->isNegation ? '!' : '' ) . '+' : '';
 	}
```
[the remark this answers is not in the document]
paladox: Fixed I think?

src/Query/QueryProcessor.php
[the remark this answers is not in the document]
paladox: So you can close this as resolved then?

src/Query/Language/Description.php
```diff
@@ -179,7 +179,7 @@ public function getQueryFeatures(): int {
 	 * reused in multiple places of one or many queries. Make new objects to reflect
 	 * changes!
 	 */
-	public function prune( &$maxsize, &$maxDepth, &$log ): Description {
+	public function prune( &$maxsize, &$maxDepth, &$log ): mixed {
```
alistair3149: Are we sure that this is not always returning Description?

alistair3149:
LGTM other than the comment above, I am not sure about widening the return type to `mixed` unless absolutely nessecary.

src/Query/Cache/ResultCache.php
```diff
@@ -152,7 +153,7 @@ public function setQueryEngine( QueryEngine $queryEngine ): void {
 	/**
 	 * @since 2.5
 	 *
-	 * @param boolean
+	 * @return bool
 	 */
 	public function isEnabled() {
```
alistair3149: Might as well set the native return type here if we are sure that it is only returning bool.

src/Query/Language/Description.php
paladox: Seems reverting back to Description also works. I guess because I updated ThingDescription prune to use ThingDescription rather than static maybe?

src/Query/Cache/ResultCache.php
paladox: Thanks, done.

PR #6628, opened 2026-04-16
Fix src/Protection phan errors

paladox:


src/Protection/ProtectionValidator.php
```diff
@@ -37,12 +37,12 @@ class ProtectionValidator {
 	private bool $changePropagationProtection = true;
 
 	/**
-	 * @var
+	 * @var array
```
alistair3149: We can drop the phpdoc since it is not adding any new information covered by the PHP type.

paladox: Thanks, done.

PR #6632, opened 2026-04-16
Fix src/Property phan errors

paladox:


src/Property/Annotators/PropertyAnnotatorDecorator.php
```diff
@@ -48,7 +48,7 @@ public function getSemanticData() {
 	 *
 	 * @since 1.9
 	 *
-	 * @return PropertyAnnotator
+	 * @return self
```
alistair3149: We have the native type here already, no need to keep the return tag

src/Property/DeclarationExaminerFactory.php
```diff
@@ -34,7 +34,7 @@ public function newDeclarationExaminerMsgBuilder(): DeclarationExaminerMsgBuilde
 	 * @param Store $store
 	 * @param SemanticData|null $semanticData
 	 *
-	 * @return DeclarationExaminer
+	 * @return UserdefinedPropertyExaminer
```
alistair3149: Same here

paladox: Thanks, done.

src/Property/Annotators/PropertyAnnotatorDecorator.php
paladox: Thanks, done.

PR #6633, opened 2026-04-16
Fix src/PostProcHandler.php phan errors

paladox:


src/PostProcHandler.php
```diff
@@ -98,7 +95,7 @@ public function getOption( $key, $default = false ) {
 	/**
 	 * @since 3.0
 	 *
-	 * @return array|string
+	 * @return array
```
alistair3149: We already have native type, no need to have the return tag since it is not adding new information

paladox: Thanks, done.

PR #6634, opened 2026-04-16
Fix src/ParserParameterProcessor.php phan errors

paladox:


src/ParserParameterProcessor.php
```diff
@@ -75,7 +72,7 @@ public function getFirstParameter(): ?string {
 	 *
 	 * @since 1.9
 	 *
-	 * @return string
+	 * @return array
```
alistair3149: We already have native type here. No need to keep the return tag

paladox: Thanks, done.

PR #6641, opened 2026-04-17
Fix src/ParserParameterProcessor.php phan errors

paladox:


src/ParserParameterProcessor.php
```diff
@@ -61,7 +61,7 @@ public function addError( $error ): void {
 	/**
 	 * @since 2.3
 	 *
-	 * @return string
+	 * @return string|null
```
alistair3149: We can drop the return tag

paladox: Thanks, done.

PR #6642, opened 2026-04-17
Fix src/ParserFunctionFactory.php phan errors

paladox:


src/ParserFunctionFactory.php
```diff
@@ -247,7 +247,7 @@ public function newConceptParserFunction( Parser $parser ): ConceptParserFunctio
 	 *
 	 * @param Parser $parser
 	 *
-	 * @return SubobjectParserFunction
+	 * @return SubobjectParserFunc
```
alistair3149: If phpdoc has the same return type as the native type, there are no need to include it

src/ParserFunctionFactory.php
```diff
@@ -293,7 +293,7 @@ public function newSubobjectParserFunction( Parser $parser ): SubobjectParserFun
 	 *
 	 * @param Parser $parser
 	 *
-	 * @return RecurringEventsParserFunction
+	 * @return RecurringEventsParserFunc
```
alistair3149: Same here

paladox: Thanks, done.

src/ParserFunctionFactory.php
paladox: Thanks, done.

PR #6643, opened 2026-04-17
Fix ParserData.php phan errors

paladox:


src/ParserData.php
```diff
@@ -273,7 +270,7 @@ public function setEmptySemanticData(): void {
 	/**
 	 * @since 2.1
 	 *
-	 * @param ParserOutput|null
+	 * @param ParserOutput|null $parserOutput
```
alistair3149: Can drop this since the param is typed

src/ParserData.php
```diff
@@ -180,7 +177,7 @@ public function setParserOptions( ParserOptions $parserOptions ): void {
 	/**
 	 * @since 3.0
 	 *
-	 * @return ParserOptions|null
+	 * @return void
```
alistair3149: Can drop this since the return is typed

src/ParserData.php
paladox: Thanks, done.

src/ParserData.php
paladox: Thanks, done.

PR #6646, opened 2026-04-17
Fix ParameterListDocBuilder.php phan errors

paladox:


src/ParameterListDocBuilder.php
```diff
@@ -17,19 +17,12 @@ class ParameterListDocBuilder {
...
 	}
 
 	/**
 	 * Returns the wikitext for a table listing the provided parameters.
-	 *
-	 * @param ParamDefinition[] $paramDefinitions
-	 *
-	 * @return string
 	 */
 	public function getParameterTable( array $paramDefinitions ): string {
```
alistair3149, line 31: We need to keep the `@param` phpdoc because it describes the array shape `ParamDefinition[]`, which is added information.

src/ParameterListDocBuilder.php
```diff
@@ -58,14 +51,9 @@ public function getParameterTable( array $paramDefinitions ): string {
 			"\n|}";
 	}
 
-	/**
-	 * @param ParamDefinition[] $paramDefinitions
-	 *
-	 * @return bool
-	 */
 	private function containsAliases( array $paramDefinitions ): bool {
```
alistair3149, line 63: Same here

paladox: Thanks, done.

src/ParameterListDocBuilder.php
paladox: Thanks, done.

PR #6662, opened 2026-04-21
Fix HierarchyLookup.php phan errors

paladox:


src/HierarchyLookup.php
```diff
@@ -169,27 +147,24 @@ public function hasSubcategory( WikiPage $category ): bool {
 
 	/**
 	 * @since 2.3
-	 *
-	 * @param Property $property
-	 *
-	 * @return DIWikiPage[]|[]
 	 */
-	public function findSubpropertyList( Property $property ) {
+	public function findSubpropertyList( Property $property ): array|false {
```
alistair3149: We need to keep the return annotation for ` DIWikiPage[]|false` if it was accurate.

src/HierarchyLookup.php
```diff
@@ -199,32 +174,32 @@ public function findSubcategoryList( WikiPage $category ) {
 
 	/**
 	 * @since 3.1
-	 *
-	 * @param WikiPage $category
-	 *
-	 * @return DIWikiPage[]|[]
 	 */
-	public function findNearbySuperCategories( WikiPage $category ) {
+	public function findNearbySuperCategories( WikiPage $category ): array {
```
alistair3149: Preserve `@return DIWikiPage[]` if it is accurate.

src/HierarchyLookup.php
```diff
@@ -199,32 +174,32 @@ public function findSubcategoryList( WikiPage $category ) {
...
 
 	/**
 	 * @since 3.0
-	 *
-	 * @param Property|WikiPage $id
-	 *
-	 * @return DIProperty[]|DIWikiPage[]|[]
 	 */
-	public function getConsecutiveHierarchyList( $id, $hierarchyType = self::TYPE_SUB ) {
+	public function getConsecutiveHierarchyList(
```
alistair3149, line 201: Preserve `@return DIProperty[]|DIWikiPage[]` if it is accurate.

src/HierarchyLookup.php
paladox: You can't do DIWikiPage[] as a php type. Only as a phpdoc.

alistair3149: Yes return annotation means the phpdoc.

paladox: Thanks, done.

src/HierarchyLookup.php
paladox: Thanks, done.

src/HierarchyLookup.php
paladox: Thanks, done.

src/HierarchyLookup.php
```diff
@@ -200,31 +178,35 @@ public function findSubcategoryList( WikiPage $category ) {
 	/**
 	 * @since 3.1
 	 *
-	 * @param WikiPage $category
-	 *
-	 * @return DIWikiPage[]|[]
+	 * @return WikiPage[]|array|Iterator
```
alistair3149: `WikiPage[]` includes `array`, so `@return WikiPage[]|Iterator` is fine

src/HierarchyLookup.php
```diff
@@ -200,31 +178,35 @@ public function findSubcategoryList( WikiPage $category ) {
...
+		);
 	}
 
 	/**
 	 * @since 3.0
 	 *
-	 * @param Property|WikiPage $id
-	 *
-	 * @return DIProperty[]|DIWikiPage[]|[]
+	 * @return Property[]|WikiPage[]|array
```
alistair3149: Same here, you can drop `array`

paladox: Thanks, done.

src/HierarchyLookup.php
paladox: Thanks, done.

Issue #6559, opened 2026-03-26
SMW performance improvement suggestions


bawolff:
I've been recently trying to make WikiApiary faster. While doing this i discovered some queries which i think are suboptimal (However, i am not an expert in the code base, so there could be reasons for these or maybe i just don't understand). Putting it here as a list in case its helpful feedback.

Keep in mind, wikiapiary was using an old version (4.1.1). Its also possible people might have changed things in weird ways making it not reflect defaults. MariaDB version: 10.11.14-MariaDB-0+deb12u2

I also haven't tested these, so of course ymmv

- Property namespace pages use OFFSET paging. This was the number one performance issue i encountered as the query was slow and bots were hitting this hard. If possible, this should be switched to a WHERE whateverFieldWeAreOrderingBy > lastResultShown. Mysql is much faster when using paging via where.
- smw_object_ids.hash A lot of slow queries seem to have the form 
```sql
SELECT smw_id,smw_title,smw_namespace,smw_hash 
FROM `smw_object_ids` 
WHERE smw_hash IN ('cb365ef5b79b68efca23569d97beb9ed05afa88d','3a353497a3f53a6a20b4d9c3a9c88c2a887a6e9b','a5079f89d2fb736190f3168c3c3f77d2010497ba','9450d63b62a652f81df84a032d3bcc60351d30ba','ab35d5e0caf08aced0581b4902ef277a808d44a2','d7732f0b9ee0616b8d3a69ab7c719030cc6f258d', ... 
```
using an index on (smw_hash, smw_id). Instead consider taking the fist 64 bits of the sha1 hash, and storing it as an unsigned long. This would make the index file shorter allowing more of it to fit in ram at once, potentially improving speed. The smw_id part of the index is also seems unnecessary. This would require application code to revalidate the hash to be sure that no collisions happen. [ If we have to keep the current structure, at least make the field be the binary representation of the hash instead of hexadecimal and make the index only index the first 6 bytes of the field)
- The smw_di_wikipage table seems to have some indexes that are entirely a prefix of another index. This seems kind of pointless and they should probably be removed. I'd suggest just having indexes on (s_id, p_id), (o_id, s_id) and (p_id, o_id, s_id)
- Queries like:
```sql
[...]

PR #6695, opened 2026-04-28
Drop redundant (o_id) index from smw_di_wikipage

alistair3149:
## Summary

- Removes the standalone `(o_id)` entry from `DIWikiPageHandler::getTableIndexes()`. It was a leftmost prefix of `(o_id, s_id)` and `(o_id, p_id)`, so the composites cover any query the standalone could serve while it imposed B-tree maintenance cost on every INSERT/UPDATE/DELETE to `smw_di_wikipage` and to every `smw_fpt_*` fixed-property table backed by the wikipage handler.
- Audited the other DataItem handlers (`DIBlobHandler`, `DIBooleanHandler`, `DIConceptHandler`, `DIGeoCoordinateHandler`, `DINumberHandler`, `DITimeHandler`, `DIUriHandler`); `DIWikiPageHandler` was the only one with prefix-redundancy.
- Existing wikis pick up the change automatically on `update.php` via the existing `TableBuilder` schema-reconciliation path. **On very large wikis, the resulting `ALTER TABLE ... DROP INDEX` may take some time.** No data migration, no application-side flag.

Refs #6559 (the broader index-reshuffle proposed in that thread is intentionally out of scope here — it depends on `EXPLAIN` evidence from real-world wikis and interacts with the existing `getIndexHint` workaround in `DIWikiPageHandler.php`).

## Test plan

- [x] `composer test -- --filter "DIWikiPageHandlerTest|TableSchemaManagerTest"` — 17 tests, 33 assertions, OK
- [x] `composer analyze` — no new errors or warnings on the modified files
- [x] CI green
- [ ] Smoke-test on a dev wiki: after `update.php`, the standalone `o_id` index is gone from `smw_di_wikipage` and from each `smw_fpt_*` table whose value type is wikipage; the five composite indexes remain intact

### DIFF
--- a/docs/releasenotes/RELEASE-NOTES-7.0.0.md
+++ b/docs/releasenotes/RELEASE-NOTES-7.0.0.md
@@ -189,6 +189,7 @@ For more detailed information, see the [compatibility matrix](../COMPATIBILITY.m
 * `#ask` queries that sort by a property value are now significantly faster on MariaDB and MySQL. The query engine restructures the SQL so the database can choose a more efficient plan; on large wikis the improvement can be orders of magnitude depending on query shape. ([#6559](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/6559))
   * Set `$smwgQUseLegacyQuery = true` in `LocalSettings.php` to fall back to the previous query shape if you encounter a regression after upgrading.
   * A redundant `DISTINCT` keyword was also dropped from the disjunction-query temp-table insert. Same result, less work for the database; no setting required.
+* Page saves are a little faster and the database uses less memory on large wikis. A redundant index on `smw_di_wikipage` and related fixed-property tables has been removed. ([#6559](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/6559))
 
 ### Internal improvements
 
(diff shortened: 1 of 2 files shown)


alistair3149:
Closing this — investigation showed the change is a no-op.

The standalone `(o_id)` index in `smw_di_wikipage` (and the corresponding `smw_fpt_*` tables) is created by the `'po'` base index in `TableSchemaManager::newPropertyTable` — line 339 calls `$diHandler->getIndexField()`, which returns `'o_id'` for `DIWikiPageHandler`. This happens *before* `getTableIndexes()` is consulted, and the `array_search` check at `TableSchemaManager.php:359` then dedupes the handler's `'o_id'` entry against the base. So the standalone `'o_id'` line in `getTableIndexes()` was already a no-op pre-PR; removing it doesn't drop a single index from any wiki.

Verified empirically by running `update.php` against a fresh dev wiki on this branch and observing the standalone ``o_id`` index still present in ``SHOW INDEX FROM smw_di_wikipage``.

bawolff's observation in #6559 is correct — the standalone ``(o_id)`` *is* prefix-redundant against ``(o_id, s_id)`` and ``(o_id, p_id)``. The actual fix needs to either (a) skip the ``'po'`` base in ``newPropertyTable`` when a handler's composites cover it, or (b) override ``getIndexField()`` for ``DIWikiPageHandler`` to point at one of the composites. Either touches schema-management core and warrants its own design pass — to be picked up separately as part of the broader #6559 work.

PR #6702, opened 2026-04-28
Fix src/ConfigLegacyTrait.php phan errors

paladox:


alistair3149:
Unrelated: The config seem to be a good candidate to remove before 7.0.0 release, almost all of them are from 3.X.

PR #6704, opened 2026-04-28
Fix src/Connection/ConnectionManager.php phan errors

paladox:
> src/Connection/ConnectionManager.php:100 PhanPluginUnreachableCode Unreachable statement detected

### DIFF
--- a/src/Connection/ConnectionManager.php
+++ b/src/Connection/ConnectionManager.php
@@ -96,8 +96,6 @@ class ConnectionManager {
 		);
 
 		$setupCheck->showErrorAndAbort( $setupCheck->isCli() );
-
-		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


src/Connection/ConnectionManager.php
```diff
@@ -96,8 +96,6 @@ private function isConnectable(): bool {
 		);
 
 		$setupCheck->showErrorAndAbort( $setupCheck->isCli() );
```
alistair3149, line 98: The process dies here, so the return type of this function might not be correct since this is not returning anything.

paladox: Oh? What should we change it to? Since we use the return value somewhere in the file?

PR #6712, opened 2026-04-29
Run phan under composer test

paladox:


alistair3149:
@gesinn-it-gea would it be possible to have the `php-ast` PHP extension in the CI? That would be helpful for running Phan.

PR #6732, opened 2026-05-03
Fix "MediaWiki.Commenting.PropertyDocumentation.MissingDocumentationPublic"

paladox:


src/Query/Query.php
```diff
@@ -153,7 +153,7 @@ public function isEmbedded(): bool {
 	 *
 	 * @param int $queryMode
```
alistair3149: Can remove this now since this is duplicated.

paladox: I've done it but in phpcs we have it disabled to report on @param not being defined.

PR #6747, opened 2026-05-04
config: Migrate path-relative defaults via path:true (4/5)

alistair3149:
## Summary

PR 4 of a 5-PR series moving SMW's `$smwg*` defaults from the bespoke `src/DefaultSettings.php` + `setupGlobals()` pipeline to MediaWiki's standard `extension.json` `config` block. **Stacked on PR #6746** (PR 3). GitHub auto-retargets to `master` as the upstream PRs merge.

The 7 `$smwgIP`-relative path settings move into `extension.json` using the manifest's `"path": true` flag, which prefixes the extension directory at registration time:

| Setting | extension.json key | value | flags |
| --- | --- | --- | --- |
| `smwgIP` | `IP` | `""` | `path: true` |
| `smwgDir` | `Dir` | `""` | `path: true` |
| `smwgConfigFileDir` | `ConfigFileDir` | `""` | `path: true` |
| `smwgExtraneousLanguageFileDir` | `ExtraneousLanguageFileDir` | `"i18n/extra"` | `path: true` |
| `smwgServicesFileDir` | `ServicesFileDir` | `"src/Services"` | `path: true` |
| `smwgMaintenanceDir` | `MaintenanceDir` | `"maintenance"` | `path: true` |
| `smwgImportFileDirs` | `ImportFileDirs` | `{ "smw": "data/import" }` | `path: true` + `merge_strategy: array_plus` |

The `merge_strategy: array_plus` on `ImportFileDirs` makes the documented user pattern from `src/Importer/README.md` (`$GLOBALS['smwgImportFileDirs']['custom-vocab'] = __DIR__ . '/custom';`) work correctly — adding a new vocabulary without dropping the default `'smw' => …/data/import` entry.

## Note on resolved-path string semantics

The original `$smwgIP . '/i18n/extra'` produced `.../SemanticMediaWiki//i18n/extra` (double slash, since `$smwgIP = dirname(__DIR__) . '/'` ended with a slash). MW's `path: true` produces `.../SemanticMediaWiki/i18n/extra` (single slash). Functionally equivalent on Unix; only matters if any consumer does literal-string comparison on these paths. Confirmed (via codebase grep) that no SMW or MW core consumer does that — every consumer treats these as filesystem paths via `is_dir()` / `realpath()` / `file_get_contents()` / etc.

## After this PR

[...]

PR #6754, opened 2026-05-04
Clarify $smwgConfigFileDir override semantics

alistair3149:
## Summary

Follow-up to #6747. [Nikerabbit pointed out](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/6747#issuecomment-4370751617) that the new `$smwgConfigFileDir` section in `docs/config.md` reads as a contradiction:

> On one hand this says values are prefixed with SMW's extension directory, on other hand it says e.g. config directory can be overwritten to upload directory.

Both statements are individually correct (the manifest's `path: true` flag prefixes the *default* with the extension directory, while a `LocalSettings.php` override is taken verbatim), but the prior wording never spelled out that distinction, so a reader naturally infers a conflict.

## Change

Reword only the body of `$smwgConfigFileDir`, making it explicit that the override is set in `LocalSettings.php` to an absolute path, with `$wgUploadDirectory` cited as one example. The default value, the `**Default:**` line, and the manifest entry are unchanged.

```diff
-`.smw.maintenance.json`). The directory must be writable. You may assign
-the same directory as `$wgUploadDirectory` or select an entirely different
-location.
+`.smw.maintenance.json`). The directory must be writable. To override the
+default, set this in `LocalSettings.php` to an absolute path, for example
+`$wgUploadDirectory` or any other writable location.
```

## Test plan

- [x] CI green (docs-only, no code paths touched)
- [x] Visual review of the rendered section on GitHub

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -209,9 +209,9 @@ them.
 ## $smwgConfigFileDir
 
 Directory used to persistently store SMW configuration files (`.smw.json`,
-`.smw.maintenance.json`). The directory must be writable. You may assign
-the same directory as `$wgUploadDirectory` or select an entirely different
-location.
+`.smw.maintenance.json`). The directory must be writable. To override the
+default, set this in `LocalSettings.php` to an absolute path, for example
+`$wgUploadDirectory` or any other writable location.
 
 **Since:** 3.0
 **Default:** the extension's root directory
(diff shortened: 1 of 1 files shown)


Nikerabbit:
I don't see in the code that it would handle absolute paths differently from relative ones. Neither in https://gerrit.wikimedia.org/g/mediawiki/core/+/b31a114afab0cc3c9e71d2feee851a45944ed809/includes/Registration/ExtensionProcessor.php#911 or in SMW code.

alistair3149:
Just verified and you are right.
Stating absolute path was overly defensive as I thought that `cwd` is unreliable.
I have updated the description to drop the mention of absolute path.

PR #6765, opened 2026-05-04
Drop legacy DML wrappers + slim Database facade

alistair3149:
## Summary

Final cleanup of SMW's database access layer migration to MediaWiki Rdbms QueryBuilder factories. Earlier work converted every internal callsite from the positional `IDatabase`-style API to `newSelectQueryBuilder()` / `newInsertQueryBuilder()` / `newUpdateQueryBuilder()` / `newDeleteQueryBuilder()` / `newReplaceQueryBuilder()`. With those callsites all migrated, the SMW database wrapper sheds the legacy passthroughs it has been keeping alive purely for backwards compatibility, and the cross-DB regex shim block in `Database::executeQuery()` is deleted now that no caller depends on it.

This PR also closes a gap in the earlier audits: the top-level `maintenance/` directory was missed (audits scoped to `src/` only), so 14 callsites across 6 maintenance scripts are migrated here as a precondition for the wrapper deletion.

## What's deleted

Methods removed from `SMW\MediaWiki\Connection\Database`:

- `select()`, `selectRow()`, `selectField()`, `estimateRowCount()`
- `insert()`, `update()`, `delete()`, `upsert()`, `replace()`
- `makeSelectOptions()` passthrough

`Database::query()` and `Database::readQuery()` accept only `string` now; the `Query|string` union has been removed (`Query::execute()` updated to call `\$this->build()` before passing the result through).

The cross-DB regex shim block in `Database::executeQuery()` is deleted in full. It used to rewrite `INSERT IGNORE` → `ON CONFLICT DO NOTHING`, `@INT` → `::integer`, `TEMPORARY` → `TEMP`, `ENGINE=MEMORY` → strip, `DROP TEMPORARY TABLE` → `DROP TABLE IF EXISTS`, `RAND()` → `RANDOM()`, `TRUNCATE TABLE` → `DELETE FROM`, etc. on the way to the underlying connection. With every emitter now producing platform-correct SQL directly, the shim is dead code.

`OptionsBuilder` (the class) is kept: `TraversalPropertyLookup` and `QueryEngine`'s legacy debug path still use it. Only the `Database::makeSelectOptions()` wrapper passthrough is removed; `QueryEngine::doExecuteDebugQueryResult()` now calls `OptionsBuilder::makeSelectOptions(\$connection, \$sqlOptions)` directly.

`Query.php` + `Database::newQuery()` are also kept: `EntityUniquenessLookup` uses `Query` as a formatter object passed to user-supplied `extraCondition` callbacks (callback signature is documented public API). Deletion needs a callback-signature change and is left for a follow-up.

## What's added

Class-level docblock on `Database` describing the four categories of methods the wrapper may expose (QueryBuilder factories, connection-routing helpers, transaction plumbing, platform-quirk escape hatches) and stating an explicit rule against new SQL-passing methods. Future maintainers see the intent up front; the wrapper's eventual removal is named in the docblock.

Short breaking-change entry in `docs/releasenotes/RELEASE-NOTES-7.0.0.md` listing the removed methods and pointing external consumers at MediaWiki core's [Manual:Database access](https://www.mediawiki.org/wiki/Manual:Database_access). The SMW wrapper is internal infrastructure scheduled for further removal; external code that migrates to MW core directly only migrates once.

## Platform-aware emission conversions (commit `55b86c7a4`)

Two callers were still relying on the shim's regex translations to produce dialect-correct SQL on Postgres / SQLite. Converted to explicit `getType()` dispatch:

- `TemporaryTableBuilder::create()` (via `getSQLCodeFor()`): adds an explicit SQLite branch that emits `CREATE TEMP TABLE ... ( id INTEGER PRIMARY KEY )` (no `UNSIGNED`, no `ENGINE=MEMORY`).
- `TemporaryTableBuilder::drop()`: replaces the bare `DROP TEMPORARY TABLE` with three-way dispatch (MySQL `DROP TEMPORARY TABLE` / Postgres `DROP TABLE IF EXISTS` / SQLite `DROP TABLE`).
- `ConceptCache::refreshConceptCache()`: extends the existing Postgres / non-Postgres split to three-way (Postgres `INSERT INTO` with LEFT JOIN dedup / SQLite `INSERT OR IGNORE` / MySQL `INSERT IGNORE`).

## Maintenance script conversions (commit `2ea4880ee`)

- `maintenance/populateHashField.php` (3 callsites: `selectField`, `select`, `update`)
- `maintenance/purgeEntityCache.php` (2 callsites: `selectField`, `select`)
- `maintenance/rebuildElasticMissingDocuments.php` (2 callsites)
- `maintenance/updateEntityCollation.php` (2 callsites; one comment added clarifying the `->where([ \$condition ])` array form requires non-empty `\$condition`)
- `maintenance/updateEntityCountMap.php` (4 callsites including one `upsert`)
- `maintenance/updateQueryDependencies.php` (1 callsite: `select` with INNER JOIN + GROUP BY)

Mechanical conversion. Behaviour preserved exactly.

## Test fixture updates

10 unit-test files modified:

- `DatabaseTest`: `testSelectMethod`, `testSelectFieldMethod`, `testSelectThrowsException` deleted (tested removed methods directly). `testQueryOnSQLite` (which exercised the deleted shim translations) replaced with a single pass-through assertion confirming SQL reaches the underlying connection unchanged. `testDoQueryWithAutoCommit` updated to drop the `expects(getConnection)->atLeastOnce()` on the read connection (no longer called now that `executeQuery()` doesn't probe `getType()`).
- `QueryTest::testExecute`: updated for the `Query::execute()` -> `\$this->build()` change so it asserts on a string-typed argument and supplies a minimal type/table/field configuration so `build()` doesn't throw.
- 8 other tests had legacy `select`/`selectRow`/etc. mocks alongside newer QueryBuilder-factory stubs (cross-impact fixes from earlier conversion work). The legacy mocks now error with `MethodCannotBeConfigured` and are removed; where the legacy mock provided coverage that the factory stub didn't (e.g. `ElasticStoreTest`'s `populateHashField` flow needing `newSelectQueryBuilder` wired on the `mw.db` connection mock plus extra row fields), the factory stub was extended.
- `DataRebuilderTest`, `DistinctEntityDataRebuilderTest`, `PurgeEntityCacheTest`: capture `\$capturedTables` and `\$whereConditions` to pin the table-routing and condition shape that the legacy `with(stringContains(...), ...)` matchers used to enforce, recovering the coverage those matchers provided.

## Verification

- `composer analyze`: clean (2122 files, 0 syntax errors, 0 PHPCS issues).
- Full unit suite: 6867 tests, 14119 assertions, 6 pre-existing skips (elasticsearch-php / postgres test environment, unrelated).
- `update.php`: clean (re-runs required after the schema-key change in `executeQuery()` removal).
- Browser smoke against the seeded dev wiki:
  - Special:Ask default sort over `[[Category:Dog_Breeds]]`: 5 results in alphabetical order.
  - Special:Ask default sort over `[[Category:Dogs]]` (parent category, exercises hierarchy temp-table machinery via HierarchyTempTableBuilder + the converted TemporaryTableBuilder): 5 results returned correctly.
  - Special:Ask `order=random`: 3 different results per consecutive run.
  - Special:Browse, Special:SearchByProperty, Special:SemanticMediaWiki: render without errors.
  - No `smw-error`, `Fatal error`, `Call to undefined method`, or unexpected SQL token mangling.
- `maintenance/purgeEntityCache.php` exercised end-to-end: fetched 194 entities via the converted SELECTs, invalidated 675 cache entries, clean exit.

## Test plan
- [x] Local unit suite green on MariaDB container
- [x] `composer analyze` clean
- [x] `update.php` clean
- [x] Browser smoke: Ask default + random sort + hierarchy traversal, Browse, SearchByProperty, SMW admin
- [x] Maintenance script smoke (purgeEntityCache.php end-to-end)
- [x] CI green


docs/releasenotes/RELEASE-NOTES-7.0.0.md
```diff
@@ -116,6 +116,7 @@ For more detailed information, see the [compatibility matrix](../COMPATIBILITY.m
 
 **Removed APIs:**
 
+* **Legacy DML methods on `SMW\MediaWiki\Connection\Database` removed.** Removed methods: `select()`, `selectRow()`, `selectField()`, `estimateRowCount()`, `insert()`, `update()`, `delete()`, `upsert()`, `replace()`, and the `makeSelectOptions()` passthrough. `Database::query()` and `Database::readQuery()` also tightened from `Query|string` to `string` only; callers that built SQL via `$db->newQuery()` should call `$query->build()` themselves and pass the resulting string. SMW's database wrapper is internal infrastructure; external code that called these methods directly on `$store->getConnection( 'mw.db' )` should migrate to MediaWiki core's database services. See [Manual:Database access](https://www.mediawiki.org/wiki/Manual:Database_access).
```
paladox, line 119: Should we not deprecated first before straight to removal? (Thinking of extensions).

alistair3149: I searched on [Codesearch](https://codesearch.wmcloud.org/things/?q=SMW%5C%5CMediaWiki%5C%5CConnection%5C%5CDatabase&files=&excludeFiles=&repos=) and it seems purely internal. Besides the Database class is a thin wrapper for MW database methods, external consumers should have used MW methods instead.

paladox: 👍

PR #6779, opened 2026-05-08
phan: Fix "PhanParamSignatureMismatch"

paladox:


src/Elastic/Connection/DummyClient.php
```diff
@@ -194,13 +194,15 @@ public function delete( array $params ): array {
...
+		return [];
 	}
 
 	/**
 	 * @see Client::index
 	 */
-	public function index( array $params ): void {
+	public function index( array $params ): array {
+		return [];
 	}
```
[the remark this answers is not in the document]
paladox: They do here https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Elastic/Connection/Client.php#L451 ?

src/Property/Annotator.php
```diff
@@ -28,7 +28,7 @@ public function getSemanticData();
 	 *
 	 * @since 1.9
 	 *
-	 * @return Annotator
+	 * @return Annotator|void
```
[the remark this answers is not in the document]
paladox: Yes they do https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/424f927b5b8fcba302c4c5e45a6aa85e44521059/src/Query/ProfileAnnotators/NullProfileAnnotator.php#L71

src/QueryPages/QueryPage.php
```diff
@@ -49,8 +49,10 @@ abstract public function getResults( $requestoptions );
 	/**
 	 * Clear the cache and save new results
 	 * @todo Implement caching for SMW query pages
+	 *
+	 * @suppress PhanTypeMissingReturnReal
 	 */
-	public function recache( $limit, $ignoreErrors = true ): void {
+	public function recache( $limit, $ignoreErrors = true ): bool {
 		/// TODO
 	}
```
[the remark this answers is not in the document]
paladox: whoops should have reverted that. Thanks, done.

src/DataValues/ValueParsers/AllowsListValueParser.php
```diff
@@ -45,9 +45,9 @@ public function clear(): void {
 	/**
 	 * @since 2.5
 	 *
-	 * @param string $userValue
+	 * @param mixed $userValue
 	 *
-	 * @return string|false
+	 * @return array|null
```
alistair3149: Are you sure that there is a null path?

src/Property/Annotator.php
alistair3149: `@return Annotator|void` isn't valid because `void` is not a phpdoc value type.
Either drop the tag or have void implementation return `null` and use `@return Annotator|null`

paladox: @coderabbitai are you able to show me the code for above for all files that need changing? but also can you tell me if any class uses this function to return Annotator?

paladox: I guess we can drop the tag.

paladox: Done.

src/DataValues/ValueParsers/AllowsListValueParser.php
paladox: Fixed I think.

PR #6789, opened 2026-05-11
phan: Fix src/Services/mediawiki.php "PhanTypeMismatchArgument" error

paladox:


src/Services/mediawiki.php
```diff
@@ -55,7 +55,7 @@
 
 	/**
 	 * @param CallbackContainerBuilder $containerBuilder
-	 * @param ImportStreamSource $source
+	 * @param string $source
```
alistair3149: It should be [`resource`](https://gerrit.wikimedia.org/g/mediawiki/core/+/6668a5fae63a3d62cc9a02f7dd62e141447812ef/includes/Import/ImportStreamSource.php#28)

paladox: Thanks, done.

PR #6839, opened 2026-05-19
Add order=none to skip result sorting in #ask queries

alistair3149:
## Problem

Every `#ask` query sorts its results, by default by page. The `sort` parameter defaults to the empty-string key, which `QueryCreator` turns into a sort-key map, and the query engine renders that as an `ORDER BY`. There is no per-query way to opt out.

On large result sets this is expensive: the database sorts the whole intermediate result set before applying the `limit`. A benchmark on a category with 1,000,000 members measured the default-sorted query at about 94 seconds to return 500 rows, versus about 52 milliseconds for the same query unsorted, where the `limit` short-circuits with no sort. The gap grows with result-set size.

## Change

Add `order=none` to `#ask`. It maps to an internal `NONE` token, and `QueryCreator::getSortKeys()` then returns an empty sort-key map, which both the SQL and SPARQL query engines already render as no `ORDER BY`. No query-engine changes are needed.

- `none` in any `order=` slot makes the whole query unsorted.
- Results are returned in an unspecified order.
- `order=none` combined with `cursor=` is rejected with a query error, because keyset pagination needs a stable order to seek past.
- The directive is recorded as a `Query::SORT_DISABLED` option so `QueryStringifier` and `QueryLinker` re-emit `order=none`. Without this, the "further results" link and the Special:Ask round-trip dropped the parameter, and a paginated query re-sorted on the next page.

## Testing

- Unit tests for `QueryCreator`, `QueryStringifier` and `QueryLinker` covering the parameter, the cursor rejection, the any-slot rule, and the serialisation round-trip.
- Exercised end-to-end against a wiki: an `#ask` with `order=none` generates SQL with no `ORDER BY` (a default query still emits one), the result page renders without error, and `order=none` is preserved through the "further results" link and Special:Ask.

Refs #6559

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jaideraf:
@kghbln FYI

jaideraf:
Somewhat documented: https://www.semantic-mediawiki.org/w/index.php?title=Help%3ASorting&type=revision&diff=91405&oldid=82002

PR #6843, opened 2026-05-19
Remove 3 pre-7.0 deprecated methods (Tier 3) and update callers

alistair3149:
## What

Tier 3 of the pre-SMW-7.0 deprecation cleanup. Removes three deprecated thin-delegator methods and repoints their callers to the documented replacement:

| Removed | Replacement | Deprecated since |
|---|---|---|
| `ParserData::pushSemanticDataToParserOutput()` | `copyToParserOutput()` | 3.0 |
| `DataValue::isEnabledFeature()` | `hasFeature()` | 3.0 |
| `TaskHandler::isEnabledFeature()` | `hasFeature()` | 3.1 |

23 call sites repointed (8 `pushSemanticDataToParserOutput`, 15 `isEnabledFeature`). Each removed method delegated directly to its replacement, so there is no behaviour change.

`ResultPrinter::isEnabledFeature()` is a **separate, non-deprecated** method (same name, different class) and is intentionally left unchanged, along with its test.

Also updated: `RELEASE-NOTES-7.0.0.md` and two `docs/examples/*.md` snippets that referenced the removed `pushSemanticDataToParserOutput()`.

## Why

Dead-code cleanup for the 7.0 major release; these methods have carried `@deprecated` annotations since 3.0-3.1. Follows on from #6841 (Tier 1+2).

## Testing

- `parallel-lint`: pass
- PHPCS: pass on all 21 changed PHP files
- PHPUnit unit suite: pass (7093 tests, 0 failures)
- Independent code review: pass
- PHPUnit integration/structure suites: not verified locally (local dev container hang). Relying on CI, which runs the full suite on a clean environment; kept in draft until CI is green.

### DIFF
--- a/docs/examples/hook.pagecontentsavecomplete.md
+++ b/docs/examples/hook.pagecontentsavecomplete.md
@@ -92,7 +92,7 @@ MediaWikiServices::getInstance()->getHookContainer()->register( 'PageContentSave
                 new Container( $containerSemanticData )
         );
 
-        $parserData->pushSemanticDataToParserOutput();
+        $parserData->copyToParserOutput();
 
         return true;
 
--- a/docs/examples/semanticdata.access.md
+++ b/docs/examples/semanticdata.access.md
@@ -44,7 +44,7 @@ foreach ( $values as $value ) {
 }
 
 // Ensures that objects are pushed to the ParserOutput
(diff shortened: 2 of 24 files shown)


paladox:
Some reason ci isn’t running

PR #6856, opened 2026-05-20
Inject services into Job classes via JobClasses ObjectFactory spec

alistair3149:
> **Stacks on #6855.** This PR targets `spike-global-service-registration` so the diff shows only the Jobs-DI changes. Merge #6855 first; this PR then auto-retargets to `master`.

## Summary

Sixth slice of the follow-up to #6827: moving SMW classes off global service-locator dependency access. This is the first framework-constrained slice; SMW `Job` subclasses cannot take arbitrary required constructor parameters because MediaWiki's job runner reconstructs them via `Job::factory()` with only `(Title, $params)`. The MW-canonical solution is `JobClasses` ObjectFactory specs with a `services` array, which #6855 unlocks by registering SMW services on `MediaWikiServices`.

## Pattern

Each `JobClasses` entry in `extension.json` is upgraded to an ObjectFactory spec; each Job's constructor takes those services as required parameters after `(Title, $params)` (ObjectFactory's extraArgs ordering).

`extension.json`:

```json
"smw.refresh": {
    "class": "SMW\\MediaWiki\\Jobs\\RefreshJob",
    "services": [ "SMW.Store" ]
}
```

`RefreshJob`:

```php
public function __construct(
    Title $title,
    array $params,
    Store $store
) {
    parent::__construct( 'smw.refresh', $title, $params );
    $this->setStore( $store );
}
```

`$params` loses its default `= []` because a required parameter follows it; the typed factory methods on SMW's `JobFactory` keep accepting the optional `$parameters = []` and pass `[]` through to MediaWiki's `JobFactory::newJob()` when omitted.

## SMW JobFactory delegation

`src/MediaWiki/JobFactory.php` previously instantiated each Job class directly. It now delegates every typed `newXxxJob()` method to `MediaWiki\JobQueue\JobFactory::newJob()` so ObjectFactory resolves the spec (and any declared services). The wrapper still exists for call-sites that want typed entry points. The `SMW.JobFactory` wiring callback in `src/Services/ServiceWiring.php` constructs `new JobFactory( $services->getJobFactory() )`.

`Job::waitOnCommandLineMode()` re-enqueues through MW's JobFactory rather than `new static(...)` so a job re-enqueued from itself picks up the services declared on its spec.

## Changes

**Jobs converted** (9 in `src/MediaWiki/Jobs/`, 2 in `src/Elastic/Jobs/`):

- `RefreshJob` (Store)
- `FulltextSearchTableUpdateJob` / `FulltextSearchTableRebuildJob` (Store)
- `PropertyStatisticsRebuildJob` (Store)
- `ParserCachePurgeJob` (no services; logger is now lazy via `LoggerAwareTrait`)
- `EntityIdDisposerJob` (Store, IteratorFactory)
- `UpdateDispatcherJob` (Store)
- `UpdateJob` (Store; partial - see below)
- `ChangePropagationDispatchJob` (Store, Cache, PropertySpecificationLookup, IteratorFactory)
- `FileIngestJob` (Store, ElasticFactory)
- `IndexerRecoveryJob` (Store, Cache)

**Dead-scaffolding removals** (test-only): several `registerObject('Store', ...)` calls in `setUp()` were observed to be unbacked by any test assertion; the tests rely on per-test re-registration of more-specific Store mocks.

## Kept as-is

- **Static methods** on `ChangePropagationDispatchJob` (`planAsJob`, `cleanUp`, `hasPendingJobs`, `getPendingJobsCount`) retain direct locator access because instance setters cannot intervene in static calls. The corresponding tests retain their `registerObject` lines.
- **`UpdateJob`** keeps a `$this->applicationFactory ??= ApplicationFactory::getInstance()` cache because it still pulls `newPageCreator`, `newPageUpdater`, `newSerializerFactory`, `newParserData`, `newContentParser`, and `getMediaWikiLogger` from the factory. A class-level comment marks this as a partial DI conversion; a later slice should convert the remaining factory methods after they are registered as services.
- **Transitive `registerObject`** in `UpdateJobTest`, `FileIngestJobTest`, `IndexerRecoveryJobTest`: when an SMW collaborator (for example `ParserData::updateStore`, `Job::insert`) reaches the locator on the Job's behalf, the mock cannot be supplied through the Job's constructor. Documented inline.

## Numbers

`registerObject(` call count across the affected test files: **43 -> 24** (-19); the remaining ones cover transitive collaborators or static-method tests.

## Test plan

- Unit suite (`semantic-mediawiki-unit`) on the stacked base: 7081 tests / 14495 assertions / 0 failures / 6 pre-existing skips.
- Integration `ServiceWiringTest` is extended to resolve every `JobClasses` command through `MediaWikiServices::getJobFactory()`, exercising the ObjectFactory spec end-to-end (14 commands).
- PHPCS clean on all changed files.
- phan and integration suites expected to be covered by CI.

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -202,7 +202,10 @@
 		"WantedProperties": "SMW\\MediaWiki\\Specials\\SpecialWantedProperties"
 	},
 	"JobClasses": {
-		"smw.update": "SMW\\MediaWiki\\Jobs\\UpdateJob",
+		"smw.update": {
+			"class": "SMW\\MediaWiki\\Jobs\\UpdateJob",
+			"services": [ "SMW.Store" ]
+		},
 		"smw.refresh": {
 			"class": "SMW\\MediaWiki\\Jobs\\RefreshJob",
 			"services": [ "SMW.Store" ]
--- a/extension.json
+++ b/extension.json
@@ -243,7 +243,10 @@
 		"smw.changePropagationClassUpdate": "SMW\\MediaWiki\\Jobs\\ChangePropagationClassUpdateJob",
 		"smw.deferredConstraintCheckUpdateJob": "SMW\\MediaWiki\\Jobs\\DeferredConstraintCheckUpdateJob",
 		"smw.elasticIndexerRecovery": "SMW\\Elastic\\Jobs\\IndexerRecoveryJob",
-		"smw.elasticFileIngest": "SMW\\Elastic\\Jobs\\FileIngestJob",
+		"smw.elasticFileIngest": {
+			"class": "SMW\\Elastic\\Jobs\\FileIngestJob",
+			"services": [ "SMW.Store", "SMW.ElasticFactory" ]
+		},
 		"smw.parserCachePurgeJob": "SMW\\MediaWiki\\Jobs\\ParserCachePurgeJob"
 	},
 	"ResourceFileModulePaths": {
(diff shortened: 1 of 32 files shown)


paladox:
This seems to have reduced code coverage quite considerably.

paladox:
@alistair3149 see above

PR #6865, opened 2026-05-21
phan: Fix src/PropertyRegistry.php "PhanTypeMismatchArgumentNullable" errors

paladox:


src/PropertyRegistry.php
```diff
@@ -371,12 +371,12 @@ public function findPropertyIdFromLabelByLanguageCode( $label, $languageCode = '
 	 * @since 2.5
 	 *
 	 * @param string $id
-	 * @param string|null $languageCode
+	 * @param string|null|false $languageCode
 	 *
 	 * @return string
 	 */
 	public function findPreferredPropertyLabelFromIdByLanguageCode( $id, $languageCode = '' ) {
-		if ( $languageCode === false || $languageCode === '' ) {
```
alistair3149, line 379: We should still reject `$languageCode === ''` to preserve the previous behavior

paladox: this is the same tho? Using ! simplified it so if it's false, '' or null it would execute this. Or did you mean something else?

alistair3149: Oops my bad, for some reason I thought `!$languageCode` is a strict check

PR #6872, opened 2026-05-22
Fix empty maintenance scripts list in Special:SemanticMediaWiki

### LINKED ISSUE #6870: There are no scripts listed in Special:SemanticMediaWiki, Maintenance scripts section
### Setup

- SMW version: 7.0.0-alpha (85e4568), but also on 6.0.1
- MW version: 1.43.8
- PHP version: 8.1.32 (fpm-fcgi)
- DB system (MySQL, Blazegraph, etc.) and version: 10.3.39-MariaDB-0ubuntu0.20.04.2 (and Elastic)

### Issue

`Special:SemanticMediaWiki` > Maintenance > maintenance scripts says "The **following** maintenance scripts require an administrator and access to the command line to be able to execute listed scripts.", but there are no scripts listed. I guess this is a bug.

Image

alistair3149:
image

## Summary

The "maintenance scripts" sub-tab of Special:SemanticMediaWiki rendered an empty list. The discovery loop in `MaintenanceTaskHandler::buildHTML()` filtered every script out.

## Cause

MediaWiki core moved `Maintenance` into the `MediaWiki\Maintenance` namespace, with `class_alias( Maintenance::class, 'Maintenance' )` shimming the legacy bare name. `ReflectionClass::getParentClass()->getName()` returns the canonical FQN, never the alias, so the previous check `$reflectionClass->getParentClass()->getName() !== 'Maintenance'` always evaluated true and `continue`d past every Maintenance subclass.

## Fix

Import `MediaWiki\Maintenance\Maintenance` and use `$reflectionClass->isSubclassOf( Maintenance::class )` instead. Drops the now-redundant null-parent guard (`isSubclassOf` returns false for classes with no parent) and naturally accepts indirect subclasses (e.g. via `LoggedUpdateMaintenance`).

Fixes #6870

## Test plan

- [x] PHPCS and PHPUnit (`MaintenanceTaskHandler`, `SpecialAdmin`) pass.
- [x] Reflection probe inside the dev container against a stub `extends MediaWiki\Maintenance\Maintenance` class: old check FAIL, new check PASS.
- [x] Smoke test in a dev wiki: Special:SemanticMediaWiki, Maintenance tab, "maintenance scripts" sub-tab lists the SMW maintenance scripts grouped under "update" / "rebuild" / ungrouped.

### DIFF
--- a/src/MediaWiki/Specials/Admin/MaintenanceTaskHandler.php
+++ b/src/MediaWiki/Specials/Admin/MaintenanceTaskHandler.php
@@ -3,6 +3,7 @@
 namespace SMW\MediaWiki\Specials\Admin;
 
 use MediaWiki\Html\Html;
+use MediaWiki\Maintenance\Maintenance;
 use MediaWiki\Request\WebRequest;
 use ReflectionClass;
 use SMW\Localizer\Message;
@@ -160,9 +161,7 @@ class MaintenanceTaskHandler extends TaskHandler implements ActionableTask {
 
 			$reflectionClass = new ReflectionClass( $class );
 
-			if (
-				!$reflectionClass->getParentClass() ||
-				$reflectionClass->getParentClass()->getName() !== 'Maintenance' ) {
+			if ( !$reflectionClass->isSubclassOf( Maintenance::class ) ) {
 				continue;
 			}
 


jaideraf:
Thanks!

PR #6905, opened 2026-05-28
Remove the SMW Utils\Logger wrapper, use LoggerFactory directly

alistair3149:
## Summary

`SMW\Utils\Logger` was a thin PSR-3 wrapper bundling two behaviours, neither of which carries weight in the current code:

- **A role-based filter** (`ROLE_DEVELOPER` / `ROLE_USER` / `ROLE_PRODUCTION`). The wrapper was always instantiated as `ROLE_DEVELOPER` in production (`ServicesFactory::getMediaWikiLogger( $channel = 'smw', $role = Logger::ROLE_DEVELOPER )`), and `ROLE_DEVELOPER` logs every message regardless of the caller's `'role'` context tag. The filter was inert.
- **Context-value transformations**: rounded `procTime` and `time` floats to five decimals, JSON-encoded any array-valued context entry, and joined array messages with `array_shift . ': ' . json_encode`. No production caller passes array context; the rounding is cosmetic in log output.

The class and its dedicated test are removed. `ServicesFactory::getMediaWikiLogger()` and the `'MediaWikiLogger'` `singleton()` / `create()` dispatch entry are removed too. SMW now resolves its PSR-3 logger directly via `MediaWiki\Logger\LoggerFactory::getInstance( 'smw' )` (or `'smw-elastic'` for the Elastic stack) at every production callsite. The `SMW.Logger` ServiceWiring entry and `ServicesFactory::getLogger()` accessor are unrelated to this PR and remain in place.

## Behaviour notes

- The `'role' => 'user'` / `'role' => 'production'` context tags that ~26 callsites still pass become observational metadata only. Monolog will record them in the log context like any other key. Stripping the tags at 26 sites was kept out of this PR's scope.
- `procTime` and `time` floats logged by callers are no longer pre-rounded to five decimals; downstream formatters see the raw float.
- The array-to-JSON context transformer and the array-message coalescer are gone. Neither has any production caller.

## Tests

- `LoggerTest` is removed (the wrapper is gone).
- `ServicesFactoryTest::testCanConstructMediaWikiLogger` is removed (the accessor is gone).
- `MwCollaboratorFactoryTest::testCanConstructConnectionProvider` no longer sets a `getMediaWikiLogger` mock expectation (the production callsite migrated to `LoggerFactory::getInstance( 'smw' )` directly).
- `DataValueServicesContainerBuildTest::testCanConstruct` drops the now-dead `registerObject( 'MediaWikiLogger', $this->logger )` line. `src/Services/datavalues.php` now resolves the logger through `LoggerFactory::getInstance( 'smw' )` directly, so the test override no longer routes anywhere.

## Test plan

- [x] PHPCS clean on touched files
- [x] `composer phan` exit 0 (caught one missed `Logger::ROLE_DEVELOPER` reference in `ServiceWiring.php`'s `SMW.Store` wiring, corrected to bare `LoggerFactory::getInstance( 'smw' )`)
- [x] `semantic-mediawiki-unit` suite green locally
- [x] Live runtime smoke against the seeded dev wiki: `Main_Page`, `Special:Ask`, `Special:Properties`, `Special:SearchByProperty` all 200, no fatals
- [x] CI matrix (MW 1.43 / 1.44 / 1.45) green


jaideraf:
@alistair3149, as you know, I am not a developer, but I have been here for a long time... I suspect mwjames did all those wrappers in the code in order to make SMW more testable. I understand your goal to bring SMW code closer to MW code, but have you considered why mwjames has made exactly the opposite? Also, could you explain a bit your overall intentions in relation to this release (or SMW in the future)?

alistair3149:
> @alistair3149, as you know, I am not a developer, but I have been here for a long time... I suspect MWJames did all those wrappers in the code in order to make SMW more testable. I understand your goal to bring SMW code closer to MW code, but have you considered why MWJames has made exactly the opposite? Also, could you explain a bit your overall intentions in relation to this release (or SMW in the future)?

Many of the wrappers in SMW were written many years ago when they were actually needed (MW core doesn't have a replacement, or testability requirements). As MW core has became more matured, some of these wrappers exist simply as a passthrough that increases complexity and indirection.

My goal is to make SMW more maintainable. One of the approach is to audit existing MW wrappers to see what can be replaced by MW core, where it is more maintained and better-tested. For each audit, I also look up the commit and issue on why they are implemented (though it usually doesn't provide much context).

As for testability, many of the SMW tests don't really test behavior. They simply exist on assert the function didn't crash and buff line coverage. There are many instances that tests fail while production behavior is correct. The work related to DI is the foundation to make SMW more testable, and perhaps in a distant future, we would have tests that are testing actual behaviors.

jaideraf:
Perfect! Thanks for the answers. 🥇

PR #6919, opened 2026-05-29
Convert EntityCache to MediaWiki BagOStuff

alistair3149:
## What

Continues the migration off the `onoi/cache` dependency to MediaWiki core's `Wikimedia\ObjectCache\BagOStuff`. This converts `SMW\EntityCache`, the wrapper that links cache entries to a WikiPage so they can be invalidated together.

## How

- The injected delegate is retyped from `Onoi\Cache\Cache` to `Wikimedia\ObjectCache\BagOStuff`. Internal calls become `fetch()` to `get()` and `save()` to `set()` (argument order unchanged); `delete()` is kept. The two-argument `set()` in `associate()` relies on `BagOStuff::set()` treating a missing expiry as no expiry, matching the previous `save()` default.
- `contains()` keeps its public boolean signature; its body becomes `get( $key ) !== false`. This is safe at all three external callers, which store non-falsy values: `ProtectionValidator` stores `'yes'`/`'no'`, `DependencyValidator` stores a `saveSub`/`overrideSub` array, and `TableStatisticsTaskHandler` reads a non-empty statistics array.
- `getStats()` is removed; it had no callers, and `BagOStuff` does not expose it.
- The `SMW.EntityCache` service now injects `getObjectCache()` (a `BagOStuff` over the configured main cache type) instead of the shared Onoi composite from `getCache()`. EntityCache namespaces its own keys, so its entries stay separate from the consumers still on the composite.

## Behavioural note

EntityCache no longer shares a backing store with the Onoi composite returned by `getCache()`. Code that wrote through the composite and read back through EntityCache (or vice versa) would no longer see a single shared store. No such coupling exists in the current code; this is noted for completeness as the per-consumer migration proceeds.

## Tests

`EntityCacheTest` is rewritten to characterize behaviour against a real `HashBagOStuff` rather than replaying mock expectations: store/retrieve/contains/delete, the `*Sub` md5 packing (`saveSub` merges into the record, `overrideSub` replaces it), `associate` anchor construction, and the `invalidate` cascade (every associated key and the anchor are deleted while an unrelated entry survives). The rewrite also revives three tests that were silently skipped because of misspelled method names. Two further construction sites that passed an in-memory Onoi cache (`CachedFactboxTest`, `DependencyValidatorTest`) move to `HashBagOStuff`.

This targets the 7.0.0 major release, so the conversion is a hard cutover with no deprecation shims.

### DIFF
--- a/src/EntityCache.php
+++ b/src/EntityCache.php
@@ -3,8 +3,8 @@
 namespace SMW;
 
 use MediaWiki\Title\Title;
-use Onoi\Cache\Cache;
 use SMW\DataItems\WikiPage;
+use Wikimedia\ObjectCache\BagOStuff;
 
 /**
  * Class provides a simple interface the link independent cache entries as
@@ -36,13 +36,11 @@ class EntityCache {
 	const TTL_WEEK = 604800; // 7 * 24 * 3600
 	const TTL_MONTH = 2592000; // 30 * 24 * 3600
 	const TTL_YEAR = 31536000; // 365 * 24 * 3600
-	private Cache $cache;
 
 	/**
 	 * @since 3.1
 	 */
-	public function __construct( Cache $cache ) {
-		$this->cache = $cache;
+	public function __construct( private readonly BagOStuff $cache ) {
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


src/EntityCache.php
```diff
@@ -36,12 +36,12 @@ class EntityCache {
...
 	const TTL_MONTH = 2592000; // 30 * 24 * 3600
 	const TTL_YEAR = 31536000; // 365 * 24 * 3600
-	private Cache $cache;
+	private BagOStuff $cache;
 
 	/**
 	 * @since 3.1
 	 */
-	public function __construct( Cache $cache ) {
+	public function __construct( BagOStuff $cache ) {
```
paladox: I wonder if this can be just `public function __construct( private BagOStuff $cache ) {`

Dropping `$this->cache = $cache;` and `private BagOStuff $cache;` above?

alistair3149: Good catch, I should have promoted it, fixing it now.

PR #6790, opened 2026-05-11
Add PHP 8.5 and MediaWiki 1.46 to CI

alistair3149:
Awaiting 1.46 image from https://github.com/gesinn-it/docker-mediawiki

PR #6953, opened 2026-06-01
7.0.0 release

alistair3149:
> [!CAUTION]
> Only merge this on **June 4, 2026**. If the release date drifts, modify the PR accordingly before merging.

Marks Semantic MediaWiki 7.0.0 as the official stable release, dated **June 4, 2026**.

- Bump `extension.json` version from `7.0.0-alpha` to `7.0.0`
- Set the release date in the 7.0.0 release notes
- Promote the 7.0.x compatibility-matrix row to the current stable release and mark the superseded 6.0.x row obsolete
- Add a parser cache defragmentation highlight to the performance section, summarising the reduced per-language fragmentation and lower eviction of still-useful entries

Held as a draft until the release date.

### DIFF
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -25,16 +25,16 @@ Increases of minimum requirements are indicated in bold.
     </tr>
     <tr>
         <th><a href="https://www.semantic-mediawiki.org/wiki/Semantic_MediaWiki_7.0.0">7.0.x</a></th>
-        <td>Development</td>
-        <td>TBD</td>
-        <td>TBD</td>
-        <td>8.1 - 8.4</td>
-        <td>1.43 - 1.45</td>
+        <td><strong>Stable release</strong></td>
+        <td>2026-06-04</td>
+        <td>2026-06-04</td>
+        <td>8.1 - 8.5</td>
+        <td>1.43 - 1.46</td>
         <td></td>
     </tr>
     <tr>
         <th><a href="https://www.semantic-mediawiki.org/wiki/Semantic_MediaWiki_6.0.0">6.0.x</a></th>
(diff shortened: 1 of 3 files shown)


jaideraf:
> The 1.46.0 stable release is expected to come out in the beginning of June 2026. [0]

What about waiting until MW 1.46 is released AND @gesinn-it-gea makes the MW 1.46 image so we can test against it? Maybe MW 1.46 could be supported, too.

[0] https://www.mediawiki.org/wiki/MediaWiki_1.46
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/6790

alistair3149:
> > The 1.46.0 stable release is expected to come out in the beginning of June 2026. [0]
> 
> What about waiting until MW 1.46 is released AND @gesinn-it-gea makes the MW 1.46 image so we can test against it? Maybe MW 1.46 could be supported, too.
> 
> [0] https://www.mediawiki.org/wiki/MediaWiki_1.46 [1] #6790

1.46 is not a blocker for the 7.0.0 unless we need to make breaking changes to support it.
It is more beneficial to release 7.0.0 stable earlier so that we can gather more feedback and issue reports, and have the downstream extensions updated.

jaideraf:
Right. By the way, it is generally not effective, but @SemanticMediaWiki/testers should test before the release happens, but we can't count on that.

jaideraf:
This fix is not mentioned in the release notes: https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/6941

alistair3149:
> This fix is not mentioned in the release notes: #6941

Added.

alistair3149:
Wiki documentations are updated.

PR #6967, opened 2026-06-03
Drop the redundant single-column o_id index on wikipage tables

alistair3149:
## Benchmark

Synthetic 3M-row `smw_di_wikipage` clone, MariaDB 11.2, 128 MB buffer pool (working set well above RAM). A/B/A controlled, toggling only the standalone `o_id` index on identical data:

| Metric | With redundant `o_id` index | Without it |
| --- | --- | --- |
| Total index storage | 600.7 MB | 532.1 MB (−68.6 MB, −11.4%) |
| Steady-state edit workload (delete a subject's rows + reinsert, 3000 subjects) | ≈56 s | ≈42 s (≈25% faster) |
| `o_id` lookup query plan | covering composite `(o_id,s_id)` / `(o_id,p_id)` | identical, the planner never picked the standalone index even when present |

The standalone `o_id` index (68.6 MB) is essentially the same size as its covering sibling `(o_id,s_id)` (69.6 MB), because the hidden InnoDB row id dominates a single-column secondary index, so it is close to pure duplication. The space and write savings scale with table size; read behaviour is unchanged.

## What

The wikipage data-item tables (`smw_di_wikipage` and the fixed page-property tables such as `smw_fpt_inst`, `smw_fpt_sobj`, `smw_fpt_subp`, `smw_fpt_subc`, `smw_fpt_pplb`, `smw_fpt_ask`) carry an auto-generated single-column object index on `o_id`. That index is a left-prefix of the composite indexes `(o_id,s_id)` and `(o_id,p_id)` the same tables already define, so it serves no lookup the composites cannot, while adding per-row write maintenance and index storage.

This drops the standalone `o_id` index on those tables. `TableSchemaManager::newPropertyTable()` now skips the single-column object ("po") index when a declared composite index already begins with that column. The rule is general but only affects the wikipage tables in practice: no other data-item type (blob, number, time, uri, and so on) declares a composite that leads with its index field, so their object index is preserved. The redirect table `smw_fpt_redi` is also unaffected, because its `o_id` index is not a prefix of any composite there.

## Why it is safe

- Every `o_id` read path (in-property lookup, category and instance joins, subproperty/subcategory hierarchy fills, prefetch and inverse lookups) filters `o_id` by equality, range, or `IN`, which a composite leading with `o_id` serves identically as its leftmost prefix while also covering the partner column.
- No `FORCE INDEX` / `USE INDEX` hint names the `o_id` index. The existing `s_id` index hint refers to the separate `(s_id,p_id)` index and is left untouched.
- The planner behaviour above confirms the standalone index was never selected; EXPLAIN plans and row estimates are identical with and without it.

## Migration

The index is removed automatically during `update.php` / store setup, through the existing obsolete-index diff in the table builders (MySQL/MariaDB, PostgreSQL, SQLite). No manual action is required.

## Scope

A second index, `(s_id,p_id)`, is also a redundant prefix of `(s_id,p_id,o_id)`, but the index MySQL auto-names `s_id` is exactly that `(s_id,p_id)` index, and that name is pinned by an existing `USE INDEX(s_id)` hint (`PropertySubjectsLookup`, plus the blob/number/uri handlers). Removing it would break the hint, so it is handled separately and not part of this change.

Refs #6559


src/SQLStore/TableBuilder/TableSchemaManager.php
```diff
@@ -393,6 +393,26 @@ private function newPropertyTable( $propertyTable, $diHandler ): Table {
...
 
+		// Drop the auto-generated single-column object index ("po") when a
+		// declared composite index already begins with that same column. In
+		// that case the standalone index is a redundant left-prefix: the
+		// optimizer satisfies every lookup from the composite, which also
+		// covers the trailing column, so keeping it only adds write and storage
+		// overhead. This affects the wikipage data-item tables, whose handler
+		// declares (o_id,s_id) and (o_id,p_id). See issue #6559.
+		if ( isset( $indexes['po'] ) ) {
+			$objectField = $indexes['po'];
```
paladox: Maybe use this variable outside the if statement so you can also use it in the if statement?

alistair3149: Done

Issue #6990, opened 2026-06-09
Error 1137: Can't reopen table 'p' on MySQL 8 with temporary tables (integration tests)


gesinn-it-gea:
## Summary

After PR #6936 removed the `use-normal-tables` flag (restoring temporary-table isolation in integration tests), the full integration suite fails on **MySQL 8** with:

\`\`\`
Wikimedia\Rdbms\DBQueryError: Error 1137: Can't reopen table: 'p'
Function: SMW\SQLStore\EntityStore\SemanticDataLookup::fetchSemanticDataFromTable
Query: SELECT p.smw_title AS `prop`, o_id AS `id0`, o0.smw_title AS `v0`, ...
  FROM `unittest_smw_di_wikipage` `smw_di_wikipage`
  JOIN `unittest_smw_object_ids` `p`  ON ((p_id=p.smw_id))
  JOIN `unittest_smw_object_ids` `o0` ON ((o_id=o0.smw_id))
  WHERE s_id = 4 AND (p.smw_iw != ':smw') AND (p.smw_iw != ':smw-delete')
\`\`\`

## Root cause

MySQL 8 raises Error 1137 when the same **temporary** table is opened more than once in a single query. The query above joins `smw_object_ids` twice under different aliases (`p` for the property name, `o0` for the object value). This is legal SQL and works fine on real (non-temporary) tables and on MariaDB, but fails on MySQL 8 when the table is a temporary table created by the test framework.

The double join is constructed in `SemanticDataLookup::fetchFromTable()`:
- Line 558: `$qb->join( SQLStore::ID_TABLE, 'p', 'p_id=p.smw_id' )` (property alias)
- Line 779: `$qb->join( SQLStore::ID_TABLE, "o$valueCount", "$fieldname=o$valueCount.smw_id" )` (value alias)

## Reproduction

Run the SMW integration test suite against **MySQL 8** (not MariaDB):

\`\`\`
DB_IMAGE=mysql:8 make install composer-phpunit
\`\`\`

Virtually every integration test that touches `SemanticDataLookup` fails.

## History

- **2014 — PR #80**: The original fix for this exact error was to force `use-normal-tables = true` in `SMWIntegrationTestCase`, avoiding temporary tables on MySQL.
[...]

PR #7001, opened 2026-06-24
Use normal tables for MySQL integration tests to avoid Error 1137

alistair3149:
## Problem

The experimental `mysql:8` CI leg (added in #6990) fails with hundreds of `Error 1137: Can't reopen table`. MySQL, unlike MariaDB, cannot reference a `TEMPORARY` table more than once in a single query, and SMW lookups self-join `smw_object_ids`. Since #6936 switched the integration tests to clone their tables as temporary tables (to fix the recurring teardown lock-wait timeout, `1205`), every self-joining lookup now errors on MySQL. MariaDB lifted this restriction years ago, so its CI legs are unaffected.

## Fix

Force normal (persistent) tables for the integration tests on genuine MySQL only, selectively restoring the pre-#6936 behaviour. MediaWiki reports both MySQL and MariaDB as the `mysql` database type, so the server is distinguished by its version string, matching core's own discriminator (`MariaDB` / `-maria-`). MariaDB legs keep temporary tables, so the lock-wait timeout #6936 addressed does not return for them.

The flag is set in `setUpBeforeClass()` because it must be in place before MediaWiki's `maybeSetupDB()` reads it.

## Validation

On a MySQL 8.0.46 stack, a JSONScript case that errored with `1137` on master passes with this change, and a multi-class SQLStore integration run completes cleanly. MariaDB behaviour is unchanged.

This is one of two changes needed for the experimental `mysql:8` leg to pass; the companion fixes an unrelated result-ordering issue in `PropertySubjectsLookup`.

### DIFF
--- a/tests/phpunit/SMWIntegrationTestCase.php
+++ b/tests/phpunit/SMWIntegrationTestCase.php
@@ -36,6 +36,42 @@ abstract class SMWIntegrationTestCase extends MediaWikiIntegrationTestCase {
 
 		// Load default settings specific to SMW
 		TestEnvironment::loadDefaultSettings( [ 'smwgQEqualitySupport' ] );
+
+		// MySQL (unlike MariaDB) cannot reference a TEMPORARY table more than
+		// once in a single query, so SMW lookups that self-join smw_object_ids
+		// fail with "Error 1137: Can't reopen table" when MediaWiki clones the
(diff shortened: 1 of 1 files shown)


alistair3149:
MW core ticket for reference: https://phabricator.wikimedia.org/T256006

PR #7010, opened 2026-07-06
Fix infinite datatable loading when DPL is on the same page

### LINKED ISSUE #7009: Dynamic Page List breaks datatable query
### Setup

- SMW version: master/7

### Issue

List of steps to reproduce (step by step, including full links if applicable):
To reproduce, try this code:
{{#ask: [[:+]]
|?Modification date
|format=table
|class=datatable
|limit=5
}}

{{#dpl:
|namespace=
}}

What happens?:
If you have a semantic query with datatable format and a DPL query on the same page, the JS for displaying the datatable loads infinitely.

What should have happened instead?:
SMW Datatable loads

HamishSlater:
When a page contains both an SMW datatable query and a DPL query, DPL's triggerEndResets() creates a nested Parser::parse() call which triggers SMW's InternalParseBeforeLinks hook. This calls Outputs::commitToParserOutput() on the nested parser's output, draining the static module buffers. The smw.tableprinter.datatable module is flushed to a discarded ParserOutput and never reaches the main page, causing the datatable spinner to load infinitely.

Change commitToParserOutput() to preserve the static buffers by default so modules survive nested parse calls and are re-committed to the top-level ParserOutput. ParserOutput::addModules() deduplicates, so re-adding the same modules is safe. commitToOutputPage() still clears the buffers as before.

Fixes: #7009

### DIFF
--- a/src/MediaWiki/Hooks/ParserAfterTidy.php
+++ b/src/MediaWiki/Hooks/ParserAfterTidy.php
@@ -9,6 +9,7 @@ use MediaWiki\Parser\ParserOutputLinkTypes;
 use MediaWiki\Permissions\RestrictionStore;
 use Psr\Log\LoggerInterface;
 use SMW\DataModel\SemanticData;
+use SMW\MediaWiki\Outputs;
 use SMW\NamespaceExaminer;
 use SMW\ParserData;
 use SMW\Services\ServicesFactory as ApplicationFactory;
@@ -132,6 +133,7 @@ class ParserAfterTidy implements ParserAfterTidyHook {
 	 */
 	public function onParserAfterTidy( $parser, &$text ) {
 		if ( !Site::isReady() ) {
+			Outputs::onParseEnd();
 			$this->doAbort();
 			return true;
 		}
@@ -162,6 +164,8 @@ class ParserAfterTidy implements ParserAfterTidyHook {
 			if ( self::$inFlightParses !== null ) {
 				unset( self::$inFlightParses[$parser] );
 			}
+
+			Outputs::onParseEnd();
 		}
 
 		return true;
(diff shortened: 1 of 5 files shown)


alistair3149 (commented):
Thanks for tracking this down. The diagnosis is spot on and this fixes the reported spinner. One thing to resolve before merge though: as written, it swaps the DPL bug for a quieter regression in batch parsing.

**The buffers are now never cleared on the parse path.**

`commitToParserOutput()` defaults `$clear = false`, and no caller ever passes `true` (`InTextAnnotationParser` and `commitToParser()` both omit it). So the static `Outputs` buffers are now only cleared by `commitToOutputPage()`, which runs on SMW special pages, never on normal article parses or jobs. Because those buffers are process global, any process that parses more than one page (the job queue, `refreshLinks.php`, `rebuildData.php`) carries the modules required by one page into every page parsed after it, and that lands in the parser cache.

Reproduced on a clean MW 1.43 + SMW install by parsing two pages in one process, swapping only `Outputs.php` between master and this branch:

master:
```
Page A (datatable ask): [ext.smw.styles, smw.tableprinter.datatable]
Page B (plain text):    [ext.smw.styles]
```
this PR:
```
Page A (datatable ask): [ext.smw.styles, smw.tableprinter.datatable]
Page B (plain text):    [ext.smw.styles, smw.tableprinter.datatable]   <-- leaked
```

Page B has no semantic markup, yet its cached `ParserOutput` picks up page A's datatable module. In a rebuild run, the last page parsed inherits every module required by every page before it. The dedup point in the description holds for re-committing to the same page, but not for this cross-page case.

**Suggested direction:** keep the buffer alive through nested parses as you do now, but clear it once the outermost parse finishes. A recursion counter inside the `InternalParseBeforeLinks` handler won't work: DPL's nested `Parser::parse()` fires and returns before the top-level hook runs, so the handler never sees a depth greater than one. Counting parse start and end across the whole lifecycle, and clearing only at the outermost level, does work, since the top-level `ParserOutput` has already received the modules by the time that parse ends. A test in both directions (survives a nested re-commit, does not leak into the next page) would lock it down.

Happy to help with the follow-up if that's useful.

alistair3149 (changes requested):

PR #7018, opened 2026-07-07
Support negated property existence in SPARQL queries

hkwi:
## Summary
- Translate `[[Property::!+]]` in the SPARQLStore query engine to a `FILTER NOT EXISTS` property pattern.
- Preserve inverse property handling and keep subproperty hierarchy expansion inside the `NOT EXISTS` block.
- Add unit coverage using dummy `Foo` properties for direct, inverse, and hierarchy cases.

## Rationale
The query parser already represents `!+` as a negated `ThingDescription`. The SPARQLStore previously treated that inner description the same as `+`, causing `[[Property::!+]]` to still require a matching property triple. This was not a limitation of SPARQL itself; it was a missing translation in `SomePropertyInterpreter`.

## Tests
- `php -l src/SPARQLStore/QueryEngine/DescriptionInterpreters/SomePropertyInterpreter.php`
- `php -l tests/phpunit/Unit/SPARQLStore/QueryEngine/DescriptionInterpreters/SomePropertyInterpreterTest.php`
- `git diff --check`

Full targeted PHPUnit was attempted through the Docker test environment, but the local build stalled on Packagist access during Composer dependency resolution.

### DIFF
--- a/src/SPARQLStore/QueryEngine/DescriptionInterpreters/SomePropertyInterpreter.php
+++ b/src/SPARQLStore/QueryEngine/DescriptionInterpreters/SomePropertyInterpreter.php
@@ -196,14 +196,33 @@ class SomePropertyInterpreter implements DescriptionInterpreter {
 			return "$subjectName $propertyName $objectName .\n";
 		}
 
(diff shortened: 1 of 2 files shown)


alistair3149:
Thanks for the PR!

PR #7032, opened 2026-07-10
Store the --auto-recovery checkpoint in the database

alistair3149:
## Problem

In 7.0.0 (#6892) the install-state metadata that used to live in the `.smw.json` file moved into the `smw_meta` database table, and `$smwgConfigFileDir` was deprecated. That migration did not cover the `--auto-recovery` code path: `AutoRecovery` still wrote a `.smw.json` file to `$smwgConfigFileDir`.

On a deployment where that directory is not writable by the maintenance user (containerized or read-only image, root-owned extension files), any rebuild script run with `--auto-recovery` aborted with `FileNotWritableException` before doing any work. The failure happened during the pre-run abort check, because `AutoRecovery::has()` (a read) created the file as a side effect.

## Fix

`AutoRecovery` now stores its checkpoint in the `smw_meta` table instead of a file, so it no longer depends on a writable `$smwgConfigFileDir`, and the read path (`has`/`get`) never writes.

Because `smw_meta` is otherwise owned by `SetupFile`, which persists install state with a full-slice sync (every write deletes any key not present in its in-memory snapshot), the checkpoint is kept deliberately separate from that slice:

- `DatabaseMetaRepo` treats the auto-recovery key as a reserved, out-of-band key: it is excluded from the install-state slice returned by `loadSmwJson`, and exempted from `saveSmwJson`'s sync-delete, so an ordinary install-state write never removes it.
- The checkpoint is read and written one row at a time through dedicated single-row `readValue`/`writeValue` methods. The per-entity checkpoint write is therefore a single-row upsert that cannot rewrite the whole slice and cannot clobber install-state keys (`upgrade_key`, `maintenance_mode`, ...) through a stale snapshot. The read comes from the primary so a resuming run always sees the checkpoint its previous run wrote.

No schema change and no `update.php` run are required; the `smw_meta` table already exists since 7.0.0.

## Testing

- Unit coverage for the reserved-key exclusion and exemption, the single-row read/write path, and the rewritten `AutoRecovery` (including a guard that the read path never writes).
- A database integration test covering the checkpoint round-tripping through `smw_meta`, distinct maintenance scripts not colliding, the read path not creating a row, and the checkpoint surviving an install-state write without leaking into the install-state slice.
- Verified live: `rebuildData --auto-recovery` on a non-writable extension directory now completes without error, stores its checkpoint in `smw_meta`, creates no `.smw.json`, and leaves the install-state rows intact.

### DIFF
--- a/docs/releasenotes/RELEASE-NOTES-7.2.0.md
+++ b/docs/releasenotes/RELEASE-NOTES-7.2.0.md
@@ -14,6 +14,7 @@ For more detailed information, see the [compatibility matrix](../COMPATIBILITY.m
 ## Bug fixes
 
 * Fixed a query showing incorrect printout values when the same property is requested through different printout contexts, such as a direct and an inverse printout or the same property with different sort options ([#7026](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/7026))
+* Fixed maintenance scripts run with `--auto-recovery` (such as `rebuildData` and `rebuildElasticIndex`) aborting with a `FileNotWritableException` on deployments where the extension directory is not writable; the recovery checkpoint is now stored in the database instead of a `.smw.json` file, so it no longer depends on a writable `$smwgConfigFileDir` ([#7030](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7030))
 
 ## Upgrading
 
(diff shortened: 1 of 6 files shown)


paladox:
@alistair3149 I'm not sure why the change that was merged is different from the diff viewing this pull.

PR #7036, opened 2026-07-15
Flush buffered stats periodically in long-running maintenance scripts

malberts:
[MediaWiki's stats library](https://www.mediawiki.org/wiki/Manual:Stats) buffers a sample in process memory on every metric emit. Web requests flush the buffer when they end, but maintenance processes never do on MediaWiki 1.43. MediaWiki 1.44 added periodic CLI flush points in Maintenance::output() and the replication waits ([T380609](https://phabricator.wikimedia.org/T380609)), but that fix was not backported to the 1.43 LTS, and those flush points fire only as a side effect of progress printing and core-managed transaction waits, which several SMW rebuild loops do sparsely or not at all. Long entity sweeps therefore grow the buffer without bound: rebuildData retains roughly 115 samples (about 15 KB) per processed entity, which exhausts a 1 GB memory limit after about 65k entities and makes full rebuilds impossible on large wikis regardless of the limit.

Older versions are unaffected in practice: on MediaWiki 1.39 the buffered stats library existed only as the experimental Metrics predecessor, with no emitters on the parser and object cache hot paths, and the legacy statsd buffer is cleared unconditionally at the periodic maintenance flush points from [T181385](https://phabricator.wikimedia.org/T181385). The exposure arrived as core migrated its metrics onto the modern stats library ([T350592](https://phabricator.wikimedia.org/T350592)): the parser cache counters in MediaWiki 1.42, and the object cache and parser output access layers in 1.43, which is what pushed per-entity accumulation to the scale measured here.

A failing run dies with the generic OOM fatal at whichever allocation happens to cross the limit; observed frames include the stats recording path itself, Monolog timestamp creation while logging during the shutdown, and SMW's own cache serializer:

> PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 4096 bytes) in /var/www/html/w/includes/libs/Stats/Metrics/BaseMetric.php on line 193
>
> PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 32768 bytes) in /var/www/html/w/vendor/monolog/monolog/src/Monolog/DateTimeImmutable.php on line 1
>
> PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 131072 bytes) in /var/www/html/w/extensions/SemanticMediaWiki/src/Utils/HmacSerializer.php on line 138

Add PeriodicStatsFlusher, which drains the buffer every 100 loop iterations, and wire it into every maintenance loop whose iteration count scales with wiki size. Maintenance scripts construct it and inject it into their worker classes via setter, so web requests are unaffected; with no stats target configured the flush is a near-free no-op.

The periodic flush also composes safely with the core flush points on MediaWiki 1.44 and later: a flush drains and clears the shared buffer, so each sample is still emitted exactly once, by whichever flush point fires first, and a flush that finds an empty buffer does next to nothing (1.45 short-circuits it explicitly). The per-loop flush still earns its place on those versions, covering the loops whose reporting pattern never passes through Maintenance::output().

Measured on a 30k-entity wiki (MediaWiki 1.43): rebuildData memory use drops from 511 MB (about 15 KB per entity, unbounded) to 66 MB (flat). At a 512M limit on the same wiki, the unpatched script dies a third of the way through the sweep while the patched script completes it.

> <sub>AI-authored — Claude Code, `Fable 5 (max)`; investigation and workflow directed by @malberts, ~10 steering rounds after PR creation; independently AI-reviewed after publication (max-effort pass, zero blockers), human review pending; TDD unit test run against real MediaWiki 1.43 classes, end-to-end verified on a seeded 30k-entity MediaWiki 1.43.9 wiki (unpatched run OOMs, patched run completes using 66 MB), CI green on the reviewed head and re-running on the final one.</sub>

<details><summary>Production notes</summary>

Design and review by `Fable 5 (max)`; implementation and measurement runs by `Opus 4.8` subagents. Two independent adversarial AI review passes (correctness/completeness and project-conventions lenses) preceded this PR; their findings are reflected in the diff. Verification included a direct A/B at a 512M memory limit on identical warm-cache state: the unpatched code dies with the OOM fatal at entity ~9,995 of 30,537, the patched code completes all entities reporting 66 MB used. A further independent review pass after publication (six read-only passes including mutation testing and a live re-verification) found zero blockers; its one advisory, dropping the configurable flush interval in favour of a private constant, is applied.

</details>

### DIFF
--- a/docs/releasenotes/RELEASE-NOTES-7.2.0.md
+++ b/docs/releasenotes/RELEASE-NOTES-7.2.0.md
@@ -17,6 +17,7 @@ For more detailed information, see the [compatibility matrix](../COMPATIBILITY.m
 * Fixed maintenance scripts run with `--auto-recovery` (such as `rebuildData` and `rebuildElasticIndex`) aborting with a `FileNotWritableException` on deployments where the extension directory is not writable; the recovery checkpoint is now stored in the database instead of a `.smw.json` file, so it no longer depends on a writable `$smwgConfigFileDir` ([#7030](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7030))
 * Fixed viewing an old revision failing with a "This ParserOutput contains no text!" error on wikis running an extension that adds parser output metadata to the page, such as PageNotice ([#7033](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7033))
 * Fixed `Special:Ask` returning an internal error when a query is run with `format=debug` ([#7035](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/7035))
+* Fixed long-running maintenance scripts (such as `rebuildData`, `rebuildFulltextSearchTable` and `dumpRDF`) exhausting memory on large wikis; MediaWiki's buffered stats samples, which accumulate in process memory for every processed entity and have no flush point on the command line, are now flushed periodically during the rebuild loops ([#7036](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/7036))
 
 ## Upgrading
 
--- a/maintenance/disposeOutdatedEntities.php
+++ b/maintenance/disposeOutdatedEntities.php
@@ -3,6 +3,7 @@
 namespace SMW\Maintenance;
 
 use MediaWiki\Maintenance\Maintenance;
+use MediaWiki\MediaWikiServices;
 use Onoi\MessageReporter\CallbackMessageReporter;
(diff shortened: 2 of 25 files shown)


malberts:
Just FYI, on the wiki where this was tested (running RebuildData), memory exhaustion went from 1GB in 17 minutes, to 1GB in 3:17 hours.

Memory usage before:
Image

After:
Image

The job still died for other reasons, but there is a definite improvement in memory usage.

Issue #5624, opened 2024-04-12
Enhance #set: parser function with display option


krabina:
Since the `#set:` parser function is not a 100% replacement for the inline annotation` [[Foo::Bar]]`, we should need to move forward in this direction. Therefore I propose to add a display functionality to the parser function.

### Background
In this issue, I am not referring to the problem of SMW installations that generally use the inline annotation to annotate content in wikitext using the wikitext editor. For this, see [#4727](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4727)

In the use case I am addressing, semantic values are set in a template, which is the general use case for many SMW installations. It is generally not a problem to avoid inline annoations in templates: 
Instead of putting 
`Your Property: [[Foo::Bar]]` in your template (maybe for a fact box of a wiki article), it is easy to do something like
`Your Property: {{{Bar|}}}{{#set:Foo={{{Bar|}}} }}`
This approach also has the advantage, that in your template you can control the output of the value `{{{Bar|}}}` anyway you like and set the property with the silent `{{#set:Foo={{{Bar|}}} }}` annotation.

### Problem
The problem, however, is that in the set parser function is not a complete replacement for the inline annotation. These are some cases (there are certainly more):
- Datatype External Identifier will render a link that points to the external formatter URI. You cannot achieve this with `[[{{{Bar|}}}]] `in the template. This will render an internal link instead of the link to the external URI.
- Datatype Keyword can be used with a [keyword link formatter rule](https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/3019), which also works only with inline annotations.
- The [property link feature](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1855) only works with inline annotations: `[[Your Property::@@@]]`
- It is cumbersome to annotate Property Pages with the silent annotation, since you usually want have something displayed there. My workaround is this, e. g. for a page Property:Link

```
* [[Has type::@@@]]: {{#show:{{FULLPAGENAME}}|?Has type}}
* [[Has preferred property label::@@@]]]: {{#show:{{FULLPAGENAME}}|?Has preferred property label}}
* [[Has property description::@@@]]: {{#show:{{FULLPAGENAME}}|?Has property description}}

{{#set:
|Has type=Url
|Has preferred property label=Link@de
|Has property description=Externer Link@de
|Has property description=External Link@en
}}
```
### Solution
[...]

PR #7039, opened 2026-07-16
Add the #property_link parser function

kghbln:
For https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5624

Adds a `{{#property_link:}}` parser function that renders a link to a property page, as a dedicated-syntax alternative to the `[[Foo::@@@]]` in-text annotation syntax (https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1855). Intercepting standard `[[…]]` links for annotations is incompatible with where the MediaWiki parser is going (Parsoid), so property links need a syntax that does not depend on it.

- `{{#property_link:Foo}}` renders exactly what `[[Foo::@@@]]` renders
- `{{#property_link:Foo|custom label}}` renders exactly what `[[Foo::@@@|custom label]]` renders, including the `|#` no-link variant

Parity is by construction: the `@@@` link building moved from a private `InTextAnnotationParser` method into a shared `PropertyLinkRenderer` used by both syntaxes. The legacy `@@@` syntax is unchanged.

Notes:

- The `@@@<lang>` language-annotation variant has no parser-function equivalent in this first version.
- The JSONScript cases first pin the legacy behavior (p-0212 additions and the legacy half of p-0466), then mirrored `{{#property_link:}}` cases assert identical strings as the parity proof (p-0215, and p-0466 for a non-English content language). Newly pinned edge cases: nonexistent properties render red links; invalid or empty property names render nothing (both syntaxes agree).
- No new user-facing messages; the i18n surface is the new `property_link` magic word.
- The regenerated JSONScript README index also catches up on a few pre-existing stale entries.


JeroenDeDauw (approved):

JeroenDeDauw:
I tested this locally and it looks good to me

Issue #7045, opened 2026-07-17
No JavaScript/QUnit tests actually run — CI wiring and test infrastructure gaps


gesinn-it-gea:
## Summary

While adding QUnit test coverage for #7044 (client-side backoff for the `@annotation` purge-reload loop), it turned out that **no JavaScript/QUnit test has ever actually executed in this repository**, neither in CI nor via the standard MediaWiki `Special:JavaScriptTest/qunit` runner. Several independent gaps stack up to cause this:

### 1. CI never runs JS tests at all

`.github/workflows/main.yml` only wires up PHP tooling (`composer analyze:ci`, `composer lint`, `composer phpunit` via the `Test`/`Static analysis`/`Coverage` jobs — see lines 60-105). There is no `npm`/`qunit` step anywhere in the workflow. Confirmed against recent runs, e.g. `gh run view 29539446344 --repo SemanticMediaWiki/SemanticMediaWiki --json jobs` lists only PHP-related jobs (`Test (MW ...)`, `Static analysis (MW ...)`, `Coverage (MW ...)`), and `gh run view <id> --log` contains zero references to `npm`/`qunit`.

### 2. `make npm-test` is a silent no-op locally

`build/Makefile`'s `npm-test`/`npm-test-coverage` targets are gated behind `ifdef NODE_JS`, and `NODE_JS?=true` is commented out in the root `Makefile` (line ~26). There is also no `package.json` at the repository root. So even a contributor who explicitly runs `make ci` locally never executes any JS test.

### 3. `extension.json`'s `QUnitTestModule` was structurally invalid

Before this was fixed on branch `fix/7044-purge-reload-backoff` (see #7044), the SMW QUnit test list was nested as:

```json
"QUnitTestModule": {
  "ext.smw.tests": { "scripts": [...], "dependencies": [...], ... }
}
```

[...]

PR #7046, opened 2026-07-17
Add client-side backoff and termination guard to page-purge reload

### LINKED ISSUE #7044: `@annotation`-driven page-purge reload has no client-side backoff or termination guard (can loop indefinitely / flickers)
## Summary

Pages with a `#ask ... |@annotation` query and `$smwgEnabledQueryDependencyLinksStore = true` can trigger a client-side reload loop via `PostProcHandler::getHtml()` + `res/smw/util/ext.smw.util.purge.js`. Every view that is not a post-edit request and for which `DependencyValidator::hasLikelyOutdatedDependencies()` returns `true` emits a `<div class="smw-postproc page-purge" data-msg="smw-purge-update-dependencies">`. `ext.smw.util.purge.js` finds any `.page-purge` element unconditionally and calls `action=purge` followed by an immediate `location.reload()` (`res/smw/util/ext.smw.util.purge.js:22-47,60-62`).

This reload has no client-side rate limiting:
- No counter, no delay, no backoff between attempts.
- If the server keeps emitting the purge div on every reload (e.g. because the referencing query's own `_ASK` entity never receives a fresh `smw_touched` in that update cycle, or because the update job backlog takes longer than a few reloads), the browser reloads the page repeatedly and immediately, producing a visible "flicker" loop with no upper bound.

## Where this differs from the related, already-fixed mechanism

#6882 repaired the *other* dependency-driven parser-cache rejection path (the one that runs without `@annotation`, via MediaWiki's normal `RejectParserCacheValue` flow) and explicitly scoped the `@annotation` write-back path as "a separate mechanism, separate question." This issue is that separate question — it only concerns the `@annotation`/`PostProcHandler`/`ext.smw.util.purge.js` path. `markTitle()`/`hasLikelyOutdatedDependencies()` remain in use here by design (per #6882's "Out of scope" note) — this issue is not proposing to consolidate the two mechanisms, only to harden the existing one.

## Proposed changes

1. **Test coverage first.** There is currently no QUnit coverage for `ext.smw.util.purge.js` (it is not registered under any `qunit` ResourceLoader test module in `extension.json`, unlike `ext.smw.api`, `ext.smw.query`, etc.). Before changing behavior, add QUnit tests covering the current `purge()` flow (click-triggered purge, `.page-purge` auto-trigger, success/fail paths) so the reload-loop fix has a regression baseline. Also check whether `PostProcHandlerTest`/`DependencyValidatorTest` exercise the repeated-purge-on-repeated-view scenario on the PHP side, and extend if not.
2. **Client-side backoff.** Instead of an immediate `location.reload()` on every `.page-purge` encounter, track retry state client-side (e.g. via `sessionStorage`, keyed by page/title) and wait an increasing interval between reloads (e.g., roughly 1s → 3s → 7s → ...) instead of reloading back-to-back. Exact sequence/formula left to implementation.
3. **Termination guard.** Cap total retry time at a fixed ceiling (proposed: 2 minutes). Once exceeded, stop reloading — do not treat this as an error state (the underlying data usually does arrive eventually via the job queue), but surface a low-key notice (similar to the existing `mw.notify(..., { type: 'info' })` calls already used in this file) telling the user that the update is taking longer than expected and a manual purge later may help.

## Scope

- In scope: `res/smw/util/ext.smw.util.purge.js`, its QUnit coverage, and PHP-side test coverage for the repeated-purge scenario in `PostProcHandler`/`DependencyValidator`.
- Out of scope: consolidating this mechanism with the one repaired in #6882.

## References

- #6882 (repaired the sibling non-`@annotation` cache-rejection path; explicitly excluded this one)
- #4974 (related but distinct: store-update timing without this reload mechanism)

gesinn-it-gea:
## Summary

The `@annotation`-driven page-purge reload (`res/smw/util/ext.smw.util.purge.js`) reloaded immediately and unconditionally on every `.page-purge` element found in the DOM, with no client-side rate limiting. If the server kept emitting the purge div across successive reloads — e.g. because the referencing query's `_ASK` entity hadn't received a fresh `smw_touched` yet, or the update job backlog hadn't caught up — the browser would reload repeatedly and immediately, producing a visible, unbounded "flicker" loop.

This mirrors #6882, which fixed the *sibling* non-`@annotation` cache-rejection path and explicitly scoped this `@annotation`/`PostProcHandler`/`purge.js` path as a separate, out-of-scope mechanism. This PR is that separate fix. `DependencyValidator::markTitle()`/`hasLikelyOutdatedDependencies()` remain in use by design, unchanged — only the client's reload behavior changes.

## Changes

- **`res/smw/util/ext.smw.util.purge.js`**: exposes its logic under `mw.smw.purge` (mirroring the existing `mw.smw.localtime` pattern) so it can be unit tested. Adds:
  - `getRetryState`/`setRetryState`/`clearRetryState` — per-title retry bookkeeping in `mw.storage.session`.
  - `computeBackoff` — an increasing delay sequence (1s → 3s → 7s → 15s → 30s, then plateauing) between auto-reload attempts.
  - `nextAttempt` — decides whether another auto-reload is still within a fixed 2-minute retry ceiling.
  - `reload` — a small indirection so tests can stub the actual `location.reload()` call.
  - `run()` now applies the above only for the auto-triggered (`.page-purge`) path; a user-initiated click-to-purge (`#ca-purge`/`.purge`) still reloads immediately, unchanged.
  - Once the retry ceiling is exceeded, the client stops reloading and shows a low-key `mw.notify(..., { type: 'info' })` notice (new message `smw-purge-update-dependencies-timeout`) instead of treating it as an error — the data usually does arrive eventually via the job queue.
- **`tests/qunit/smw/util/ext.smw.util.purge.test.js`** (new): QUnit coverage for the retry-state helpers, the backoff sequence, the retry-budget decision, and `run()`'s click-triggered vs. auto-triggered behavior (including the termination-guard path).
- **`tests/phpunit/Unit/PostProcHandlerTest.php`** / **`tests/phpunit/Unit/DependencyValidatorTest.php`**: new tests documenting the server-side condition that motivates the client fix — `hasLikelyOutdatedDependencies()` has no reset/expiry by design, so it (and therefore the emitted `.page-purge` div) stays `true` across as many successive requests/reloads as it takes for the underlying dependency to actually become fresh. Bounding the number of reloads is therefore solely the client's responsibility.
- **`i18n/en.json`** / **`i18n/qqq.json`**: new `smw-purge-update-dependencies-timeout` message.
- **`extension.json`**: fixes `QUnitTestModule` being structurally invalid — it was wrapped in a spurious `"ext.smw.tests"` key instead of the module definition sitting directly under `QUnitTestModule` (MediaWiki's `ExtensionProcessor` auto-derives the RL module name as `test.{extension-name}` and only recognizes the un-wrapped form), and it depended on a non-existent `"ext.smw.data"` module. This was a necessary prerequisite for the new QUnit tests (and, it turns out, every pre-existing QUnit test in this repo) to load and run at all — see **#7045** for the full detail on the broader JS-test-infrastructure gap this surfaced (CI never runs JS tests, no `package.json`, `NODE_JS` disabled in the `Makefile`, plus a few further pre-existing, unrelated QUnit failures now visible for the first time).

## Test plan

- [x] `composer test` (lint, phpcs, phan, phpunit:unit, phpunit:integration) passes in a local `make install`'d container — 7319/7319 unit tests pass (7 pre-existing, unrelated skips).
- [x] New `PostProcHandlerTest`/`DependencyValidatorTest` cases pass.
- [x] New QUnit tests verified locally via Docker + Playwright against `Special:JavaScriptTest/qunit?component=SemanticMediaWiki`: the pure retry-state/backoff/budget logic (8 of the 12 new tests) passes directly; the remaining 4 `run()` tests initially failed only because `window.location.reload` is non-reassignable in that environment, fixed by introducing the `mw.smw.purge.reload` indirection point (a test-mockability change, not a logic change).
- [ ] Manual verification of the actual flicker-loop scenario (`#set +@annotation` query + `$smwgEnabledQueryDependencyLinksStore = true` + a persistently-outdated dependency) is left to reviewers with a live wiki, since reproducing the update-job-backlog timing reliably in an automated test is impractical.

Closes #7044
Follow-up filed as #7045

### DIFF
--- a/extension.json
+++ b/extension.json
@@ -583,35 +583,34 @@
 		}
 	},
 	"QUnitTestModule": {
-		"ext.smw.tests": {
-			"scripts": [
-				"tests/qunit/smw/ext.smw.test.js",
-				"tests/qunit/smw/data/ext.smw.dataItem.wikiPage.test.js",
-				"tests/qunit/smw/data/ext.smw.dataItem.uri.test.js",
-				"tests/qunit/smw/data/ext.smw.dataItem.time.test.js",
-				"tests/qunit/smw/data/ext.smw.dataItem.property.test.js",
-				"tests/qunit/smw/data/ext.smw.dataItem.unknown.test.js",
-				"tests/qunit/smw/data/ext.smw.dataItem.number.test.js",
-				"tests/qunit/smw/data/ext.smw.dataItem.text.test.js",
-				"tests/qunit/smw/data/ext.smw.dataValue.quantity.test.js",
-				"tests/qunit/smw/data/ext.smw.data.test.js",
-				"tests/qunit/smw/api/ext.smw.api.test.js",
-				"tests/qunit/smw/query/ext.smw.query.test.js",
-				"tests/qunit/smw/util/ext.smw.localtime.test.js"
-			],
-			"dependencies": [
-				"mediawiki.language.months",
-				"ext.smw",
-				"ext.smw.tooltip",
-				"ext.smw.query",
-				"ext.smw.data",
-				"ext.smw.api",
-				"ext.smw.localtime"
-			],
-			"position": "top",
-			"localBasePath": "",
-			"remoteExtPath": "SemanticMediaWiki"
-		}
+		"scripts": [
+			"tests/qunit/smw/ext.smw.test.js",
+			"tests/qunit/smw/data/ext.smw.dataItem.wikiPage.test.js",
+			"tests/qunit/smw/data/ext.smw.dataItem.uri.test.js",
+			"tests/qunit/smw/data/ext.smw.dataItem.time.test.js",
+			"tests/qunit/smw/data/ext.smw.dataItem.property.test.js",
+			"tests/qunit/smw/data/ext.smw.dataItem.unknown.test.js",
+			"tests/qunit/smw/data/ext.smw.dataItem.number.test.js",
+			"tests/qunit/smw/data/ext.smw.dataItem.text.test.js",
+			"tests/qunit/smw/data/ext.smw.dataValue.quantity.test.js",
+			"tests/qunit/smw/data/ext.smw.data.test.js",
+			"tests/qunit/smw/api/ext.smw.api.test.js",
+			"tests/qunit/smw/query/ext.smw.query.test.js",
+			"tests/qunit/smw/util/ext.smw.localtime.test.js",
+			"tests/qunit/smw/util/ext.smw.util.purge.test.js"
+		],
+		"dependencies": [
+			"mediawiki.language.months",
+			"ext.smw",
+			"ext.smw.purge",
+			"ext.smw.tooltip",
+			"ext.smw.query",
+			"ext.smw.api",
+			"ext.smw.localtime"
+		],
+		"position": "top",
+		"localBasePath": "",
+		"remoteExtPath": "SemanticMediaWiki"
 	},
 	"SpecialPages": {
 		"ExportRDF": "SMW\\MediaWiki\\Specials\\SpecialOWLExport",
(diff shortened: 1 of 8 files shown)


gesinn-it-gea:
CodeRabbit's finding above (async ceiling test only asserting `reloadCalled === false`, not that no future reload timer was scheduled) is addressed in #7047 rather than here: this PR's `tests/qunit/smw/util/ext.smw.util.purge.test.js` is ported to `tests/node-qunit/ext.smw.util.purge.test.js` there (part of #7045's node-QUnit migration), and the fix landed on the ported file — see dd560431c.

PR #7072, opened 2026-07-28
Update recursion methods to check for special titles

### LINKED ISSUE #7055: Deprecated: Use of MediaWiki\Parser\Parser::getPage without a Title set was deprecated in MediaWiki 1.34.
### Setup

- SMW version: 7.2.0-alpha [(c449282)]
- MW version: 1.43.9
- PHP version: 8.3
- DB system and version: 10.11.14-MariaDB / Elasticsearch 7.10.2

### Issue

When using the [`templatefile format`](https://www.semantic-mediawiki.org/wiki/Help:Templatefile_format) the result is broken and comes with a warning: 

```
Deprecated: Use of MediaWiki\Parser\Parser::getPage without a Title set was deprecated in MediaWiki 1.34. 
[Called from SMW\Parser\RecursiveTextProcessor::recursivePreprocess in /src/Parser/RecursiveTextProcessor.php at line 193] 
in w/includes/debug/MWDebug.php on line 385
```
I don't know when it breaks (if it is a 7.0.0+ bug), but [this page](https://wikincat.org/wiki/Especial:Executar_consulta/MARCXMLexporter) was working some time ago. It was producing a **valid XML** file (just click on "**Preparar registros**" button, then "**Baixar registros bibliográficos**"). I know the wiki code is complex, and it is hard to explain, but the goal is to produce a [MARCXML file](https://www.loc.gov/standards/marcxml/Sandburg/sandburg.xml) using SMW, PF (`#arraymap:`), RegexFunctions (`#rreplace:`).

PS: This is the result [in one page](https://wikincat.org/w/index.php?title=Predefini%C3%A7%C3%A3o:Marcxml&fullpagename=B000006), it should be like that inside the exported file.

jaideraf:
GitHub Copilot helped me. Partially fix https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7055

### DIFF
--- a/src/Parser/RecursiveTextProcessor.php
+++ b/src/Parser/RecursiveTextProcessor.php
@@ -190,7 +190,7 @@ class RecursiveTextProcessor {
 	 */
 	public function recursivePreprocess( $text ): string {
 		// not during parsing, no preprocessing needed, still protect the result
-		if ( !$this->parser || !$this->parser->getOptions() || !$this->parser->getPage() ) {
+		if ( !$this->parser || !$this->parser->getOptions() || $this->parser->getTitle()->isSpecial( 'Badtitle' ) ) {
 			return $this->recursiveAnnotation ? $text : '[[SMW::off]]' . $text . '[[SMW::on]]';
 		}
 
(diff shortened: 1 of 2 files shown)


alistair3149:
Thanks for looking into it.
I implemented a full fix in #7080 that supersedes this PR.

PR #7080, opened 2026-08-07
Restore template expansion for queries resolved outside a page parse

### LINKED ISSUE #7055: Deprecated: Use of MediaWiki\Parser\Parser::getPage without a Title set was deprecated in MediaWiki 1.34.

alistair3149:
Fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7055

`RecursiveTextProcessor` bails out of preprocessing when the parser it was handed has no page context. That
happens whenever a query is resolved outside a page parse, for example a `templatefile` download from
`Special:Ask`, because `ResultPrinter::copyParser()` hands out the shared main parser and `TemplateExpander`
primes it with the `Special:Badtitle/Missing` placeholder. Template calls come back unexpanded, and a
`Parser::getPage without a Title set` deprecation notice is written into the response body, which corrupts
file downloads.

That clause has only been live since #6735. It was written as `!$this->parser->getTitle()`, and
`Parser::getTitle()` has a non-nullable `Title` return type across the whole supported MediaWiki range, so it
was always false. #6735 was right that the condition was impossible, but replaced it with
`!$this->parser->getPage()` rather than removing it, which turned a dead branch into a live one.

Removing the clause leaves the parser options check that has been doing the real work all along, and restores
the behaviour of 6.0.1 and every release before it. The leading `!$this->parser` test goes with it, since the
property is typed `Parser` and always assigned in the constructor. Only the export path changes; rendering a
query on a page, and `format=template` on `Special:Ask`, produce identical output before and after.

The unit tests pin the guard condition, and the integration test pins the state it exists for by priming a
real parser through `TemplateExpander` exactly as the file export does. Both fail without the production
change.

This removes the notice SMW itself was emitting, but expansion on a page-less parser can still reach
`Parser::addTrackingCategory()` and `Parser::msg()`, which call `Parser::getPage()` unconditionally, so
content that trips a tracking category or a parser limit still produces the same notice from core.
`ResultPrinter::copyParser()` returning the wiki's shared main parser is the underlying defect and is left
for a separate change.

Supersedes #7072, which removes the notice by reading the page through the deprecated `Parser::getTitle()`
accessor but leaves the branch live, so templates stay unexpanded. Thanks to @jaideraf for the report and for
tracking the notice down to the accessor.

### DIFF
--- a/docs/releasenotes/RELEASE-NOTES-7.2.1.md
+++ b/docs/releasenotes/RELEASE-NOTES-7.2.1.md
@@ -12,6 +12,7 @@ For more detailed information, see the [compatibility matrix](../COMPATIBILITY.m
 * Fixed property and category pages sometimes reporting the wrong Semantic MediaWiki protection status, where a page's change-propagation lock and its "Is edit protected" status could be confused for one another ([#4344](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4344))
 * Fixed property and category pages sometimes remaining permanently locked for a change propagation update even when no such update was pending, leaving them uneditable; affected pages are editable again, and routine category changes such as setting a parent category no longer trigger the lock ([#4344](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4344))
 * Fixed change propagation failing to complete for a property or category connected to a very large number of pages, where selecting the affected pages could exhaust the available memory ([#4344](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4344))
+* Fixed the `templatefile` result format leaving templates unexpanded in the downloaded file, and the PHP deprecation notice that corrupted it ([#7055](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7055))
 
 ## Upgrading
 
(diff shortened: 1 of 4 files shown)


jaideraf:
It works! Thank you! 🥇

PR #7093, opened 2026-08-18
EntityLookup: cast inverse-property lookup result to array

### LINKED ISSUE #7092: TypeError in EntityLookup::getPropertyValues() when querying an inverse property with many matches
## Description

Requesting an inverse property (e.g. `?-Has subobject`) via `#ask` can throw a `TypeError` during page save/re-parse:

```
TypeError: SMW\SQLStore\EntityStore\EntityLookup::getPropertyValues(): Return value must be of type array, SMW\Iterators\MappingIterator returned
```

## Root cause

`EntityLookup::getPropertyValues()` is typed to always return `array`. For inverse properties it delegates to `getPropertySubjects()`:

```php
if ( $property->isInverse() ) { // inverses are working differently
    $noninverse = new Property( $property->getKey(), false );
    $result = $this->getPropertySubjects( $noninverse, $subject, $requestOptions );
}
...
return $result;
```

`getPropertySubjects()` has no return type declaration and can return a `MappingIterator`, since `PropertySubjectsLookup::fetchFromTable()` always wraps its result in one:

```php
public function fetchFromTable( ... ) {
    $res = $this->doFetch( ... );
    $res = $this->iteratorFactory->newMappingIterator(
        $this->iteratorFactory->newResultIterator( $res ),
        [ $this, 'newFromRow' ]
    );
    return $res;
}
```

The `: array` return type on `getPropertyValues()` was added in #6531 ("refactor: improve return type coverage") without accounting for this branch.

## Stack trace (abridged)

```
#0 SQLStore.php: SQLStore->getPropertyValues()
#1 Query/Result/ItemFetcher.php: ItemFetcher->legacyFetch()
#2 Query/Result/FieldItemFinder.php: FieldItemFinder->fetchContent()
...
#17 ParserFunctions/AskParserFunction.php: AskParserFunction->doFetchResultsFromFunctionParameters()
```

## Steps to reproduce

1. Create a property used as `?-SomeProperty` (inverse printout) in an `#ask` query.
2. Ensure the query returns a subject with enough matches that `PropertySubjectsLookup::fetchFromTable()` runs its normal path.
3. Save/re-parse the page containing the query (e.g. via the API, or any hook that forces a re-parse on save, such as TemplateData's `onMultiContentSave`).
4. Observe the `TypeError`.

## Fix

A PR is attached that casts the `getPropertySubjects()` result to an array in the inverse-property branch of `getPropertyValues()`, and adds a regression test (`testGetPropertyValues_Property_Inverse_ReturnsArray_WhenLookupReturnsMappingIterator`) that mocks `PropertySubjectsLookup::fetchFromTable()` returning a real `MappingIterator`, matching production behaviour (the existing `testGetPropertyValues_Property_Inverse` test masked the bug by mocking a plain array instead).

gesinn-it-gea:
Fixes #7092

## Summary

- `getPropertyValues()` is typed to return `array`, but its inverse-property branch assigns it the return value of `getPropertySubjects()`, which can be a `MappingIterator` (since `PropertySubjectsLookup::fetchFromTable()` always wraps its result in one).
- This raises a `TypeError` in production whenever an inverse property (e.g. `?-Has subobject`) is queried and the lookup takes its normal iterator path — reproduced by asking for an inverse property on a page with a matching result set.
- Cast the result to an array right after the assignment, only when it isn't already one.

## Test plan

- Added `testGetPropertyValues_Property_Inverse_ReturnsArray_WhenLookupReturnsMappingIterator`, which mocks `PropertySubjectsLookup::fetchFromTable()` returning a real `MappingIterator` (rather than a plain array, as the pre-existing `testGetPropertyValues_Property_Inverse` test does) — this reproduces the exact production `TypeError` before the fix, and passes after it.
- Ran the full `tests/phpunit/Unit/SQLStore/EntityStore/` suite (203 tests, 434 assertions) — all green, no regressions.

### DIFF
--- a/docs/releasenotes/RELEASE-NOTES-7.2.1.md
+++ b/docs/releasenotes/RELEASE-NOTES-7.2.1.md
@@ -16,6 +16,7 @@ For more detailed information, see the [compatibility matrix](../COMPATIBILITY.m
 * Fixed Elasticsearch and OpenSearch sort keys losing characters on wikis that set `$smwgEntityCollation` to a `uca-*` value, which made entries that differ only in a number sort as if that number were absent ([#7079](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7079))
 * Fixed the `templatefile` result format leaving templates unexpanded in the downloaded file, and the PHP deprecation notice that corrupted it ([#7055](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7055))
 * Fixed pages with a very large tooltip failing to render with a `pcre.backtrack_limit exhausted` error; the tooltip's `title` attribute is now capped at 1024 characters, so readers without JavaScript see a truncated native tooltip ([#7076](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7076))
+* Fixed a fatal error when a query asked for an inverse property, such as the `?-Has subobject` printout, which stopped the page from being saved or rendered ([#7092](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7092))
 
 ## Upgrading
 
--- a/src/EntityLookup.php
+++ b/src/EntityLookup.php
@@ -63,7 +63,7 @@ interface EntityLookup {
 	 * @param Property $property
 	 * @param RequestOptions|null $requestOptions
 	 *
-	 * @return DataItem[]|array|Iterator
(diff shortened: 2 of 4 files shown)


alistair3149:
LGTM, just added follow-up commits to address some doc and test gap.

PR #7094, opened 2026-08-18
Convert smw_hash hex values in batches during update.php

### LINKED ISSUE #7091: No update today, MySQL has gone away
### Setup

- SMW version: 7.2.1 (FTS), upgrading from 6.0.0 (FTS)
- MW version: 1.43.9, upgrading from 1.43.5
- PHP version: 8.2 (litespeed)
- MariaDB 10.11

### Issue

My attempts to upgrade SMW from v6.0.0 to v7.2.1-alpha recently stranded. Every time I run `/maintenance/run.php update`, the script aborts, after 100 seconds or so, with

```
Checking table smw_object_ids ...
   Table already exists, checking structure ...
   ... field smw_id is fine.
   ... field smw_namespace is fine.
   ... field smw_title is fine.
   ... field smw_iw is fine.
   ... field smw_subobject is fine.
   ... field smw_sortkey is fine.
   ... field smw_sort is fine.
   ... field smw_proptable_hash is fine.
   ... changing type of field smw_hash from 'VARBINARY(40)' to 'BINARY(20)' ... Wikimedia\Rdbms\DBQueryDisconnectedError from line 1196 of [...]/includes/libs/rdbms/database/Database.php: A connection error occurred during a query. 
Query: ALTER TABLE `o2a5_smw_object_ids` CHANGE `smw_hash` `smw_hash` BINARY(20) AFTER smw_proptable_hash
Function: SMW\SQLStore\TableBuilder\MySQLTableBuilder::doUpdateFieldType
Error: 2006 MySQL server has gone away

#0 [...]/includes/libs/rdbms/database/Database.php(1182): Wikimedia\Rdbms\Database->getQueryException()
#1 [...]/includes/libs/rdbms/database/Database.php(1156): Wikimedia\Rdbms\Database->getQueryExceptionAndLog()
#2 [...]/includes/libs/rdbms/database/Database.php(647): Wikimedia\Rdbms\Database->reportQueryError()
#3 [...]/includes/libs/rdbms/database/DBConnRef.php(127): Wikimedia\Rdbms\Database->query()
#4 [...]/includes/libs/rdbms/database/DBConnRef.php(302): Wikimedia\Rdbms\DBConnRef->__call()
#5 [...]/extensions/SemanticMediaWiki/src/SQLStore/TableBuilder/MySQLTableBuilder.php(239): Wikimedia\Rdbms\DBConnRef->query()
#6 [...]/extensions/SemanticMediaWiki/src/SQLStore/TableBuilder/MySQLTableBuilder.php(200): SMW\SQLStore\TableBuilder\MySQLTableBuilder->doUpdateFieldType()
#7 [...]/extensions/SemanticMediaWiki/src/SQLStore/TableBuilder/MySQLTableBuilder.php(128): SMW\SQLStore\TableBuilder\MySQLTableBuilder->doUpdateField()
#8 [...]/extensions/SemanticMediaWiki/src/SQLStore/TableBuilder/TableBuilder.php(143): SMW\SQLStore\TableBuilder\MySQLTableBuilder->doUpdateTable()
#9 [...]/extensions/SemanticMediaWiki/src/SQLStore/Installer.php(203): SMW\SQLStore\TableBuilder\TableBuilder->create()
#10 [...]/extensions/SemanticMediaWiki/src/SQLStore/SQLStore.php(429): SMW\SQLStore\Installer->install()
#11 [...]/extensions/SemanticMediaWiki/src/Store.php(493): SMW\SQLStore\SQLStore->setup()
#12 [...]/includes/installer/DatabaseUpdater.php(595): SMW\Store::setupStore()
#13 [...]/includes/installer/DatabaseUpdater.php(552): MediaWiki\Installer\DatabaseUpdater->runUpdates()
#14 [...]/maintenance/update.php(195): MediaWiki\Installer\DatabaseUpdater->doUpdates()
#15 [...]/maintenance/includes/MaintenanceRunner.php(703): UpdateMediaWiki->execute()
#16 [...]/maintenance/run.php(51): MediaWiki\Maintenance\MaintenanceRunner->run()
#17 {main}
```

- The script was run on a copy of the live database. Other semantic extensions were upgraded, too, to stay compatible with SMW.
- `smw_object_ids` contains about 1456000 rows, according to special page statistics for the live database (`Special:SemanticMediaWiki&action=stats%2Ftable`).
- Every 20 seconds or so, an error message is being logged: `Backend log: PHP Deprecated:  Caller from SMW\\DatabaseMetaRepo::loadSmwJson ignored an error originally raised from SMW\\DatabaseMetaRepo::loadSmwJson: [1146] Table '[...].[...]_smw_meta' doesn't exist in [...]/includes/debug/MWDebug.php on line 385`. (Script did not get round to creating the table)

## `max_allowed_packet`?
- By itself, a 2006 error is too generic to be useful but a common response to 2006 errors is the suggestion to increase `max_allowed_packet` (default: 64MB, but 16MB on the client CLI). The default of 16MB should normally be sufficient but if too low a value were the issue, I may be stuck because I don't have the privileges needed to change it myself.
- I encountered the same error message seven months ago - well before SMW 7 and before #6715 - when I ran an upgrade in a Docker container and ended up raising `max_allowed_packet` to 1GB. At the same time, I also increased `innodb_ft_cache_size` and `innodb_ft_total_cache_size`, so it's hard to say which of these was ultimately decisive in making `update.php` run again.
- I fear I may have to revert to my original plan, which is to run maintenance scripts locally and transfer the updated database back to the server (which I did try but I made the mistake of using a more recent version of MariaDB and creating an incompatible database as a result).

It feels a little ironic that what's needed to make `smw_hash` more compact to improve overall performance is what currently appears to be a bottleneck.

alistair3149:
Fixes https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7091

The hex-to-binary conversion that has to precede the VARBINARY(40) to
BINARY(20) change of `smw_hash` ran as a single whole-table UPDATE. On a wiki
with over a million entities that is a statement running for minutes: it
rewrites every row of `smw_object_ids` and maintains the `smw_hash` index for
each one, in one transaction.

Anything that drops the connection inside that window aborts `update.php` with
`DBQueryDisconnectedError` and "2006 MySQL server has gone away". The statement
then rolls back in full, so the next run starts from scratch and fails the same
way. MediaWiki replays a query after a lost connection only when it failed
within three seconds (`Database::DROPPED_CONN_BLAME_THRESHOLD_SEC`), so a
statement of this length is never recovered.

Walk the ID table in `smw_id` ranges instead, committing after each range and
reporting progress. The `LENGTH(smw_hash) = 40` predicate already made the
conversion idempotent, so committed ranges are skipped and an interrupted run
resumes where it stopped. Each range also stays under MediaWiki's replay
threshold, which turns a transient disconnect into a retry rather than a fatal
error.

Until now "every hex row was converted" held by construction, because one
unbounded statement could not miss a row; it now depends on the walk covering
them all. The caller narrows the column immediately afterwards and would
truncate anything left behind, so count the remaining hex rows once the walk
ends and stop the upgrade with an explanation rather than let the schema change
run. Every read the migration makes goes to the primary, so replication lag can
neither hide rows from the walk nor fail a conversion that actually succeeded.

Measured against a 1,605,632 row `smw_object_ids` on MariaDB 11.2: the
conversion goes from one 328 second statement to about a second per range, and
the whole `update.php` run from 7m48s to 4m10s. Batching is faster
overall, not a trade-off, because InnoDB purges and flushes as it goes instead
of carrying one large undo image into the table copy that follows.

That table copy is unavoidable. MariaDB rejects `ALGORITHM=INPLACE` for this
type change, so the `ALTER TABLE` still rebuilds the table; it is the smaller
half of the migration and gets quicker once the conversion no longer leaves a
large transaction behind.

### DIFF
--- a/docs/releasenotes/RELEASE-NOTES-7.2.1.md
+++ b/docs/releasenotes/RELEASE-NOTES-7.2.1.md
@@ -17,6 +17,7 @@ For more detailed information, see the [compatibility matrix](../COMPATIBILITY.m
 * Fixed the `templatefile` result format leaving templates unexpanded in the downloaded file, and the PHP deprecation notice that corrupted it ([#7055](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7055))
 * Fixed pages with a very large tooltip failing to render with a `pcre.backtrack_limit exhausted` error; the tooltip's `title` attribute is now capped at 1024 characters, so readers without JavaScript see a truncated native tooltip ([#7076](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7076))
 * Fixed a fatal error when a query asked for an inverse property, such as the `?-Has subobject` printout, which stopped the page from being saved or rendered ([#7092](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7092))
+* Fixed `update.php` aborting on wikis with a large entity table, where the `smw_hash` conversion ran as one long database statement that a dropped connection discarded in full; it now converts in batches and resumes where an interrupted run stopped ([#7091](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7091))
 
 ## Upgrading
 
--- a/src/SQLStore/TableBuilder/Examiner/HashField.php
+++ b/src/SQLStore/TableBuilder/Examiner/HashField.php
@@ -3,7 +3,9 @@
 namespace SMW\SQLStore\TableBuilder\Examiner;
 
 use Onoi\MessageReporter\MessageReporterAwareTrait;
+use RuntimeException;
 use SMW\Maintenance\populateHashField;
+use SMW\MediaWiki\Connection\Database;
 use SMW\SQLStore\SQLStore;
 use SMW\Utils\CliMsgFormatter;
 use Wikimedia\Rdbms\DBError;
@@ -18,6 +20,17 @@ use Wikimedia\Rdbms\RawSQLValue;
 class HashField {
 
 	use MessageReporterAwareTrait;
+
+	/**
+	 * Number of `smw_object_ids` rows one conversion statement covers.
+	 *
+	 * Small enough that a single statement finishes in about a second even
+	 * on a wiki with millions of entities, which is what keeps the migration
+	 * clear of connection and query timeouts.
+	 *
+	 * @since 7.2.1
+	 */
+	public const CONVERSION_BATCH_SIZE = 5000;
 
 	/**
 	 * @var ?PopulateHashField
@@ -51,8 +64,7 @@ class HashField {
 	 * The LENGTH check distinguishes hex (40) from already-converted
 	 * binary (20) and empty values.
 	 *
-	 * Always runs to completion regardless of row count: this is a
-	 * single server-side UPDATE on MySQL/Postgres, and skipping it would
+	 * Always runs to completion regardless of row count: skipping it would
 	 * leave 40-byte hex values that the subsequent ALTER TABLE cannot
 	 * narrow to BINARY(20) without truncation.
 	 *
@@ -62,12 +74,7 @@ class HashField {
 		$cliMsgFormatter = new CliMsgFormatter();
 		$connection = $this->store->getConnection( 'mw.db' );
 
-		$count = (int)$connection->newSelectQueryBuilder()
-			->select( 'COUNT(*)' )
-			->from( SQLStore::ID_TABLE )
-			->where( 'LENGTH(smw_hash) = 40' )
-			->caller( __METHOD__ )
-			->fetchField();
+		[ 'count' => $count, 'first' => $first, 'last' => $last ] = $this->hexRowBounds( $connection );
 
 		if ( $count === 0 ) {
 			return;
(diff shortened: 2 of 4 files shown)


Nikerabbit:
>  The caller narrows the column immediately afterwards and would
truncate anything left behind

I do not think this is true. The schema change would fail rather than truncate. I had this problem updating my live sites and it failing when page views / job queues were adding new rows with old hash format. I had to make things read-only and stop job processing to be able to complete the schema change.

alistair3149:
> > The caller narrows the column immediately afterwards and would
> > truncate anything left behind
> 
> I do not think this is true. The schema change would fail rather than truncate. I had this problem updating my live sites and it failing when page views / job queues were adding new rows with old hash format. I had to make things read-only and stop job processing to be able to complete the schema change.

Both statements are correct, it depends on whether the DB is in strict mode, i.e. `sql_mode` (or [`$wgSQLMode`](https://www.mediawiki.org/wiki/Manual:$wgSQLMode)).
| `sql_mode` / `$wgSQLMode` | `ALTER` outcome |
| --- | --- |
| `''` ([MW default](https://www.mediawiki.org/wiki/Manual:$wgSQLMode)) | ✅ `Warning 1265` - Truncates |
| `ANSI` | ✅ `Warning 1265` - Truncates |
| `STRICT_TRANS_TABLES`  | ❌ `ERROR 1406` - Data too long, table unchanged |
| `STRICT_ALL_TABLES`  | ❌ `ERROR 1406` - Data too long, table unchanged |
| `TRADITIONAL`  | ❌ `ERROR 1406` - Data too long, table unchanged |

I didn't notice it initially because dev container in SMW only exercises the truncation path.
I have filed #7098 to make the doc more accurate.

PR #7098, opened 2026-08-19
Correct what happens to a leftover hex smw_hash at the schema change

alistair3149:
Follows-up to https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/7094

`HashField` stated that the `VARBINARY(40)` to `BINARY(20)` change truncates any
hex value the conversion left behind. That is only what happens under a
permissive SQL mode. MediaWiki sets `sql_mode` from `$wgSQLMode`, whose default
is the empty string, and there the `ALTER` succeeds with warning 1265 and
rewrites the value to the first 20 bytes of its hex text. A wiki that sets a
strict mode, or sets `$wgSQLMode = null` and inherits a strict server default,
gets error 1406 from the same `ALTER` and no change at all.

Both outcomes lose the hash, so the check that stops the upgrade stays as it is.
Only the explanation was wrong: in a docblock, in the message an operator reads
when the check fires, and in a test comment. State both outcomes in the
docblock, and in the operator message say only that the rows would not survive,
since the distinction does not change what the operator has to do.

The claim predates #7094 in the `migrateHexHashes` docblock, which is where the
wording came from, so correct it there too.

Reported by @Nikerabbit

### DIFF
--- a/src/SQLStore/TableBuilder/Examiner/HashField.php
+++ b/src/SQLStore/TableBuilder/Examiner/HashField.php
@@ -57,16 +57,16 @@ class HashField {
 	}
 
 	/**
-	 * Convert hex-encoded smw_hash values to raw binary.
+	 * Convert hex-encoded `smw_hash` values to raw binary.
 	 *
-	 * Must run BEFORE the column type changes from VARBINARY(40) to
-	 * BINARY(20), because the ALTER would truncate 40-byte hex strings.
-	 * The LENGTH check distinguishes hex (40) from already-converted
-	 * binary (20) and empty values.
+	 * Must run before the installer narrows the column from `VARBINARY(40)`
+	 * to `BINARY(20)`. A row still holding hex at that point loses its hash;
+	 * {@see self::assertNoHexRowsRemain()} says how. The LENGTH check
+	 * distinguishes hex (40) from already-converted binary (20) and empty
+	 * values.
 	 *
-	 * Always runs to completion regardless of row count: skipping it would
-	 * leave 40-byte hex values that the subsequent ALTER TABLE cannot
-	 * narrow to BINARY(20) without truncation.
+	 * Unlike `check()`, this is never skipped on the grounds of row count.
+	 * There is no threshold above which leaving the values in hex is safe.
 	 *
 	 * @since 7.0
 	 */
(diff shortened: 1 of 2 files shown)


tests/phpunit/Unit/SQLStore/TableBuilder/Examiner/HashFieldTest.php
```diff
@@ -311,8 +311,9 @@ function () {
 	}
 
 	public function testMigrateHexHashes_StopsBeforeSchemaChangeWhenHexRowsSurvive() {
-		// The caller narrows smw_hash to BINARY(20) right after this runs,
-		// which would truncate anything the walk missed.
+		// A non-zero count after the conversion must abort the run. Reporting
+		// and carrying on would hand a hex row to the schema change, which the
+		// hash does not come through.
```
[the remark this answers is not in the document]
alistair3149: Most of the tests here are about the migration path though, so it should be somewhat clear already by the time this comment is read.

src/SQLStore/TableBuilder/Examiner/HashField.php
```diff
@@ -132,12 +132,17 @@ private function hexRowBounds( Database $connection ): array {
...
-	 * `smw_id` walk having covered them all. The caller narrows the column
-	 * to `BINARY(20)` immediately afterwards, which truncates whatever is
-	 * left, so confirm the conversion really is complete and stop the
-	 * upgrade with an explanation if it is not.
+	 * The walk converts the `smw_id` range that held hex hashes when
+	 * `hexRowBounds()` measured it. Another process still running an older
+	 * Semantic MediaWiki can write a hex row after that, outside the range,
+	 * so completeness has to be measured rather than assumed.
+	 *
+	 * On MySQL and MariaDB no SQL mode gets a leftover hex value through the
```
Nikerabbit: I do not fully understand what this is trying to say. Is "no SQL" mode a thing, or "none of the possible SQL modes" or "if sql mode variable is not set"? What does "getting a leftover hex value" mean?

alistair3149: It means "none of the possible SQL modes", which was confusing and redundant. I have dropped it and rewrote the comment to make it more clear.

Nikerabbit (approved):

Issue #6129, opened 2025-05-10
Guard against RuntimeException errors in Full-Text Search queries


D-Groenewegen:
### Setup
- MW 1.39.6
- SMW 4.1.3
- PHP 8.1

### Issue

There is a (long-standing) issue with SMW's implementation of Full-Text Search that can cause queries to exit with a 'fatal error of type RuntimeException'. 

The issue is with the end result of the evaluation of strings to produce tokens that can be used in a MATCH condition. To take an example from #6128, 

`{{#ask: [[ MyProperty ::~O'Se* ]] }}`

`O'Se` is evaluated as two strings - wrongfully, but that's not the point here (see #6128 instead). Both of them are discarded because they are too short to be considered tokens (`< $smwgFulltextSearchMinTokenSize`). We end up with a stray asterisk in a clause which causes the RuntimeException:

`(MATCH(t1.o_text) AGAINST ('*' IN BOOLEAN MODE) AND t1.p_id=16)`

It is not just asterisks, but also 'leftover' boolean operators (+/-) and combinations of them that can cause the query to fail ungracefully:

`{{#ask: [[ MyProperty ::~ +Col* +O'Se* ]] }}`

which is represented in the WHERE clause as

`(MATCH(t1.o_text) AGAINST ('+col* +*' IN BOOLEAN MODE) AND t1.p_id=16)`

where the first token turns out fine, but what follows is what ultimately chokes the request.

So all in all, a couple of additional checks would be required to prevent those RunExceptions from happening. In the case of single phrases, such as `[[ MyProperty ::~ a* ]]`, SMW already provides a fallback to a LIKE condition. The same should probably happen for 'non-tokens' that are part of multiple phrases.

D-Groenewegen:
Should be fixed with #7101

PR #7101, opened 2026-08-26
Better guard against invalid syntax / database errors when queries use Full-text Search (MySQL/MariaDB)

D-Groenewegen:
This patch addresses database parsing issues with SMW queries using [Full-text Search](https://www.semantic-mediawiki.org/wiki/Help:Full-text_search) (FTS) in MySQL/MariaDB (SQLite is out of scope).

In some cases, the current `TextSanitizer` for FTS will fail to catch syntax errors in MATCH / AGAINST with the IN BOOLEAN MODE modifier. In others, it is even responsible for creating them, as reported for instance in #6129. If the database is not able to ignore or just misinterpet malformed syntax, it errors out - since SMW 7, the result is a hard database query error. This problem is particularly evident when the storage engine of the `*_smw_ft_search` table is InnoDB (MyISAM is more lenient).
* Incidentally, the default SQL engine set in`$smwgFulltextSearchTableOptions`is MyISAM, but oddly enough, I've noticed that InnoDB may end up being used instead.

### Syntax errors to watch out for and targeted by this update

* Most or many of those that potentially cause database errors are listed and described on https://dev.mysql.com/doc/refman/9.7/en/fulltext-boolean.html - including sequences like `+*` and `-+` 
* Add to the above standalone operators divorced from any token
* the at sign (see below)

One possible cause for these is a fragility inherent in the current implementation:

* array-based 'tokenization' into tokens and operators removes the distinction between leading and trailing operators as used in the original input.
* the removal of tokens in `filterTokens()` can have the side-effect of leaving behind a bit of debris, such as bare operators as well as infelicitous combinations of operators.

This patch provides a safety net (`sanitizeFilteredText()`) at the end of `sanitize()`, but also intervenes at earlier stages, for instance to remove unsupported leading/trailing ops.

### Related changes

* Removed `'* '` from the list of the replacements after `filterTokens()`, which may have been intended to reattach a leading asterisk, but (a) leading asterisks are unsupported as wildcards (and are now removed anyway) and (b) when combined with the first replacement, it can cause unintended contractions, e.g. _`*apple* pear*`_ would become `*apple*pear*`. The unit tests have been corrected accordingly.
* Removed the at sign (`@`) from the exemptions list (and list of replacements) since it can cause the DB to error out. It has no special meaning except in InnoDB's `@distance`, which is not supported by SMW. Added a case to the unit test ('at sign as delimiter').
* Removed the comma from the list of exemptions since it has no special meaning except as a delimiter. The regex-based tokenizer ensures the comma is always treated as such. The unit test case `'+foo*, *bar'` ('plus wildcard combo') has been adjusted.
* While MySQL's documentation says that "if a word is specified with the truncation operator \[i.e. \*\], it is not stripped from a boolean query, even if it is too short or a stopword", SMW discards these words/tokens nevertheless. A fix is provided to not discard them, which should also diminish the possibility of leaving behind solitary truncators. Note that it does not factor in that SMW provides a fallback to LIKE if the entire input string is too short to represent a token.
* Fixed a type declaration in `tokenize()`
* I tried to keep inline comments to a minimum but the complexity and fragility of the code called for at least some clarifications.

### Unit tests
Existing cases in `TextSanitizerTest` have been corrected where necessary and new ones were added ('two letters with wildcard', 'single asterisk wildcard', etc.).

### Variables to take into account when changing `TextSanitizer`

* `TextSanitizer::sanitize()` serves a double duty: it is called by `MySQLValueMatchConditionBuilder` (with `$isSearchTerm` = true) to answer queries; by `SearchTableRebuilder` (with `$isSearchTerm` = false) to populate the database. All of the care taken to ensure that operators are properly in place shouldn't affect the latter situation when `$isSearchTerm` is false.
* Fixes will target queries IN BOOLEAN MODE, the default mode, but although undocumented, SMW also supports IN NATURAL LANGUAGE MODE and WITH QUERY EXPANSION using shorthands. To my knowledge, special operators are not part of their syntax, but syntax errors such as a standalone asterisk can similarly cause database errors.
* InnoDB and MyISAM are the most common MySQL/MariaDB engines, but it would be good to know if anyone uses ARIA and has encountered related issues that are peculiar to ARIA.

### DIFF
--- a/docs/releasenotes/RELEASE-NOTES-7.2.2.md
+++ b/docs/releasenotes/RELEASE-NOTES-7.2.2.md
@@ -13,6 +13,9 @@ For more detailed information, see the [compatibility matrix](../COMPATIBILITY.m
 * Fixed a failed data update or setup state write hiding its real cause when rolling the transaction back also failed ([#7107](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7107))
 * Fixed the tooltip of `{{#info:}}` and `{{#property_link:}}` not working on `Special:ExpandTemplates` when a context title is given ([#7109](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/7109))
 * Fixed links and other markup in a property description not being rendered in the property tooltip ([#5494](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5494))
+* Fixed full-text search queries failing with a database error when the search term left a boolean operator without a term, such as `[[Has text::~O'Se*]]`, where both parts of the term are too short to be indexed and only the wildcard remained ([#6129](https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/6129))
+* Fixed full-text search queries failing with a database error when the search term contained an at sign, as in `[[Has text::~name@example.org]]`
+* Full-text search now keeps a term that carries a wildcard even when it is shorter than `$smwgFulltextSearchMinTokenSize` or is a stop word, matching how MySQL and MariaDB treat the truncation operator. A term such as `[[Has text::~to* be]]` previously searched only for "be" and now also searches for everything starting with "to", so such queries can return more results than before.
 
 ## Upgrading
 
--- a/src/SQLStore/QueryEngine/Fulltext/TextSanitizer.php
+++ b/src/SQLStore/QueryEngine/Fulltext/TextSanitizer.php
@@ -134,8 +134,8 @@ class TextSanitizer {
 		// trailing operators ('*') and operators expected in
 		// both positions ('"')
 		$filteredText = str_replace(
-			[ ' *', '* ', ' "', '" ', '+ ', '- ', '@ ', '~ ', '*+', '*-', '*~' ],
-			[ '*', '*', '"', '"', '+', '-', '@', '~', '* +', '* -', '* ~' ],
+			[ ' *', ' "', '" ', '+ ', '- ', '@ ', '~ ', '*+', '*-', '*~' ],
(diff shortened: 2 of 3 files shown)


D-Groenewegen:
[Edited]

MariaDB's official documentation reads as if commas are the delimiters required for IN NATURAL LANGUAGE MODE : "There are no special operators, and searches consist of one or more comma-separated keywords" (https://mariadb.com/docs/server/ha-and-performance/optimization-and-tuning/optimization-and-indexes/full-text-indexes/full-text-index-overview).
Compare this to MySQL's broader support for delimiters: "determines where words start and end by looking for certain delimiter characters; for example,   (space), , (comma), and . (period)" (https://dev.mysql.com/doc/refman/9.7/en/fulltext-natural-language.html)

Thought this would have been a blocker. It appears, however, that MariaDB's wording is just misleading here and spaces are in fact, treated as proper delimiters, i.e. by virtue of not qualifying as true tokens, in both MyISAM and InnoDB tables. Phew!

D-Groenewegen:
Ready for review, AFAIC
@coderabbitai resume

D-Groenewegen:
Ready as can be.

D-Groenewegen:
Looking great! That's even thorougher and cleaner than I had in mind originally, especially when it comes to those longer invalid sequences. There are some cases I created and checked privately and they still pass after your additional changes.

Just for future reference:
- CodeRabbit brings up the example of `apple "pear` becoming `apple"pear` after a re-join. While the outcome could have been prettier, MariaDB at least does not complain because it treats the solitary quote as a delimiter.

